Issue #954, opened 2014-01-17
Migrate to Django's new test runner


pricem:
Split from Django 1.6 support issue (#618): Django 1.6 uses a new test runner which uses the discovery feature of Python's unittest module.  This changes the requirements for test importing.  See https://docs.djangoproject.com/en/1.6/releases/1.6/#new-test-runner

Because of the large number of changes required and the potential merge conflicts, this issue has not yet been addressed yet (we are using a legacy test runner), and should be done so in the next few months (after stable release 4).

willgearty:
I believe this was fixed in #1779.

Issue #618, opened 2013-03-22
Support Django 1.5


pricem:
Hopefully this isn't going to be a big deal, but you never know.

benjaminjkraft:
Is anyone taking point on moving us to newer django?  Support for Django 1.4 is expected to end in [October 2015](https://docs.djangoproject.com/en/dev/internals/release-process/#lts-releases) and at the rate we've been putting out stable releases, it's entirely possible we'll only put out a single stable release between now and then, and it would be good to have this well-tested before then.

joeleg:
I am currently investigating the remaining issues with the FormWizard

PR #955, opened 2014-01-17
Django 1.6 support (WIP)

pricem:
Changes needed to move to Django 1.6, including:
* Update Python library versions in requirements.txt
* Changed {% url %} template tag syntax
* Removed simplejson, now using built-in json library
* Move to external localflavor library
* Cache program modules in ProgramModuleObj form instead of subclasses
* Changed test runner to mimic old Django behavior (see #954)
* Many regression bug fixes

Work in progress - see discussion of issue #618 and individual commits.


pricem:
Joel has been working on this; he converted over the FormWizard code along with a few other things.  I just merged in the latest main branch code and added a few tweaks to try and clean up the conflicts.

esp/esp/settings.py
```diff
@@ -51,6 +51,7 @@
 if len(DATABASES['default']['PASSWORD']) == 0:
     try:
         DATABASES['default']['PASSWORD'] = DATABASE_PASSWORD
+        DATABASES['default']['PASSWORD'] = DATABASE_PASSWORD
```
mgersh: Why was a copy of this line added? I can't think of any reason other than it being accidental.

esp/esp/themes/theme_data/circles/templates/main.html
```diff
@@ -113,7 +113,7 @@
 </div>
 
 {% load preview %}
-{% miniblog_for_user None as announcements 6 %}
+{% miniblog_for_user AnonymousUser as announcements 6 %}
```
mgersh: I'm not sure what this is supposed to be doing, but I'm wondering if it does what was intended or something else due to #1519

esp/esp/dbmail/cronmail.py
```diff
@@ -45,7 +45,7 @@
 
 import os
 
-@transaction.autocommit
+@transaction.atomic
```
mgersh: Unlike the others, this and the other change to the same file are actual changes in behavior. I haven't taken the time to think about which one is what we want. Relevant: #1666

mgersh:
I reviewed the code that's already here (except for the dataviews changes since they're irrelevant now, and the program modules changes since I'm not sure about relevance), and I'm about to push an update, including a merge of main and a few of my own minor fixes.

Questions besides the comments I've already made:
-Do we need anything in middleware to be inside transactions? If not, we should remove TransactionMiddleware.
-Do we need the program module infrastructure changes anymore? #618 implies that changing them to proxy models, which we've done, would solve the problem. I haven't actually reviewed that code yet; I'm capable of doing so but would rather not if we can just revert it, since it doesn't look like it would otherwise be an improvement.

Bugs I've found:
-Something weird is going on with the admin panel formatting, but only on GET requests, not POST requests. I have no idea why.
-Tests won't run on my dev server
-The debug toolbar doesn't show up

General status: I haven't done major testing. I've gone through the changelogs and checked things myself, not including things deprecated but not yet removed. Locally I'm keeping a list of items from the changelogs that I'm not sure we're done with and will do another pass through those sometime that's not now, then put whatever's left here for others to look at.

If someone else wants to do major testing, I recommend either putting this on telekinesis or starting up a fabric/vagrant dev server separate from your main one, which avoids the switching django version issues.

mgersh:
Another bug: manage.py update now fails. I've narrowed it down to the clean_pyc, the other commands work individually.

Error message: "CommandError: No --path specified and settings.py does not contain BASE_DIR"

mgersh:
Remaining things that I'm not yet happy with, other than stuff mentioned above and general need for testing:
-As of Django 1.5, request.POST only contains data posted with content-types multipart/form-data or application/x-www-form-urlencoded. We use request.POST all over the place, and I don't know how likely it is that some of them are another content-type, so we'll want to make sure to check those when testing.
-In 1.6 there were ORM changes that might affect custom SQL, so we should test everywhere that's used.
-From the 1.6 changelog: "If your CSS/JavaScript code used to access HTML input widgets by type, you should review it as type='text' widgets might be now output as type='email', type='url' or type='number' depending on their corresponding field type." I'm not familiar enough with CSS/Javascript to check this.
-Passing timeout=None to memcached now means "never expire"; it used to be "use default timeout", which for us is a day. I changed instances of 0, which used to also mean "use default timeout" and now means "expire immediately", to None. At this point we could either leave this as is or put in more work to restore the old behavior. Thoughts?

Semi-related suggestion: since we're about to have to test everything anyway, maybe this branch would be a good time to fix #1519. Does anyone think that's a bad idea?

esp/esp/settings.py
mgersh: (I assumed this was an accident and fixed it)

benjaminjkraft:
> Do we need anything in middleware to be inside transactions? If not, we should remove TransactionMiddleware.

We should double-check that ReversionMiddleware will be fine with this; I assume it will.

> Passing timeout=None to memcached now means "never expire"; it used to be "use default timeout", which for us is a day. I changed instances of 0, which used to also mean "use default timeout" and now means "expire immediately", to None. At this point we could either leave this as is or put in more work to restore the old behavior. Thoughts?

I think "never expire" may be better behavior for us anyway; @luac do you have any opinion on this?

benjaminjkraft:
Yuck -- it looks like in django 1.6+, [ReversionMiddleware](https://django-reversion.readthedocs.org/en/release-1.8.1/api.html#revisionmiddleware) has to put the revision data in a separate transaction from the actual data itself (even with `ATOMIC_REQUESTS` on).

It seems like if we want to get around this globally we have to roll our own middleware that would effectively emulate the old behavior.  This would be a little messy, although not infeasible; neither middleware in question is particularly complicated.  (Although it sounds like one of the reasons they excluded transaction handling from middleware is that it had some issues.)

Or we can just remove the TransactionMiddleware and hope that this doesn't cause any problems.  My guess is that in our current usage of reversion, it would be okay to just not worry about it, since we just use it for QSD and TemplateOverrides, and neither of those are likely to cause a mess if a transaction fails -- requests that change them are likely not doing anything else.  I think the main failure mode is that if something goes wrong, the revision that was in progress might be munged with the revision following, which doesn't seem like the worst thing.  But if/when we start using reversion a lot more (for instance in new resources), we may want to think more carefully about this.

tl;dr it's kind of a mess, but we're probably okay for now, and can cross this bridge with the new resources stuff, or when we otherwise start using reversion more.

esp/esp/django_settings.py
```diff
@@ -242,7 +242,9 @@
 
 SESSION_EXPIRE_AT_BROWSER_CLOSE=True
 
-SESSION_ENGINE="django.contrib.sessions.backends.cached_db"
+SESSION_ENGINE = 'django.contrib.sessions.backends.cached_db' #which is persistent storage
+
+ATOMIC_REQUESTS = True
```
benjaminjkraft: I think if we're switching to `ATOMIC_REQUESTS` that deactivates `TransactionMiddleware`, so we may as well remove it now.  (The docs are unclear, so this is assuming that I read the [source](https://github.com/django/django/blob/stable/1.6.x/django/middleware/transaction.py) right.)

benjaminjkraft:
> From the 1.6 changelog: "If your CSS/JavaScript code used to access HTML input widgets by type, you should review it as type='text' widgets might be now output as type='email', type='url' or type='number' depending on their corresponding field type." I'm not familiar enough with CSS/Javascript to check this.

I just pushed a commit that I believe addresses all of the instances of this.  (Basically, just adding a `[type=url]`, `[type=email]`, and `[type=number]` everywhere we have `[type=text]` that could be referring to a django-generated form.)  I haven't yet tested it because I haven't yet set up a dev server for 1.6.

mgersh:
I think my preferred solution for Reversion is switching to reversion.create_revision() instead of middleware. We're not using it in very many places right now, so it'll be easy to add those. If we want to use it more in the future, we can add decorators to those views as we go, and since program module views all get routed through the same place we'd only have to think separately about the views that aren't program modules.

benjaminjkraft:
Yeah, that sounds like a better idea.  I'll have to do some more thinking about what exactly we want out of versioning for the resources stuff.

esp/esp/django_settings.py
```diff
@@ -51,6 +51,7 @@
 # Must be unique for every site hosted
 CACHE_PREFIX="ESP"
 
+ALLOWED_HOSTS = ['*']
```
benjaminjkraft: Django seems to [think](https://docs.djangoproject.com/en/1.8/ref/settings/) this is a bad idea for security reasons, so we may want to have this set in `local_settings` per-set (and possibly have it default to *.learningu.org).  I _think_ our current apache and cache configurations on both *.learningu.org and esp.mit.edu protect us from malicious host-setting, but I wouldn't want to bet on it, and those configs change and are a bit more fragile than this setting.

esp/esp/local_settings.py.travis
```diff
@@ -46,6 +46,7 @@ Learning Unlimited, Inc.
 
 SITE_INFO = (1, 'testserver.learningu.org', 'ESP Test Server')
 CACHE_PREFIX = "Test"
+SECRET_KEY = 'b36cfdfb78aba27ddbde330b23352bc6d40973b4'
```
benjaminjkraft: Will this need to get added to other local_settings files too?

benjaminjkraft:
I reviewed the changes you made (45e3c36..492c5b6); the code looks fine to me.  I left a couple of small comments.

@luac if you have time you might want to glance at the program modules stuff since I don't really know the details of what's going on there, although it seems like at this point the change is fairly small.  Relatedly, do we also need to add an `app_label` to the actual PMO model?

esp/esp/django_settings.py
```diff
@@ -369,6 +371,7 @@ def custom_show_toolbar(request):
         ('esp.utils.shell_utils', '*'),
         )
 
+#   Set test runner to behave like pre-1.6 versions of Django
 #   Exclude apps from testing
 TEST_RUNNER = 'esp.utils.testing.ExcludeTestSuiteRunner'
```
benjaminjkraft: Note that we will need to update this for 1.8; I'm happy to not worry about it for this PR though.

benjaminjkraft:
> In 1.6 there were ORM changes that might affect custom SQL, so we should test everywhere that's used.

My guess based on looking at the queries is that we're okay, but the docs aren't very clear about what the changes were, so we will need to test anything that uses custom SQL carefully.  (I think this is mainly the catalog, the scheduler, and the scrolling class list.)

The docs also say

>  In addition some queries will now produce different results. An example is exclude(condition) where the condition is a complex one (referencing multijoins inside Q objects).

We should check if this affects any of the weird query behaviors we've hit (#1718, #1655, [this thing](https://github.com/learning-unlimited/ESP-Website/blob/main/esp/esp/utils/query_builder.py#L63) in `QueryBuilder`, and possibly others I'm forgetting about).  My guess is that if anything it either fixes our issues, or means that we don't need the workarounds we used, but we may want to make sure to check that nothing breaks.

benjaminjkraft:
> As of Django 1.5, request.POST only contains data posted with content-types multipart/form-data or application/x-www-form-urlencoded. We use request.POST all over the place, and I don't know how likely it is that some of them are another content-type, so we'll want to make sure to check those when testing.

Looks like jQuery defaults to the latter (which is also the default for HTML form submission), and I can't find anywhere where we manually set it, so unless we're doing Ajax that doesn't go through jQuery (e.g. in the scheduler?), we're probably okay.

esp/esp/dbmail/cronmail.py
benjaminjkraft: Interesting question.  I think what we want is `atomic` on the first one, and then the second one should either autocommit, or have `atomic` wrappers somewhere further inside (e.g. around the `mailtxt.send()`).  This way, if `MessageRequest` processing fails, we just back out entirely, because that's all happening in the DB, but if `TextOfEmail` processing fails, we keep whatever we've done already, so that we don't hit the people we've already emailed again, if we can avoid it.

Whichever ones we switch to `atomic` should also have some of their exception-catching removed, because catching exceptions can prevent the transaction from rolling back correctly.  So it may be time to just roll a fix for #1666 into this PR.

luac:
The program module stuff looks like it's just reverting a change that was made earlier in this branch which is now irrelevant because of #1232.

luac:
Regarding having caches never expire: I can think of two things. First, I assume memcached kicks old entries from the cache once it reaches some max memory level, but we should check that it's configured sanely and won't hog the entire server over time.

Second, if there are caching bugs it'll make them persist for longer, which can be bad if we don't notice them. (On the flip side, maybe we could write a tool to catch caching bugs by iterating through all the entries, recomputing and comparing them, at low-traffic times on some periodic schedule. Hmm. I want this now.)

esp/esp/local_settings.py.travis
mgersh: Good catch. Fabric/Vagrant dev servers already have it, including old ones, but we'll need to make sure the deploy process for production sites includes adding one, and add it to whatever script we're using these days to create new sites. And all you people still using legacy dev setup will have to fix it on your dev servers.

esp/esp/django_settings.py
mgersh: I was about to say "this isn't a regression, let's fix it some other time" but then I noticed how easy to fix it is.

mgersh:
I'll add `app_label` to the main PMO definition--I didn't even realize it didn't have one already, since nothing was obviously breaking. As lua said, the rest of the program module stuff in my commits was just a (really annoyingly conflict-ful) revert.

I checked #1718 on this branch when I was trying to figure out the easiest workaround for it, and it hasn't changed. I went to check #1655 just now, and it's also still broken.

While I was checking #1655, the comm panel decided that it needs to run the catalog query, which I noticed because it's so slow, and I don't remember noticing that when checking #1718 so it's probably a regression I introduced somehow.

I like the caching bug detector idea a lot. memcached on esp.mit.edu has a limit of 1 GB, and on diogenes has a limit of 64 MB (which is the default, or at least was the default at some point), so I don't think we need to worry about the cache hogging memory.

mgersh:
Latest commit fixes the Reversion and ALLOWED_HOSTS issues, doesn't fix dbmail or the SECRET_KEY issue.

benjaminjkraft:
I pushed commits to fix dbmail's transaction management.  I haven't yet tested them; if I have time I will do that tonight, but I might not have time.  (I will be away from email for 3 weeks, so if anyone thinks the code needs changes, feel free to just make the changes.)

esp/esp/themes/theme_data/circles/templates/main.html
mgersh: On my second attempt at figuring out what miniblog_for_user does, nope, this is fine. I'm going to also take out the checking for "None" from that code since it won't work anymore and we don't want people trying to use it.

mgersh:
I went through and reviewed all the new commits that aren't mine. One question about the css one: there seems to be some remaining [type=text] without the others in bootstrap-1.0.0.min.css. Does it matter? Are we even using that file? If so, should we maybe consider non-ancient Bootstrap? I don't know enough about that part of our setup to be able to tell.

The email stuff all looks okay to me, although I haven't tested it myself. processed_by seems a little weird but no weirder than it was already.

mgersh:
All that's left to do now, I think:
* Someone needs to review my last two commits
* A whole bunch of stuff needs testing, both generally and specific issues listed above 
* We should start keeping a deploy checklist somewhere so we can put "generate a secret key" on it, since that will have to be done by hand for existing sites and legacy dev servers
* The cache timeouts will need fixing if we decide we want to keep current behavior

benjaminjkraft:
In re bootstrap: I think we're not using that file, as far as I can tell?  It looks like everything includes bootstrap from `{{ settings.CDN_ADDRESS }}/bootstrap` which is LU's Cloudfront CDN.  I don't know what version of bootstrap we have there, or how we update it.  It looks like it's from 2012, so we should probably try to update it at some point, but I think we're probably okay for this pull request?

Your commit eb13bad looks good.  (I don't see another one?)

esp.mit.edu memcached's current oldest key is about 10 days old (the cache has not been flushed for almost a month), and we're currently using 940MB of memory.  (If keys expire, it in general does not evict them until it has something new to put there.)  The behavior might change a bit if the keys didn't have expiration times, but that suggests to me that if we let things be unlimited, they would not last much longer than that in general.  I'd suggest we try just letting the lifetimes be infinite, and see what the oldest key is after a month or two as a check.  (For reference, you can get the oldest key by running `echo stats items | nc localhost 11211 | grep age | cut -f 3 -d ' ' | sort -n | tail -1`.)

mgersh:
I've run through general testing using an old stable release checklist as a guideline, and made sure I hit everything that uses extra(). All issues I've found have been fixed if they're regressions between main and this branch, or reported in separate issues if not. So I think this is about ready, once I fix the test error that I noticed just after typing the first part of this sentence, and other people do whatever other testing you think is needed.

The other commit I was talking about was 5d597bf. I just pushed four more. Note that c378794 is hidden above a comment because I made it several days ago.

benjaminjkraft:
I'm not convinced that all the transaction management in custom forms is actually necessary, since some of them just contain a single statement that I think should be atomic anyway, but I'm happy to leave it, since I don't think there's a performance penalty either.

The remaining commits look good to me.  I'm going to do a little more testing for the sake of completeness, and then merge later this afternoon, unless anyone else wants more time to test or otherwise look this over.

mgersh:
I copied a deploy checklist from a previous stable release and made some edits so we can start keeping track, even though this isn't a stable release yet. Current version is at https://docs.google.com/document/d/1oDsmCjvJ245b7m_EvrKzz-YGwGdKrQM0hWXvEnIIEZU/edit.

benjaminjkraft:
I found a minor regression in middleware, which I fixed.  I think after someone looks over that, and we merge main in and check that that doesn't obviously break anything, we should be good to go.

mgersh:
Okay, I'm happy with the middleware change.

benjaminjkraft:
And main did cause one issue from the `simplejson` to `json` switch, which I fixed.

mgersh:
I think we're ready to merge once tests pass!

benjaminjkraft:
And they didn't.  (Fixed.)

benjaminjkraft:
Thanks @pricem and @joeleg for starting this and @mgersh for picking it up and finishing it!

Issue #1097, opened 2014-03-01
Old onsite morph gets stuck at required modules


jmoldow:
If, for example, the medical form is required, you can't get around it to register a student. Granting a medical form bit requires being an admin, which you are not while morphed.

There should either be a morphed-only part of these views to manually add the required record to bypass it, or non-profile required modules shouldn't take effect while onsite morphing.

PR #1313, opened 2014-09-16
Allow morphed users to bypass required modules

pricem:
This small change addresses #1097 (the latter option: "non-profile
required modules shouldn't take effect while onsite morphing).

Submitting as a separate pull request so someone can sanity check the
logic.  I'm not 100% sure this is a good idea; it means there is more
responsibility for the onsite volunteer to make sure the student is
authorized for whatever changes they are making (the Web site won't slow
them down).  It also means that the student has to walk the onsite
volunteer through the whole profile form if, for some reason, they have no
profile.


esp/esp/program/modules/base.py
```diff
@@ -212,6 +213,8 @@ def findModule(request, tl, one, two, call_txt, extra, prog):
                 other_modules = moduleobj.findCategoryModules(False)
                 for m in other_modules:
                     m.request = request
+                    if request.user.is_morphed() and not isinstance(m, RegProfileModule):
```
luac: Maybe this should be `request.user.updateOnsite(request)`? (This is the call that is used to determine whether to skip deadlines when you're using the onsite morph, as opposed to normal morph.)

luac:
Also, there is already a message at the top of at least the medical form: https://github.com/learning-unlimited/ESP-Website/blob/main/esp/templates/program/modules/formstackmedliabmodule/medliab.html#L9

However, it doesn't show when morphed, and anyway the bypass module only works for admins, not onsite.

luac:
Addressed my comment above. With this change, I think this is probably fine behavior; if a student is at the program checked in and they have a profile, anything on the onsite interface is probably okay.

benjaminjkraft:
Okay, this makes sense to me -- we could perhaps be clearer about this in documentation but the convention is certainly that onsite morphing has admin powers whereas regular morphing is ~identical to what the user sees, and this fits with that.

PR #1370, opened 2014-10-11
Add generic thanks page for when a mainpage is closed

benjaminjkraft:
Sometimes it is useful to have a program module deadline open even if the
corresponding /MainPage deadline is closed.  However, many program modules
redirect to the main page on completion, which causes the user to get confused
if that page is in fact a deadline error.  Now modules can optionally have the
redirect be changed to a generic thanks page if the core module is closed.  I
added this to the teacher events module, which is the one MIT wanted it for.

This is a bit hacky, so suggestions of better ways to do it are welcome.  Or we may just not want it in main.


benjaminjkraft:
Closing this and moving to an issue, #1577.

PR #1415, opened 2014-10-19
Use the same API for the two-phase lottery even with a single priority

benjaminjkraft:
Stanford was having an issue where the selected priority classes were not loading in the two-phase timeblock edit interface.  This was because `/json///lottery_preferences` was returning a different format when there was only one priority level.  I changed it to return the same format.  This is almost certainly not the right fix in general; if we want to use that format always we should remove the rest of the function; but first we should test that it works in any other interfaces and configurations that are in use.

So don't merge this; I wanted to put it in a pull request so we can figure out what the right fix is and do that instead.


luac:
I pushed a change that takes the second approach.

benjaminjkraft:
Okay, that looks good to me.  (We talked in person, and the sections will come in the same order from both calls because they get sorted by id ascending by django, so unless a section gets added in the middle, in which case an AssertionError is probably a good thing, the merge should work fine.)

PR #1489, opened 2014-11-30
Small Vagrant improvements

szhu:


jmoldow:
The string stuff looks good to me, but since I don't have Vagrant on my computer, I don't know about the ``bootstrap.sh`` addition or the ``vb.name`` addition. Can someone who has worked with Vagrant take a look at this PR? @pricem

mgersh:
I took a look at this.

For the vb.name, I like the idea, but it causes problems when you want to have more than one dev server, which is necessary for testing dev setup like I'm doing right now. If someone knows how to add a timestamp to the name, that would do it, otherwise we can leave as is and let the developer change the name if they care.

The bootstrap.sh errored when trying to install dictionaries-common, which various things depend on, and I haven't been able to figure out why. I didn't have this issue when testing new dev setup without these changes a few days ago, and I don't think there are changes to dependencies since this branch that would explain it.

benjaminjkraft:
I think some of these changes were made by @btidor in #1834, and the others will need to be merged on top of that.  @btidor do you think this should also be merged, and if so any interest in updating it?

benjaminjkraft:
Talked to @mgersh about this; it seems like the bootstrap.sh changes are less necessary after #1834, so if we want to get them working we can just do that later.  It seems like the `vb.name` thing is not desirable at least in the current form.  I'm going to cherry-pick e90f7a8 into a new pull request; if anyone wants to fix up either of the other two changes, go for it.

szhu:
Thanks!

btidor:
Okay:
* e90f7a8 / GH-1853: Looks good to me!
* b25d688: I'm inclined not to merge this. It doesn't make a lot of sense to just run `update_deps`, particularly as we're consolidating up the setup process. Having Vagrant kick off the entire setup process might be nice, but the way to do this is to run `fab` with a Vagrant provisioner. This isn't ideal because Vagrant doesn't allow user interaction during provisioning, which places additional constraints on the way we write our `fabfile`.
* b18ab0c: Setting `vb.name` would make things just a little bit nicer. I might look into the implications of this later.

Thank you, @szhu!

PR #1490, opened 2014-11-30
Add berkeley-alpha theme

szhu:
Screenshots:

| home-loggedout |
| --- |
| contact-loggedin |


benjaminjkraft:
I'm going to work on cleaning this up.  Since it looks like @szhu's branch was deleted, I'm going to make a new branch, and a new PR.

Incidentally, I think the license is not an issue -- I think we should just have chapters provide their own images for this theme, just as they provide their own logos for a number of themes.  (But I do think we should avoid having anything in the repo that's not under the same license.)

PR #1502, opened 2014-12-12
Fix cur_grade to be relative to the relevant program's start date

### LINKED ISSUE #1216: Two-Phase Student Registration doesn't use student's rising grade
Specifically, it doesn't respect the increment_default_grade_level program tag (introduced in b6f6cf0e2a657b1d205cfd52eee22bb2bbc4be7c).

luac:
This should fix #1216. Also includes some other cleanup around grade and schoolyear logic.

### DIFF
--- a/esp/esp/program/models/__init__.py
+++ b/esp/esp/program/models/__init__.py
@@ -1047,8 +1047,9 @@ class Program(models.Model, CustomFormsLinkModel):
     @cache_function
     def incrementGrade(self): 
         """
-        Increments the effective school year of the program.
+        Get the value of the "increment_default_grade_levels" tag.
 
+        This tag increments the effective school year of the program.
         Also affects how grade ranges for this program are displayed,
         to say "rising Xth grade" rather than just X.
 
--- a/esp/esp/program/models/__init__.py
+++ b/esp/esp/program/models/__init__.py
@@ -1046,6 +1046,14 @@ class Program(models.Model, CustomFormsLinkModel):
 
     @cache_function
     def incrementGrade(self): 
+        """
+        Increments the effective school year of the program.
+
+        Also affects how grade ranges for this program are displayed,
+        to say "rising Xth grade" rather than just X.
+
+        See ESPUser.program_schoolyear.
+        """
         return int(Tag.getBooleanTag('increment_default_grade_levels', self, False))
     incrementGrade.depend_on_row(lambda: Tag, lambda tag: {'self' :  tag.target})
     
--- a/esp/esp/program/models/__init__.py
+++ b/esp/esp/program/models/__init__.py
@@ -509,7 +509,7 @@ class Program(models.Model, CustomFormsLinkModel):
         # We can restore this one later if someone really needs it. As it is, I wouldn't mind killing
         # lists['all_former_students'] as well.
         del lists['all_students']
-        yog_12 = ESPUser.YOGFromGrade(12)
+        yog_12 = ESPUser.YOGFromGrade(12, ESPUser.program_schoolyear(self))
         # This technically has a bug because of copy-on-write, but the other code has it too, and
         # our copy-on-write system isn't good enough yet to make checking duplicates feasible
         lists['all_current_students'] = {'description': 'Current students in all of ESP',
--- a/esp/esp/users/models/__init__.py
+++ b/esp/esp/users/models/__init__.py
@@ -923,6 +923,8 @@ are a teacher of the class"""
             schoolyear = curyear
         else:
             schoolyear = curyear + 1
+        if program is not None:
+            schoolyear += program.incrementGrade() # adds 1 if appropriate tag is set; else does nothing
         return schoolyear
 
     @cache_function
@@ -937,8 +939,6 @@ are a teacher of the class"""
             if regProf and regProf.student_info:
                 if regProf.student_info.graduation_year:
                     grade =  ESPUser.gradeFromYOG(regProf.student_info.graduation_year, ESPUser.current_schoolyear(program))
-                    if program:
-                        grade += program.incrementGrade() # adds 1 if appropriate tag is set; else does nothing
 
         return grade
     #   The cache will need to be cleared once per academic year.
--- a/esp/esp/users/models/__init__.py
+++ b/esp/esp/users/models/__init__.py
@@ -902,17 +902,8 @@ are a teacher of the class"""
         return len(User.objects.filter(username=username.lower()).values('id')[:1]) > 0
 
     @staticmethod
-    @cache_function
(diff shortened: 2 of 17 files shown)


esp/esp/program/models/__init__.py
```diff
@@ -1046,6 +1046,14 @@ def getShirtInfo(self):
 
     @cache_function
     def incrementGrade(self): 
+        """
+        Increments the effective school year of the program.
```
jmoldow: This is not the behavior of the function. The function returns a value that the caller is responsible for adding to the school year. The function returns either 0 or 1, depending on the Tag increment_default_grade_level.

esp/esp/users/models/__init__.py
```diff
@@ -920,9 +917,23 @@ def current_schoolyear(program=None):
...
+        else:
+            # "now" is actually whenever the program ran or will run
+            now = program.dates()[0]
+            schoolyear = ESPUser.current_schoolyear(now)
+        schoolyear += program.incrementGrade() # adds 1 if appropriate tag is set; else does nothing
+        return schoolyear
+    program_schoolyear.__func__.depend_on_row(lambda: Tag, lambda tag: {'program': tag.target})
+
+    @cache_function
+    def getYOG(self, program = None):
```
jmoldow: Can you add a docstring here?

When I was reading the code changes above, I was confused why ``getYOG`` would take a ``program`` argument. It's clear from looking at the implementation, but for clarity, the behavior should be documented.

esp/esp/users/models/__init__.py
```diff
@@ -931,14 +942,26 @@ def getGrade(self, program = None):
                 regProf = RegistrationProfile.getLastForProgram(self,program)
```
jmoldow: nitpick: space after comma

esp/esp/users/models/__init__.py
jmoldow: nitpick: no spaces around the assignment operator for default argument values.

esp/esp/users/models/__init__.py
```diff
@@ -931,14 +942,26 @@ def getGrade(self, program = None):
...
-                    grade =  ESPUser.gradeFromYOG(regProf.student_info.graduation_year, ESPUser.current_schoolyear(program))
-                    if program:
-                        grade += program.incrementGrade() # adds 1 if appropriate tag is set; else does nothing
+                    return regProf.student_info.graduation_year
+        return None
+    getYOG.get_or_create_token(('self',))
+    getYOG.depend_on_row(lambda: StudentInfo, lambda info: {'self': info.user})
 
+    @cache_function
+    def getGrade(self, program = None):
```
jmoldow: nitpick: no spaces around the assignment operator for default argument values.

esp/esp/users/models/__init__.py
```diff
@@ -902,12 +902,9 @@ def isUserNameTaken(username):
         return len(User.objects.filter(username=username.lower()).values('id')[:1]) > 0
 
     @staticmethod
-    def current_schoolyear(program=None):
-        if program == None:
+    def current_schoolyear(now=None):
```
jmoldow: docstring

esp/esp/users/models/__init__.py
```diff
@@ -920,9 +917,23 @@ def current_schoolyear(program=None):
             schoolyear = curyear + 1
         return schoolyear
 
+    @staticmethod
     @cache_function
-    def getGrade(self, program = None):
-        grade = 0
+    def program_schoolyear(program):
```
jmoldow: docstring

esp/templates/errors/program/wronggrade.html
```diff
@@ -14,7 +14,7 @@
 <br />
 <em>Sorry, you do not appear to match the required grade range for this program. Please browse through
 some of our other <a href="/learn/index.html">programs</a>, though!</em><br />
-If your grade, <strong>{{request.user.getGrade}}</strong>, is incorrect, please <a href="/myesp/profile" title="ESP Profile Editor">edit your profile</a> or <a href="mailto:{{ DEFAULT_EMAIL_ADDRESSES.default }}">contact an administrator</a> to change it.
+If your graduation year, <strong>{{yog}}</strong>, is incorrect, please <a href="/myesp/profile" title="ESP Profile Editor">edit your profile</a> or <a href="mailto:{{ DEFAULT_EMAIL_ADDRESSES.default }}">contact an administrator</a> to change it.
```
jmoldow: "high school graduation year"

mgersh: This should probably just link to grade change requests now.

jmoldow:
> I wasn't sure about deprecating Program.incrementGrade because it still does something (namely it causes the text to say "rising 9th graders - graduating 12th graders" instead of just "9 - 13") but maybe it's less confusing to have only one way of tweaking a program's school year instead of two.

We might want to rename that function (as well as the Tag it uses), make it return a boolean instead of an integer, and have it only affect the catalog text (as opposed to also being used to change the schoolyear).

Right now, if you want to use the tag to change the catalog text, you can't also use your new feature, because then the year will be incremented twice, and everything will be off by 1.

We should consolidate to just one system (yours), and change the name / behavior of ``incrementGrade`` to reflect its usage as only a way of changing the catalog text. And maybe make ``incrementGrade`` always return False when your system isn't being used, since it doesn't make sense to pretend to increment the grade.

jmoldow:
Code looks good. I just left some comments about documentation, and as I said above, I think we should deprecate the use of ``incrementGrade`` and replace it with something that deals solely with changing the catalog text.

esp/templates/errors/program/wronggrade.html
jmoldow: I'm not sure we want to make it more obvious that you can change your grade and get into a program that you don't have access to? Shrug.

mgersh: Then why did we make grade change requests a feature in the first place?

luac:
So my initial thought when adding the schoolyear override tag was that I thought it would be necessary for programs to have a schoolyear associated with them independently of `Program.dates`. I wasn't aware that the current behavior of `current_schoolyear` when called with a program was already to use `Program.dates` to calculate the schoolyear. That calculation has worked fine for all programs so far, except for summer programs where we've used `incrementGrade`. So, on second thought, I don't think we need my new tag.

I think it'll be easier to just keep `incrementGrade` as it is now, rather than migrate it over to two new tags.

luac:
I removed the schoolyear override tag and fixed the docstrings. I also added a cache dependency that I missed, which is that `program.dates()` may change if you add new timeslots to the program.

PR #1505, opened 2014-12-19
Cancelling classes email

gauravjsingh:
This emails students with SSIs in a class when the last section of the class is cancelled. It also creates a new template for this email. I tested that it emails students with a valid SSI that are not otherwise being emailed.


esp/esp/program/models/class_.py
```diff
@@ -1013,20 +1013,27 @@ def cancel(self, email_students=True, include_lottery_students=False, explanatio
...
         email_title = 'Class Cancellation at %s - Section %s' % (self.parent_program.niceName(), self.emailcode())
+        ssi_email_title = 'Class Cancellation at %s - Class %s' % (self.parent_program.niceName(), self.parent_class.emailcode())
         if email_students:
             email_content = render_to_string('email/class_cancellation.txt', context)
             template = Template(email_content)
             #   Send e-mail to each student
-            for student in self.students(student_verbs):
+            q_ssi = Q(studentsubjectinterest__subject=self.parent_class) & nest_Q(StudentSubjectInterest.is_valid_qobject(), 'studentsubjectinterest')
+            ssi_students = ESPUser.objects.filter(q_ssi).exclude(id__in=self.students(student_verbs)).distinct()
+            students = [(i,True) for i in self.students(student_verbs)] + [(i,False) for i in ssi_students]
```
benjaminjkraft: This seems like it will cause students with both SRs and SSIs to get two emails.

esp/esp/program/models/class_.py
```diff
@@ -1013,20 +1013,27 @@ def cancel(self, email_students=True, include_lottery_students=False, explanatio
         else:
             student_verbs = ['Enrolled']
 
+        email_ssis = email_students and all([sec.isCancelled() for sec in self.parent_class.get_sections() if sec!=self])
```
benjaminjkraft: We should probably also email SSIs iff we are including lottery students.

esp/templates/email/class_cancellation_body.txt
```diff
@@ -1,4 +1,5 @@
...
+We regret to inform you that class {% templatetag openblock %} if classreg {% templatetag closeblock %}section {{ sec.emailcode }}{% templatetag openblock %} else {% templatetag closeblock %}{{ sec.parent_class.emailcode }}{% templatetag openblock %} endif {% templatetag closeblock %}: {{ sec.title }} at the upcoming {{ prog.niceName }} has been cancelled.  {% templatetag openblock %} if classreg {% templatetag closeblock %}{% if sec.friendly_times|length_is:1 %}The class had been scheduled for {{ sec.friendly_times.0 }}, so you can switch into another class during that period.  You can do this on our Web site or on-site at {{ prog.program_type }}.{% else %}You can sign up for other classes on our Web site or on-site at {{ prog.program_type }}.{% endif %}{% templatetag openblock %} else {% templatetag closeblock %}You can sign up for other classes on our Web site or on-site at {{ prog.program_type }}.{% templatetag openblock %} endif {% templatetag closeblock %}
```
benjaminjkraft: I'm not super-happy with the duplication of the else-block here, but I'm not sure that there's a good way to fix it with the current double-render strategy.  Also, I think admins especially at places that aren't MIT may be confused by seeing template tags in the email they get.  Since the only thing that changes is `classreg` it might be worth just having the admin email include both bodies successively if we emailed students with SSIs.  This should allow removing all the `{% templatetag openblock %}` stuff and simplify the conditionals in this template a bit.

gauravjsingh:
Do you think it would be better without double rendering? I can remove that, and it might make other things simpler.

benjaminjkraft:
Yeah, I think the double-rendering is hacky, overcomplicated, unnecessary, and doesn't really have any benefit :-) .

benjaminjkraft:
Ok, looks good now.

Issue #1131, opened 2014-03-06
Registration receipt should have more information


benjaminjkraft:
The thing that students see when they go to /confirmreg currently is not very useful if they want it as a receipt of any kind.  Ideally it should include:
*The chapter name, and ideally also the program name in larger letters near the top.
*Any costs they are charged for (the program fee and any optional costs), and the amount owed (including any credit card payments, if applicable)
*A block of text configurable without template overrides if chapters want more.

jmoldow:
Whoever looks at this ticket might want to see if there's anything useful that can be taken from the template that was removed in #1515.

hwatheod:
Also, it would be useful if there were some way to specify dates to relative a program within the receipt, and have these computed automatically.  Currently, for each Stanford program, I copy the confirm/confirmemail receipts and then manually change the relevant dates (program dates, reg close date).

willgearty:
Some things that should definitely be added:

- A button to print the receipt
- Integrate with themed pages (and then use javascript to print only the receipt like in https://github.com/learning-unlimited/ESP-Website/pull/3469)
- A button to go back to the registration page
- Student schedule should be formatted like in https://github.com/learning-unlimited/ESP-Website/pull/3376
- complete dates for the program (although this is already in the schedule)
- the date that (class) registration closes? ("You have until X/Y/Z to change your classes")

Some things that could be added (the student schedule UI, https://github.com/learning-unlimited/ESP-Website/pull/3627, could be used to modify whether these are shown):

- Barcode
- Amount owed
- Items purchased

PR #1515, opened 2014-12-23
Remove an unused template

benjaminjkraft:


jmoldow:
Since this is dead code, I'll merge the pull request. However, for #1131, someone should take a look at this template and see if there's anything useful in here that should be incorporated into a better default registration receipt.

PR #1524, opened 2014-12-30
Remove old accounting

mgersh:
This stuff has been deprecated since the stable release a year ago. I think it's safe to get rid of it now. This also clears the way for removing the remaining bits of DataTree that we still have.

When deploying this, you'll need to fake the initial migration for accounting_core, and maybe also for accounting_docs.


esp/esp/accounting_core/models.py
```diff
@@ -32,243 +32,3 @@
   Phone: 617-379-0178
   Email: web-team@learningu.org
 """
```
benjaminjkraft: Is there a reason to leave the header?

esp/esp/accounting_docs/models.py
```diff
@@ -32,265 +32,3 @@
   Phone: 617-379-0178
   Email: web-team@learningu.org
 """
```
benjaminjkraft: Likewise.

benjaminjkraft:
Is removing all the code from the migrations the right thing to do here?  (I'm legitimately not sure.)

mgersh:
I don't think removing the code from the migrations makes sense. We still need the ones I'm adding here until they've all been run, and leaving those around without the earlier ones seems likely to break something. Looks like we kept all the migrations around when removing shortterm.

benjaminjkraft:
I meant removing all the logic from 0005 -- but if that's what was done for shortterm that's fine.

mgersh:
Oh. I just did that because it's a migration in accounting (the new one) that depends on models from accounting_docs still existing, and getting rid of the code seemed easier than fixing it to properly use a frozen ORM.

benjaminjkraft:
Ah, I get it.  Makes sense.

benjaminjkraft:
Ok, this looks good to me.  Since I don't know much about the accounting system, I'll leave it for another day or two in case anyone else wants to take a look before merging.

pricem:
When is the appropriate time to remove the accounting_core and accounting_docs apps from INSTALLED_APPS?  (We have to deal with these questions for the datatree, so I figured I may as well ask.)

mgersh:
They need to stay there until after the new migrations have been run on all the sites, and I think it's fine to remove them from INSTALLED_APPS and delete the remaining files once that happens.

pricem:
Cool.  So, in other words, in the next release cycle.  Is anyone planning to give the datatree app the same treatment you gave accounting_{core, docs} this time around, or did I just sign myself up for that?

jmoldow:
I don't recall anyone having volunteered for that yet. Feel free to go for it, @pricem.

Issue #1364, opened 2014-10-09
Class flags: fix display issues with "the flag"


benjaminjkraft:
if "the flag" is selected, but not at the top level, it wraps in such a way that it looks weird.  This should be fairly easy to fix.

benjaminjkraft:
Fixed/moot due to #1525.

Issue #1352, opened 2014-10-09
autocomplete for user filter for class flags on /classsearch


benjaminjkraft:
It should be possible to search for class flags made/modified by a particular user, or by the user doing the search.

benjaminjkraft:
Added in #1525.

benjaminjkraft:
Er, somewhat added as "search for a substring of the username", but not with any autocomplete.  I'll leave this open to add an actual user select autocomplete.

ny1411:
Hello @willgearty  
Could you let me know if this issue is open? 

I'm interested in contributing to this repository, as it would be beneficial when submitting **GSOC 2026** proposals to your org.

willgearty:
@ny1411  Thanks for your interest in contributing to our project! I've assigned you to this issue. Please make sure to check out our [instructions for getting started](https://github.com/learning-unlimited/ESP-Website/wiki#getting-started) and follow our [contribution rules](https://github.com/learning-unlimited/ESP-Website/wiki#rules). Let me know if you have any questions.

ny1411:
@willgearty i have completed the project setup
docker setup is completed and the website runs on local system

can i get the issue description? 
the current issue description is slightly unclear to me

ny1411:
hey @willgearty 
i tried to understand the issue with its title (autocomplete for user filter for class flags on /classsearch) and planned how i can implement it (using the help of llms)

here's what llm and i planned:
### Key changes:
1) Backend filter:
- Create a new QueryBuilder input class (e.g., `AjaxUserInput`) that emits user ID/username and can specify an autocomplete endpoint.
- Add a `SearchFilter` to `ClassSearchModule.query_builder()` named “flag_creator” using that input, filtering `flags__created_by` (exact match) and keeping existing flag time inputs optional.
- Add a lightweight JSON endpoint under ClassSearchModule (e.g., `ajaxflagusers`) returning `{id, username, full_name}` for users who have created flags in the current program, limited and case-insensitive on username (per answers). Guard with `needs_admin`.

2) Frontend UI:
- Implement a React input component in `query-builder.jsx` to render an `<input>` with jQuery UI autocomplete using the provided endpoint; store selected user ID (fallback: show error if user not selected).
- Wire the new component into `window` for QueryBuilder to use.

3) Wiring:
- Ensure `class_search.html` still includes jQuery UI (already bundled) and loads the new component (same bundle).
[...]

Issue #1367, opened 2014-10-10
Keyboard shortcuts for class flags


benjaminjkraft:
I'm not sure what the right way to do this is, but it would be useful.

PR #1525, opened 2014-12-30
Flag search rewrite

### LINKED ISSUE #1316: Class flag filters for "any of the flags", "none of the flags"
Since flags are not displayed on the dashboard, the only way to get a list of all classes with _any_ flag is to go to the manage flags page, select "any of", then select "the flag" once for every possible flag.

There should be a top-level filter for "any of the flags", which will filter classes that have at least one assigned flag.

The same goes for "none of the flags".

Requested by Stanford.

### LINKED ISSUE #1359: Rewrite class flags frontends to use React
This is something I'm not sure if I want to do.  If I were to write class flags today I would definitely want to use [React.js](https://facebook.github.io/react/) which makes building dynamic interfaces not suck -- you tell it how you want things to look based on some internal state, and how to update that state, and it does all the DOM updating for you.  It might also have some performance benefits, both in saving sending lots of HTML to and from the server and in making the client faster.  I'm not sure if it's worth rewriting class flags to use it at this point, but it might be.

(I'd also love to use React for more things on the site in general, because it's great.  But this particular type of interface seems especially suited to using React.)

### LINKED ISSUE #1360: Class flag query builder: make "not" separate
This might or might not be a good idea; I've resisted because I can't figure out how to make a good interface.  But it might still be a good idea as the list of filters gets longer.

### LINKED ISSUE #1366: Filter class flags by title

benjaminjkraft:
~~This isn't done yet (so don't merge), but I figured I'd put the code up if anyone has comments.~~  I rewrote what was formerly the class flag search with a much more general backend, which is in `esp.utils.query_builder`.  This should be a lot easier to extend to add new fields and make other modifications, and it's general enough that we should be able to build other search interfaces without too much trouble.

Notes:
* I might eventually put it into a separate library -- the query builder core is not very tightly bound to the ESP website or even django really -- but for now it's here.
* The client-side part is all in React; I also added another JS library called [lodash](http://lodash.com/) which gives various useful functional primitives like `_.map` that work in all browsers.
* Fixes #1359.  Also fixes #1360, #1366, #1316, and makes #1364 moot.  #1352 should be solvable but will be a bit tricky; I'll probably try to convert ajax_autocomplete to a React component.
* I wrote python tests, both for the query builder core and the module that uses it to make a class search.  I do intend to write JS tests sometime, but I might wait until the new scheduler gets merged, and then we can use the same testing tools for all our JS.

Known TODOs:
- [x] fix styling because it's unusably ugly right now
- [x] add documentation
  - [x] docstrings, docstrings everywhere (add some, update some)
  - [x] documentation of data interchange formats
  - [x] documentation of how to write a new input
  - [x] documentation of the interface of the QueryBuilder class
- [x] remove old class flag search page and redirect to this one
- [x] add new features including various open class flags issues
  - [x] Fix #1366
  - [x] Fix #1316
  - [x] teacher's username
- [x] tests
  - [x] python
  - ~~JS?~~


benjaminjkraft:
Ok, I've added some tests, and this is now done and ready for review.

luac:
Some UX comments/issues from trying it out on my dev server:
* The "not" button is still confusing; when it's not selected, the fact that the word is there makes it read as if you're searching for the negation of your query when you're not.
* Under "any of"/"all of", the + and - buttons look the same visually as the "not" button, yet they do structurally different things, which is confusing. Somewhat similar for the optional inputs under "the flag"/"any flag".
* When you search for "title containing" or "teacher username containing", there seems to be some extraneous text that shows up to the right of the dropdown.
* On the query builder page, the bottom page border in the fruitsalad theme is cut off. I think this is caused by unclosed HTML tags.

Additional thoughts (not regressions from the existing class flag search, so it might be better to merge this pull request first and iterate on these afterward):
* If you want to just modify your query, you can click the back button, but it's not ideal; I feel like I would want the query builder UI to be duplicated at the top of the search results page so that it's easy to just search again.
* If this is supposed to be a general class search page, the order of the items in the dropdown is unintuitive to me, specifically that "the flag" is near the top and "title contains" is near the bottom.
* Also, having "any of"/"all of" be the first items in the dropdown adds cognitive overhead to parsing what the items in the dropdown are supposed to be.
* This interface is a bit heavyweight just for a simple title search; I wonder if there is a way to shortcut simpler cases like that.

I'll look at the code later.

esp/esp/utils/query_builder.py
```diff
@@ -0,0 +1,333 @@
...
+
+        query-builder.jsx will need to be included in the page separately.
+        """
+        context = {
+            # use a uid in case we ever want to have multiple QBs on the same
+            # page.
+            'uid': random.randrange(0, 1 << 30),
+            'json_spec': json.dumps(self.spec()),
+        }
+        return render_to_string("utils/query_builder.html", context)
```
luac: Why `render_to_string` rather than template include? (I don't actually know what is more Django-idiomatic here)

esp/esp/utils/query_builder.py
```diff
@@ -0,0 +1,333 @@
...
+
+    def as_english(self, value):
+        if value in self.options:
+            return "with %s '%s'" % (self.english_name, self.options[value])
+        else:
+            raise ESPError('Invalid choice %s for input %s'
+                           % (value, self.field_name))
+
+
+class TrivialInput(object):
```
luac: "Trivial" seems like a misnomer, maybe "Constant" or "Fixed"?

luac:
Looked over it, looks basically fine to me.

Some of the code seems a bit awkward in order to deal with BooleanOp being a special built-in filter type; any reason why this isn't a class? (I see one problem with that which is that in order to implement it correctly I think you need to have an `as_queryset` method on the filter level, not just `as_q`, but it seems like you can make that change without losing anything and this also lets you make the `as_queryset` method on `QueryBuilder` smaller.)

esp/esp/utils/query_builder.py
benjaminjkraft: I think a template include might be more idiomatic, but I don't like it, because it requires more care in passing pieces around.  It might also be possible to do it as an inclusion tag of some sort, which would be nice in that it could mark the output as safe, so you wouldn't have to do `{{QB.render|safe}}`.  I can take a look at doing that.

esp/esp/utils/query_builder.py
benjaminjkraft: Good idea, `ConstantInput` sounds better.

benjaminjkraft:
> * The "not" button is still confusing; when it's not selected, the fact that the word is there makes it read as if you're searching for the negation of your query when you're not.

Yeah, this is why I did things the way I did in the original one, where the dropdown had a separate option for each thing negated.  I could go back to that, but I don't really like it either.  Any other ideas?

> Under "any of"/"all of", the + and - buttons look the same visually as the "not" button, yet they do structurally different things, which is confusing. Somewhat similar for the optional inputs under "the flag"/"any flag".

Hmm, good point.  I'll see if I can figure out a way to make it more clear.

> When you search for "title containing" or "teacher username containing", there seems to be some extraneous text that shows up to the right of the dropdown.

Whoops, will fix.

> On the query builder page, the bottom page border in the fruitsalad theme is cut off. I think this is caused by unclosed HTML tags.

I'll take a look.

> If you want to just modify your query, you can click the back button, but it's not ideal; I feel like I would want the query builder UI to be duplicated at the top of the search results page so that it's easy to just search again.

Yeah, I'd like to fix this, and I think it's very feasible, but will require some code, so probably not in this PR.

> If this is supposed to be a general class search page, the order of the items in the dropdown is unintuitive to me, specifically that "the flag" is near the top and "title contains" is near the bottom.

Any suggested sort order?

> This interface is a bit heavyweight just for a simple title search; I wonder if there is a way to shortcut simpler cases like that.

An idea I had was to have an alternate simple search, which would do the AND of at most one of each filter, and would just be an HTML form like the current user search.  Again, it will take some code, but I think it's feasible for the future.

esp/esp/utils/query_builder.py
luac: In terms of passing context around, it looks like you can do `{% include template.html with foo=bar only %}` to avoid polluting namespaces. There is also a thing where you can use `[% include %}` with custom objects that define a `render` method, but unfortunately looks like that's Django 1.7 only.

luac:
> Yeah, this is why I did things the way I did in the original one, where the dropdown had a separate option for each thing negated. I could go back to that, but I don't really like it either. Any other ideas?

So one idea is to make "not..." an option in the dropdown that causes a second nested dropdown to appear, similar to a one-way and/or. I just thought of a disadvantage of that though which is that if you've already selected a filter, if you then want to invert it, it's inconvenient because you have to kill your filter and start over in order to select "not...". (A similar problem happens with the other boolean ops right now, where if you have a query with one clause and you want to add an extra clause to it, you have to start over and select "all of...".)

esp/esp/utils/query_builder.py
benjaminjkraft: It's still annoying IMO, because you have to include the right template with the right variable name.  (I don't really care about namespace pollution since the template is so simple.)  But it's true that the way I did it is kind of un-idiomatic so maybe I'll do this anyway, if a template tag thing doesn't work.

luac: That's true. I'm fine with keeping it the way it is.

benjaminjkraft: Doing this with an (inclusion) tag turns out to be a piece of cake, so I did that.

benjaminjkraft:
> Some of the code seems a bit awkward in order to deal with BooleanOp being a special built-in filter type; any reason why this isn't a class?

I need to think this through a bit more, but I think it probably makes sense to do.

benjaminjkraft:
Ok, I fixed most of these, but I still want to think more about a few.
[x] Replace `.render()` method with something more idiomatic
[x] Rename `TrivialInput`
[ ] Make `BooleanOp` a filter type
[ ] Make `+` and `-` buttons more visually different
[x] Remove extraneous text from `TextInput`s
[x] Fix unclosed tag
[ ] Find a better sort order

Things for a future PR:
* Put your query at the top of the page
* Simple search option

benjaminjkraft:
Another idea for the "not": have a dropdown with options "with" (default) and "without", right before the big dropdown.  (I would also remove the "with" from the text at the top.)

benjaminjkraft:
I don't think making `BooleanOp` a filter type makes sense, actually -- I think it's going to add more complexity than it's worth.  The basic reason is that `BooleanOp` exists at a parallel point in the hierarchy to `Filter`, which is currently a single class that gets instantiated repeatedly.  `Filter` can't really call out to `BooleanOp`, or at least it would have to bend over backwards to do so, so we would now have to have two points where we dynamically pick what class to use.  Since they implement different APIs, this would be awkward Python-side, and since we have to find and instantiate the classes manually, this would be awkward in React.  So I don't think it would actually simplify the code; I think it save some boilerplate in React but add some more, and would just add boilerplate in Python.

benjaminjkraft:
For the sort order I did something arbitrary but maybe more sensible:
screenshot from 2015-03-08 14 00 18

benjaminjkraft:
Ok, pushed the remaining changes/fixes; I think all of the comments that aren't for a future PR are addressed.  I can put those into issues.

benjaminjkraft:
And fixed the tests.

luac:
Cool. A couple of tiny additional comments if you feel like fixing them:
* It would be a convenience if the "Go" button was type="submit", I think that makes it so that you can go by hitting enter anywhere inside the form.
* Some of the english text is still weird, e.g. "You asked for classes none of(with the status 'accepted') and (without the flag with type 'Reviewed by admin 2' and before 03/09/2015 00:00)"

benjaminjkraft:
I'll take a look at the button.  I think for the English text as long as it's vaguely comprehensible, I'm not going to worry about it since it will be moot when I get around to #1620.

benjaminjkraft:
I tried changing the button, but it doesn't actually work because there's no HTML `<form>` (because the POST data has to get constructed from javascript anyway, so I didn't bother).  At some point I can take a look at adding it as a keyboard shortcut in JS (already #1367) but probably not worth blocking this PR on it.

luac:
Cool. Will merge once tests pass.

PR #1526, opened 2015-01-01
Some improvements to ResourceType handling

mgersh:
My main goal here was to make teacherreg stop get_or_creating default resource types, since the program creation form already handles those and I think it's more reasonable to be able to delete one without having to set a different default resource types tag for the specific program. I also made some improvements to things that confused me while I was working on it; see the commit messages.

Note: This can potentially cause the resource types form in teacherreg to get reordered from how they were before. If you liked the order before and don't like it after merging this, you can mess with priority_default to fix it. Don't use 0, though, it's a special case somewhere in the code and I haven't bothered to figure out why.

### DIFF
--- a/esp/esp/program/models/__init__.py
+++ b/esp/esp/program/models/__init__.py
@@ -784,6 +784,7 @@ class Program(models.Model, CustomFormsLinkModel):
         else:
             return None
 
+    @cache_function
     def getResourceTypes(self, include_classroom=False, include_global=None):
         #   Show all resources pertaining to the program that aren't these two hidden ones.
         from esp.resources.models import ResourceType
@@ -801,7 +802,9 @@ class Program(models.Model, CustomFormsLinkModel):
         else:
             Q_filters = Q(program=self)
         
-        return ResourceType.objects.filter(Q_filters).exclude(id__in=[t.id for t in exclude_types])
+        return ResourceType.objects.filter(Q_filters).exclude(id__in=[t.id for t in exclude_types]).order_by('priority_default')
+    getResourceTypes.depend_on_model(lambda: ResourceType)
+    getResourceTypes.depend_on_model(lambda: Tag)
 
     def getResources(self):
         from esp.resources.models import Resource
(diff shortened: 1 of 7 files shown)


jmoldow:
Looks good to me, though I'd prefer that someone else also give this a look over before merging.

Some questions:

- In some places you added ``order_by('priority_default')``, and in some places you removed it. Why?

- I don't know what you're referring to with "Unfortunately, I didn't get rid of them entirely because of a weird open classes hack." Can you explain here, and possibly add a comment somewhere in the code if there isn't already?

esp/esp/program/models/__init__.py
```diff
@@ -1925,6 +1928,7 @@ def install():
     print "Installing esp.program initial data..."
     install_class()
 
-# The following are only so that we can refer to them in caching Program.getModules.
+# The following are only so that we can refer to them in caching
 from esp.program.modules.base import ProgramModuleObj
 from esp.program.modules.module_ext import ClassRegModuleInfo, StudentClassRegModuleInfo
+from esp.resources.models import ResourceType
```
benjaminjkraft: I'm assuming you tried adding the import at the top and got circular import errors?

benjaminjkraft:
Looks good to me.  We will probably want to delete the non-static request system at some point, but maybe that can happen as a part of the resources overhaul if it ever happens.

esp/esp/program/models/__init__.py
mgersh: Yes.

mgersh:
The adding/removing of order_by(priority_default) was moving it to inside of a cached function. The most recent commit slightly changes how that's handled in teacherreg.

esp/esp/program/modules/handlers/resourcemodule.py
```diff
@@ -371,7 +371,7 @@ def resources(self, request, tl, one, two, module, extra, prog):
         if 'timeslot_form' not in context:
             context['timeslot_form'] = TimeslotForm()
         
-        context['resource_types'] = self.program.getResourceTypes().exclude(priority_default=0).order_by('priority_default')
+        context['resource_types'] = self.program.getResourceTypes().exclude(priority_default=0)
```
benjaminjkraft: I think the extra `.exclude()` may prevent this from actually using the cache, but I think that's okay in this case, since it's just a single query in an admin-facing view.

mgersh: Oh, whoops, I meant to get rid of the extra .exclude() in that last commit. It's not actually doing anything useful.

Issue #1098, opened 2014-03-01
Onsite grid can't register new students with existing accounts


jmoldow:
The user search in the onsite grid seems to only look for students with existing profiles for the program. This means that a new student, with an existing account, must be registered through the old interface (or at least must fill out the profile in the old interface before switching to the grid interface).

PR #1527, opened 2015-01-04
Onsite grid: support moving new students into program

pricem:
Joel did this work a while ago, and I only recently got to reviewing and pushing it.  This is for ticket #1098, which turned out to be a bit more complicated than it seemed.  

The main effect is that the autocomplete box on the class changes grid will show you a list of students registered for the current program (as before), but after that it will show you students who have accounts but are not associated with the current program.  These students are highlighted in red.  When you click one of those students, you'll get a dialog box asking you to check that the student has actually signed up for the program in person (i.e. onsite new registration) and has paid and filed their forms.  Then it will attempt to copy the student's most recent profile into the current program.  If this fails, it tells you to use the onsite new registration feature (old-style interface).  Otherwise, you can go on scheduling classes as usual.

To do this, we had to make the autocomplete list fetch from a JSON view, instead of having the page preloaded with the list of students in the current program.  The JSON view provides the name and ID along with a flag showing whether they have a profile for the current program.

Other tweaks include:
- Fix a z-index bug in the onsite grid (list of students would show up "underneath" the grid)
- Some cleanup of the onsite new registration code


benjaminjkraft:
How will this interact with #1659?

luac:
I tried merging this in with #1659, and surprisingly there were no merge conflicts in the JS, but I think there might be significant functionality conflicts. I'll take a look.

jerrywu64:
Is this still relevant?

luac:
Oh, shoot. I guess I totally forgot to do this.

luac:
I don't have a dev server nowadays so I think it's slightly less convenient for me to do this, also #1659's been merged now so it's less inconvenient for others to do it too. I'll still do it if no one else wants to, but I'm also willing to relinquish this ticket.

luac:
Okay so actually it seems like all the javascript works and didn't conflict logically with #1659 at all! It's not that surprising because it's really not that much and fairly localized changes. The actual server-side logic has rotted a bit, though. Whoops.

UI comments:

The autocomplete seems to be much laggier now that it's requesting from the server. Not sure how much that is because my dev server is slower than prod, though. It seems like it might be better to request the entire set of students once, and search that client-side; or if that's too big, maybe there's some way to search among registered students client-side first, and only request from the server if no registered students match the search term.

Also, the search criterion feels unintuitive, because the results are displayed as "Lastname, Firstname", but the autocomplete view expects it to be "Firstname Lastname".

luac:
Ok, I think this is in a working state right now.

My only qualm is again the server-side autocomplete; I don't like to get in the way of merging this long-overdue pull request, but I think there is potentially a usability regression here. Thoughts?

benjaminjkraft:
I like the approach of requesting from the server only if there's no match.  I think I'd delay merging for that, since probably looking up students in autocomplete is sufficently more frequent than registering new students that the UX for the former is more important, but I could certainly be convinced otherwise -- small chapters may prefer the other way.

luac:
Ok, that approach seems implementable enough. I'll probably do it this weekend.

luac:
Alright, pushed a change that does this. I also cleaned up the server autocomplete a bit and made it more consistent with the client autocomplete wrt sorting and stuff.

I think this is ready to merge, I'll give it a day if someone wants to review my changes?

benjaminjkraft:
I didn't review super closely but the recent changes look fine to me.  Thanks for finishing this up!

Issue #1288, opened 2014-08-24
Update Apache conf template


jmoldow:
When setting up a new production site, the server setup script will use a template to create an Apache conf for the site.

This template has become out-of-date with what we actually use with the live sites. We should look through the apache ```*.conf``` file for the current chapter sites, and make necessary updates.

Two examples of things I know we have to do:
- Make the ```VirtualHost```s listen on port ```81```.
- Add the redirect from http to https.

We could also consider moving the template to its own file, outside of the server setup bash script.

btidor:
This seems to have been fixed by #1529, unless it fell out of date yet again.

benjaminjkraft:
It fell out of date again, and is updated again in #1963.

Issue #1287, opened 2014-08-24
Use local_settings.py.default in dev server setup.


benjaminjkraft:
@luac and I were talking about this: right now, the dev server setup script asks for a bunch of configuration options, and then generates a `local_settings.py` from a hardcoded template and those options.  We should definitely be using the `local_settings.py.default` in some way, since we generally forget to update the dev server setup script.  Probably the easiest thing is just to not bother doing the whole configuration thing, and just have the script copy `local_settings.py.default` to `local_settings.py`, and tell the user to go update it, and it can have some comments telling them what the variables that they're filling in mean, and useful defaults for the ones that most people don't care to change.

pricem:
From discussion on #1529: I think the current goal is to centralize all of the settings.py components in one file which would be used by different setup mechanisms.  This would require handling some differences in requirements between the production setup (new_site.sh), the development setup (Vagrant), and the Travis setup, and any other variants I'm forgetting.

benjaminjkraft:
This specific issue is fixed and/or made moot by #1529.  I'll open a new one to more generally merge the various `local_settings` templates we still have.

Issue #1286, opened 2014-08-24
Update local_settings.py.default


jmoldow:
When we manually ```update local_settings.py```, we sometimes forget to update ```local_settings.py.default```, so that future site installs also get to take advantage of the new setting. We should look through the ```local_settings.py``` files of the current chapter sites and dev servers, and see if there is anything worth adding.

One example of something I know we have to do: add the ```SECURE_PROXY_SSL_HEADER``` setting, so that ```request.is_secure()``` works on newer sites. Though, in this particular case, it's also arguable that we could maybe move that to ```django_settings.py```. In either case, this setting needs to be added to the repository.

Requires #1287. Or, without #1287, the template in the server setup script should also be updated.

benjaminjkraft:
Fixed and/or made moot by #1529.

Issue #1285, opened 2014-08-24
Production server setup needs to handle dbmail


jmoldow:
When setting up a production site, the ```dbmail_cron.py``` script needs to be added to ```/etc/crontab```, and the site needs to be added to the ```exim4``` configuration. This is not currently handled by the setup script, and needs to be done manually. When this manual step is forgotten, the new site is misconfigured, and cannot send or receive mail. This should be done automatically by the server setup script, when installing a production environment.

Related to #965, #1142, and #1284.

benjaminjkraft:
Fixed by #1529 and #1963.

PR #1529, opened 2015-01-04
Server setup cleanup/updates

pricem:
This pull request does 2 things:
* Brings new_site.sh up to date with what we have on the LU server (and also makes it work on a "fresh" server, like one created using our Vagrant configuration)
* Removes the dev server setup scripts.  See comments for b019843.

It does not add any functionality to the new_site.sh script (e.g. exim4 setup as discussed in #1285), nor does it add production site capabilities to our Fabric tasks, but both of those would be logical next steps.

Relevant tickets (i.e. those which we could potentially close if these changes are accepted) include #1039, #1040, #1284, #1286, #1287, #1288.


esp/useful_scripts/server_setup/new_site.sh
```diff
@@ -312,20 +315,13 @@ DATABASE_NAME = '$DBNAME'
...
-email_choices = (
-    ('general', 'General Inquiries'),
-    ('web',     'Web Site Problems'),
-    )
-# Corresponding email addresses                                                                                                                                 
-email_addresses = {
-    'general': '$GROUPEMAIL',
-    'web':     '$SITENAME-websupport@learningu.org',
-    }
+SECRET_KEY = '`${CURDIR}/random_password.sh`'
```
benjaminjkraft: Is there a reason we're moving towards having the default configs inline rather than in a separate file?  Now might be a nice time to move away from that, which I think we've discussed doing.

esp/useful_scripts/server_setup/new_site.sh
```diff
@@ -365,28 +371,13 @@ then
...
-fi
-
 # Apache setup
 # To reset: remove appropriate section from Apache config
 if [[ "$MODE_APACHE" || "$MODE_ALL" ]]
 then
     cat >>$APACHE_CONF_FILE <<EOF
 #   $INSTITUTION $GROUPNAME (automatically generated)
-WSGIDaemonProcess $SITENAME processes=1 threads=1 maximum-requests=1000
+WSGIDaemonProcess $SITENAME processes=2 threads=1 maximum-requests=1000
```
benjaminjkraft: We may want to use 4 or 8 procs as default, I know some sites have had issues because 2 isn't a lot if one gets blocked for a while on, say, generating schedules.

esp/useful_scripts/server_setup/new_site.sh
```diff
@@ -397,19 +388,12 @@ WSGIDaemonProcess $SITENAME processes=1 threads=1 maximum-requests=1000
     #   Caching - should use Squid if performance is really important
     # CacheEnable disk /
 
+    #   Redirect HTTP requests to HTTPS for security - uncomment to use
+    # Include /etc/apache2/sites-available/esp_sites/https_redirect.conf
```
benjaminjkraft: Any reason we're not defaulting to this?

esp/useful_scripts/server_setup/random_password.sh
```diff
@@ -0,0 +1,48 @@
...
+# ==> Password will consist of alphanumeric characters.
+LENGTH="8"
+# ==> May change 'LENGTH' for longer password.
+
+
+while [ "${n:=1}" -le "$LENGTH" ]
+# ==> Recall that := is "default substitution" operator.
+# ==> So, if 'n' has not been initialized, set it to 1.
+do
+	PASS="$PASS${MATRIX:$(($RANDOM%${#MATRIX})):1}"
```
benjaminjkraft: Bash `$RANDOM` is not secure (cf. http://stackoverflow.com/a/8988902/904174).  Is there a reason not to just do `openssl rand -base64 6` or similar?

benjaminjkraft: Also, I think Django's `SECRET_KEY` can and should be much longer than 8 characters -- iirc CSRF and other protections depend on it, and 62^8 is not huge.

esp/useful_scripts/server_setup/new_site.sh
pricem: Ideally, we would have one (separate) file containing a template for settings.py.  The problem is that the different uses of settings.py files have different requirements: I think the Fabric setup uses a Python string template, whereas the new_site.sh script is using Bash substitution, and the Travis one is hardcoded.  I can try to reconcile these if you think that's in scope for this pull request.

benjaminjkraft: I didn't realize it was going to be significant additional work, so that's fine -- we can just leave that issue open.  I do think it would be a good idea to move to a single template at some point, but no reason to hold up this PR if it's a significant additional thing.

esp/useful_scripts/server_setup/new_site.sh
pricem: I changed the default to 4.  If anyone has intuition saying the overhead of extra processes isn't a problem (keeping in mind that this number is multiplied by the number of sites on the server), it could definitely be increased.

esp/useful_scripts/server_setup/new_site.sh
pricem: I had it commented out since I was afraid to make assumptions about the server configuration (e.g. that https_redirect.conf exists in the right place), but it looks like we make plenty of those assumptions elsewhere.  I think the right solution is to just put the code inline, since it's only 4 lines.

esp/useful_scripts/server_setup/random_password.sh
pricem: Good call; I removed the script and used the command you suggested.

esp/useful_scripts/server_setup/new_site.sh
pricem: Sounds good; I left a note so this could be considered as part of #1287.

esp/useful_scripts/server_setup/new_site.sh
benjaminjkraft: There's probably a way to make sure the processes get killed if they're idle for too long.  I'm not entirely sure whether the `inactivity-timeout` option to `mod_wsgi`'s `WSGIDaemonProcess` is exactly what we want -- it's unclear to me if that is for the whole process group or for individual processes -- but it should work.  We should probably have an Apache global `MaxClients` setting anyway, so AIUI this would only be necessary to prevent one site from eating all the processes and choking another.

pricem: My interpretation of the documentation and Internet commentary is that the number of processes you specify using ``processes`` is a constant, although the number of threads is allowed to vary up to the limit you specify. (It looks like ``inactivity-timeout`` is for restarting processes, not shutting them down.)  I can't tell what the MaxClients setting would do besides starving the WSGI processes of requests, given that we have 1 thread per process (which I'm guessing is still required by our code... though it would be interesting to test).  So I don't think we can do things much differently than the current configuration.  

Let's just allocate 4+ processes to everyone and put in more memory when we get the chance.  Right now we have a total of 126 WSGI processes allocated, and ps_mem.py reports that Apache2 has a total of 135 processes using 5.6 GiB of memory.

benjaminjkraft: Ok, that sounds reasonable.  It may be that Apache's WSGI module just doesn't quite do what we want, unfortunately, but that sounds like a good thing to do for now.

esp/useful_scripts/server_setup/new_site.sh
```diff
@@ -269,13 +271,14 @@ CACHE_PREFIX = "${SITENAME}ESP"
...
         'bounces': 'learningubounces@gmail.com',
-        'support': '$SITENAME-websupport@learningu.org',
+        'support': '$GROUPEMAIL',
         'membership': '$GROUPEMAIL',
         'default': '$GROUPEMAIL',
         }
 ORGANIZATION_SHORT_NAME = '$GROUPNAME'
 INSTITUTION_NAME = '$INSTITUTION'
 EMAIL_HOST = '$EMAILHOST'
+EMAIL_HOST_SENDER = EMAIL_HOST
```
benjaminjkraft: I'm guessing this should be `EMAIL_HOST_SENDER = $EMAIL_HOST`

esp/useful_scripts/server_setup/new_site.sh
```diff
@@ -397,19 +389,15 @@ WSGIDaemonProcess $SITENAME processes=1 threads=1 maximum-requests=1000
...
 
+    #   Redirect HTTP requests to HTTPS for security
+    RewriteEngine On
+    RewriteCond %{HTTP:X-Forwarded-Proto} !^https
+    RewriteCond %{REQUEST_URI} !^/media
+    RewriteRule ^(.*)$ https://%{SERVER_NAME}%{REQUEST_URI} [R,L]
+
     #   Static files
     Alias /media $BASEDIR/esp/public/media
-    <Location /media>
```
benjaminjkraft: Are these options no longer necessary?

benjaminjkraft:
Generally looks good to me modulo the comments I made; I don't have an easy way to test it, and it can probably just get more testing as we use it.

benjaminjkraft:
Just leaving a note here that after #955 gets merged there will be another place to change the random password generation method (namely the setting of `SECRET_KEY`).

esp/useful_scripts/server_setup/new_site.sh
btidor: I don't think so: `EMAIL_HOST` is defined in the line above.

benjaminjkraft: Oh, I see, I got confused because this is actually python, but with bash variable interpolation.  You're right.

PR #1553, opened 2015-01-06
Remove meta author tag from fruitsalad theme

benjaminjkraft:
At the request of Tony Valderrama.  It shows up in search engines and stuff as
the author, but in fact they just wrote the theme.  Since determining the
actual "author" of a page could be hard, let's just remove it entirely.


btidor:
++ and LGTM

PR #1522, opened 2014-12-24
Query optimizations

benjaminjkraft:
Thanks to #1516, I got nerd sniped a bit and did some query-optimization.  Should hopefully speed up the dashboard a bit.  Depends on #1502, mostly just to avoid merge conflicts.

PR #1554, opened 2015-01-07
Cache loading

luac:
This is a change to how caches are initialized, as a step toward removing the `cache_loader` app.

The practical effect of this is a slight change in the interface; rather than
```
func.depend_on_row(lambda: Model, ...)
```
you now write
```
func.depend_on_row('app.Model', ...)
```

The old way still works but raises a DeprecationWarning. <strike>For now, all of the code still uses the old way. The branch `cache-loading-test` updates the existing code to the new way, but this conflicts with outstanding pull requests (#1502, #1522, #1524, #1526), so I was hoping that those would be merged first. But I can include it in this pull request if people think that would be better.</strike> (This is now included.)

The technical detail is that, rather than relying on the cache_loader app to force-import all the other apps so that it loads last, we use Django's `get_model` to resolve the model name in a call to `depend_on_row`, and if it fails (because the model isn't loaded yet), attach a handler to the `class_prepared` signal that processes it once it is loaded. This is the same technique that Django itself uses to resolve e.g. `ForeignKey('app.Model')`.


luac:
I can (and should, now that you mention it) update the existing cache tests to use the new syntax/mechanism.

Other than that, it's hard because the cache loading step happens at app initialization, which when you run `manage.py test` is already done once the test cases start. It is the case currently that if you mistype a cache dependency or something like that, it blows up with a failed assertion immediately; but I can put in a test that runs through the same assertions a second time in case the assertions in the current code path get misplaced or changed.

benjaminjkraft:
Ok, that sounds good.

luac:
Added.

benjaminjkraft:
Ok, LGTM.

Issue #1465, opened 2014-11-08
Class changes grid shows walk-in classes


jmoldow:
I can't remember if this was always the case, and I don't know if this is a feature or a bug. I'm guessing "it wasn't" and "bug", but I don't know.

The grid also lets you register for walk-in classes, which is definitely a bug.

PR #1558, opened 2015-01-12
Ticket 1465

joeleg:


esp/public/media/scripts/onsite/ajax_status.js
```diff
@@ -772,13 +795,9 @@ function populate_classes()
...
         var new_category = data.catalog.categories[i];
-        /*
-        //  Skip "open class" category
-        if (new_category.id == open_class_category.id)
-            continue;
-        */
+
         data.categories[new_category.id] = new_category;
-        if (settings.categories_to_display.indexOf(new_category.id) == -1)
+        if (settings.is_initial && (settings.categories_to_display.indexOf(new_category.id) == -1))
```
pricem: I added this check to make the grid consistent with the settings when the data was refreshed (99e15d7 prevented the refresh from clobbering your settings, but it would still show all categories in the table regardless of what those settings were).

esp/public/media/scripts/onsite/ajax_status.js
```diff
@@ -586,6 +588,10 @@ function render_table(display_mode, student_id)
             var section = data.sections[data.timeslots[ts_id].sections[i]];
             var parent_class = data.classes[section.class_id];
             
+            //  Hide walk-in seminars from display.
+            if (parent_class.category__id == open_class_category.id)
+                continue;
+
```
pricem: Following 30a888f, I added this to exclude the walk-in classes from the table as well as from the category options... which I think was the intent of #1465, but I'm open to correction. (There is an argument for keeping walk-in classes in the table so onsite volunteers can easily inform students of their choices; however, that would be an argument for showing walk-in activities among the category checkboxes also.)

esp/public/media/scripts/onsite/ajax_status.js
```diff
@@ -17,7 +17,9 @@ var settings = {
     override_full: false,
     disable_grade_filter: false,
     compact_classes: true,
-    categories_to_display: []
+    categories_to_display: [],
+    selected_categories: [],
+    is_initial: true,
```
pricem: I'm a bit rusty with Javascript syntax; is this trailing comma going to be a problem for any browsers that people actually use?  (Wasn't a problem for me.)

benjaminjkraft: This [SO post](http://stackoverflow.com/questions/7246618/trailing-commas-in-javascript) and this [compatibility table](https://rauschma.github.io/js-feature-matrix/) suggest that it is an issue for IE≤8, but not for modern browsers.  Unfortunately I have a tendency to do this regularly without thinking, so there may be other things on the site that are broken too :-( .

luac:
Is this ready to merge? It would be good to have these fixes in for MIT Spark.

esp/public/media/scripts/onsite/ajax_status.js
luac: Since it's the onsite grid, I don't think IE≤8 is a big deal.

esp/public/media/scripts/onsite/ajax_status.js
```diff
@@ -725,8 +731,23 @@ function toggle_categories() {
...
+    var checkedCategories = [];
+    if(settings.is_initial) {
+        checkedCategories = settings.categories_to_display.slice();
+        settings.is_initial = false;
+    } else {
+        $j('#category_list input:checked').each(function() {
+            checkedCategories.push(extract_category_id($j(this).attr('id')));
+        });
+    }
+
```
luac: I don't quite understand what the intent of all this code is -- why are we clearing the `#category_list` and re-rendering it again at all?

luac:
Okay, I looked over this and pushed some adjustments -- it now shows a checkbox for toggling walk-ins, but it's unchecked by default so that the walk-ins don't show. As well, the "show/hide all" buttons ignore the walk-ins checkbox, so that you have to specifically click the walk-ins checkbox in order to show them; I wasn't sure if this would be the best behavior but it made the most sense to me.

esp/public/media/scripts/onsite/ajax_status.js
```diff
@@ -243,6 +243,10 @@ function update_checkboxes()
             {
 		      studentcheckbox.attr("disabled", "disabled");
             } 
+            else if (data.classes[section.class_id].category__id = open_class_category.id)
```
jmoldow: Should this be equality ``==`` instead of assignment ``=``?

luac: Ah yes, it should. (That's weird, I thought I tested this bit and it worked as intended.)

PR #1561, opened 2015-01-14
Ticket 1448

joeleg:


esp/esp/program/tests.py
```diff
@@ -121,9 +175,12 @@ def testUserSearchFn(self):
         self.assertStringContains(response['location'], "/manage/userview?username=testuser123543")
 
         # Full name, unique
-        response = c.get("/manage/usersearch", { "userstr": "Admin User" })
```
luac: I don't think we should take out this test case; otherwise, the tests never hit the code path for full name searches.

esp/esp/program/views.py
```diff
@@ -32,12 +32,14 @@
   Phone: 617-379-0178
   Email: web-team@learningu.org
 """
+from operator import __or__ as OR
```
luac: Is this import correct? I see `operator.or_` used below, but not `OR`.

esp/esp/program/tests.py
luac: Never mind, I didn't notice that it was hit several lines below in the unchanged part.

luac:
Note this is branched off part of #1558 so that should be merged fisrt or this should be rebased.

Issue #475, opened 2012-11-14
Scheduling Checks doesn't catch teachers teaching two classes at once


benjaminjkraft:
The title says it all; this is true as far as I can tell.

benjaminjkraft:
Fixed in #1564.

PR #1564, opened 2015-01-20
Changed _all_class_sections to include walk-ins and added _all_nonwalkin...

jerrywu64:
..._sections to be the original version. Updated some functions as appropriate. Addresses #475. 

I acknowledge I have created duplicated code; I can try to get rid of that if needed. I'd also like more opinions on what functions should include walk-ins and which shouldn't.

I'm also concerned because there's a comment mentioning memoization, and I'm not sure if this refers to select_related/prefetch_related or if there's something I'm supposed to do?

### DIFF
--- a/esp/esp/program/modules/handlers/schedulingcheckmodule.py
+++ b/esp/esp/program/modules/handlers/schedulingcheckmodule.py
@@ -223,7 +223,7 @@ class SchedulingCheckRunner:
                     return self.all_sections
                else:
                     self.all_nonwalkins = list(qs)
-                    self.listed_nonwalkins = true
+                    self.listed_nonwalkins = True
                     return self.all_nonwalkins
           
 
--- a/esp/esp/program/modules/handlers/schedulingcheckmodule.py
+++ b/esp/esp/program/modules/handlers/schedulingcheckmodule.py
@@ -124,6 +124,7 @@ class SchedulingCheckRunner:
 
           #things that we'll calculate lazilly
           self.listed_sections = False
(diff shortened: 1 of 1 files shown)


esp/esp/program/modules/handlers/schedulingcheckmodule.py
```diff
@@ -198,12 +198,29 @@ def _timeslot_dict(self, slot=lambda: 0):
...
           else:
                qs = self.p.sections()
+               #filter out non-approved classes
+               qs = qs.exclude(status__lte=0)
+               qs = qs.exclude(resourceassignment__isnull=True)
+               #filter out lunch
+               qs = qs.exclude(parent_class__category__category=u'Lunch')
+               qs = qs.select_related('parent_class', 'parent_class__parent_program', 'parent_class__category')
+               qs = qs.prefetch_related('meeting_times', 'resourceassignment_set', 'resourceassignment_set__resource', 'parent_class__teachers')
+               self.all_sections = list(qs)
```
luac: I think the "memoize" comment refers to this; when `_all_class_sections` is called, the result is stored in `self.all_sections` so that it doesn't need to be recomputed next time. But this doesn't work if the two functions both use the same variable.

luac:
Re: duplicated code, it seems like it should be easy enough to get rid of it by changing it into a single function with an option `include_walkins=True` or something. You'll need to do something else about the memoize part though (see the line note).

benjaminjkraft:
LGTM, merging.  (We discussed which checks should use this in person.)

Issue #1266, opened 2014-07-13
Grid-Based Onsite pulls resources from a strange place....


shulinye:
![screenshot from 2014-07-13 10 23 01](https://cloud.githubusercontent.com/assets/2553939/3564789/55dccb50-0a99-11e4-9b41-c4e862c8024c.png)

willgearty:
Fixed by #1565

PR #1565, opened 2015-01-21
Restricted resource pull on rooms by the classchange grid to actually on...

jerrywu64:
...ly be classrooms.

Addresses #1266.


benjaminjkraft:
Are you sure this view is only used by this page?  A lot of the cases where pages pull more data than they need it's because they share views.

esp/esp/program/modules/handlers/onsiteclasslist.py
```diff
@@ -149,8 +149,9 @@ def counts_status(self, request, tl, one, two, module, extra, prog):
     @needs_onsite
     def rooms_status(self, request, tl, one, two, module, extra, prog):
         resp = HttpResponse(mimetype='application/json')
-        data = ClassSection.objects.filter(status__gt=0, parent_class__status__gt=0, parent_class__parent_program=prog).select_related('resourceassignment__resource__name').values_list('id', 'resourceassignment__resource__name', 'resourceassignment__resource__num_students')
-        simplejson.dump(list(data), resp)
+        data = ClassSection.objects.filter(status__gt=0, parent_class__status__gt=0, parent_class__parent_program=prog).select_related('resourceassignment__resource__name').values_list('id', 'resourceassignment__resource__name', 'resourceassignment__resource__num_students', 'resourceassignment__resource__res_type__name')
+        data = [d for d in data if d[3] == 'Classroom']
```
benjaminjkraft: Why do this in python rather than in the DB?  It would probably be faster to do in the DB.

mgersh:
The view in question is actually in the onsite class list module, not the json data module, so it's probably safe to change. Git grep appears to confirm this.

jerrywu64:
I don't see why this function, if it's actually relevant to the status of rooms as advertised, would want to pull all resources anyway.

benjaminjkraft:
Ok, sounds fine.

esp/esp/program/modules/handlers/onsiteclasslist.py
jerrywu64: It's not noticeably slower on my devserver (for the Splash 2014 grid) in the current state compared to not making any modifications, and I also couldn't think of a way to do the restriction in the DB (since we need to restrict the returned tuples in the list, not the section objects) without using ResourceAssignment.objects.[...] which is painfully slow for some reason (unless I just messed up and what I tried outright doesn't work). If you have an alternate solution, I can try it out, but I don't think it'll make a big difference anyway.

benjaminjkraft: Does adding `resourceassignment__resource__res_type__name='Classroom'` to the filter not just work?  That will filter the returned rows in the query.  It might not be slow as-is, but inevitably these things get made more complicated over time, and in general it will be at least as fast in the DB, so I think it's just nicer to keep everything in the DB when possible

jerrywu64: Wow, I really need to actually try these things in the future instead of assuming they don't work. I stand corrected; I'll change that by tonight.

esp/esp/program/modules/handlers/onsiteclasslist.py
```diff
@@ -149,7 +149,7 @@ def counts_status(self, request, tl, one, two, module, extra, prog):
     @needs_onsite
     def rooms_status(self, request, tl, one, two, module, extra, prog):
         resp = HttpResponse(mimetype='application/json')
-        data = ClassSection.objects.filter(status__gt=0, parent_class__status__gt=0, parent_class__parent_program=prog).select_related('resourceassignment__resource__name').values_list('id', 'resourceassignment__resource__name', 'resourceassignment__resource__num_students')
+        data = ClassSection.objects.filter(status__gt=0, parent_class__status__gt=0, parent_class__parent_program=prog, resourceassignment__resource__res_type__name="Classroom").select_related('resourceassignment__resource__name').values_list('id', 'resourceassignment__resource__name', 'resourceassignment__resource__num_students', 'resourceassignment__resource__res_type__name')
```
benjaminjkraft: I think you can now remove the addition to the `.values_list()`, since the client doesn't need it.

PR #1574, opened 2015-01-30
Remove non-static resource requests

benjaminjkraft:
This was an alternate system that was added (edit: maybe it was the original system and the other one was added?), but most chapters that use resources don't use it, and I think the ones that do don't actually care, they're just using it because it's the default.  In addition, the non-static system appears broken.  I'd like to remove it, because it will be easier to do the resources schema changes if we don't have to maintain two systems.  As far as I can tell, this doesn't need any migration, since it's stored in the same models, it will just display things differently.


benjaminjkraft:
Nuking this to rename the branch.

Issue #1576, opened 2015-02-03
Catalog sort is weird if there are two categories with the same symbol


benjaminjkraft:
It should sort by symbol then the full category name, or something else like that.  Workaround is to manually set the `catalog_sort_fields` tag.  Reported by @jerrywu64.

jerrywu64:
The default sorting looks like it's in ~line 238 or so in esp/programs/models/class_.py, in particular the list definition of order_args. At MIT at least, the catalog_sort_fields tag already existed and overrides this sorting.

PR #1579, opened 2015-02-04
Added sorting by category name as well as category symbol.

jerrywu64:
Addresses #1576 at least in the default case. In particular, if the tag catalog_sort_fields is defined (which it is for MIT) this won't actually have any visible effects but whatever. 

I also haven't actually tested this change directly since I don't see an obvious way to do so. However, based on changing the catalog_sort_fields tag, this should fix the issue assuming it doesn't break anything.

### DIFF
--- a/esp/esp/program/models/class_.py
+++ b/esp/esp/program/models/class_.py
@@ -235,7 +235,7 @@ class ClassManager(Manager):
         if order_args_override:
             order_args = order_args_override
         else:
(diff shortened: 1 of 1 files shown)


benjaminjkraft:
Can't you test it by just deleting the tag on your dev server?  The change looks good but I don't see why you can't test it.

luac:
Tested on my dev server, looks fine.

Issue #1164, opened 2014-03-21
Line Item Types for SplashInfoModule


jmoldow:
The lunch selection module does not directly use line item types. `SplashInfo.save()` pulls the lunch costs from a `Tag` (splashinfo_costs), and then creates `Transfer` objects with those costs. Rather than a `Tag`, the costs should come from the associated `LineItemTypes`.

The current accounting system will work here if all lunch options are the same price. To support variable cost lunch options, #1119 will have to also be addressed.

pricem:
Joel is starting to work on this.  The current plan is to do essentially what this ticket says (now that #1119 is done), and just switch SplashInfoForm (used by SplashInfoModule) over to using LineItemOptions.  That's easier said than done; we'll see if it's messy and/or requires thinking about other parts of the code.

For what it's worth, I'm pretty sure this ticket is referring to the SplashInfoModule, and not the StudentLunchSelectionModule.  The latter is for choosing which timeslot to reserve for lunch on your schedule, rather than what kind of food to get (maybe these modules should be merged?).

jmoldow:
> For what it's worth, I'm pretty sure this ticket is referring to the SplashInfoModule, and not the StudentLunchSelectionModule.

Yes, that is what I meant. Thanks for correcting me. I will update the title of the issue.

joeleg:
It seems like there is a lot of overlapping functionality between extracosts and the splashinfomodule and I wonder if we should just repurpose/extend the extracosts module.

PR #1581, opened 2015-02-06
Ticket 1450

ghost:


esp/esp/program/models/__init__.py
```diff
@@ -50,19 +50,27 @@
...
 from esp.utils.expirable_model import ExpirableModel
 from esp.utils.formats import format_lazy
 from esp.qsdmedia.models import Media
 
+PROFILE_MAX_AGE_DAYS = 5
+
+class ProgramAccessError(ESPError_Log):
+        def __init__(self, msg):
+            self.msg = msg
+            super(ProfileNotFoundException, self).__init__(msg)
```
pricem: This is mismatched with the class name ProgramAccessError.  Also, I think the tab indentation is off.

esp/esp/program/models/__init__.py
```diff
@@ -1315,10 +1323,17 @@ def save(self, *args, **kwargs):
     @cache_function
     def getLastForProgram(user, program):
         """ Returns the newest RegistrationProfile attached to this user and this program (or any ancestor of this program). """
-        if user.is_anonymous():
-            regProfList = RegistrationProfile.objects.none()
-        else:
-            regProfList = (RegistrationProfile.objects
+        has_access = Permission.user_has_perm(user, '/Profile', program)
+        registration_profile = None
```
pricem: This line probable isn't necessary since it's set in both branches of the if statement below.

pricem:
I should note that this pull request also includes cleanup of SplashInfo (#1164).

esp/esp/program/models/__init__.py
```diff
@@ -50,19 +50,27 @@
...
 from esp.db.fields import AjaxForeignKey
-from esp.middleware import ESPError, AjaxError
+from esp.middleware import ESPError, AjaxError, ESPError_Log
 from esp.tagdict.models import Tag
 from esp.users.models import ContactInfo, StudentInfo, TeacherInfo, EducatorInfo, GuardianInfo, ESPUser, shirt_sizes, shirt_types, Record
 from esp.utils.expirable_model import ExpirableModel
 from esp.utils.formats import format_lazy
 from esp.qsdmedia.models import Media
 
+PROFILE_MAX_AGE_DAYS = 5
```
pricem: I don't see where this is used.

esp/esp/program/models/__init__.py
```diff
@@ -1326,16 +1341,21 @@ def getLastForProgram(user, program):
...
-                if (datetime.now() - regProf.last_ts).days <= 5:
-                    regProf.save()
+        
+        if len(profile_list) < 1:
+            registration_profile = RegistrationProfile.getLastProfile(user)
+            registration_profile.program = program
+
+            if registration_profile.id is not None:
+                registration_profile.id = None
+                if (datetime.now() - registration_profile.last_ts).days <= 5:
```
pricem: This may be where you intended to use PROFILE_MAX_AGE_DAYS (sorry about the extra comment).

esp/esp/program/modules/forms/splashinfo.py
```diff
@@ -33,32 +33,25 @@
...
-    program/modules/splashinfomodule/splashinfo.html
-"""
+from esp.accounting.models import LineItemType
+from esp.accounting.controllers import IndividualAccountingController
+
+
+LUNCH_SATURDAY = 'lunchsat'
+LUNCH_SUNDAY = 'lunchsun'
+DISCOUNT_SIBLING = 'siblingdiscount'
+SIBLING_NAME = 'siblingname'
```
pricem: Why are these variables if the field names are still hard coded into the form?  I would suggest that either the field names themselves become dynamic or the code goes back to using ```'lunchsat'```, ```'lunchsun'```, etc.

pricem:
Once #1561 is merged, I would merge main into this to make sure there aren't any merge conflicts or extraneous code (since I think this was branched off of that).

esp/esp/program/modules/forms/splashinfo.py
joeleg: The values were *previously* hardcoded - I created the variables for reuse and so this doesn't change a whole lot, materially, except that it's probably better to reference variables, rather than string literals. In terms of making the values dynamic, this opens up the discussion a bit around the feature requirements. i.e. If we want to provide more flexibility beyond the weekend lunch and sibling discount - what would that be?(and is this redundant, when considering that it could be handled by studentreg?)

benjaminjkraft:
Closing in favor of #1773.

PR #1590, opened 2015-02-12
Add group name to Stripe payments

jmoldow:
This allows for better sorting and filtering of Stripe
transactions by chapter.

### DIFF
--- a/esp/esp/program/modules/handlers/creditcardmodule_stripe.py
+++ b/esp/esp/program/modules/handlers/creditcardmodule_stripe.py
@@ -210,6 +210,8 @@ class CreditCardModule_Stripe(ProgramModuleObj):
 
         context = {'postdata': request.POST.copy()}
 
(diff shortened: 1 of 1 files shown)


esp/esp/program/modules/handlers/creditcardmodule_stripe.py
```diff
@@ -210,6 +210,8 @@ def charge_payment(self, request, tl, one, two, module, extra, prog):
 
         context = {'postdata': request.POST.copy()}
 
+        group_name = Tag.getTag('full_group_name') or '%s %s' % (settings.INSTITUTION_NAME, settings.ORGANIZATION_SHORT_NAME)
```
benjaminjkraft: We may want to just generally have a universal canonical way of getting the full group name, but I think this is fine for now.

benjaminjkraft:
Looks good to me.  I re-ran tests due to unrelated Travis errors; will merge when they pass.

benjaminjkraft:
Merging, since Travis seems to still be broken, and I don't think this has any test coverage anyway.

PR #1592, opened 2015-02-13
Fix bug with selecting option in AjaxForeignKeyNewformField

luac:
See the second commit message for all the gory details.

This fixes a browser-dependent bug with AjaxForeignKeyNewformField. To reproduce: on Firefox, go to /manage///check_availability or /manage///medicalbypass, search for a user, select the user, then click away from the field (like somewhere else on the page), then click submit. It should (incorrectly) behave as if no user had been selected.

Also worth testing, to make sure that this change didn't break it: if you select a user, then clear the textbox, then click submit, it should (correctly) behave as if no user had been selected.


benjaminjkraft:
In the K12School autocomplete on `/learn///profile`, it is now impossible to select a school.  Otherwise, the changes seem to work for me, and the code makes sense.  (I wonder if now would be a good time to merge `AjaxForeignKeyFieldBase` and `AjaxForeignKeyWidget`, but it certainly wouldn't hurt.)  I still want to do some more poking around to see if the jQuery UI version bump is likely to break anything.

benjaminjkraft:
For the jQuery UI version bump, I think we're probably safe on removed things.  I think we will hit some deprecations on the accordion and tabs widgets (see the deprecations in the changelog), so we should move to removing those.  We may also want to check whether our jQuery plugins are happy with the version or if we need to version-bump those too.  But I think that's all deprecation, not removal.

luac:
Seems the K12School problem was actually related to the jQuery UI version bump. I'm not sure what exactly, still looking into it.

luac:
Seems like the problem is double-including jQuery UI: http://stackoverflow.com/a/15088879

I fixed the one instance of this in the profile; I wasn't sure if we wanted to make a separate issue/PR to get rid of the rest of the double-includes (and making sure that nothing else is broken).

benjaminjkraft:
I think we shouldn't merge this if it breaks other things... unless I'm misunderstanding the issue.

luac:
The specific issue is when you have an autocomplete widget and include jQuery UI twice. I was thinking that these two things didn't occur together elsewhere, but a cursory search suggests otherwise. I'll fix this later.

luac:
Alright, I updated this removing the rest of the double includes of jQuery UI that I could find on the website, and checked that this resolved the issue with the relevant pages (profile, coteachers, make admin).

benjaminjkraft:
Ok, in that case I think this is fine to merge.  We should definitely take a look at the deprecation warnings, because the new scheduler bumps us to 1.11, in which those things are gone, which breaks the comm panel among other things.  But that doesn't have to be before merging this.

Issue #828, opened 2013-11-27
Classes on the half-hour


benjaminjkraft:
MIT this year tried starting its walk-in activities on the half-hour rather than on the hour; they still had to be some multiple of an hour in length, but were offset from other classes by half an hour.  It would be nice to have some way of supporting this on the website.  We don't just want to have extra timeblocks, because that will confuse student registration and possibly also the autoscheduler.  What we did was to make them listed as an hour longer in the website, but that was confusing.

I'm not sure what the best way to handle this is, but it would be useful to have actual web support for it.

PR #1596, opened 2015-02-14
Excluded walk-ins from teachers who like running

jerrywu64:
Fairly self-explanatory. I don't think there's an issue open for this right now though.

### DIFF
--- a/esp/esp/program/modules/handlers/schedulingcheckmodule.py
+++ b/esp/esp/program/modules/handlers/schedulingcheckmodule.py
@@ -510,7 +510,7 @@ class SchedulingCheckRunner:
          teachers = self.p.teachers()['class_approved'].distinct()
          for teacher in teachers:
              sections = ClassSection.objects.filter(
-                 parent_class__in=teacher.getTaughtClassesFromProgram(self.p).filter(status=10).distinct(),status=10).distinct().order_by('meeting_times__start')
+                 parent_class__in=teacher.getTaughtClassesFromProgram(self.p).filter(status=10).distinct(),status=10).exclude(parent_class__category__id=self.p.open_class_category.id).distinct().order_by('meeting_times__start')
              for i in range(sections.count()-1):
                  try:
                      time1 = sections[i+1].meeting_times.all().order_by('start')[0]


benjaminjkraft:
Will this work if there's no open class category?

jerrywu64:
open_class_category might end up as None, I think? Regardless, open class category is used in a whole bunch of scheduling checks-- such as the previous implementation of all_class_sections, so I expect this change won't break anything that wasn't already broken.

jmoldow:
Why would we do this? Why should it matter whether the class is a walkin or not?

jerrywu64:
Walk-ins start on the half hour, so you aren't actually going to be in a hurry if you need to travel to a walk-in from somewhere else on campus.

jmoldow:
There is no special property of walk-ins that means they have to be scheduled on the half hour. This is a fad that MIT happens to be doing right now. It wasn't always that way, might not continue to be that way indefinitely, and is not that way at other chapters that use the walk-in seminar feature. This will break this scheduling check for, e.g. Stanford.

Can't you guys just ignore results that include walk-ins, rather than changing the code?

benjaminjkraft:
It's really nice to remove false positives from scheduling checks, it makes it a lot easier to iterate the checks.  Maybe we should have a half hour walkins setting on the CRMI to control this, because we're eventually going to need that anyway -- I think at MIT the half-hour walkins thing is here to stay.

(OOC, do other chapters even use scheduling checks?  Some are already fairly specific to the way MIT does things, even if other chapters could theoretically use them.)

mgersh:
I'm personally not a fan of putting workaround hacks like this in the codebase, even if we're the only ones using it, especially given that walk-ins are not a huge fraction of the "teachers who like running" false positives. Actually implementing #828 is the long-term solution, and I'm guessing that someone will actually do that within the next couple of years, but not soon enough that we'll remember we put this here and remove it.

benjaminjkraft:
Seems like once #1687 is done this will be less of an issue, so I'm going to close this until such time as we have a real half-hour walk-in solution across the site.  If people still think otherwise, do say so.

PR #1598, opened 2015-02-15
Fix Travis failure related to pip install command.

luac:
The output format of pip changed, causing grep to fail and Travis to
treat that as a build error. Modify it to not do the grep thing at
all, since the only reason we had it in the first place was to output
some progress information so that Travis wouldn't think we stalled,
and that doesn't seem to be happening anymore.


benjaminjkraft:
Whoo, tests are green so I'd say this works :-) .  Merging.

PR #1606, opened 2015-02-19
UI for flushing individual caches

luac:
Fix the /cache/view_all view, which displays a list of all cached functions, and add the ability to flush individual cached functions.

### DIFF
--- a/esp/esp/cache/views.py
+++ b/esp/esp/cache/views.py
@@ -39,6 +39,6 @@ from esp.web.util.main import render_to_response
 
 @admin_required
 def view_all(request):
-    caches = sorted(all_caches.values(), key=lambda c: c.name)
+    caches = sorted(all_caches, key=lambda c: c.name)
     cache_data = [{'pretty_name': cache.pretty_name, 'hit_count': cache.hit_count, 'miss_count': cache.miss_count} for cache in caches]
     return render_to_response('cache/view_all.html', request, {'caches': cache_data})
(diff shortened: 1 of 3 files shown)


esp/esp/cache/views.py
```diff
@@ -36,9 +36,17 @@
...
 from esp.users.models import admin_required
 from esp.web.util.main import render_to_response
+from django.shortcuts import redirect
+from django.core.urlresolvers import reverse
 
 @admin_required
 def view_all(request):
-    caches = sorted(all_caches.values(), key=lambda c: c.name)
+    caches = sorted(all_caches, key=lambda c: c.name)
     cache_data = [{'pretty_name': cache.pretty_name, 'hit_count': cache.hit_count, 'miss_count': cache.miss_count} for cache in caches]
```
benjaminjkraft: I kow you didn't add this line, but it's completely uncesseary -- can you remove it (replace `cache_data` with `caches` below)?

benjaminjkraft:
It would be nice if it were more obvious that the table is sortable.  It would also be nice if it didn't overflow the page css (on the fruitsalad theme) to have long cache names as is often the case.  But I'm fine with merging without those if there's not an easy fix.

esp/esp/cache/views.py
luac: I just tried making this change, and it broke something. I think maybe the template gets confused because it thinks the cache objects are callable?

benjaminjkraft: (We talked about this in person.)  Yep, I remember this now.

benjaminjkraft:
Ok, LGTM.

Issue #1607, opened 2015-02-23
esint.sty not included in in-line latex generation


bhorowitz:
Website throws math generation error when attempting to render latex from esint.sty due to not importing package. Package contains lots of integrals of general use (surface integrals, etc.) and should be included if possible.

bhorowitz:
I supposedly put in a pull request for the extra few letters... I might have done it wrong cause it is the first time I ever tried to submit a pull request to someone else repository.

benjaminjkraft:
Fixed by #1608.

PR #1608, opened 2015-02-23
Update inlinelatex.py

bhorowitz:
Including esint.sty for more integrals. Will need to update requisite server requirements to include esint.sty in latex generator.


benjaminjkraft:
Seems reasonable to me; fixes #1607.

PR #1609, opened 2015-02-24
Add script to email asking about class cap increases

benjaminjkraft:
Used by MIT sometimes.  Previously done with a mail merge outside the website, but this seemed easier, and is hopefully easier to reuse.


gkanwar:
Suggestion: we could make this a check box option in the class registration form so we don't have to email

benjaminjkraft:
If we don't want to do that, we could also make a page with "click here to
approve" links; even if they don't directly modify the size they could at
least just go somewhere that another script can import.

benjaminjkraft:
Merged main to fix tests.

mgersh:
Looks fine, and it worked for this Spark. I'll go ahead and merge this. We can do something fancier later if we decide it would be useful.

PR #1615, opened 2015-03-06
Fix openclasses

luac:
Fix some issues with the learn///openclasses view. (It was 500ing, and also using the wrong template.)

### DIFF
--- a/esp/esp/program/modules/handlers/onsiteclasslist.py
+++ b/esp/esp/program/modules/handlers/onsiteclasslist.py
@@ -377,7 +377,7 @@ class OnSiteClassList(ProgramModuleObj):
     def classlist_public(self, request, tl, one, two, module, extra, prog):
         return self.classList_base(request, tl, one, two, module, extra, prog, options={}, template_name='allclass_fragment.html')
 
-    def classList_base(self, request, tl, one, two, module, extra, prog, options=None, template_name=None):
+    def classList_base(self, request, tl, one, two, module, extra, prog, options=None, template_name='classlist.html'):
         """ Display a list of all classes that still have space in them """
 
         #   Allow options to be supplied as an argument to the view function, in lieu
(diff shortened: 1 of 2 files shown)


esp/esp/program/modules/handlers/studentclassregmodule.py
```diff
@@ -776,7 +776,7 @@ def openclasses(self, request, tl, one, two, module, extra, prog):
 
         module = prog.getModule('OnSiteClassList')
         if module:
-            return module.classList_base(request, tl, one, two, module, 'by_time', prog, 'allclass_fragment.html')
+            return module.classList_base(request, tl, one, two, module, 'by_time', prog, options={}, template_name='allclass_fragment.html')
```
benjaminjkraft: Is there a reason to pass `{}` instead of `None` to `options`?

luac: Passing `None` takes you to a landing page for choosing the options, which isn't what we want here because this one uses a different template which doesn't use those options anyway.

benjaminjkraft: Ok, seems reasonable.

esp/esp/program/modules/handlers/onsiteclasslist.py
```diff
@@ -448,7 +448,7 @@ def classList_base(self, request, tl, one, two, module, extra, prog, options=Non
         else:
             context['use_categories'] = True
         
-        return render_to_response(self.baseDir()+'classlist.html', request, context)
+        return render_to_response(self.baseDir()+template_name, request, context)
```
benjaminjkraft: Does this mess up `classlist_public`, which appears to set the previously unused `template_name` argument (to something that's not `classlist.html`)?  (Or does it fix it?)

luac: I didn't check `classlist_public`, but it's the same template as `openclasses`, and this fixes that.

benjaminjkraft: Ah, ok, I see -- this is the one you said might be the same.  Cool.

benjaminjkraft:
Looks good to me, merging.

Issue #1158, opened 2014-03-17
Add program name as a bracketed tag in subject lines of class-list emails


gkanwar:
Currently emails to a class list (X1234-students@esp.mit.edu, for example) don't mention the program name anywhere. This can get confusing if multiple programs are running simultaneously. It would also be helpful for students to be able to identify program emails from the subject line. We should add something like "[Splash]" to the beginning of the subject.

PR #1618, opened 2015-03-08
Add program name to the subject line of messages to Mailman lists

mgersh:
This won't work for already-existing lists at the time this is merged, because it applies the setting at list creation.


mgersh:
Fixes #1158

benjaminjkraft:
LGTM.

PR #1625, opened 2015-03-09
Remove custom schedule templates

mgersh:
@benjaminjkraft pointed out that this exists. Now that our codebase isn't just used by MIT, it doesn't make sense to have. Whichever of Stanford, Yale, or Chicago gets to program 47 first would have been very confused.


benjaminjkraft:
LGTM.

PR #1627, opened 2015-03-14
Display user data in custom form responses

### LINKED ISSUE #547: Custom Form Responses Views Should Display User Data
Each response to a custom form is saved with a foreign key to the user who submitted the response. In the viewResponse view (and also in the getExcelData view), a column is added to the response data for displaying the user id of the user. This is not helpful if you want to look at the response data and quickly determine who submitted what responses.

The following data should also be displayed (or optionally displayable):
 * username
 * first name
 * last name
 * email

Do we want this information to always be displayed? Or should it be viewable only by clicking checkboxes after loading the view?

If the form is specified as being only for teachers, do we also want to have TeacherInfo fields displayed (or optionally displayable)? If the form is specified as being only for students, do we want to also have StudentInfo and student/guardian/emergency ContactInfo fields displayed (or optionally displayable)?

jmoldow:
Show usernames, names, and email addresses in custom form responses. Previously only the user id was visible, which didn't make for a good UI.

Fixes #547.

### DIFF
--- a/esp/esp/customforms/DynamicForm.py
+++ b/esp/esp/customforms/DynamicForm.py
@@ -11,7 +11,7 @@ from django.contrib.localflavor.us.forms import USStateField, USPhoneNumberField
 from esp.customforms.forms import NameField, AddressField
 from esp.customforms.DynamicModel import DMH
 from esp.utils.forms import DummyField
-from esp.users.models import ContactInfo
+from esp.users.models import ContactInfo, ESPUser
 from esp.cache import cache_function
 from esp.program.models import Program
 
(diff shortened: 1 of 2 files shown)


esp/esp/customforms/DynamicForm.py
```diff
@@ -646,14 +649,23 @@ def getResponseData(self, form):
                 # Include this field only if it isn't a dummy field
             elif generic_fields[ftype]['typeMap'] is not DummyField:
                 response_data['questions'].append([qname, field['label'], ftype])
-            
+
+        users = ESPUser.objects.filter(id__in=map(lambda response: response['user_id'], responses)).distinct()
```
luac: I believe `ESPUser.objects.in_bulk` does the same thing as this, but this seems fine too.

jmoldow: TIL! I like that better than what I did. I'll make that change.

benjaminjkraft:
Looks good to me, and worked for MIT.

PR #1631, opened 2015-03-15
Disable resource type cascading deletes

jmoldow:
This will prevent an issue that recently occurred at Stanford, where the ``ResourceType``s were accidentally deleted, taking all of the teacher ``ResourceRequest``s with it.


jmoldow:
Oops, spelling fail. The branch should be protect-resources, not protest-resources... opening a new PR.

PR #1633, opened 2015-03-19
Disable resource type cascading deletes

jmoldow:
This will prevent an issue that recently occurred at Stanford, where the ``ResourceType``s were accidentally deleted, taking all of the teacher ``ResourceRequests`` with it.


benjaminjkraft:
Cool, LGTM.  You may want to document this somewhere as well, so that people don't (deliberately) try to delete a ResourceType and then get confused why it doesn't work.  Eventually we probably want any non-admin-panel interface where one might try to do so to give a reasonable error message.

PR #1637, opened 2015-03-23
Fix admin panel filtering for Resource

mgersh:
Filtering Resources by program would miss some resources, because it
got the program by checking res_type__program. Some ResourceTypes are global,
so this doesn't catch everything. This commit switches to checking
event__program. ResourceRequest has a similar problem, but can't be
reliably fixed in the same way because of target and target_subj, and that one
wasn't causing a problem in MIT ESP's setup so I left it as is.


benjaminjkraft:
LGTM.

Issue #1290, opened 2014-08-24
Profile update should only be asked for once per program


jmoldow:
Students and teachers need to update their profile before registering for a new program. However, for some reason the site will also ask you update your profile if >N days (I don't know what N is) have passed since the last time you updated it. This seems unnecessary, and is annoying to students, teachers, and admins.

We should update the profile code to not ask you to re-update the profile for a program, once you've done it already.

There might be complications in the case where multiple programs are happening at once. When a profile is saved, a new RegistrationProfile object is created with most_recent_profile set to True, and the previous RegistrationProfile has most_recent_profile set to False. There might be program code that looks for the profile with most_recent_profile=True that also is for that specific program. Either this code will have to be changed, or we might have to accept the fact that a user will be prompted to update their profile if their most recent profile isn't for that program.

Requested by Yale, and probably also desired by many users.

PR #1638, opened 2015-03-23
ESPUserManager improvements

jmoldow:
Make ``ESPUserManager`` derive from Django's ``UserManager``, so that it gets utility methods such as ``create_user()``. Also add a new utility method, ``create_users_for_program()``.

### DIFF
--- a/esp/esp/users/models/__init__.py
+++ b/esp/esp/users/models/__init__.py
@@ -40,7 +40,7 @@ import simplejson as json
 from django import forms, dispatch
 from django.conf import settings
 from django.contrib.auth import logout, login, REDIRECT_FIELD_NAME
-from django.contrib.auth.models import User, AnonymousUser, Group
+from django.contrib.auth.models import User, AnonymousUser, Group, UserManager
 from django.contrib.localflavor.us.forms import USStateSelect
 from django.contrib.localflavor.us.models import USStateField, PhoneNumberField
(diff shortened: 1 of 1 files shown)


benjaminjkraft:
Why put this on the manager rather than as a useful_script?  It seems like something very special-purpose that might be better kept as a script or set of functions in useful_scripts.

esp/esp/users/models/__init__.py
```diff
@@ -126,8 +126,83 @@ def save(self, *args, **kwargs):
...
+            # Set the `last_ts` field of the profile to be a long time ago. If
+            # the `last_ts` field is too recent, the user won't be prompted to
+            # edit their profile when they log in and register for the program,
+            # and they'll forever have a blank, useless profile.
+            #
+            # NOTE(jmoldow): But we can't set `last_ts` with `create()` or
+            # `save()`, because the custom `RegistrationProfile.save()` always
+            # sets that field using `datetime.now()`. So we must use
+            # `update()`.
+            RegistrationProfile.objects.filter(id=profile.id).update(last_ts=(datetime.now() - timedelta(days=20)))
```
luac: Is this the right thing to do / does this actually cause the user to be prompted to edit their profile for that program? See confusion in #1290.

If so, is there a way to have this not be a magic constant?

benjaminjkraft:
Closing in favor of #1831.

PR #1651, opened 2015-03-24
Remove the "security schedule" printable

mgersh:
It's broken. Specifically, it doesn't display in a way that you
could actually use for printing, and it includes timeblocks that
aren't actual class timeblocks. I figure that if anyone were still
using it, they probably would have mentioned that it's broken.


benjaminjkraft:
Whoo removing old code!

PR #1652, opened 2015-03-24
Remove irrelevant scheduling checks

mgersh:
We don't have high school only blocks anymore, so these aren't doing anything useful.


jmoldow:
LGTM

PR #1656, opened 2015-03-27
Program module install improvements

mgersh:
Running install() now updates module_properties for program modules that have changed, and deletes modules that don't exist anymore. Also, the code was assuming that (handler, type) pairs are unique. I fixed the one case where this wasn't true, since there was no reason for there to be two modules there.

I tested this by running it on my dev server and checking that only the modules that should be deleted were deleted, nothing bad happened to a program that had one of them enabled, and I was able to change the link_title of a module.

### DIFF
--- a/esp/esp/program/modules/handlers/onsiteclasslist.py
+++ b/esp/esp/program/modules/handlers/onsiteclasslist.py
@@ -61,17 +61,12 @@ def hsl_to_rgb(hue, saturation, lightness=0.5):
 class OnSiteClassList(ProgramModuleObj):
     @classmethod
     def module_properties(cls):
-        return [ {
-            "admin_title": "Show All Classes at Onsite Registration",
-            "link_title": "List of All Classes",
-            "module_type": "onsite",
-            "seq": 31,
-            }, {
+        return {
             "admin_title": "Show Open Classes at Onsite Registration",
             "link_title": "List of Open Classes",
             "module_type": "onsite",
             "seq": 32,
-            } ]
+            }
 
     @cache_function
     def section_data(sec):
--- a/esp/esp/program/modules/models.py
+++ b/esp/esp/program/modules/models.py
@@ -36,18 +36,6 @@ Learning Unlimited, Inc.
 from esp.program.modules.handlers import *
 from django.db.models import Q
 
-#from django.contrib import admin
-#from django.db import models 
-#from esp.program.models import Program
-#
-#class DBReceipt(models.Model):
-#    """ Per-program Receipt templates """
-#    program = models.OneToOneField(Program)
-#    receipt = models.TextField()
-#
-#admin_site.register(DBReceipt)
-
-
 def updateModules(update_data, overwriteExisting=False, deleteExtra=False, model=None):
     """
     Given a list of key:value dictionaries containing fields from the
@@ -89,7 +77,7 @@ def updateModules(update_data, overwriteExisting=False, deleteExtra=False, model
         for (datum, (mod, created)) in mods:
             ids.append(mod.id)
 
-        #ProgramModule.objects.exclude(id__in=ids).delete()
+        ProgramModule.objects.exclude(id__in=ids).delete()
 
     for (datum, (mod, created)) in mods:
         #   If the module exists but the provided data adds fields that 
@@ -114,6 +102,6 @@ def install(model=None):
     for module in modules:
         table_data += module.module_properties_autopopulated()
 
-    updateModules(table_data, model=model)
+    updateModules(table_data, overwriteExisting=True, deleteExtra=True, model=model)
     
 from esp.program.modules.module_ext import *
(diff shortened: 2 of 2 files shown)


jmoldow:
I'm okay with most of this, but I'd like to discuss the change to use ``overwriteExisting=True``.

The fields on ``ProgramModule`` are:
```python
    link_title = models.CharField(max_length=64, blank=True, null=True)

    # Human-readable name for the Program Module
    admin_title = models.CharField(max_length=128)

    #   A module can have an inline template (whose context is filled by prepare())
    #   independently of its main view.
    inline_template = models.CharField(max_length=32, blank=True, null=True)

    # One of teach/learn/etc.; What is this module typically used for?
    module_type = models.CharField(max_length=32)

    # self.__name__, stored neatly in the database
    handler    = models.CharField(max_length=32)

    # Sequence orderer.  When ProgramModules are listed on a page, order them
    # from smallest to largest 'seq' value
    seq = models.IntegerField()

    # Must the user supply this ProgramModule with data in order to complete program registration?
    required = models.BooleanField()
```

- ``handler`` will never change (if it does, the module will be deleted and recreated). 
- ``link_title`` is what appears to teachers/students in teacherreg/studentreg in the list of steps, so chapters may wish to customize this text.
- ``admin_title`` is the way admins identify modules in the admin panel / manage/onsite mainpages, so chapters may with to customize this text.
- ``seq`` and ``required`` may be customized by chapters that want their program module setup to always be created correctly for each new program, without always having to muck with ProgramModuleObjs.
- ``inline_template`` and ``module_type`` should probably always match the database. But I also don't think admins ever change these values anyway.

For ``link_title``, ``admin_title``, ``seq``, and ``required``, I think we'd be doing a disservice to chapters by overwriting their values. And I don't think the values of these fields are ever important to us. If they were that important to be controlled values, we'd remove them from the database and put them in code only (but I don't think we should do that).

For ``inline_template`` and ``module_type``, I guess we can overwrite them? But it doesn't seem important to me.

Basically, I think we should go back to using ``overwriteExisting=False``, or that we should change the behavior to only overwrite ``inline_template`` and ``module_type``.

mgersh:
I thought about this before making the change, and I disagree. My original motivation for this was noticing that I couldn't change the `link_title` of a module, not because I want to do that now, but because I want to write a module now with the intent of extending its behavior in the future and changing its description to match, in order to be more useful to admins trying to use the module. This is the sort of thing that I could write a migration to change, but I shouldn't have to.

I also doubt that anyone is deliberately using the ability to change defaults. MIT isn't, at least not consciously. If someone can think of examples that the chapter actually cares about (not just someone set it that way once a long time ago), that could convince me that we shouldn't overwrite `seq` and `required`, but I do know that very few chapters have the website knowledge to do it without asking us. In general, it seems much more likely that a dev would want to change defaults than that a chapter would want to change theirs, and if we really need to support both, the system we have just isn't built for that.

benjaminjkraft:
I think MIT has used this once or twice, but it's currently a pain to use, because you have to make the change for all programs on the `ProgramModule` (with the exception of `seq`, which you can also change per-program on the `ProgramModuleObj`).  If we actually want to make this customizable, it should be customizable per-program; until such time I don't think it's worth doing acrobatics to keep it theoretically customizable in a way that I'd argue isn't very useful.

Actually, while we're at it, why have these fields at all?  Why not just pull them from the code directly at runtime?  Obviously we need a model so that `ProgramModuleObj`s can exist, but it could just be the handler and nothing more.  But that change is a bigger one and may be beyond the scope of this PR.

mgersh:
seq isn't the only per-program customizable one, required also is, and MIT
sometimes changes required for individual programs. There's also
required_label, which we also frequently customize, and that one isn't on
the ProgramModule at all. But I don't think we tend to change the other
fields, or change anything on the ProgramModule, and in most cases where
someone would want to change the others, it would be for clarification and
that should be possible to do in the code.

benjaminjkraft:
In the interest of arguing about things with actual data, I checked what chapter sites actually have changes.  I put the diff in http://pastebin.com/JTbcHT0c (along with the script I used, if you care).  Some observations:
* A lot of chapters have customized `required` and `seq`; these we should probably just not override.
* A lot of chapters have changes that appear to just be things not getting updated after the code changes.  See for example most of the changes on Chicago's site.
* A few chapters have actually customized things.  In some cases, we should probably just make the change for everyone (e.g. "Support More Splash!" rather than "Optional Donation" on Berkeley).  In others, it's an actual change (e.g. "Payment type (and additional purchases)" for "Student Optional Fees" on Princeton).

In short, there are a few customizations we might want to keep, but most of them should get updated.  So we may want to think about how to do this so that we can make updates and chapters can still override them, or something like that.

mgersh:
Since it does look like some chapters are actually using this feature, particularly on the StudentExtraCosts module which is actually used differently by different chapters, I'll take that part out of this pull request so the rest of it can get merged. We still need an update mechanism, though. My current idea is to make the ProgramModule field values null/blank by default, and putting something in them is an override.

benjaminjkraft:
While we're at it, we may want to set `editable=False` on the things that admins definitely shouldn't edit, like `handler`, just to make it obvious.

benjaminjkraft:
Ok.  I'm going to merge this, and make an issue to discuss further.

PR #1663, opened 2015-03-30
Django 1.6 form wizard

joeleg:


pricem:
Is there anything different here relative to #955?  (Actually, it looks like #955 is a few commits ahead.)

PR #1669, opened 2015-04-10
Ticket 1468

joeleg:


benjaminjkraft:
This looks good to me.  It's not entirely clear that the functionality we're using here is considered a stable API (cf. https://code.djangoproject.com/ticket/17985), but if it changes the worst that happens is probably that this change no longer does anything, so we can cross that bridge when we come to it.  Since we generally assume that anyone with admin panel access has more or less full DB access, this is not a security issue, but if that ever changes we may want to reevaluate.

PR #1674, opened 2015-04-16
Fix bug when program doesn't have dates yet

mgersh:
When you first create a program, it doesn't have any timeblocks yet, so program.dates() returns an empty list. This was causing every program module view to error.


benjaminjkraft:
LGTM.

Issue #1374, opened 2014-10-11
All teacher, student names in manage pages should hyperlink to user view


jmoldow:
For example, lists of teachers in the dashboard, manage class pages, class flags pages, and scheduling checks. Or students in the credit card viewpay module.

Requested by Stanford.

willgearty:
Links added to viewpay in https://github.com/learning-unlimited/ESP-Website/pull/2979

willgearty:
Teachers also already have links on the manage class page

PR #1677, opened 2015-04-20
Link teacher and volunteer names to /userview

ghost:
Adds links to `/userview` for teacher and volunteer names on various manage pages.

First step towards fixing #1374.

### DIFF
--- a/esp/esp/program/modules/handlers/teachereventsmodule.py
+++ b/esp/esp/program/modules/handlers/teachereventsmodule.py
@@ -195,7 +195,7 @@ class TeacherEventsModule(ProgramModuleObj):
         training_times = self.getTimes('training')
         
         for ts in list( interview_times ) + list( training_times ):
-            ts.teachers = [ x.user.first_name + ' ' + x.user.last_name + ' <' + x.user.email + '>' for x in self.entriesBySlot( ts ) ]
+            ts.teachers = [ x.user for x in self.entriesBySlot( ts ) ]
         
         context['prog'] = prog
         context['interview_times'] = interview_times
--- a/esp/esp/program/urls.py
+++ b/esp/esp/program/urls.py
@@ -1,4 +1,5 @@
 from django.conf.urls.defaults import *
+""
 
 urlpatterns = patterns('',
                        # manage stuff
@@ -6,7 +7,7 @@ urlpatterns = patterns('',
                        (r'^manage/newprogram/?$', 'esp.program.views.newprogram'),
                        (r'^manage/submit_transaction/?$', 'esp.program.views.submit_transaction'),
                        (r'^manage/pages/?$', 'esp.program.views.manage_pages'),
-                       (r'^manage/userview/?$', 'esp.program.views.userview'),
+                       url(r'^manage/userview/?$', 'esp.program.views.userview', name='userview'),
                        (r'^manage/deactivate_user/?$', 'esp.program.views.deactivate_user'),
                        (r'^manage/activate_user/?$', 'esp.program.views.activate_user'),
                        (r'^manage/usersearch/?$', 'esp.program.views.usersearch'),                       
--- a/esp/esp/users/models/__init__.py
+++ b/esp/esp/users/models/__init__.py
@@ -180,7 +180,7 @@ class ESPUser(User, AnonymousUser):
         return self.get_full_name() or self.username
 
     def get_absolute_url(self):
-        return '{0}?username={1}'.format(resolvers.reverse('userview'), self.username)
+        return '{0}?username={1}'.format(urlresolvers.reverse('userview'), self.username)
       
     @classmethod
     def create_membership_methods(cls):
--- a/esp/esp/users/models/__init__.py
+++ b/esp/esp/users/models/__init__.py
@@ -176,6 +176,12 @@ class ESPUser(User, AnonymousUser):
 
         self.create_membership_methods()
 
+    def __unicode__(self):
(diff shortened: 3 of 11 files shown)


esp/templates/program/modules/volunteermanage/main.html
```diff
@@ -29,8 +29,9 @@
     <tr>
     <td>{{ vr.timeslot.pretty_time }}</td>
     <td>{{ vr.timeslot.description }}: {{ vr.num_offers }} / {{ vr.num_volunteers }} volunteers <br />
-        <div id="offers" style="color: blue; font-size: 0.9em;">{% for offer in vr.get_offers %}
-        {{ offer.name }}, {{ offer.email }}, {{ offer.phone }} <br />
+        <div id="offers" style="color: blue; font-size: 0.9em;">
+        {% for offer in vr.get_offers %}
+        <a href="{{ offer.user.get_absolute_url }}">{{ offer.name }}</a>, {{ offer.email }}, {{ offer.phone }} <br />
```
benjaminjkraft: Will this work if they aren't a user on the website?  I think ideally it would be linkified only if there is an `offer.user`, which is not always.

esp/esp/users/templatetags/userlist.py
```diff
@@ -0,0 +1,11 @@
+from django.template import Library
+from django.utils.safestring import mark_safe
+
+register = Library()
+
+def userlist(var, delimiter):
+	atag = lambda x:'<a href="{0}">{1}</a>'.format(x.get_absolute_url(),unicode(x))
+	userlist_html = '{0} '.format(delimiter).join([atag(i) for i in var])
+	return mark_safe(userlist_html)
```
benjaminjkraft: Do we know that users can't set their name to contain HTML tags?  We should probably make sure it gets escaped on the way into this tag so that we can safely mark it as safe on the way out.

esp/esp/users/models/__init__.py
```diff
@@ -252,7 +258,7 @@ def getOld(self):
         return self.__olduser
 
     def name(self):
-        return u'%s %s' % (self.first_name, self.last_name)
+        return self.get_full_name()
```
benjaminjkraft: Cool, TIL.

esp/esp/program/urls.py
```diff
@@ -1,4 +1,5 @@
 from django.conf.urls.defaults import *
+""
```
benjaminjkraft: is this intentional?

esp/esp/program/modules/handlers/teachereventsmodule.py
```diff
@@ -195,7 +195,7 @@ def teacher_events(self, request, tl, one, two, module, extra, prog):
         training_times = self.getTimes('training')
         
         for ts in list( interview_times ) + list( training_times ):
-            ts.teachers = [ x.user.first_name + ' ' + x.user.last_name + ' <' + x.user.email + '>' for x in self.entriesBySlot( ts ) ]
```
benjaminjkraft: I don't like losing the email addresses in this.  I think the intention is that you can just copy-paste the entire block into an email to field, and send to all teachers with a given teacher training.  So it would be nice to keep that, possibly by having an option to the userlist tag that lets you also display email addresses.

benjaminjkraft:
Left some comments but generally looks good.  Note for when it gets merged: it might still be worth keeping #1374 open because there are a lot more instances of this; this is a great start though.

esp/public/media/scripts/program/modules/adminclass.js
```diff
@@ -247,45 +247,46 @@ function createClassRow(clsObj)
...
-    for (var i = 0; i < clsObj.sections.length; i++)
-    {
-	var section = sections[clsObj.sections[i]];
-	section_link_list = section_link_list.concat("<a href='/teach/"+base_url+"/select_students/"+section.id+"'>Sec. "+section.index+"</a><br />");
-    }
+  var teacher_link_list = "";
+  for (var i in teacher_list) {
+      var teacher = json_data.teachers[teacher_list[i]];
+      var teacherName = teacher.first_name + " " + teacher.last_name;
+      teacher_link_list = teacher_link_list.concat('<a href="/manage/userview?username=' + teacher.username + '">' + teacherName + '</a><br />');
```
benjaminjkraft: Note to self or whomever ends up cleaning this up: here's another XSS issue (which may exist in the original too :/ ) .

esp/templates/program/modules/volunteermanage/main.html
willgearty: This should be fine now that accounts are created for all volunteers?

benjaminjkraft: That's not guaranteed for all configurations, as I recall.

willgearty: Oh, really? I thought any volunteers without accounts had accounts made for them?

benjaminjkraft: Nope!  If we're not requiring login, we just take their name and email and leave it at that.

willgearty: Now I'm really confused, because #2060 says that accounts are autocreated for any volunteers without accounts.

benjaminjkraft: Oh, huh, maybe this works differently than I thought!

willgearty:
Closing in favor of #3205

PR #1678, opened 2015-04-24
Some performance improvements to student schedule generation

benjaminjkraft:
Doesn't solve all the issues (I didn't touch the IAC in particular) but it
should still be a good deal faster.

### DIFF
--- a/esp/esp/program/models/class_.py
+++ b/esp/esp/program/models/class_.py
@@ -590,10 +590,11 @@ class ClassSection(models.Model):
         return False
 
     def start_time_prefetchable(self):
-        """Like self.start_time(), but can be prefetched.
+        """Like self.start_time().start, but can be prefetched.
 
-        If self.meeting_times.all() has been prefetched, this will not hit the
-        DB.
+        Gets the start time of a class.  If self.meeting_times.all() has been
+        prefetched, this will not hit the DB.  If it has not been prefetched,
+        this will not help.
         """
         mts = self.meeting_times.all()
         if mts:
--- a/esp/esp/program/models/class_.py
+++ b/esp/esp/program/models/class_.py
@@ -588,6 +588,18 @@ class ClassSection(models.Model):
             if time_passed > timedelta(0):
                 return True
         return False
+
+    def start_time_prefetchable(self):
+        """Like self.start_time(), but can be prefetched.
+
+        If self.meeting_times.all() has been prefetched, this will not hit the
+        DB.
+        """
+        mts = self.meeting_times.all()
+        if mts:
+            return min(mt.start for mt in mts)
+        else:
+            return None
    
     def start_time(self):
         if self.meeting_times.count() > 0:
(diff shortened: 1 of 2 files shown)


mgersh:
I tested this by generating .tex schedules on a fairly small testcase (HSSP Summer 2013, only schedules for students who started a financial aid application). It ran roughly twice as fast, but I found some discrepancies between the output from main and the output from this branch. In particular, there are a few cases where a class shows up on a student's schedule twice in this version.

esp/esp/program/models/class_.py
```diff
@@ -588,6 +588,18 @@ def already_passed(self):
             if time_passed > timedelta(0):
                 return True
         return False
+
+    def start_time_prefetchable(self):
+        """Like self.start_time(), but can be prefetched.
```
mgersh: The comment here is misleading--it makes me think that you can drop this in anywhere start_time() is used, but they return different kinds of things. (Also, are there other places in the code that might want start_time_prefetchable() instead?) Maybe the name of one or the other wants to be different too.

Question: why use meeting_times.all() here instead of getting it from the cache with get_meeting_times()?

benjaminjkraft: In practice when this is being used, meeting_times should be prefetched, in which case getting it directly avoids a roundtrip to memcache, which is fairly quick since the memcache is on the same server, but is still nontrivial if you're doing it 50k times (as for example MIT Splash schedules do).  I'll update the comment, though.

esp/esp/program/modules/handlers/programprintables.py
```diff
@@ -888,23 +887,46 @@ def studentschedules(self, request, tl, one, two, module, extra, prog, onsite=Fa
...
+        all_classes = all_classes.prefetch_related('meeting_times')
+        classes_by_id = {cls.id: cls for cls in all_classes}
+
+        sr_pairs = all_classes.values_list('id', 'studentregistration__user')
+        classes_by_student = collections.defaultdict(list)
+        for cls_id, user_id in sr_pairs:
+            classes_by_student[user_id].append(classes_by_id[cls_id])
+
+        for user_id in classes_by_student:
+            # Sort the classes.  We don't want to use
```
mgersh: Don't want to use what?

esp/esp/program/modules/handlers/programprintables.py
```diff
@@ -888,23 +887,46 @@ def studentschedules(self, request, tl, one, two, module, extra, prog, onsite=Fa
...
+        for cls_id, user_id in sr_pairs:
+            classes_by_student[user_id].append(classes_by_id[cls_id])
+
+        for user_id in classes_by_student:
+            # Sort the classes.  We don't want to use 
+            classes_by_student[user_id].sort(
+                key=lambda cls: (cls.start_time_prefetchable(), cls.title()))
+
+        times_compulsory = Event.objects.filter(program=prog, event_type__description='Compulsory').order_by('start')
+        for t in times_compulsory:
```
mgersh: I know you didn't write this code, but I'm confused about what it's doing. Any idea?

benjaminjkraft: (We talked in person, but FTSOTR, I think it's roughly trying to fake a class that the template will accept for a compulsory timeblock, such as a required lunch.)

PR #1688, opened 2015-05-04
Make 'survey' and 'surveys' aliases of each other

mgersh:
Because I got annoyed that the manage and teach/learn URLs are different.


benjaminjkraft:
LGTM.

PR #1696, opened 2015-05-28
Completely delete the datatree, userbits, and some less interesting things

mgersh:
See individual commits for what the other things are. I didn't bother touching the really old useful_scripts or a few comments that still reference DataTree, but everything else is gone.

After deleting the datatree, I decided to check and make sure that messing with migrations hadn't broken new dev server setup. Turns out that it had very badly broken new dev server setup, so the last few commits are fixing that. It works now.


benjaminjkraft:
LGTM.  I agree that dataviews can be removed; if there is at some point a desire to resurrect and fix it, the code is always in git, but there's no point maintaining it in the meantime.

PR #1701, opened 2015-06-03
Improve sanity of custom SQL in Ajax scheduler json view

### LINKED ISSUE #1698: Ajax scheduler breaks on four-digit timeblock id
It's caused by a bit of custom SQL in the json data module which assumes that all ids are three digits. MIT's database just hit 1000, and Yale will get there before the end of fall and possibly a lot sooner. See schedule_assignments(), https://github.com/learning-unlimited/ESP-Website/blob/main/esp/esp/program/modules/handlers/jsondatamodule.py#L138

The quick fix is adding a fourth '9' in the obvious place, and it works even on programs with a mix of three- and four-digit ids. It's probably a good idea to just do that now for at least MIT and Yale, but really we should find some better way of writing this.

mgersh:
Fixes #1698 by avoiding the cast to a string, since what we really wanted was a list of ints anyway.

### DIFF
--- a/esp/esp/program/modules/handlers/jsondatamodule.py
+++ b/esp/esp/program/modules/handlers/jsondatamodule.py
@@ -136,12 +136,9 @@ class JSONDataModule(ProgramModuleObj, CoreModule):
     @needs_admin
     @cached_module_view
     def schedule_assignments(prog):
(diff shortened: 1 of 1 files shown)


benjaminjkraft:
LGTM.

PR #1704, opened 2015-06-07
Ressurect old schedule printing script

### LINKED ISSUE #1630: Fix the old print schedules script
It's occasionally useful if you want to spool a printer headlessly.  This requires a local commit that I put on mit-prod, and putting the script back in the directory, plus some local changes that I have.

benjaminjkraft:
At some point the old `poll_schedules.sh` script got deleted in favor of the new interface.  While the new one is better, occasionally it's useful to be able to print headlessly, which the new interface doesn't do by construction.  This adds the old script back, with some slight modifications, along with a change to the schedule-printer module to make it work.  Fixes #1630.


mgersh:
This worked for Spark. LGTM.

Issue #1695, opened 2015-05-27
Collapse announcements area when there are no announcements


luac:
See the blank space at the top of https://esp.mit.edu/teach/index.html. Screenshot:
![announcements](https://cloud.githubusercontent.com/assets/1305639/7846143/d27da23e-0489-11e5-9cc8-cade1f1960d2.png)

Requested by MIT ESP: "The desired behaviour, I think, would be that it goes away or shrinks when there is no message."

May be specific to fruitsalad theme, I don't know what announcements look like on other themes.

PR #1705, opened 2015-06-07
Changed fruitsalad theme to hide annoucements box if there are no ann…

jerrywu64:
…ouncements

Addresses #1695. However, it looks like MIT is using template overrides; the way I tested it is I pasted the modified file into the template override for main.html and it works.

Also, I feel like it would look better if we had a header, so similar to the program pages, instead of just not having anything there, but I don't know what to put so I left a comment there instead.

### DIFF
--- a/esp/esp/themes/theme_data/fruitsalad/templates/main.html
+++ b/esp/esp/themes/theme_data/fruitsalad/templates/main.html
@@ -143,9 +143,6 @@
 <div class="corners"><div class="ul"></div><div class="ur"></div></div>
 {% if request.path|equal:"/teach/index.html" or request.path|equal:"/learn/index.html" %}
 
-{# announcements #}
-<div id="divannouncements">
-<ul>
 {% load preview %}
 
 {% if request.path|equal:"/teach/index.html" %}
@@ -153,6 +150,13 @@
 {% else %}
     {% miniblog_for_user AnonymousUser as announcements learn 3 %}
 {% endif %}
+
+{% if not announcements.announcementList|length_is:0 %}
+
+{# announcements #}
+<div id="divannouncements">
+<ul>
+
 
 {% for announce in announcements.announcementList %}
        <li>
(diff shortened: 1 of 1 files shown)


benjaminjkraft:
(Note that MIT's template override on `main.html` is just the autogenerated one from themes.)

esp/esp/themes/theme_data/fruitsalad/templates/main.html
```diff
@@ -143,9 +143,6 @@
 <div class="corners"><div class="ul"></div><div class="ur"></div></div>
 {% if request.path|equal:"/teach/index.html" or request.path|equal:"/learn/index.html" %}
 
-{# announcements #}
-<div id="divannouncements">
-<ul>
 {% load preview %}
 
 {% if request.path|equal:"/teach/index.html" %}
```
benjaminjkraft: I think this should also be inside the conditional, since I think it's supposed to act similarly to announcements, although it's not used anymore (at least not by MIT).

benjaminjkraft: Oh wait I read more carefully now, nevermind.

esp/esp/themes/theme_data/fruitsalad/templates/main.html
```diff
@@ -175,6 +179,9 @@
   <a href="/myesp/home/">more announcements...</a></div>
 {% endif %}
 {% else %}
+<!-- Default blank if there's no announcements -->
+{% endif %}
+{% else %}
 <h1 class="title">{% block content_title %}&nbsp;{% endblock %}</h1>
```
jmoldow: If you move this line above ``{% if request.path|equal:"/teach/index.html" or request.path|equal:"/learn/index.html" %}`` on line 144, then teach/index.html and learn/index.html will display their QSD page titles. Then, when the announcements area is hidden, the page will look exactly the same as all other QSD pages.

jmoldow:
> Also, I feel like it would look better if we had a header, so similar to the program pages, instead of just not having anything there, but I don't know what to put so I left a comment there instead.

I left a comment regarding this.

Otherwise, this change looks good to me.

jerrywu64:
Done, also removed some empty else clauses. I think the expected behavior if there are announcements is that they'll appear under the header?

jmoldow:
Well, the expected behavior used to be "there is no title on those two pages". The new expected behavior can be whatever we want. I thought it made sense for the title to be on top of the announcements, so that's why I suggested what I did. Your change looks good.

PR #1706, opened 2015-06-08
Extra student schedule info

tsutton:
This does two things:
- makes a utility method, end_time, in ClassSection, which returns the Event with the latest ending time among the meeting times of the section
- passes an extra parameter per student, student.last_classes, to the student class schedule template. This parameter is a list of classes, one per day from the program, of the student's last class on that day.

There is also some template override work I did which displays the last classes, which isn't recorded here anywhere but I can put if it would be useful.


esp/esp/program/models/class_.py
```diff
@@ -699,6 +698,12 @@ def assign_start_time(self, first_event):
         if availability:
             for room in current_rooms:
                 self.assign_room(room)
+
+    def end_time(self):
+        """Returns the meeting time for this section with the latest end time"""
+        all_times = self.meeting_times.order_by("-end")
+        if all_times: return all_times[0]
+        else: return []
```
luac: When do we have sections with no meeting_times? It seems like if we hit this case it's a type error later on anyway.

tsutton: Sections have no meeting_times when they aren't scheduled (for example because they are rejected). If they aren't scheduled, they shouldn't end up on student schedules, so that shouldn't be a problem for the other code in this branch.

luac:
Looks like this conflicts with #1678.

luac:
I pushed a fix for the merge conflicts. I was hesitating because I wasn't sure how it would interact with the performance improvements from #1678, although looking at it closer my guess is that it won't affect much.

esp/esp/program/modules/handlers/programprintables.py
```diff
@@ -922,6 +922,20 @@ def get_student_schedules(request, students, prog, extra='', onsite=False):
             # get list of valid classes
             classes = classes_by_student[student.id]
 
+            #get the student's last class on each day
+            last_classes = []
+            days = {}
+            for cls in classes:
+                date = cls.end_time().end.date().isocalendar()
```
benjaminjkraft: Wait, is `ClassSection.end_time()` a function that exists?  I can't find it.

benjaminjkraft:
I think it will actually cause performance problems -- you're going to end up running a query per class for the meeting_times.  I think it should do something similar to what #1678 did with `start_time_prefetchable` instead.  If you don't want to do this, we can make a separate issue for it and merge this until such time.

luac:
Added `end_time_prefetchable`, and resolved another merge conflict, this time against 936175099c6d7b78b4da2b972019f98dfddee2cf.

luac:
(PSA since I just learned this: you can get a 3-way diff for a merge conflict using `git diff [commit] [commit]^1 [commit]^2`: http://stackoverflow.com/questions/16573555/show-conflict-diff-part-of-a-merge)

benjaminjkraft:
Thanks @luac.  This looks good to me.

Issue #1675, opened 2015-04-20
Show is_superuser field in ESPUser admin panel


jmoldow:
The ESPUser admin panel shows most of the fields, including is_active and is_staff, but is missing is_superuser. That field should also be included.

This makes it hard to add/remove administrative permissions. If someone doesn't know about the makeadmin page, and tries to give permissions via the admin panel, they won't know that the user doesn't have is_superuser, and it'll look like the account was successfully given admin privileges, when it is actually missing something important (without is_superuser, you can't see anything on the admin panel). And when removing an admin, with the is_superuser field missing, it's not possible to fully revoke admin privileges, but again this won't be apparent to the user.

mgersh:
Fixed in #1708

PR #1708, opened 2015-06-08
Added is_superuser back to ESPUser on admin panel

jerrywu64:
There was a comment in the code that we don't use is_superuser, but #1675 disagrees. I haven't bothered checking either way.


mgersh:
Yeah, our own code doesn't really use it but apparently Django's does. Looks good, merging.

Issue #1658, opened 2015-03-28
Teacher check-in fails on accounts that are Administrators but not Teachers


mgersh:
It won't let you check in a teacher who isn't a Teacher. Instead it gives an error message that they're not a teacher for this program. We probably don't want to remove the ability for Administrator accounts to go through teacher reg, so ideally teacher check-in should work for them too to be consistent. Student check-in probably has the same issue.

This was discovered at Northeastern, where some admins had accounts that were only Administrators for some reason.

jerrywu64:
When would we want an admin account to be checked in as a student?

I note that onsitecheckinmodule.rapidcheckin() explicitly checks to make sure the student isn't also a teacher. I don't know the exact rationale for this but I also don't see a reasonable use case in which this would be bad. Presumably if we insist upon testing student check-in, for example, we should use a student non-admin account. Unless there are plans to allow admins to register as students and take classes?

mgersh:
On second thought, yeah, I don't know why we would want to check in a non-student as a student, so that one is probably fine.

mgersh:
Fixed in #1709

PR #1709, opened 2015-06-08
Added check so that admin accounts which are not teacher accounts can…

jerrywu64:
… be teacher-checked-in.

Addresses #1658. I didn't change the behavior of the student check-in module, however; see the discussion on #1658.


mgersh:
Why add another check instead of removing it? If they're registered to teach, I think we always want them to be able to check in, even if something really weird happened to allow that in the first place.

jerrywu64:
Good point, changed.

Issue #1644, opened 2015-03-23
"view on site" buttons in the admin panel


luanthe:
Requested at MIT Spark wrap-up: ""The "view on site" button on the admin panel for a user doesn't seem to work (https://esp.mit.edu/admin/users/espuser/60201/).""

luac:
It also may (or may not) be useful to add more of these for other models; I think it just involves defining a `get_url()` method or some such on the model.

jmoldow:
++

Though only for models where it would make sense. We have very few models that have canonical urls on the website.

jerrywu64:
Darn this is annoying, I'm pretty sure I could get this to work if we were on something more recent than Django 1.4. 

If other people want to look into this, see #1710 for examples of what supposedly you can do. Things I found out:
1) The way Django 1.4 does it is that if your model has the get_absolute_url attribute, the View on site button appears, and goes to a redirect page; see https://github.com/django/django/blob/stable/1.4.x/django/contrib/admin/templates/admin/change_form.html for the relevant template. I think the redirect is supposed to append the result of get_absolute_url to whatever is in Sites, but this could be false.
2) The User class in django.contrib.auth.models.py which ESPUser is extending comes with a get_absolute_url() but for some reason my attempts to override it did not work. Maybe I was doing something stupid, you can try it yourself.

benjaminjkraft:
Yeah, after a bunch of source-diving, it looks like the way it works it ends up converting the model to a ContentType and back, which effectively strips off the proxy model.  The easy solution to this is probably to make the `/users/.../` URL that it redirects to just be another redirect to our userview page.  Or it might work differently with newer django.

PR #1710, opened 2015-06-09
Added 'View on site' buttons for various models on the admin panel

jerrywu64:
Relevant to #1644 although I couldn't actually get it to work for ESPUser. See discussion there.

Note that for me the link causes me to leave my devserver, so be careful not to break anything if you test this. I think it's appending the specified url to whatever is in Sites but I'm not sure.

There also might be other places we could add these links, so feel free to suggest.


benjaminjkraft:
Why doesn't it work for ESPUser?  Does it not work to override `get_absolute_url()`?  I don't see relevant discussion on #1644.

jerrywu64:
Yeah, I was still typing it.

benjaminjkraft:
Got it.  LGTM, merging.  The only other place I can think to add it is surveys (linking to the manage surveys page), but idk if that would ever get used.

Issue #381, opened 2012-07-03
Class Change Request Form Module


jmoldow:
The class change request form currently exists in the website, but it is not a proper program module. We should change this so we don't have to hard-code it into urls.py anymore.

PR #1711, opened 2015-06-13
Turn the class changes request form into a program module

mgersh:
I mostly just moved the code and didn't attempt serious cleanup. Actual changes:
* Deleted some code which reinvents things we have decorators for
* Fixed an import error introduced in #1223 by switching to the equivalent method in ESPUser
* Fixed a bug which has apparently always been there: if a student submits the form and then resubmits it later, their old requests would get expired, and new requests for the same classes wouldn't get created even if they were still requesting them

Fixes issue #381.


mgersh:
Whoops, looks like I got distracted and left some things half-implemented here. I'm also not sure if that test failure was Travis being sad or an actual bug. Looking at those now.

mgersh:
Okay, ran tests locally and got the same result, and also that it was definitely introduced by this code. I'm not having any success making sense of the error, though.

jmoldow:
Looks good. Thanks for doing this!

PR #1712, opened 2015-06-22
Ticket 1454

ghost:


esp/esp/program/modules/handlers/admincore.py
```diff
@@ -213,13 +213,12 @@ def deadline_management(self, request, tl, one, two, module, extra, prog):
         context = {}
 
         #   Set a flag on each perm for whether it has ended
+        current_date = datetime.now()
         for perm in perms:
-            if perm.end_date is None or  perm.end_date > datetime.now():
-                perm.open_now = True
-            else:
-                perm.open_now = False
+            did_start = perm.start_date <= current_date
```
benjaminjkraft: This should also allow `None`, since that nominally means "has always been open", although I think it is used less.  But also, I think you can just use `perm.is_valid()` instead (it's inherited from `ExpirableModel`).

joeleg: Ok, will make the change - thanks for the suggestion.

benjaminjkraft:
Pushed some fixes, and tested; if someone could take a quick glance and merge, that would be great.

PR #1715, opened 2015-06-23
Remove catalog sort by _num_students for performance reasons

mgersh:
This is removing a feature that's still in use, but I think it's a good tradeoff. The catalog query is an abomination, it gets run all the time during FCFS student reg, and removing _num_students from it helps a lot: down from 2000-3300ms to 1100-1300ms on a really big program on my dev server. Having this sorting seems relatively unimportant compared to having FCFS be less slow. If we do remove it, we'd have to edit the tags on the bc and chicago sites (which appear to just be copies of what the default sort order was before #1576 was fixed).

Thoughts?

### DIFF
--- a/esp/esp/program/models/class_.py
+++ b/esp/esp/program/models/class_.py
@@ -213,16 +213,12 @@ class ClassManager(Manager):
         #   Retrieve the content type for finding class documents (generic relation)
         content_type_id = ContentType.objects.get_for_model(ClassSubject).id
         
-        select = SortedDict([( '_num_students', 'SELECT COUNT(DISTINCT "program_studentregistration"."user_id") FROM "program_studentregistration", "program_classsection" WHERE ("program_studentregistration"."relationship_id" = %s AND "program_studentregistration"."section_id" = "program_classsection"."id" AND "program_classsection"."parent_class_id" = "program_class"."id" AND ("program_studentregistration"."start_date" IS NULL OR "program_studentregistration"."start_date" <= %s) AND ("program_studentregistration"."end_date" IS NULL OR "program_studentregistration"."end_date" >= %s))'),
-                             ('teacher_ids', 'SELECT list(DISTINCT espuser_id) FROM program_class_teachers Where program_class_teachers.classsubject_id=program_class.id'),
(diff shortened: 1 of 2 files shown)


luac:
Another option is to sort on the DerivedField `enrolled_students`, which should be the same thing as `_num_students`*, and since it's a real DB field it might be faster to query that instead.

(I think `enrolled_students` is sometimes inconsistent with `num_students` due to caching race conditions or something, but for this purpose it might be better than nothing (assuming I'm right about it being not-slow)?)

benjaminjkraft:
Ooh, good idea.  I think using enrolled_students might be a good idea if we want to sort the catalog to show classes with space first; but in any case that should be both easy to query and not require a mess of custom SQL and special handling.

(The code looks fine to me; I haven't tested it yet though.)

mgersh:
That wouldn't just work as-is, `enrolled_students` is a field on ClassSection and here we're trying to sort ClassSubjects.

benjaminjkraft:
Ah.  It should still be reasonably possible to do an aggregate to the ClassSection to grab the total enrollment, and I think that would be a good deal faster than the subquery it's doing now.

mgersh:
Result: performance of the enrolled_students version is about halfway between this version and the original. (Exact numbers appear to be different today than they were last night. Performance testing is hard.)

I'm also not at all sure that this even makes sense as a heuristic for showing classes with space first. In my experience, smaller classes tend to fill up and classes with tons of students tend to be huge classes with space for even more, so the sort should go the opposite of the direction it's currently going.

benjaminjkraft:
Yeah, probably if we wanted to do that the sort would end up subtracting from total capacity or something.  But in any case, the way it is now is not that useful, so any alternate functionality can be a separate thing.

mgersh:
So, summary: my preference is for the version currently in this PR. If people want to look at the other version, I can make a separate PR with that.

mgersh:
Update: I was wrong earlier, because performance testing is hard. When checking more carefully, the enrolled_students version is basically just as fast as this. Closing this and making a new PR with the other one.

PR #1717, opened 2015-06-23
Change catalog _num_students subquery to a faster aggregate

mgersh:
See discussion on #1715. This version doesn't remove any features, it just makes it faster.


benjaminjkraft:
This looks great -- one more line of custom SQL defeated!

Issue #1687, opened 2015-05-02
Scheduling checks: a way to hide false positives


luac:
This could be an alternative resolution for issues such as those discussed in #1596, #1597.

mgersh:
Implementation idea: each item has something you can click on to hide it, which happens entirely client-side. It doesn't persist across refreshes, but it probably reduces the need for refreshing, since you can click to hide something once you're done handling it instead of refreshing to clear out clutter.

benjaminjkraft:
I like this idea.  It might not actually be that hard to persist it across reloading a single check without refreshing the page.  In theory we could persist beyond that using cookies, but that might be more work.  One disadvantage is that you can't share with someone else, so that the same person has to be keeping track of the same check.  But that's usually what happens anyway.

benjaminjkraft:
Another thing: we should make sure to have an unhide button.  Or, it might be easier to just put the hidden ones in very light gray, so that they are still readable, but not as obvious.  Or it might be better to just not try to keep beyond reloads for this reason.

PR #1721, opened 2015-07-02
React-ify scheduling check output tables

jerrywu64:
Addresses #1687.

Some notes (some of which are mentioned in in-code comments):
I didn't touch HTMLSCFormatter. It looks to me like there isn't actually any code which uses it (except for scheduling checks obviously except those now use JSONFormatter), but idk if it'll be useful in the future.
JSONFormatter possibly contains redundant code, but maybe the code is useful for making sure things are formatted properly-- I didn't check preconditions/postconditions carefully enough to tell.
I'm using JSON.parse() which apparently isn't supported in old browsers, idk if we care about that for a management page.
Changes in public/media/styles/scheduling_checks.css are needed, but that's in .gitignore; stuff I added is below. (My devserver originally didn't even have this file, so I don't know what other people's  scheduling_checks.css files look like or if this conflicts.)
Scheduling check tables now enable sorting (upon header click) and greying-out of a row (upon row click, *provided that you make the appropriate change in scheduling_checks.css*. I didn't add any buttons to this effect, though (in my css I did do cursor:pointer-- although it's for the whole table so this is kind of silly), and the sorting can be kinda sketchy.
Sorting and greying does not persist across closing and reopening the scheduling check, let alone refreshing.

Additions to scheduling_checks.css:
table { border-spacing: 0; }
td { padding: 10px; cursor: pointer; }
th { cursor: pointer; }
.rowGreyed {
color: grey;
}
.headerSelected {
  color: blue;
}

People probably have opinions on the changes made, so feel free to comment. Although I note that if you want slightly different behavior upon clicking a row, if it's doable in css, you can probably just do it in css on your end.

### DIFF
--- a/esp/esp/program/modules/handlers/schedulingcheckmodule.py
+++ b/esp/esp/program/modules/handlers/schedulingcheckmodule.py
@@ -47,68 +47,6 @@ class RawSCFormatter:
     def format_list(self, l, options={}, help_text=""):
         return l
 
-class HTMLSCFormatter: # This might be dead, but idk maybe someone still wants it
-    #requires: d, a two level dictionary where the the first set of
-    #   keys are the headings expected on the side of the table, and
-    #   the second set are the headings expected on the top of the table
-    def format_table(self, d, options={}, help_text=""):
-        if isinstance(d, list):
-            return self._format_list_table(d, options['headings'], help_text=help_text)
-        else:
-            return self._format_dict_table(d, options['headings'], help_text=help_text)
-
-    def format_list(self, l, help_text=""):
-        output = self._table_start(help_text)
-        for row in l:
-            output += self._table_row([row])
-        output += "</table>"
-        return output
-
-    def _table_start(self, help_text=""):
-        output = ''
-        if help_text:
-             output += '<div class="help-text">%s</div>' % help_text
-        output += '<table cellpadding=10>'
-        return output
-
-    def _table_headings(self, headings):
-        #column headings
-        next_row = ""
-        for h in headings:
-            next_row = next_row + "<th><div style=\"cursor: pointer;\">" + str(h) + "</div></th>"
-        next_row = next_row + "</tr></thread>"
-        return next_row
-
-    def _table_row(self, row):
-        next_row = ""
-        for r in row:
-            #displaying lists is sometimes borked.  This makes it not borked
-            if isinstance(r, list):
-                r = [str(i) for i in r]
-            next_row += "<td>" + str(r) + "</td>"
-        next_row += "</tr>"
-        return next_row
-        
-    def _format_list_table(self, d, headings, help_text=""):
-        output = self._table_start(help_text)
-        output = output + self._table_headings(headings)
-        for row in d:
-            ordered_row = [row[h] for h in headings]
-            output = output + self._table_row(ordered_row) 
-        output = output + "</table>"
-        return output
-
-    def _format_dict_table(self, d, headings, help_text=""):
-        headings = [""] + headings[:]
-        output = self._table_start(help_text)
-        output = output + self._table_headings(headings)
-
-        for key, row in sorted(d.iteritems()):
-            ordered_row = [row[h] for h in headings if h]
-            output = output + self._table_row([key] + ordered_row)
-        output += "</table>"
-        return output
-
 # Builds JSON output for an object with attributes help_text, headings, and body.
 class JSONFormatter:
     #requires: d, a two level dictionary where the the first set of
--- a/esp/esp/program/modules/handlers/schedulingcheckmodule.py
+++ b/esp/esp/program/modules/handlers/schedulingcheckmodule.py
@@ -64,18 +64,8 @@ class JSONFormatter:
         output["headings"] = [] # no headings
         
         # might be redundant, but it makes sure things aren't in a weird format
-        body = []        
-        for row in l:
-            body.append(self._table_row([row]))
-        output["body"] = body
+        output["body"] = [self._table_row([row]) for row in l]
         return json.dumps(output)
-
-    def _table_headings(self, headings): # in case headings are a dict
-        #column headings
-        next_row = []
-        for h in headings:
-            next_row.append(str(h))
-        return next_row
 
     def _table_row(self, row):
         next_row = []
@@ -89,25 +79,16 @@ class JSONFormatter:
     def _format_list_table(self, d, headings, help_text=""): #needs verify
         output = {}
         output["help_text"] = help_text
-        output["headings"] = self._table_headings(headings)
-        body = []
-        for row in d:
-            ordered_row = [row[h] for h in headings]
-            body.append(self._table_row(ordered_row)) # maybe?
-        output["body"] = body
+        output["headings"] = map(str, headings)
+        output["body"] = [self._table_row([row[h] for h in headings]) for row in d]
         return output
 
     def _format_dict_table(self, d, headings, help_text=""): #needs verify
         headings = [""] + headings[:]
         output = {}
         output["help_text"] = help_text        
-        output["headings"] = self._table_headings(headings)
-        
-        body = []
-        for key, row in sorted(d.iteritems()):
-            ordered_row = [row[h] for h in headings if h]
-            body.append(self._table_row([key] + ordered_row)) #maybe
-        output["body"] = body
+        output["headings"] = map(str, headings)
+        output["body"] = [self._table_row([key] + [row[h] for h in headings if h]) for key, row in sorted(d.iteritems())]
         return output
         
 class SchedulingCheckRunner:
--- a/esp/esp/program/modules/handlers/schedulingcheckmodule.py
+++ b/esp/esp/program/modules/handlers/schedulingcheckmodule.py
@@ -38,7 +38,7 @@ class SchedulingCheckModule(ProgramModuleObj):
     class Meta:
         proxy = True
 
-#For formatting output.  The default is to use HTMLSCFormatter, but someone writing a script
+#For formatting output.  The default is to use JSONFormatter, but someone writing a script
 #may want to use RawSCFormatter to get the original data structures
 class RawSCFormatter:
     def format_table(self, l, options={}, help_text=""):
@@ -47,7 +47,7 @@ class RawSCFormatter:
     def format_list(self, l, options={}, help_text=""):
         return l
 
-class HTMLSCFormatter:
+class HTMLSCFormatter: # This might be dead, but idk maybe someone still wants it
     #requires: d, a two level dictionary where the the first set of
     #   keys are the headings expected on the side of the table, and
     #   the second set are the headings expected on the top of the table
@@ -109,9 +109,72 @@ class HTMLSCFormatter:
         output += "</table>"
         return output
 
+# Builds JSON output for an object with attributes help_text, headings, and body.
+class JSONFormatter:
+    #requires: d, a two level dictionary where the the first set of
+    #   keys are the headings expected on the side of the table, and
+    #   the second set are the headings expected on the top of the table
+    def format_table(self, d, options={}, help_text=""):
+        if isinstance(d, list):
+            return json.dumps(self._format_list_table(d, options['headings'], help_text=help_text))
+        else:
+            return json.dumps(self._format_dict_table(d, options['headings'], help_text=help_text))
+
+    def format_list(self, l, help_text=""): # needs verify
+        output = {}
+        output["help_text"] = help_text
+        output["headings"] = [] # no headings
+        
+        # might be redundant, but it makes sure things aren't in a weird format
+        body = []        
+        for row in l:
+            body.append(self._table_row([row]))
+        output["body"] = body
+        return json.dumps(output)
+
+    def _table_headings(self, headings): # in case headings are a dict
+        #column headings
+        next_row = []
+        for h in headings:
+            next_row.append(str(h))
+        return next_row
+
+    def _table_row(self, row):
+        next_row = []
+        for r in row:
+            #displaying lists is sometimes borked.  This makes it not borked
+            if isinstance(r, list):
+                r = [str(i) for i in r]
+            next_row.append(str(r))
+        return next_row
+        
+    def _format_list_table(self, d, headings, help_text=""): #needs verify
+        output = {}
+        output["help_text"] = help_text
+        output["headings"] = self._table_headings(headings)
+        body = []
+        for row in d:
+            ordered_row = [row[h] for h in headings]
+            body.append(self._table_row(ordered_row)) # maybe?
+        output["body"] = body
+        return output
+
+    def _format_dict_table(self, d, headings, help_text=""): #needs verify
+        headings = [""] + headings[:]
+        output = {}
+        output["help_text"] = help_text        
+        output["headings"] = self._table_headings(headings)
+        
+        body = []
+        for key, row in sorted(d.iteritems()):
+            ordered_row = [row[h] for h in headings if h]
+            body.append(self._table_row([key] + ordered_row)) #maybe
+        output["body"] = body
+        return output
+        
 class SchedulingCheckRunner:
 #Generate html report and generate text report functions?lingCheckRunner:
-     def __init__(self, program, formatter=HTMLSCFormatter()):
+     def __init__(self, program, formatter=JSONFormatter()):
           """
           high_school_only and lunch should be lists of indeces of timeslots for the high school
           only block and for lunch respectively
--- a/esp/esp/program/modules/handlers/schedulingcheckmodule.py
+++ b/esp/esp/program/modules/handlers/schedulingcheckmodule.py
@@ -67,13 +67,17 @@ class JSONFormatter:
         output["body"] = [self._table_row([row]) for row in l]
         return json.dumps(output)
 
+
     def _table_row(self, row):
         next_row = []
         for r in row:
             #displaying lists is sometimes borked.  This makes it not borked
             if isinstance(r, list):
                 r = [str(i) for i in r]
-            next_row.append(str(r))
+            if isinstance(r, int):
+                next_row.append(r)
+            else:
+                next_row.append(str(r))
         return next_row
         
     def _format_list_table(self, d, headings, help_text=""): #needs verify
--- a/esp/public/media/scripts/program/modules/scheduling_checks.jsx
+++ b/esp/public/media/scripts/program/modules/scheduling_checks.jsx
@@ -39,7 +39,7 @@ var SchedulingCheck = React.createClass({
       timestamp: "never",
       tableState: {
         greyed: {},
-        sort: false,
+        sort: -1,
         reverse: false
       }
     };
@@ -47,6 +47,18 @@ var SchedulingCheck = React.createClass({
   
   updateTableState: function (state) {
     this.state.tableState = state;
+  },
+  
+  resetTable: function () {
+    this.setState({
+      tableState: {
+        greyed: {},
+        sort: -1,
+        reverse: false
+      },
+      open: false
+      });
+    
   },
 
   handleClick: function () {
@@ -125,6 +137,7 @@ var SchedulingCheck = React.createClass({
       <div className="scheduling-check-title">
         <span onClick={this.handleClick}>{this.props.title}</span>
         <RefreshButton onClick={this.loadData} />
+        <ResetButton onClick={this.resetTable} />
       </div>
       <div className="scheduling-check-body">
         {body}
@@ -148,6 +161,20 @@ var RefreshButton = React.createClass({
   },
 });
 
+/**
+ * Calls its onClick prop to reset table greying/sorting
+ */
+var ResetButton = React.createClass({
+  propTypes: {
+    onClick: React.PropTypes.func.isRequired,
+  },
+
+  render: function () {
+    return <button onClick={this.props.onClick} className="reset-button">
+      Reset
+    </button>;
+  },
+});
 
 // Modified from react-json-table example code.
 var SelectTable = React.createClass({
@@ -174,9 +201,7 @@ var SelectTable = React.createClass({
     // clone the rows
     items = this.props.rows.slice();
     
-    items = _.sortBy(items, function( item ){
-         return item[ me.state.sort ];
-      });
+    items = _.sortBy(items, me.state.sort);
     
     if (this.state.reverse) items.reverse();
     
--- a/esp/public/media/scripts/program/modules/scheduling_checks.jsx
+++ b/esp/public/media/scripts/program/modules/scheduling_checks.jsx
@@ -16,6 +16,15 @@ var SchedulingCheckList = React.createClass({
  *
  * This might or might not be loaded yet; clicking the heading will load the
  * data from the server and expand it.
+ * 
+ * There is supposedly a functionality in which clicking on a table row will cause it
+ * to be greyed out, however whether this actually happens depends on your
+ * scheduling_checks.css (clicking could possibly do nothing, or possibly
+ * do other things).
+ *
+ * Likewise, clicking on a table header will sort by that column. You can also
+ * add formatting in scheduling_checks.css so that the selected header will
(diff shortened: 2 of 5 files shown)


esp/esp/program/modules/handlers/schedulingcheckmodule.py
```diff
@@ -47,7 +47,7 @@ def format_table(self, l, options={}, help_text=""):
     def format_list(self, l, options={}, help_text=""):
         return l
 
-class HTMLSCFormatter:
+class HTMLSCFormatter: # This might be dead, but idk maybe someone still wants it
```
benjaminjkraft: I think it's specific enough to this context, and simple enough, that you can just nuke it.  If anyone else wants it they can just grab it from git.

jerrywu64:
Sorting and greying should now persist across hide/unhide and the refresh button, but not refreshing the page itself. This includes if you press the refresh button and something changes. 

Expected behavior is that if the data in a row changes, it reverts to the default state (e.g. not greyed if it was greyed before). 

However, I also expect that if for whatever reason a scheduling check has two duplicate rows, it'll be impossible to apply different formatting to them, i.e. if you try to grey one, it'll also grey the other. Do people think this is an issue?

esp/public/media/scripts/program/modules/scheduling_checks.jsx
```diff
@@ -80,11 +93,32 @@ var SchedulingCheck = React.createClass({
     } else if (!this.state.data) {
       body = <div className="placeholder">loading...</div>;
     } else {
+      var data = JSON.parse(this.state.data); // Might not work on old browsers
+      var table;
+      if (data.headings.length == 0) {
+        var settings = {
+          header: false
+        };
+      table = <SelectTable rows = {data.body} settings = {settings} saveState = {this.state.tableState} />;
```
benjaminjkraft: This should be indented one more.

esp/public/media/scripts/program/modules/scheduling_checks.jsx
```diff
@@ -114,3 +148,71 @@ var RefreshButton = React.createClass({
     </button>;
   },
 });
+
+
+// Modified from react-json-table example code.
+// Required props: rows, saveState (with sort and greyed attributes), settings (with header attribute)
```
benjaminjkraft: Putting this in a PropTypes will make React automatically check it for you, and also make it more readable.

esp/public/media/scripts/program/modules/scheduling_checks.jsx
```diff
@@ -80,11 +93,32 @@ var SchedulingCheck = React.createClass({
     } else if (!this.state.data) {
       body = <div className="placeholder">loading...</div>;
     } else {
+      var data = JSON.parse(this.state.data); // Might not work on old browsers
+      var table;
+      if (data.headings.length == 0) {
+        var settings = {
```
benjaminjkraft: I'd just inline this in the `<SelectTable>`, I think that makes it easier to see what's happening.  I'd also recommend just passing a single header attribute, instead of a settings object, since there is just one setting.  If the number later becomes unmanageable, you can cross that bridge then.

esp/public/media/scripts/program/modules/scheduling_checks.jsx
```diff
@@ -80,11 +93,32 @@ var SchedulingCheck = React.createClass({
...
+          header: false
+        };
+      table = <SelectTable rows = {data.body} settings = {settings} saveState = {this.state.tableState} />;
+      } else {
+        var columns = [];
+        for (i = 0; i < data.headings.length; i++) {
+          if (!!data.headings[i]) {
```
benjaminjkraft: Why the double-negation?

esp/public/media/scripts/program/modules/scheduling_checks.jsx
```diff
@@ -114,3 +148,71 @@ var RefreshButton = React.createClass({
...
   },
 });
+
+
+// Modified from react-json-table example code.
+// Required props: rows, saveState (with sort and greyed attributes), settings (with header attribute)
+var SelectTable = React.createClass({
+  getInitialState: function(){
+    // We will store the sorted column and whether each row is greyed out
+    return {sort: this.props.saveState.sort, greyed : this.props.saveState.greyed};
```
benjaminjkraft: Minor thing: your spacing around the colons is inconsistent.

esp/public/media/scripts/program/modules/scheduling_checks.jsx
```diff
@@ -28,6 +37,10 @@ var SchedulingCheck = React.createClass({
       open: false,
       failed: false,
       timestamp: "never",
+      tableState: {
```
benjaminjkraft: Why is this in state?  It doesn't seem like it ever gets modified.

esp/public/media/scripts/program/modules/scheduling_checks.jsx
```diff
@@ -114,3 +148,71 @@ var RefreshButton = React.createClass({
...
+    // We will store the sorted column and whether each row is greyed out
+    return {sort: this.props.saveState.sort, greyed : this.props.saveState.greyed};
+  },  
+  render: function(){
+    var me = this,
+        // clone the rows
+        items = this.props.rows.slice()
+    ;
+    // Sort the table
+    if( this.state.sort ){
```
benjaminjkraft: [`_.sortBy`](https://lodash.com/docs#sortBy) will do this in one line.

esp/public/media/scripts/program/modules/scheduling_checks.jsx
```diff
@@ -114,3 +148,71 @@ var RefreshButton = React.createClass({
...
+      items.sort( function( a, b ){
+         return a[ me.state.sort ] > b[ me.state.sort ] ? 1 : -1;
+      });
+    }
+      
+    if (this.props.columns == undefined) {
+      return <JsonTable
```
benjaminjkraft: Rather than writing out the whole thing twice, it [looks like](https://github.com/arqex/react-json-table/blob/master/rjt.js#L93) if you just pass the undefined through, things should work correctly.

esp/public/media/scripts/program/modules/scheduling_checks.jsx
```diff
@@ -114,3 +148,71 @@ var RefreshButton = React.createClass({
...
+      />;
+    }
+  },
+  
+  getSettings: function(){
+      var me = this;
+      // We will add some classes to the selected rows and cells
+      return {
+        headerClass: function( current, key ){
+            if( me.state.sort == key )
```
benjaminjkraft: I'm a fan of explicit else clauses, but we don't have a style guide so you can ignore me if you want.

esp/public/media/scripts/program/modules/scheduling_checks.jsx
```diff
@@ -114,3 +148,71 @@ var RefreshButton = React.createClass({
...
+          if( me.state.greyed[item] )
+            return current + ' rowGreyed';
+          return current;
+        },
+        header: this.props.settings.header
+      };
+  },
+    
+  onClickHeader: function( e, column ){
+    this.props.saveState.sort = column;
```
benjaminjkraft: You shouldn't modify props, ever.  Anything that can change must go in state.

esp/public/media/scripts/program/modules/scheduling_checks.jsx
```diff
@@ -114,3 +148,71 @@ var RefreshButton = React.createClass({
...
+    this.setState( {sort: column} );
+  },
+  
+  onClickRow: function( e, item ){
+    this.state.greyed[item] = !this.state.greyed[item];
```
benjaminjkraft: I think this is not a recommended approach, although it might work, it's better to actually put the change in the `setState` call.

esp/public/media/scripts/program/modules/scheduling_checks.jsx
```diff
@@ -114,3 +148,71 @@ var RefreshButton = React.createClass({
...
+    this.props.saveState.greyed[item] = this.state.greyed[item];
```
benjaminjkraft: Again, don't mutate props.

esp/templates/elements/html
```diff
@@ -46,6 +46,7 @@
 
     <script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/react/0.13.3/react.js"></script>
     <script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/react/0.13.3/JSXTransformer.js"></script>
+    <script type="text/javascript" src="https://rawgit.com/arqex/react-json-table/master/build/react-json-table.min.js"></script>
```
benjaminjkraft: Since this is only used on one page, I'd just put it there.  Also, I would include a specific commit rather than master, just in case there is at some point a breaking change.

esp/esp/program/modules/handlers/schedulingcheckmodule.py
```diff
@@ -47,71 +47,72 @@ def format_table(self, l, options={}, help_text=""):
...
     def _table_row(self, row):
-        next_row = ""
+        next_row = []
         for r in row:
             #displaying lists is sometimes borked.  This makes it not borked
             if isinstance(r, list):
                 r = [str(i) for i in r]
-            next_row += "<td>" + str(r) + "</td>"
-        next_row += "</tr>"
+            next_row.append(str(r))
```
benjaminjkraft: This is going to result in things having extra quotes, around them, no?  (`str(["a"]) == "['a']"`)

esp/esp/program/modules/handlers/schedulingcheckmodule.py
```diff
@@ -47,71 +47,72 @@ def format_table(self, l, options={}, help_text=""):
...
-        output += "</table>"
-        return output
-
-    def _table_start(self, help_text=""):
-        output = ''
-        if help_text:
-             output += '<div class="help-text">%s</div>' % help_text
-        output += '<table cellpadding=10>'
-        return output
+            body.append(self._table_row([row]))
```
benjaminjkraft: May be clearer to just do `output["body"] = [self._table_row([row]) for row in l]` and likewise in `_format_list_table`.

esp/esp/program/modules/handlers/schedulingcheckmodule.py
```diff
@@ -47,71 +47,72 @@ def format_table(self, l, options={}, help_text=""):
...
+        output["body"] = body
+        return json.dumps(output)
 
-    def _table_headings(self, headings):
+    def _table_headings(self, headings): # in case headings are a dict
```
benjaminjkraft: I would just remove this function: at this point it's literally `map(str,headings)` which you can just inline.

benjaminjkraft:
Cool!  I don't particularly care about the exact CSS; I'd suggest maybe making the gray a little lighter, but I don't feel strongly.  I left a bunch of inline comments on the code style, but I think it overall looks good.

> I'm using JSON.parse() which apparently isn't supported in old browsers, idk if we care about that for a management page.

That's fine, I don't think there are many admins using [IE 7](http://caniuse.com/#feat=json).

> Changes in public/media/styles/scheduling_checks.css are needed, but that's in .gitignore; stuff I added is below. (My devserver originally didn't even have this file, so I don't know what other people's scheduling_checks.css files look like or if this conflicts.)

You want to look at `esp/public/media/default_styles/scheduling_checks.css`, and edit that.  Most dev servers just link `esp/public/media/styles/` to that directory; some older setups have the directories actually copied.

> However, I also expect that if for whatever reason a scheduling check has two duplicate rows, it'll be impossible to apply different formatting to them, i.e. if you try to grey one, it'll also grey the other. (Although I haven't actually tested this.) Do people think this is an issue?

I can't think of a reason this would happen (other than someone forgetting a `.distinct()` in a query), so I think it's fine.

esp/esp/program/modules/handlers/schedulingcheckmodule.py
jerrywu64: Empirically, no (I think e.g. lunch blocks is an example where you would expect that to happen). I think it's because javascript is rendering each cell separately.

benjaminjkraft: Yes, it does happen.
> ['Sat Nov 22: 12:05 to 12:55 PM', 'Sat Nov 22: 1:05 to 1:55 PM']
> ['Sun Nov 23: 12:05 to 12:55 PM', 'Sun Nov 23: 1:05 to 1:55 PM']

But I guess that's existing behavior, and the alternative could be confusing, so it's fine.

esp/public/media/scripts/program/modules/scheduling_checks.jsx
jerrywu64: Oops, was trying to cast to boolean, but I guess it already did that.

mgersh:
The reason why @jerrywu64's dev server has a weird media setup is because it's fabric/Vagrant on Window, see https://github.com/learning-unlimited/ESP-Website/blob/main/esp/fabfile.py#L125

benjaminjkraft:
Ah, ok.  (Another reason we should get rid of this setup, but I digress.)  In that case, yeah, just actually copy the directories.

esp/public/media/scripts/program/modules/scheduling_checks.jsx
```diff
@@ -80,11 +93,32 @@ var SchedulingCheck = React.createClass({
     } else if (!this.state.data) {
       body = <div className="placeholder">loading...</div>;
     } else {
+      var data = JSON.parse(this.state.data); // Might not work on old browsers
+      var table;
+      if (data.headings.length == 0) {
```
benjaminjkraft: This case and the other two I can find in this diff will be okay, but it's safer to more or less always use `===` instead of `==` in JS.  (See the [`==` equality table](https://dorey.github.io/JavaScript-Equality-Table/).)

esp/public/media/scripts/program/modules/scheduling_checks.jsx
jerrywu64: The current operation is that modifying the prop will also modify the check's saveState, but I'll find a different mechanism if this isn't good practice.

benjaminjkraft: Yeah, the point of the way things work in React is to have the flow of data be clear, so e.g. props are a thing that you know never change.  If the parent component needs to know it, then it should be the one with the event handler, which it then passes down as a prop.  It requires a little boilerplate, but it makes it clear what's going on, and also allows React to be clever by assuming that props never change.

esp/public/media/scripts/program/modules/scheduling_checks.jsx
jerrywu64: (this is no longer the case)

esp/public/media/scripts/program/modules/scheduling_checks.jsx
```diff
@@ -114,3 +147,86 @@ var RefreshButton = React.createClass({
...
+    // We will store the sorted column and whether each row is greyed out
+    return {sort: this.props.saveState.sort, greyed: this.props.saveState.greyed, reverse: this.props.saveState.reverse};
+  },  
+  render: function(){
+    this.props.updateTableState(this.state);
+    var me = this;
+    // clone the rows
+    items = this.props.rows.slice();
+    
+    items = _.sortBy(items, function( item ){
```
benjaminjkraft: I believe you can actually do `_.sortBy(items, me.state.sort)` and it will automatically do the lookup.

jerrywu64:
Updates on css:

Something like
.reset-button {
     margin-left: auto;
     float: right;
}

If you're sorting a column in descending order, if you've added formatting for sorted columns it'll still use that, but you can also override the formatting, e.g.
.sortReversed {
  color: red;
}
(although this will have to go after the formatting for .headerSelected).

Also, note that sorting currently doesn't "stack": yes, we're using a stable sort, but if you sort first by one column then by a second, it'll use the original ordering as tiebreaker for the second column instead of the first ordering.

This could conceivably be fixed by saving the current set of rows in state, but reverse sorting may produce unwanted behavior, e.g. if you sort by column 1 and then by column 2, it'll give the reverse of sort-by-2-using-1-as-tiebreaker, which has the opposite tiebreaker order from sort-descending-by-2-using-1-as-tiebreaker. Or maybe that's the behavior people want.

benjaminjkraft:
Can you put the css into `esp/public/media/default_styles/scheduling_checks.css`?  If you then copy that into `esp/public/media/styles` locally it should work the way it will on an actual server.

For multicolumn sorting, I would just not worry about it for now.

esp/public/media/scripts/program/modules/scheduling_checks.jsx
```diff
@@ -28,8 +37,29 @@ var SchedulingCheck = React.createClass({
...
       timestamp: "never",
+      tableState: {
+        greyed: {},
+        sort: -1,
+        reverse: false
+      }
     };
   },
+  
+  updateTableState: function (state) {
```
benjaminjkraft: You want to use `this.setState`, not modify `this.state` directly.

esp/public/media/scripts/program/modules/scheduling_checks.jsx
```diff
@@ -114,3 +160,98 @@ var RefreshButton = React.createClass({
...
+    columns: React.PropTypes.array,
+    updateTableState: React.PropTypes.func.isRequired
+  },
+    
+  getInitialState: function(){
+    // We will store the sorted column and whether each row is greyed out
+    return {sort: this.props.saveState.sort, greyed: this.props.saveState.greyed, reverse: this.props.saveState.reverse};
+  },  
+  render: function(){
+    this.props.updateTableState(this.state);
```
benjaminjkraft: I think this might cause problems -- while rendering the SelectTable, it's going to modify the state of the parent, which will force another rerender.  This also really shouldn't be necessary in the first place -- why do you think you need it?

esp/public/media/scripts/program/modules/scheduling_checks.jsx
```diff
@@ -114,3 +160,98 @@ var RefreshButton = React.createClass({
...
+    var me = this;
+    // clone the rows
+    items = this.props.rows.slice();
+    
+    items = _.sortBy(items, me.state.sort);
```
benjaminjkraft: You can now reference `this` directly.

esp/public/media/scripts/program/modules/scheduling_checks.jsx
```diff
@@ -114,3 +160,98 @@ var RefreshButton = React.createClass({
...
+            return current + ' rowGreyed';
+          } else {
+            return current;
+          }
+        },
+        header: this.props.header
+      };
+  },
+    
+  onClickHeader: function( e, column ){
```
benjaminjkraft: These two handlers are where you want to pass the state updates upwards, not in `render`.

benjaminjkraft:
Thanks.  I left a few more comments -- mostly just some issues with how you're modifying state.  Sorry this is taking so many rounds -- I know React can be a bit tricky to get used to, but I promise most of the constraints exist for good reason.

esp/public/media/scripts/program/modules/scheduling_checks.jsx
```diff
@@ -28,8 +37,29 @@ var SchedulingCheck = React.createClass({
...
+  },
+  
+  resetTable: function () {
+    this.setState({
+      tableState: {
+        greyed: {},
+        sort: -1,
+        reverse: false
+      },
+      open: false
```
benjaminjkraft: I find it confusing that reset closes the box -- is there a reason for that?

jerrywu64: I could possibly try to implement it so that it's unnecessary, but currently I have it so that rendering the child updates the parent's record of the child's state. Reset is supposed to force the child to be recreated, but if I don't close the box the child's state will "update" the parent's record and then we will be sad.

benjaminjkraft: Read below -- I don't think having rendering the child update the parent's state is a good strategy anyway, and I don't think it's any harder to do things right.

esp/public/media/scripts/program/modules/scheduling_checks.jsx
jerrywu64: I did that initially; it didn't work. The parent gets passed an outdated version of the state (in particular, the most recent state change isn't processed). There's a thing on https://facebook.github.io/react/docs/component-api.html which seems to be relevant, about how it doesn't immediately mutate the state.

esp/public/media/scripts/program/modules/scheduling_checks.jsx
jerrywu64: Child passes the parent a copy of its state so that it can be resurrected properly between closing/opening the box or refreshing. I tried just saving a copy of the table once (instead of creating a new SelectTable) but it didn't seem to work; it seems like the state dies if I close the box even if the table is saved. I read something about the state being meaningless if the component isn't mounted, which if it means what I think it means it makes sense that the state died, as opposed to me just having incorrectly tried to preserve the SelectTable.

esp/public/media/scripts/program/modules/scheduling_checks.jsx
benjaminjkraft: So I think what you want to do is to have all of the state exist in the SchedulingCheck, and none in the SelectTable -- SchedulingCheck can just pass what rows are grayed and such down to SelectTable as a prop.  Then the event handler here just calls the callback from SchedulingCheck to update its state, which will then rerender a new SelectTable with a different set of rows grayed.  (In fact, React is smart and will do the minimal set of DOM updates necessary to accomplish this.)

Either that or I don't understand the issue -- you are setting the new state to whatever you want it to be; React may batch updates, but if you make a change that change will get there eventually.

esp/public/media/scripts/program/modules/scheduling_checks.jsx
benjaminjkraft: Yeah, you don't want to be trying to "preserve the state [of the SelectTable]".  Any state that you want to stick around shouldn't be on the SelectTable at all, it should be on the SchedulingCheck.

jerrywu64:
Let me know if we would rather use react than react-with-addons, I'm using react-with-addons for update(). Those portions can be rewritten to be done manually if needed, the objects I'm updating aren't very big.

benjaminjkraft:
I don't have a particular problem with using react-with-addons.  It might be easier all around to not have `tableState` be a separate object, and have it just be three more keys in `state`, in which case you wouldn't need the complicated update thing, but I don't particularly care, so feel free to leave it as-is.

esp/public/media/scripts/program/modules/scheduling_checks.jsx
```diff
@@ -114,3 +187,89 @@ var RefreshButton = React.createClass({
...
+      reverse: React.PropTypes.bool.isRequired
+    }).isRequired,
+    header: React.PropTypes.bool.isRequired,
+    columns: React.PropTypes.array,
+    clickHeader: React.PropTypes.func.isRequired,
+    clickRow: React.PropTypes.func.isRequired
+  },
+    
+  getInitialState: function(){
+    return {sort: this.props.saveState.sort, greyed: this.props.saveState.greyed, reverse: this.props.saveState.reverse};
```
benjaminjkraft: This state is no longer getting modified, so you may as well just remove it, and change the references to get everything directly from props.

benjaminjkraft:
In any case, I just left a few comments, but I think other than those couple of minor issues, this is good.  Thanks for your patience with all my comments :-) .

jerrywu64:
Cool. I'm not going to bother changing tableState to be different, although yes the current setup doesn't quite make sense.

And nah, thanks for your help; I should really get used to using things more in line with how they're supposed to be used instead of hacking things together (and probably more generally figuring out how things are supposed to be used in the first place), and also checking over my code so I don't have the problem in which stuff is set up a certain way because of a previous implementation attempt that has since been scrapped oops.

benjaminjkraft:
Cool!  I think this looks good now, merging.

PR #1724, opened 2015-07-15
Fix incorrect link to class docs in catalog

mgersh:
target_url was removed in #856, and all the other class docs interfaces use the code I've changed it to here.


benjaminjkraft:
Indeed they do.  Merging.

PR #1726, opened 2015-07-17
Fix error when sending class changes emails

mgersh:
Django complained that bcc has to be a list or a tuple. This version worked.


benjaminjkraft:
LGTM.

PR #1727, opened 2015-07-17
Added statement descriptor to Stripe Module

### LINKED ISSUE #1725: Customize statement descriptor for credit card charges on Stripe
Currently, our credit card charges show up as "Learning Unlimited" on the credit card statement.  Some people don't recognize this name, and recently we had a disputed credit card payment because of this.

It would be better to customize the descriptor per program (e.g. "Yale Splash") which is something that students and their parents are more likely to recognize.  This can be done as described here:

https://support.stripe.com/questions/when-i-charge-a-customer-what-will-they-see-on-their-card-statements

"If you want to set a per-charge (dynamic) descriptor, you can do so by utilising the statement_descriptor parameter, which is available when creating a charge and on a plan."

bhorowitz:
Added statement descriptor to be the group's name's first 22 characters. Perhaps a more elegant way might be to have the descriptor be a setting in local_settings.py, and to use the group_name if none is given there.

Fixes #1725.

### DIFF
--- a/esp/esp/program/modules/handlers/creditcardmodule_stripe.py
+++ b/esp/esp/program/modules/handlers/creditcardmodule_stripe.py
@@ -262,6 +262,7 @@ class CreditCardModule_Stripe(ProgramModuleObj):
                         currency="usd",
                         card=request.POST['stripeToken'],
                         description="Payment for %s %s - %s" % (group_name, prog.niceName(), request.user.name()),
+                        statement_descriptor=group_name[0:22], #stripe limits statement descriptors to 22 characters
                         metadata={
                             'ponumber': request.POST['ponumber'],
                         },


jmoldow:
Looks good. My only question is, is this parameter available to us? The page Ted linked to says its only available when you're "on a plan". I don't remember us ever subscribing to a plan. I can double-check when I get home.

I forget, do you have a dev server you can test this with? If not, I can try to test it.

bhorowitz:
Hmmm, yeah, it isn't quite clear what that page is trying to get at as it seems like a "plan" in their documentation is referring to using Stripe to processes reoccurring subscription costs. Their API documentation for statement_descriptor doesn't seem to mention plans though, and simply lists it a core argument of the Charge module: https://stripe.com/docs/api#create_charge

I don't currently have a dev server running on my system.

benjaminjkraft:
It looks like the "plan" they're referring to is just that when you get the data about your current subscription it will include your default statement_descriptor, in the same way you could include it in a charge.  So I think we're fine.  I also quickly tested this with the stripe test keys, but it's hard to tell what it would have shown up as in reality, since it didn't actually make a charge.  Merging.

hwatheod:
A parent from Splash UNC has reported that they are still seeing "Learning Unlimited" on their credit card statement.  But it doesn't look like there's anything wrong with our code.  I have contacted Stripe customer service about this.

PR #1741, opened 2015-08-18
Update the MathJax JavaScript include link

### LINKED ISSUE #1736: MathJax doesn't load
I assume we need to update the URL, or something like that.  It's 404ing for me.

jmoldow:
Use the official MathJax CDN (cdn.mathjax.org) link, as documented at [1].

According to dig, this is a CloudFlare CDN.

This change is necessary because the old link is no longer available, and returns a 404 response.

Fixes #1736.

[1] <http://docs.mathjax.org/en/latest/start.html#using-the-mathjax-content-delivery-network-cdn>


benjaminjkraft:
LGTM.

Issue #1309, opened 2014-09-14
Lists of programs should always be sorted in reverse


jmoldow:
In the admin panel, lists of programs should always be sorted in reverse (either from high id number to low id number, or from later dates to earlier dates). This includes:
- Program filters for other models.
- Foreign key field widgets that point at Program.

benjaminjkraft:
Fixed by #1742 (but feel free to reopen if you find more instances of it).

PR #1742, opened 2015-08-18
Fix for ticket #1309

ghost:
- Added more fields to program list
- Added search on name
- Default sort order is now by id, name


esp/esp/program/models/__init__.py
```diff
@@ -281,6 +281,7 @@ class Program(models.Model, CustomFormsLinkModel):
     class Meta:
         app_label = 'program'
         db_table = 'program_program'
+        ordering = ['-id','name']
```
benjaminjkraft: Not sure what the point of the `name` here is, since `id`s should be unique.  It might make some queries slower, I'm not sure.

benjaminjkraft:
One comment, otherwise LGTM.

esp/esp/program/models/__init__.py
joeleg: The idea was to sort 'chronologically' and also by name. I can remove name if it does not seem to be useful.

benjaminjkraft: Right, I support sorting "chronologically".  But there are no ties to break on `id`, so unless I'm misunderstanding something, adding a second sort cannot possibly change the sort order.

benjaminjkraft:
Looks like someone else already fixed the thing I had commented about in `main`, so I just took that in the merge, and will merge this once tests pass.

PR #1744, opened 2015-08-22
Django 1.8 support (broken)

mgersh:
Here's the work I currently have on Django 1.8. It's very much unfinished. I ran into a lot of old hacks which I attempted to change into new hacks instead of actually fixing them, because I didn't want to take forever, but if you feel inspired to do something nicer, please do.

Current state: Nothing here is tested except for updating requirements, and there are still some known to-do items, but the current blocking factor is that manage.py refuses to run any commands. This is because of a bad interaction between the Django 1.7 app-loading refactor and argcache. I'm not sure if there's a simple fix or if it will require deeper changes. @luac, you probably should look at this if you can.

Also I realize that the diff on this PR is a disaster because I deleted all the migrations. For now, I recommend pulling this branch and looking at diffs locally if you want to see anything.


mgersh:
More thoughts: it's likely that the best thing to do involves switching our infrastructure to use the new ready() method (https://docs.djangoproject.com/en/1.8/ref/applications/#django.apps.AppConfig.ready). I'm still not sure how hard that would be or if there's an easier temporary fix.

luac:
Oops. Yeah, I remember at some point I actually wanted to be able to use ready() but we weren't on 1.7 yet. I'll take a look.

mgersh:
A more organized description of what still needs to be done, now that it's not 2am:

Argcache:
- [ ] Fix it

Migrations and related issues:
- [ ] Figure out what needs to be done for our custom Fields to work right with the new migrations, and do it
- [ ] Create new initial migrations
- [ ] Re-create the zipcode initial data migration
- [ ] Debug any issues that might come up (I know that there are a few minor inconsistencies between the code and the esp.mit.edu database schema, probably left over from the pre-South days, and some documentation suggested that this might cause errors. I don't know if any other sites are likely to have issues.)

Tests:
- [ ] Run tests, and actually fix the test runner situation. What I did here is probably not correct but I felt like trying it anyway.

Manual testing:
- [ ] Test all the features I messed with here
- [ ] Do a full manual test of the website
- [ ] Test deploy on database dumps from various sites if we expect any issues

Note: again, I didn't make an attempt to remove all uses of anything that is deprecated but not removed yet. We'll need to look at those eventually, not necessarily right now.

luac:
Okay so, as a hack, I changed the `except LookupError` to `except (LookupError, AppRegistryNotReady)` to see what would happen. It looks like it at least got further in loading but ran into another issue

"There is no South database module 'south.db.postgresql_psycopg2' for your database. Please either choose a supported database, check for SOUTH_DATABASE_ADAPTER[S] settings, or remove South from INSTALLED_APPS."

luac:
Okay, I found what I think is a slightly better workaround, so I've pushed that. Still stuck on the South error.

mgersh:
Not sure what you mean by "forgot to manage.py update" when the problem is that we can't run manage.py?

The story of my debugging adventures so far:

I locally commented out everything involving south in an attempt to see what's going on, even though those will need a more careful look later, and...CustomFormsCache._populate() broke in the same way that cache loading was breaking earlier. It's a get_models, not a get_model, so it doesn't have the same workaround.

So I commented out the call to cf_cache._populate(), and then the get_app_configs() in cache_loader did the same thing.

Then I gave up and just commented out cache_loader from INSTALLED_APPS and, after a ton of deprecation warnings, got a different confusing error that definitely has nothing to do with argcache.

Conclusion: maaaaaybe we're better off fixing at least some of this the right way instead of trying to work around Django. This might be doable by just throwing the contents of cache_loader/__init__.py into a ready()?

luac:
Sounds good. I think cache_loader really just wants to be the ready() method of cache.

mgersh:
Looks like at this point, tests aren't going to get any farther until we deal with the rest of the migrations situation, so that's next.

mgersh:
Why I had to delete esp/__init__.py: http://stackoverflow.com/questions/23854677/django-finds-tests-but-fail-to-import-them

mgersh:
Right now there are 50 tests erroring, but most of them are the same error. We have a total of 152 tests running, and there were 178 before. I think the difference might just be external apps' tests which don't run with ours anymore, but I'm not sure.

I'm not going to do more work on this for a while, I need to focus on other things. Feel free to pick things up from here.

mgersh:
More details about what's going on right now:

* I updated the checklist above with what needs to be done. Fixing the current test failures and re-adding the zipcodes can be done in either order, probably both should be done before anyone tries manual testing.
* Most of the test failures right now are coming from template issues. If you want to fix them, https://docs.djangoproject.com/en/1.8/ref/templates/upgrading/ is probably the right place to look.
* A new South migration was just added in main. This means that we'll need to resolve the conflict. I think the best way to do this is by rebasing this work on top of main right before it's ready to merge (in case anybody adds any more migrations), then re-generating and re-testing the new migrations.

benjaminjkraft:
I should really go to sleep, but I think I have a strong lead on the template thing -- I think it's an issue with our `esp.web.util.template.cache_inclusion_tag`.  In particular, that contains a pile of code that's basically copy-pasted from the django source for `inclusion_tag`, with some changes to cache it.  Ideally, we could find a way to do things that doesn't require that, but I recall that being ~impossible as of a couple of years ago*, so we may have to just update it to new copy-pasted code.  (@luac, I think you've looked at this more recently -- do you have any idea if we can make this less of a mess?)  I don't know exactly what the issue is but I'm pretty sure that's the issue: I followed the `context.template` getting passed around, and the place where it gets dropped is where we enter the `cache_inclusion_node`.

Also, I noticed an issue where css files were 500ing.  (I'm guessing this was unnoticed due to the template errors.)  This appears to be due to our `esp.middleware.whitespacemiddleware`, which is making some assumptions about request structure that are no longer true of requests for static files.  My inclination is to remove it rather than fixing it -- it only removes whitespace from the beginning and end of the request body, which seems rather trivial to me.  If you think otherwise, say so -- it's probably not *that* hard to update.

*In particular, when we cache an inclusion tag, we don't want to just cache the returned context -- we want to cache the whole rendered thing.  This means we have to deal with the guts of `inclusion_tag` to inject caching code in the right places.

mgersh:
Oh, that was a copy-paste? That explains a lot. I knew that piece of code was the problem but didn't understand what it was doing. I've found a few other copy-pastes during this project, but they were all clearly commented so I could easily update the copy-paste instead of trying to figure out the details of what's wrong, while apparently this one you just have to know about.

I'm fine with getting rid of the whitespace middleware. This isn't the first time we've run into a problem with it recently, and it seems like more trouble than it's worth.

benjaminjkraft:
Ok, I just pushed a bunch of fixes to various template-related stuff for 1.8.  The highlights:
- used the forwards-compatible way of specifying our template settings (59335f2)
- removed StripWhitespaceMiddleware (see above for discussion; a4da023)
- updated our template loaders. (836645a) I removed our `CachedLoader` because I don't think we need it.  @mgersh,I clobbered 7bc9db5 by just removing that section of code because I am pretty convinced we just don't need it at all.
- updated `cache_inclusion_tag`. (a65a465 and 8357032) This was by far the hardest part, because this code is based on a copy-paste from the django source, which needed to get updated.  It's also a fairly complex piece of code, because template rendering is complex, and caching it right gets tricky.  I'd love to have a fairly careful code review on this (@mgersh? @luac?); I think everything is fairly well-commented, but poke me if you want more explanation anywhere.  Also, unfortunately, this will probably need to be updated again in 1.9, although I think the updates won't be too difficult (since the change is mostly just moving code around, and now I think I understand most of the relevant concerns).

I'm going to see if I can manage to write some tests for some of this stuff, especially `cache_inclusion_tag`.  But that will be in a separate commit.

mgersh:
I looked at the parts of that recent bunch of commits that aren't related to cache_inclusion_tag, and those all look good. Also, now that the template stuff is fixed, general manual testing is no longer blocked! I'm currently still working on porting customforms, and then I'll write the zipcode migration unless someone else wants to claim it.

I think once I've finished those last two things, it'll be a reasonable time to rebase this on main, resolve migrations, and not merge any more South migrations ever. Then we can do manual testing on the final branch. If anyone objects, say something.

benjaminjkraft:
:+1: I just looked through the older commits that I didn't write, which I'm not sure were reviewed; they generally looked good, just a few minor comments.  Hopefully we are getting close!

mgersh:
Looks like 1.8.5, which has @benjaminjkraft's fix for the SimpleLazyObject pickling error in it, will likely be released before this is ready to merge. Thoughts on whether we'll still want the changes in f1204b1 and a81d57a? It's a somewhat scary infrastructure change that won't be necessary anymore, but it's probably overall a good thing anyway.

luac:
If it weren't for manual testing being more convenient to do all at once, I'd say split it into a separate pull request and merge it separately. But we should include it in the manual testing.

benjaminjkraft:
I don't have a strong opinion, but I'd say keep it around.  I'm not the happiest with the `SimpleLazyObject` check in a81d57a7, but I think `ESPUser(SimpleLazyObject(lambda: User(...)))` was broken before my changes -- I guess we just don't use that anywhere (since we already make sure `request.user` is a (lazy) `ESPUser`).

(And agreed with @luac that unfortunately we want to keep it in this if we merge it, to make sure it gets tested.)

If we do delete it, we may want to clean up the removal of the `__eq__` method and the tests.

mgersh:
Look, we have a clean travis build!

Also, I think I've figured out the correct way to apply migrations on an existing database. First, fake initial migrations for auth, contenttypes, and reversion. Then actually migrate those apps. Then fake everything else.

The next things to do now are rebase on main, generate a new set of initial migrations, and test everything. Probably the easiest way to do this involves merging the stable release branch into mit-prod so the server has up-to-date migrations and taking a database dump to test with once that's done.

mgersh:
Sounds like we want to leave the ESPUser changes as they are. Works for me.

I noticed while checking migrations that our custom update contenttypes thing isn't working as intended. The whole point of that copy-paste is so we don't have to confirm deletion of stale contenttypes, and it made me confirm when I migrated. I'm not really sure why we wanted this in the first place, so I'm inclined to just delete it instead of trying to fix it. Any objections?

When I do the rebase, I can fix the code duplication in the AppConfigs, and I'll bump us to the latest Django point release. The change in DynamicModel.py that you thought was unnecessary really is, I made it before I'd entirely figured out what was going on, I'll fix that too.

I think that's all the comments in this branch that aren't resolved yet. If anything needs doing that isn't mentioned in this comment or my previous one, I've forgotten about it and you should remind me.

benjaminjkraft:
> I noticed while checking migrations that our custom update contenttypes thing isn't working as intended. The whole point of that copy-paste is so we don't have to confirm deletion of stale contenttypes, and it made me confirm when I migrated. I'm not really sure why we wanted this in the first place, so I'm inclined to just delete it instead of trying to fix it. Any objections?

This just means one more manual confirm step on each stable release?  That's fine with me.

PR #1746, opened 2015-08-22
Pipe input directly to lessc to avoid making a temp file.

luac:
I vaguely remember people in the past encountering some sort of
permissions issue involving temporary files during themes setup. I
don't know if it's still prone to happen or if this is even the
relevant code, but this might address that issue and seems to
otherwise work the same as before.

### DIFF
--- a/esp/esp/themes/controllers.py
+++ b/esp/esp/themes/controllers.py
@@ -236,19 +236,13 @@ class ThemeController(object):
             tf1.write(less_data)
             tf1.close()
 
-        (less_output_fd, less_output_filename) = tempfile.mkstemp()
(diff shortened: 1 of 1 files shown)


benjaminjkraft:
Thoughts on whether it's worth keeping the old behavior if `DEBUG` (or `THEME_DEBUG`, or something else) is true to aid in debugging?  I guess if you're debugging it's easy enough to just make a local change to print the text to stdout or whatever.

luac:
There's already a `THEME_DEBUG` conditional that will write `less_data` out into another file.

benjaminjkraft:
Got it.  LGTM then.

PR #1748, opened 2015-08-23
Change some Stripe configs to prevent errors

jmoldow:
Add extra pre-charge validation by enabling ``data-zip-code``. And prevent browser state from causing problems by disabling ``data-allow-remember-me``.

Documentation of these configurations is at [1].

For more context, see the thread on websupport@learningu.org with the subject

  [ ESP CC ] Credit card error on yale.learningu.org: 3246

[1] <https://stripe.com/docs/checkout#integration-simple-options>


benjaminjkraft:
Seems fine to me.  Do we want to backport this?

PR #1749, opened 2015-08-23
In Stripe module, skip donation code when disabled

jmoldow:
Previously, if the Stripe module setting ``offer_donation`` was set to ``False``, only the accounting controller and template would be affected.

Now, the donation setup that happens in the module will also be skipped when ``offer_donation`` is ``False``.

### DIFF
--- a/esp/esp/program/modules/handlers/creditcardmodule_stripe.py
+++ b/esp/esp/program/modules/handlers/creditcardmodule_stripe.py
@@ -113,11 +113,12 @@ class CreditCardModule_Stripe(ProgramModuleObj):
         #   Check for a 'donation' line item type on this program, which we will need
         #   Note: This could also be created by default for every program,
         #   in the accounting controllers.
-        (lit, created) = LineItemType.objects.get_or_create(
-            text=self.settings['donation_text'],
-            program=self.program,
-            required=False
-        )
+        if self.settings['offer_donation']:
+            (lit, created) = LineItemType.objects.get_or_create(
+                text=self.settings['donation_text'],
+                program=self.program,
+                required=False
+            )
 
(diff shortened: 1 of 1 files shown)


benjaminjkraft:
Just to check, this will all work right if a chapter enables the module without donations, and later in the program turns on donations, right?

esp/esp/program/modules/handlers/creditcardmodule_stripe.py
```diff
@@ -158,8 +159,9 @@ def payonline(self, request, tl, one, two, module, extra, prog):
         payment_type = iac.default_payments_lineitemtype()
         sibling_type = iac.default_siblingdiscount_lineitemtype()
         grant_type = iac.default_finaid_lineitemtype()
-        donate_type = iac.get_lineitemtypes().get(text=self.settings['donation_text'])
-        context['itemizedcosts'] = iac.get_transfers().exclude(line_item__in=[payment_type, sibling_type, grant_type, donate_type]).order_by('-line_item__required')
+        offer_donation = self.settings['offer_donation']
+        donate_type = iac.get_lineitemtypes().get(text=self.settings['donation_text']) if offer_donation else None
+        context['itemizedcosts'] = iac.get_transfers().exclude(line_item__in=filter(None, [payment_type, sibling_type, grant_type, donate_type])).order_by('-line_item__required')
```
benjaminjkraft: > `filter(None, ...)`

Huh, TIL.

jmoldow:
> Just to check, this will all work right if a chapter enables the module without donations, and later in the program turns on donations, right?

It should. The ``check_setup()`` function gets run on every request. This will create the needed ``LineItemType`` before anything else happens.

benjaminjkraft:
Cool, LGTM.

PR #1750, opened 2015-08-23
Reword the deadline-learn.html template

jmoldow:
This is a local change that was made by Yale. Someone else should look at this and decide whether or not the rewording is better, and worth merging.

### DIFF
--- a/esp/templates/errors/program/deadline-learn.html
+++ b/esp/templates/errors/program/deadline-learn.html
@@ -11,7 +11,7 @@
 
 <strong>
 <p>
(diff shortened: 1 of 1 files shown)


mgersh:
I think we've already fixed most, if not all, of the cases where non-student accounts get the deadline error instead of the "you're not a student" error. I vaguely remember there being one which was a bit more tricky to fix for some reason, but I'd still rather fix it instead of putting this version on main.

benjaminjkraft:
I'm fine with merging this, or not.  I think @mgersh is right that the real way to fix these is to make sure we give the right error.  The question is whether it's better to be more general just in case or to be more specific so users give us more useful information.

benjaminjkraft:
I am leaning against doing this.  Since it hasn't had any discussion in a few months, I'm going to close it; feel free to reopen if anyone thinks we should make this change.

PR #1751, opened 2015-08-23
Add created_at field to dbmail models

jmoldow:
Add `created_at` field to `MessageRequest` and `TextOfEmail`.

With the knowledge of how old each request is, cronmail can now filter out un-sent, out-of-date messages.

In a perfect world, this would be unnecessary. But with each release, we improve our dbmail error handling, and can sometimes accidentally pick up and send very old, previously un-sent messages. This commit adds a safe-guard against that.

It is also nice to be able to see when a message was created.

Tested successfully on Yale's site. Without this, 6156 old `TextOfEmails` would get sent. With it, none of them do. But new comm panel emails still get sent successfully.

### DIFF
--- a/esp/esp/dbmail/models.py
+++ b/esp/esp/dbmail/models.py
@@ -32,6 +32,7 @@ Learning Unlimited, Inc.
   Phone: 617-379-0178
   Email: web-team@learningu.org
 """
+import re
 import sys
 
 from django.db import models, transaction
@@ -114,7 +115,16 @@ class ActionHandler(object):
             return getattr(obj, key)
         
         return obj.get_msg_vars(self._user, key)
-    
+
+
+_MESSAGE_CREATED_AT_HELP_TEXT = """
+    The time this object was created at. Useful for informational
+    purposes, and also as a safety mechanism for preventing un-sent
+    (because of previous bugs and failures), out-of-date messages from
+    being sent.
+"""
+_MESSAGE_CREATED_AT_HELP_TEXT = re.sub(r'\s+', ' ', _MESSAGE_CREATED_AT_HELP_TEXT.strip())
+
 
 class MessageRequest(models.Model):
     """ An initial request to broadcast an e-mail message """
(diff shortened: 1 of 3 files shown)


benjaminjkraft:
This seems like a reasonable safety check.  It might be a good idea to put in a comment somewhere saying that this is a safety check, and is not intended to be the main mechanism for determining whether/when an email gets sent.

esp/esp/dbmail/models.py
```diff
@@ -355,6 +362,14 @@ class TextOfEmail(models.Model):
     send_from = models.CharField(max_length=1024) # Valid email address
     subject = models.TextField() # E-mail subject; plain text
     msgtext = models.TextField() # Message body; plain text
+
+    # Use `default` instead of `auto_now_add`, so that when a
+    # :class:`TextOfEmail` is created from a :class:`MessageRequest`, the
+    # `created_at` value may be copied over at creation time. If `auto_now_add`
+    # were `True`, the value of `now()` would always be used, even if we wanted
+    # to use something else.
+    created_at = models.DateTimeField(default=datetime.now, null=False, blank=False, editable=False, auto_now_add=False)
```
benjaminjkraft: I would suggest just not setting a default for this.  If we set a default, it will make it harder to notice error where we intended to set this manually, and didn't.  I think `TextOfEmail`s only ever get created by `MessageRequest.process`, so it's not like this will require a bunch of updating existing code.

esp/esp/dbmail/models.py
```diff
@@ -159,6 +159,12 @@ class MessageRequest(models.Model):
                     "should receive the message.")
     sender = models.TextField(blank=True, null=True) # E-mail sender; should be a valid SMTP sender string 
     creator = AjaxForeignKey(ESPUser) # the person who sent this message
+
+    # Use `default` instead of `auto_now_add`, so that the migration creating
+    # this field can set times in the past, and to keep consistency with the
+    # `TextOfEmail.created_at` field.
+    created_at = models.DateTimeField(default=datetime.now, null=False, blank=False, editable=False, auto_now_add=False)
```
benjaminjkraft: This default seems fine, though.

benjaminjkraft:
Code looks good other than the default issue I mentioned inline.

esp/esp/dbmail/models.py
jmoldow: Will push a fix for this. Thanks.

jmoldow:
> 

This seems like a reasonable safety check. It might be a good idea to put in a comment somewhere saying that this is a safety check, and is not intended to be the main mechanism for determining whether/when an email gets sent.

Is there somewhere specific you recommend, or some wording that is lacking in the current comments? Right now I already have comments before both places that do this filtering, and in the migration, saying that we are doing this to prevent out-of-date messages from being sent.

benjaminjkraft:
I'd put it on the model field(s), too, since that's where I would look first if I were trying to figure out what this did, and also where I might happen across it and decide to use it for something else, say.  (It could be either a comment or `help_text`.)

esp/esp/dbmail/models.py
jmoldow: Done.

jmoldow:
Done.

PR #1753, opened 2015-08-24
In Stripe module, skip donation code when disabled

jmoldow:
Request to backport PR #1749 (already merged into main) into stable release 6.

Previously, if the Stripe module setting ``offer_donation`` was set to ``False``, only the accounting controller and template would be affected.

Now, the donation setup that happens in the module will also be skipped when ``offer_donation`` is ``False``.


benjaminjkraft:
Can you redo this with `git cherry-pick -x` to make it easier to keep track
of what's from where?

jmoldow:
This isn't a cherry pick. I made the original commit off of the merge base (branch origin/mit-prod-merge-base). This is a PR of the exact same commit, just into a different base (origin/stable-release-6).

benjaminjkraft:
Ah, never mind then!

PR #1754, opened 2015-08-24
Add created_at field to dbmail models

jmoldow:
Request to backport PR #1751 (already merged into main) into stable release 6. This is an important part of the release: without this, unsent messages will be sent by each site after the upgrade is complete.

Add `created_at` field to `MessageRequest` and `TextOfEmail`.

With the knowledge of how old each request is, cronmail can now filter out un-sent, out-of-date messages.

In a perfect world, this would be unnecessary. But with each release, we improve our dbmail error handling, and can sometimes accidentally pick up and send very old, previously un-sent messages. This commit adds a safe-guard against that.

It is also nice to be able to see when a message was created.

Tested successfully on Yale's site. Without this, 6156 old `TextOfEmails` would get sent. With it, none of them do. But new comm panel emails still get sent successfully.

### DIFF
--- a/esp/esp/dbmail/cronmail.py
+++ b/esp/esp/dbmail/cronmail.py
@@ -42,6 +42,10 @@ from django.template.loader import render_to_string
 
 from django.conf import settings
 
+
+_ONE_WEEK = timedelta(weeks=1)
+
+
 @transaction.autocommit
 def process_messages(debug=False):
     """Go through all unprocessed messages and process them.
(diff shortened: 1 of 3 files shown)


esp/esp/dbmail/cronmail.py
```diff
@@ -42,6 +42,10 @@
 
 from django.conf import settings
 
+
+_ONE_WEEK = timedelta(weeks=1)
+
+
 @transaction.autocommit
```
jmoldow: @benjaminjkraft, can you confirm that it is correct to leave this here? In main, there is no decorator. But you didn't remove this during your backport.

benjaminjkraft: Yes.  In 1.6+, roughly the same behavior is the default.  (We have middleware to change that, but it doesn't apply here.)

PR #1755, opened 2015-08-24
Update .gitignore with Mac ignore files

jmoldow:


benjaminjkraft:
It's worth just putting these in a global gitignore file (see for example [mine](https://github.com/benjaminjkraft/dotfiles/blob/master/.gitignore), but I don't mind having them here too just in case.  Merging.

Issue #1198, opened 2014-04-09
Large Splashes Can't Generate Schedules PDF


jmoldow:
In spring 2014, at least two large Splash programs (Clark and Stanford) had trouble generating the student schedules.
- The Clark .pdf file was not downloadable from the website. I was able to download the .tex file from the website, and use that to manually create the .pdf file on the server using `gen_latex()`.
- Neither the Stanford .pdf file nor .tex file were downloadable from the website. I was able to manually create the .tex file on the server, and was also able to generate the .tex file by running `render_to_latex`. The .pdf file was also able to be generated this way, but it took a very long time. The bottleneck was running `ps2pdf`.

Stanford has been running large programs for a while now, and I don't think Clark's Splash is the biggest non-Stanford Splash ever, so I don't know why this is happening now. Maybe we changed something that made these files larger even if the content isn't larger. Or maybe we reconfigured diogenes wrong when we reinstalled the OS, and it can't handle large files as well.

benjaminjkraft:
Putting here what I put in email: I think at least some of the issues with things getting stuck between generating TeX and returning a PDF to the user come from `dvips`, and thus would be fixed by the change to `pdflatex` in #1509.  Probably not the only issue, but it's at least one that should just require a backport to solve.

willgearty:
Update from Stanford:
I haven't been able to generate a pdf of student schedules via the website for the past 2 years (it usually takes 10+ minutes, then crashes).
I've resorted to generating the latex code and then compiling that locally.

benjaminjkraft:
Hmm, maybe we have a new issue here then!  It'll take some more investigation to see why things are breaking now; I think all the information above was a different issue fixed in #1759, as it turned out.

mariolsen0112358:
MIT Splash 2016 had issues downloading the schedules PDF (on Windows and Mac) with the same 10+ minutes load -> crash issue. However, Favyen managed it first try without issues on whatever (Linux?) he uses and said he saw "some connection refused errors from varnish in the log" after our failed attempts.

There was some speculation that it might be due to a special character on one of the schedules? Or just sheer size of the doc. In any case, smaller groups ("Students with Saturday dinner" as opposed to "Students who filled out medliab") loaded fine.

willgearty:
Yeah, we haven't had trouble downloading smaller subsets of the student schedules, either.

[...]

PR #1759, opened 2015-08-25
Remove use of subprocess.PIPE in gen_latex()

jmoldow:
`stdout=PIPE` was used as a shortcut to hide the output of latex generation. However, according to the documentation for `subprocess.call()` [1], this is unsafe, as "that can deadlock based on the child process output volume".

This explains some of the symptomps of #1198. Namely, a hanging pdflatex process, for what should normally be a <10 second operation.

Instead, the default arguments to `gen_latex()` now open a file handle to `os.devnull`, and pass that to `subprocess.call()`.

Fixes, either partially or entirely, #1198 (depending on whether database deadlocks from accounting are still an issue).

[1] <https://docs.python.org/2/library/subprocess.html#subprocess.call>

### DIFF
--- a/esp/esp/web/util/latex.py
+++ b/esp/esp/web/util/latex.py
@@ -34,7 +34,6 @@ Learning Unlimited, Inc.
 """
 """ This module will render latex code and return a rendered display. """
 
-from contextlib import nested
 import os.path
 import os
 from subprocess import check_call, STDOUT
@@ -47,7 +46,7 @@ from django.http import HttpResponse
 
 TEX_TEMP = tempfile.gettempdir()
 TEX_EXT  = '.tex'
-devnull_sentinel = object()
+_devnull_sentinel = object()
 
 def render_to_latex(filepath, context_dict=None, filetype='pdf'):
     """ Render some tex source to latex. This will run the latex
@@ -74,7 +73,7 @@ def render_to_latex(filepath, context_dict=None, filetype='pdf'):
     
     return gen_latex(rendered_source, filetype)
 
-def gen_latex(texcode, type='pdf', remove_files=False, stdout=devnull_sentinel, stderr=STDOUT):
+def gen_latex(texcode, type='pdf', remove_files=False, stdout=_devnull_sentinel, stderr=STDOUT):
     """Generate the latex code.
 
     :param texcode:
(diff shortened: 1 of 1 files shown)


esp/esp/web/util/latex.py
```diff
@@ -107,18 +109,24 @@ def gen_latex(texcode, type='pdf', remove_files=False, stdout=PIPE, stderr=STDOU
     :rtype:
         HttpResponse
     """
+    files = [stdout, stderr]
```
benjaminjkraft: I think this is more complicated than necessary.  You can just pass the same file handle to both inputs if `stdout` and `stderr` are both `devnull_sentinel`, and only open it once.  This avoids the whole `nested()` thing, which I found a bit confusing.

esp/esp/web/util/latex.py
```diff
@@ -107,18 +109,24 @@ def gen_latex(texcode, type='pdf', remove_files=False, stdout=PIPE, stderr=STDOU
...
+    with nested(*(open(os.devnull, 'w') for f in files if f is devnull_sentinel)) as devnull_files:
+        for devnull_file in devnull_files:
+            files[files.index(devnull_sentinel)] = devnull_file
```
benjaminjkraft: Also, I don't think this will work correctly if you actually have `stdout==stderr==devnull_sentinel`.

esp/esp/web/util/latex.py
```diff
@@ -46,6 +47,7 @@
 
 TEX_TEMP = tempfile.gettempdir()
 TEX_EXT  = '.tex'
+devnull_sentinel = object()
```
benjaminjkraft: If importing this from elsewhere is supported, I'd just name it `devnull`, since the fact that it's a sentinel object is an implementation detail.  If it's not (which would support), it should start with an underscore.

esp/esp/web/util/latex.py
jmoldow: Done.

esp/esp/web/util/latex.py
jmoldow: Made it private. If a caller wants to, say, have ``stdout`` go to a file but ``stderr`` go to ``devnull``, the caller should explicitly open a handle to ``devnull``, pass that for ``stderr``, and ensure proper file closing, rather than use the sentinel.

benjaminjkraft:
Thanks, LGTM.

PR #1761, opened 2015-09-01
Schedule performance improvements from BC's site

benjaminjkraft:
These were local changes on BC's site; I don't know who made them, but these
are the parts that it looks like we want and don't already have.  The timeslots
thing should help a little; the tag matters less since it's cached but we might
as well do it less stupidly.

### DIFF
--- a/esp/esp/program/modules/handlers/programprintables.py
+++ b/esp/esp/program/modules/handlers/programprintables.py
@@ -881,10 +881,10 @@ Volunteer schedule for %s:
             students = list(ESPUser.objects.filter(filterObj.get_Q(restrict_to_active=False)).distinct())
 
         students.sort()
-        return ProgramPrintables.get_student_schedules(students, prog, extra, onsite)
+        return ProgramPrintables.get_student_schedules(request, students, prog, extra, onsite)
 
     @staticmethod
-    def get_student_schedules(students, prog, extra='', onsite=False):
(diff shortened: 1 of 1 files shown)


esp/esp/program/modules/handlers/programprintables.py
```diff
@@ -881,10 +881,10 @@ def studentschedules(self, request, tl, one, two, module, extra, prog, onsite=Fa
             students = list(ESPUser.objects.filter(filterObj.get_Q(restrict_to_active=False)).distinct())
 
         students.sort()
-        return ProgramPrintables.get_student_schedules(request, students, prog, extra, onsite)
+        return ProgramPrintables.get_student_schedules(students, prog, extra, onsite)
 
     @staticmethod
-    def get_student_schedules(request, students, prog, extra='', onsite=False):
```
jmoldow: ``request`` is used in the function.

I only removed it locally because, IIRC, I had also temporarily removed the line that used ``request``, and was trying to make it easier for me to generate schedules from the command line. But that part of the change should be reverted.

jmoldow:
I wrote this. Looks good to me (don't know if I can approve my own code even though I didn't make the PR), except for my comment about ``request``.

benjaminjkraft:
Fixed.  I think as long as we're both fine with the changes we can merge (once tests pass), and I'm fine with it.

PR #1762, opened 2015-09-01
Bc schedule perf

benjaminjkraft:
Copy of #1761 to stable-release-6.


jmoldow:
Not super important to have this, but I'll merge it anyway, in case chapters run in to schedule generation performance problems we can upgrade their code easily.

Issue #1765, opened 2015-09-03
QSD .edit.html pages shouldn't 500 if you are logged out


benjaminjkraft:
They should display error text, but they should 401 or 200 or something.  It's not a server error.

PR #1766, opened 2015-09-03
Make QSD .edit.html pages raise 403 instead of ESPError

luac:
Addresses #1765. Make this case consistent with the other permission
checks in this file, which raise 403.

### DIFF
--- a/esp/esp/qsd/views.py
+++ b/esp/esp/qsd/views.py
@@ -44,7 +44,7 @@ from os.path import basename, dirname
 from datetime import datetime
 from django.core.cache import cache
 from django.template.defaultfilters import urlencode
-from esp.middleware import ESPError, Http403
+from esp.middleware import Http403
 from esp.utils.no_autocookie import disable_csrf_cookie_update
 from django.utils.cache import add_never_cache_headers, patch_cache_control, patch_vary_headers
 from django.views.decorators.vary import vary_on_cookie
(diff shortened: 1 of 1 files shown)


benjaminjkraft:
Thanks!  Looks like there's another one down in `ajax_qsd` that is explicitly returning an `HttpResponse(status=500)` that could also get updated.  (I can't comment inline because it's not in the diff.)

esp/esp/qsd/views.py
```diff
@@ -179,7 +179,7 @@ def qsd(request, url):
 
         # Enforce authorizations (FIXME: SHOW A REAL ERROR!)
         if not have_edit:
-            raise ESPError("You don't have permission to edit this page.", log=False)
+            raise Http403, "You don't have permission to edit this page."
```
jmoldow: Does `Http403` not support being thrown with ``raise Http403("message")`` ?

luac: It does, in fact as far as I can tell they're exactly equivalent.

I only did it this way because the rest of the file does it that way; it's probably better to actually update all of them to the other way to avoid mixup with `raise (Http403, "message")`, but I didn't feel like updating the other uses of it.

luac:
Good catch. I fixed that one but it's untested because I didn't know how to get it to reach that case.

benjaminjkraft:
You can probably open an inline QSD page, log out in another tab, and then try to edit it, and see in the console what the error is?

PR #1768, opened 2015-09-10
Fix display of small input items in customforms

btidor:
Checkboxes and radio buttons now appear on the same line as the label, instead of on a separate line.

**Before**

before

**After**

after


benjaminjkraft:
:+1:

PR #1772, opened 2015-09-21
Form to choose fields for all_classes.csv

### LINKED ISSUE #1214: CSV dumps of teacher/class data
Directors seem to want this fairly often and we tell them to ask a webmin. It would be great to have an automated way to do it, ideally one that's configurable.

ghost:
This adds a form for admins to choose which fields should be included in the `all_classes.csv` spreadsheet.

Fixes #1214.


willgearty:
I've brought this PR back to life! I've made sure it works on main, and I've tested it with all of the fields.

image

kkbrum:
Looks good, but I think a "move all to other column" button might be useful above the selector pane so that you can select all/ deselect all more easily!

Issue #1450, opened 2014-11-03
RegistrationProfiles created for users without permission to register for program


jmoldow:
Ideally, a `RegistrationProfile` for a program should only be created if the user has actually filled out the profile form for the program (which implies they have permission to register). However, `getLastForProgram()` tries to be smart, and it will copy an old profile from a different program. This can create unprivileged profiles in the following way:
- Go to /studentreg url.
- `findModule()`is called.
- Since /studentreg is the main view of a core module, the list of required modules is iterated through, and their `isCompleted()` methods are called.
- regprofilemodule is always required.
- Its `isCompleted()` method calls `getLastForProgram()`.

There are a couple of potential ways to fix this. I don't know which is the most correct solution.
- `getLastForProgram` never saves profile copies. That is done manually when it is actually needed.
- `getLastForProgram` takes an optional argument, which defaults to `False`, which determines whether saving should occur. The argument is only passed as True in limited circumstances.
- `getLastForProgram` checks if the user has the /Profile permission. If the user lacks that permission, either it doesn't save, or it raises some Exception.

Another solution, which is more effort and doesn't address the root cause, would be to have some way of checking permissions for a module without calling the view function (this is the only way currently since permissions are implemented via decorators). Then we can check for permissions before calling `isCompleted()`.

Issue reported by @hwatheod. This could explain why there are always many more profiles for a program than class registrations.

benjaminjkraft:
I think I'm a fan of having `getLastForProgram` just not save the program ever.  Any time we actually want to save real data we'll be saving it again anyway, and we already don't save it if it's recent.  But maybe this has performance implications I don't quite understand.  Does anyone any idea if it would?

If not I'll modify/replace #1773 to do that.

Dhirenderchoudhary:
@willgearty if this issue is not resolved please assign it to me.

PR #1773, opened 2015-09-22
Updated ticket-1450 branch to be current with main

ghost:
Ticket 1450 - Clean merge with main


benjaminjkraft:
Moving discussion of the part for #1450 to that issue; I might cherry-pick some stuff from here or I might just rewrite since this has gotten quite old.  I'm not sure of the purpose of the SplashInfo stuff here but it's not obvious to me that it's not a breaking change for chapters who are using the module, so I don't want to merge it without knowing more about that.  If anyone knows (@pricem?) we can pull that into a separate PR.

PR #1779, opened 2015-10-05
Django 1.8 support (rebased for new migrations)

mgersh:
Updated version of #1744. Changes:
* Bumped Django version from 1.8.3 to 1.8.5 (a84cf3c)
* Deleted the new South migration, and re-generated new Django migrations (cbff413, 45941be)
* Entirely rewrote the commit which adds all the AppConfigs (38c36a6). Yes, having to subclass it everywhere just to add the name is silly, but I think that's Django's fault, not mine.
* Addressed a comment on the customforms port (9964e3a)
* Added a new commit to delete the update content types thing (16c7290)
* Resolved some trivial merge conflicts

The MIT server is up to date on all migrations now. There's a database dump sitting in my home directory on the server if you want to test with it. All we have left to do now is manual testing.


mgersh:
I just ran through a basic test of everything. Things generally look good. I know I didn't catch all the more obscure features, and this is big enough that it should definitely have more eyes on it before we merge, but as of now I believe there are no major glaring regressions in commonly-used features.

mgersh:
I figured out what's causing most of the deprecation warning spew. It's the esp.miniblog.views import that I added in #1517. It's there because the argcache error we were getting claimed that that's the correct way to fix it, so I did, and it fixed it. @luac or anyone else, ideas for a more reasonable way to fix it?

luac:
Try moving the import to `miniblog/models.py`, or adding a `miniblog/apps.py` that imports it upon `ready()`?

luac:
Might be a good idea to update the comment in sad_face.py as well.

mgersh:
Huh, I've been getting that test error on my dev server but not on Travis for a while now. Now it's apparently happening on Travis too but only sometimes (at this point, 'pr' and 'push' should be running the same code, but only 'pr' failed). What I know so far is that it happens when the program tests try to delete a user during cleanup and it's related to custom forms. I'm worried that there could be a bad interaction between custom forms and new app loading, but I don't know, that's just a wild guess.

luac:
It looks like the Travis error comes about because the customforms tests create a new model for the new form, and this happens in its own transaction, but then there are various places in Django's app loader where lists of models and fields are cached, so then those become stale once the transaction gets rolled back which causes problems for subsequent tests.

I did some source/debugger diving and found the caches that needed busting in order to get rid of the error. What I ended up doing was adding these lines at the end of ViewUserInfoTest.setUp:
```
from django.apps import apps
del apps.get_app_config('customforms').models['response_1']
apps.get_models.cache_clear()
self.user._meta._populate_directed_relation_graph()
self.user._meta._get_fields_cache.clear()
```

I'm not sure what the real fix for this should be. Possibly the customforms tests should handle this in tearDown.

benjaminjkraft:
You both have thought about this more than I have, so I could be totally off-base here, but I would think ideally rolling back a transaction involving custom forms should do the right thing, even if it's not in the context of this test.  (The thing is buggy enough that rolling back a transaction is a thing that might happen, and I wouldn't want it to break the site.)  So perhaps the big custom form creation transaction should be wrapped in a `try`/`finally` that makes sure that if there's an error (or perhaps also if there isn't) it dumps the django models cache?

luac:
It's not that there's an error when creating the form, it's that after the form is created and all its tests are done, the database gets rolled back in preparation for the next tests. This is a situation that is basically ~never expected to happen otherwise, so the test should just clean up after itself.

Wrapping the custom form creation in some sort of try/finally is a good idea, but it won't fix the testing issue.

mgersh:
I think at this point this is reasonably stable, and I'd like to get it merged into main and do any further testing from there. The longer we wait, the more the branches will diverge, and it's inconvenient that there's no reasonable way to do work involving migrations right now. This is at least as well-tested as 1.6 (#955) was when we merged. It's probably not ready to be released to sites yet, but neither is current main.

Any objections?

benjaminjkraft:
:rocket: I agree.  I also looked over the last few commits that I don't think were reviewed, and they look fine to me.  I still don't understand entirely what's happening with custom forms, but I think you two have that covered.

benjaminjkraft:
Hearing no objections, merging.  Thanks @mgersh and @luac for all your work on this!  :star2:

PR #1783, opened 2015-10-06
Upgrade jquery-ui to v1.11.2 (for ajax-spike)

uakfdotb:
Upgrade jquery-ui to v1.11.2 and fix resulting commpanel bug.

This is meant to help transition to the ajax-spike branch, which uses jquery-ui v1.11.2 in the new AJAX scheduler.

### DIFF
--- a/esp/public/media/default_styles/jquery-ui/jquery-ui.css
+++ b/esp/public/media/default_styles/jquery-ui/jquery-ui.css
@@ -1,101 +1,1008 @@
-/*
- * jQuery UI CSS Framework 1.8.18
- *
- * Copyright 2011, AUTHORS.txt (http://jqueryui.com/about)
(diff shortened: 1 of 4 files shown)


uakfdotb:
While this fixes the commpanel, there may be additional uses of jquery-ui in other modules that also break due to the upgrade. I did a search for this particular bug (activate on accordion) and didn't find any, but there are other changes as well.

The old version of jquery-ui was v1.9.2.

uakfdotb:
Going through http://jqueryui.com/upgrade-guide/1.10/ may be helpful, for example we use changestart in lottery_student_reg/student_reg.js, but this is removed in v1.10 according to the upgrade guide, so that is probably broken. (I'll try and look into it tomorrow.)

v1.11 doesn't have as many significant changes.

PR #1784, opened 2015-10-06
Remove unused program/modules/studentregphase2 JS

uakfdotb:
Source commit: 3b6272dbada220b0c07320887ea249c0cd3fff61

Seems to have been added for twophase structure, but not used in current twophase. Other files from the commit referenced above have been correctly renamed from studentregphase2 to studentregtwophase.


benjaminjkraft:
LGTM but it would be good to confirm with @luac or @gkanwar that these are unneeded.

gkanwar:
LGTM, sorry for the dead code.

PR #1786, opened 2015-10-07
Add resource requests to the class search page

### LINKED ISSUE #1584: Class Search page doesn't show resource requests
The results in the class search page don't include the class's resource requests, which is annoying.  (This is more or less independent of #1525.)

mgersh:
Fixes #1584


benjaminjkraft:
:+1:

PR #1791, opened 2015-10-08
Add hosed teachers scheduling check

### LINKED ISSUE #1790: "Hosed teachers" scheduling check
By "hosed teachers", I mean teachers who will be teaching for all or almost all of the times they're available. Somebody asked me today if it's a scheduling check, and I thought that sounded like a good idea. It has a somewhat different purpose from the others but would work well in the same format.

mgersh:
It's a weird scheduling check because it's intended for use before scheduling, not during/after. But I think it's useful and works well in this format. Fixes #1790.

### DIFF
--- a/esp/esp/program/modules/handlers/schedulingcheckmodule.py
+++ b/esp/esp/program/modules/handlers/schedulingcheckmodule.py
@@ -157,7 +157,8 @@ class SchedulingCheckRunner:
           ('teachers_who_like_running', 'Teachers who like running'),
           ('hungry_teachers', 'Hungry Teachers'),
           ('no_overlap_classes', "Classes which shouldn't overlap"),
-          ('special_classroom_types', "Special Classroom Types")
+          ('special_classroom_types', 'Special Classroom Types'),
+          ('hosed_teachers', 'Hosed Teachers')
      ]
 
      #################################################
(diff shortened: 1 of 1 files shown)


esp/esp/program/modules/handlers/schedulingcheckmodule.py
```diff
@@ -553,3 +554,33 @@ def special_classroom_types(self):
          return self.formatter.format_table(mismatches,
                                             {'headings': HEADINGS},
                                             help_text=self.special_classroom_types.__doc__)
+
+     # This isn't really a scheduling check. It's a check that's useful
+     # to run before scheduling. But it works well with the format and
+     # this way everyone else doesn't have to rediscover the round_to
+     # argument to ESPUser.getTaughtTime() every year.
```
benjaminjkraft: TIL.  I always did something even jankier.

esp/esp/program/modules/handlers/schedulingcheckmodule.py
```diff
@@ -553,3 +554,33 @@ def special_classroom_types(self):
...
+         Teachers who have registered almost as many hours of classes
+         as hours of availability. Intended to be run before scheduling,
+         and will not change as classes are scheduled.
+         """
+         teachers = self.p.teachers()['class_submitted']
+         hosed = []
+         for teacher in teachers:
+             # This will break if we ever start having class blocks
+             # that aren't an hour long
+             availability = len(teacher.getAvailableTimes(self.p, ignore_classes=True))
```
benjaminjkraft: I'd suggest adding up the durations of these rather than the lengths, and comparing that to the total duration.  It should be equally simple, and I think some chapters do do non-hour-long class blocks.

esp/esp/program/modules/handlers/schedulingcheckmodule.py
```diff
@@ -553,3 +554,33 @@ def special_classroom_types(self):
...
+             class_hours = teacher.getTaughtTime(program=self.p, round_to=1).seconds/3600
+             delta = availability - class_hours
+             # Arbitrary formula, seems to do a good job of catching the cases I care about
+             if delta <= 0 or class_hours/float(delta) >= 2:
```
benjaminjkraft: `class_hours/float(availability) <= 2/3` is the same and I find easier to understand.

benjaminjkraft:
LGTM, just a few minor comments.

esp/esp/program/modules/handlers/schedulingcheckmodule.py
mgersh: Yeah, I kinda suspected that, it's buried pretty deep in ESPUser and I don't remember ever noticing it before.

esp/esp/program/modules/handlers/schedulingcheckmodule.py
mgersh: I think making this generalize well would be more tricky than I feel like bothering with. None of the big chapters have blocks that aren't an hour long for their Splashes, and distinguishing cases like Stanford's 45-minute-long "hours" from hypothetical 30-minute or 90-minute blocks seems annoying when we really do want to be working in whole hours. Things will get incredibly messy if we try to work with real lengths because of the thing where long classes aren't n*50 minutes long, they're n*60-10.

esp/esp/program/modules/handlers/schedulingcheckmodule.py
mgersh: It's not the same if you write it like that :stuck_out_tongue: 

But yes, I see your point, will fix.

esp/esp/program/modules/handlers/schedulingcheckmodule.py
benjaminjkraft: Ok, happy to leave this then.

benjaminjkraft:
And another :ballot_box_with_check:

PR #1793, opened 2015-10-09
Sort rooms by building number in the scheduler.

luac:
Not sure if we want this, or if it's applicable to other chapters' room numbers, but I was getting annoyed at how 26 came before 4.


uakfdotb:
Can't think of a reasonable case where this would fail.

PR #1794, opened 2015-10-10
Docker-based dev environment

luac:
This is a thing I did some work on a year ago. It allows a dev server environment to be set up in a [Docker](https://www.docker.com/) container, as an alternative to a VM. This has an advantage for people running Linux because the container can share the same kernel and resources as the host, so it's more lightweight. (For people on other OSes, Docker provides a Linux VM image that it can run inside, but then it's still a VM.)

This setup is modeled after the Vagrant/Fabric setup, and it mostly works the same, with some small differences.

Instructions for testing this out (some of this is in comments in the Dockerfile as well):
- Install Docker
- `sudo docker pull learningu/esp-website-base:sr6` (or `sudo docker build -t learningu/esp-website-base .` to build your own; it's basically just a stock Ubuntu image with dependencies installed, I just made and pushed one)
- `sudo docker run -d -v /path/to/local/repository:/home/vagrant/devsite --name esp-website -p 127.0.0.1:8000:8000 -p 127.0.0.1:2222:22 learningu/esp-website-base` (replace /path/to/local/repository with the path to your repo; it will be shared with the container)
- `fab --set container=docker vagrant_dev_setup`
- `fab --set container=docker run_devserver` (check that you get the dev server on localhost:8000)
- `fab --set container=docker manage:shell_plus` (check that the shell works)

I haven't yet looked into importing a database dump. I suspect the Vagrant encrypted database setup isn't straightforward to port over, but it might be possible to do something else instead.


benjaminjkraft:
Ooh, exciting!  Hopefully this can make our dev setup a bit easier.  Two broad questions:

1. What parts of this stack are managed by Vagrant vs. Fabric vs. Docker?  It seems like we're growing a bit of a proliferation of tools here, and if they're all necessary that's okay but it'd be nice if they weren't.  This is especially true if we ever want to move our production servers to Docker, which seems not-unlikely, although it's certainly not worth a lot of engineering effort around.
2. What installation is needed to get this running on a blank Ubuntu install?  Do you still need all the same tools, or can you just install docker and `git clone` and let the container do the rest?

luac:
1. Docker manages the image, and Fabric talks to the image. Vagrant isn't used at all here, except that I made the names of things match with the Vagrant install so that the fabfile would be interoperable. It made sense at the time.
2. What do you mean by "all the same tools"? Docker, Fabric, and git clone should be enough. All the other dependencies, Django etc, should live inside the container.

benjaminjkraft:
1. Good to know.

2. I seem to recall that even with Vagrant, we've had trouble getting things installed because there are still a bunch of things we have to install on the host.  I don't remember exactly what, since I don't personally use a VM, so I don't know if this solves the problem.

luac:
I think it's mostly Fabric (Python and SSL dependencies), and I thought the difficulties with that were mainly on Windows hosts.

luac:
Also, Fabric isn't strictly necessary; you can always just SSH into the container. I think this is possible with the Vagrant VMs too.

benjaminjkraft:
I tried setting this up, and it looks good to me.  I think it would be worth noting in the docs that you shouldn't run this on the same clone of the website from which you run a local dev server, since it sets its own `local_settings.py` on your clone.

benjaminjkraft:
(Also, one of the tests is failing; I'm assuming this is because it's on an old version of main or something?  It's the one on the branch rather than the merge, so I'm a bit less worried.)

luac:
Yeah, this is an old branch.

luac:
I looked into doing the encrypted partition. I have something that works, but it has some issues. In paritcular, Docker containers have to be run as privileged in order to mount anything, and it's kind of sketchy because it creates new devices on the host (and can name-collide with existing devices on the host).

It might be more polite to create and mount the encrypted partition on the host and make that visible inside the container. But this would require installing cryptsetup on the host and the same mess of devices.

luac:
I'm still of the opinion that dev servers should use fake or redacted data instead of encryption, but.

benjaminjkraft:
Can you just make the partition name include, like, the container hash or something, to avoid the collision issue?

luac:
I guess. Apparently `$HOSTNAME` inside the container is a 12-char hash of the container ID (unless it's set explicitly with `docker run -h`, which I'm not concerned about), so the name could be based on that.

Should I do that?

benjaminjkraft:
Sounds good to me.  We could also just make the name a bit more specific --
it wouldn't surprise me if somebody out there already has a
`/dev/mapper/encrypted`, but, say,
`/dev/mapper/esp-website-dbdump-encrypted` or whatever might be fine.  But
the container hash seems simpler and safer.

luac:
So I just learned that there is a Vagrant plugin for LXC containers https://github.com/fgrehm/vagrant-lxc, which addresses most of what I wanted this for (mainly, the not-VirtualBox part), and probably requires fewer changes and less maintenance to support it and VirtualBox at the same time.

Given that, I might try to set that up and scrap this branch if that works.

benjaminjkraft:
Closing this for now since even if we go with this method it will require significant changes per #1834.

PR #1803, opened 2015-10-11
Add (optional) Sentry integration

btidor:
Enable sites to log errors to Sentry. This shouldn't interfere with the existing email/serverlog logging. In addition, the Sentry error ID is displayed on the 500 page, so we can connect user reports to their tracebacks.

### DIFF
--- a/esp/esp/settings.py
+++ b/esp/esp/settings.py
@@ -78,7 +78,7 @@ MANAGERS = ADMINS
 
 TEMPLATE_DIRS = (
     os.path.join(PROJECT_ROOT, 'templates'),
-    
+
 )
 
 DEFAULT_HOST = SITE_INFO[1]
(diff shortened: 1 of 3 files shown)


esp/esp/settings.py
```diff
@@ -115,3 +115,18 @@
...
+    # and PrettyErrorEmailMiddlware) will remain enabled and will receive
+    # exceptions before Raven does.
+    import raven
+
+    INSTALLED_APPS += (
+        'raven.contrib.django.raven_compat',
+    )
+    RAVEN_CONFIG = {
+        'dsn': SENTRY_DSN,
+        'release': raven.fetch_git_sha(os.path.dirname(os.path.dirname(os.path.dirname(__file__)))),
```
benjaminjkraft: Why the triple `os.path.dirname`?  If the point is to move up a few directories, I think `os.path.abspath(os.path.join(__file__, '..', '..'))` is a lot clearer.

benjaminjkraft:
Ooh, cool!  Just to check: I'm assuming that if the log requests to Sentry fail, we'll just ignore or log the error, and not 500, right?

esp/esp/settings.py
benjaminjkraft: (Also, I think `PROJECT_ROOT` is already what you want, although at the moment it's set manually rather than with a relative path.)

btidor: We actually want the directory one above `PROJECT_ROOT`. I changed it to use the same `os.path.join` that we use to construct `PROJECT_ROOT`.

btidor:
Yep! It looks like Raven suppresses connection errors by default (but note that if Raven receives an error, we're going to be displaying a 500 message to the user no matter what).

benjaminjkraft:
> Yep! It looks like Raven suppresses connection errors by default (but note that if Raven receives an error, we're going to be displaying a 500 message to the user no matter what).

Is that true even if it's a `logging.error`/`logging.warning` rather than a `raise FooBarException`?  Will Sentry/Raven log those? (I don't think we use those to log non-fatal errors/warnings currently, but I'd like us to.)

btidor:
No. If we start using the `logging` module, there's a way for Raven to hook in. If we don't use logging now, I'd rather that go in a separate pull request, later.

benjaminjkraft:
Ok, sounds good.

PR #1805, opened 2015-10-11
Make Fabric dev setup a bit more fail-safe

luac:
Handle a couple of exception cases.
- Add fab command to overwrite existing encrypted partition (in case someone loses the encryption password)
- Make link_media do the right thing if run multiple times (instead of erroring out because the symbolic link already exists)

### DIFF
--- a/esp/fabfile.py
+++ b/esp/fabfile.py
@@ -131,8 +131,9 @@ def link_media():
         return
 
     with cd('%s/esp/public/media' % REMOTE_PROJECT_DIR):
-        run('ln -s -T default_images images || true')
-        run('ln -s -T default_styles styles || true')
+        with settings(warn_only=True):
+            run('ln -s -T default_images images')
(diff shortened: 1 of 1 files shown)


btidor:
+1 Looks good to me.

esp/fabfile.py
```diff
@@ -206,10 +215,17 @@ def load_db_dump(dbuser, dbfile):
...
     with use_vagrant():
-        create_encrypted_partition()
+        ensure_encrypted_partition()
         load_encrypted_database(dbuser, dbfile)
 
 @task
+def recreate_encrypted_partition():
+    """ Blow away any previous encrypted partition and create a new one. """
+
+    with use_container():
```
benjaminjkraft: This doesn't exist until #1794 is merged, right?

luac: Oh, shoot.

PR #1811, opened 2015-10-12
Add script to generate list of all classes ever

benjaminjkraft:
I didn't write this, I'm just putting it into git.  It has been used by MIT ESP
to generate lists of classes for use on t-shirts.  It is currently stored on
the MIT ESP wiki, which is a little silly.


mgersh:
A lot of things are silly about this, but given what it is, it's not worth anyone's time messing with it so I'm just going to merge.

PR #1813, opened 2015-10-13
Some new scheduling checks

mgersh:
* Classes that are scheduled, but aren't approved
* Classes that have gaps in them (will also catch classes that are longer than they're supposed to be)
* Fix to hungry teachers

The first three commits here are already in mit-prod-ajax-spike and were used during scheduling.


esp/esp/program/modules/handlers/schedulingcheckmodule.py
```diff
@@ -233,6 +235,24 @@ def classes_which_cover_lunch(self):
                          l.append(s)
           return self.formatter.format_list(l)
 
+     def classes_too_long(self):
+         output = []
+         for sec in self._all_class_sections():
+             start_time = sec.start_time_prefetchable()
+             end_time = sec.end_time_prefetchable()
+             length = end_time - start_time
+             if length.total_seconds()/float(3600) - float(sec.duration) > 0.3:
```
benjaminjkraft: Is it worth also catching classes that are too short?

benjaminjkraft:
:heavy_check_mark: Just one suggestion.

esp/esp/program/modules/handlers/schedulingcheckmodule.py
mgersh: Done.

benjaminjkraft:
:white_check_mark:

PR #1814, opened 2015-10-13
changed form action for search_form to use https when getting data fr…

### LINKED ISSUE #1787: Google search URL is HTTP
Reported by @dehnert:

> I noticed today that the search box on the top right of the ESP website submits to http://www.google.com/search. AFAICT, both the ESP website and Google are HTTPS-only and redirect HTTP connections to HTTPS -- can that search box point to an HTTPS URL? (At the moment, it gives an "insecure connection" warning message for me, which is especially annoying.)

I believe this is an issue in the fruitsalad theme.

gsvolt:
fixes #1787 where we want to fetch results from https://www.google.com instead of http://www.google.com


uakfdotb:
Thanks.

PR #1816, opened 2015-10-14
Feature report: program_size_max=0 is useful

mgersh:
Let's document it so the student reg mainpage can load faster. This eliminates a really slow query on large MIT programs, which in the past have just had program_size_max set to some huge number.


luac:
Why is this a nullable field? What's the difference between null and 0?

mgersh:
There's no difference, null is treated the same as 0. Note that it doesn't have blank=True, the program creation form and the admin panel won't actually let you set it to null.

luac:
Okay. Seems like it would make more sense to have blank=True and use null as the only "no cap" value, but it's not a big difference.

mgersh:
I'd rather not go messing with this to try to get rid of the 0 option right now.

benjaminjkraft:
This seems fine to me; if we want to do anything else we can make an issue at that point.

PR #1817, opened 2015-10-14
Cache have_paid() to speed up the student reg mainpage

mgersh:
The longer-term fix is probably to rework the IAC, since it also causes performance problems elsewhere. But this helps.

Please check the cache dependency, I'm not sure I did it right.


esp/esp/program/modules/handlers/studentregcore.py
```diff
@@ -62,10 +63,12 @@ def module_properties(cls):
             "seq": -9999
             }
 
+    @cache_function
     def have_paid(self, user):
         """ Whether the user has paid for this program.  """
         iac = IndividualAccountingController(self.program, user)
         return (iac.amount_due() <= 0)
+    have_paid.depend_on_row('accounting.Transfer', lambda transfer: {'user': transfer.user})
```
benjaminjkraft: I think it needs to also depend on financial aid grants.

benjaminjkraft:
Thanks, LGTM.

PR #1818, opened 2015-10-14
Monospace Font for QSD

wnavarre:
Set standalone QSD edit page to use a monospace font for editing the HTML/markdown.

### DIFF
--- a/esp/templates/qsd/qsd_edit.html
+++ b/esp/templates/qsd/qsd_edit.html
@@ -64,7 +64,7 @@
   <p>Please use <a href="http://daringfireball.net/projects/markdown/syntax">
      Markdown</a> whenever possible. (Although HTML will work.)
   </p>
-  <p><textarea class="wide" name="content" class="markdown" rows="50" style="font-family: monospace;">{{ content|escape }}</textarea></p>
+  <p><textarea class="wide" name="content" class="markdown" rows="50">{{ content|escape }}</textarea></p>
   <p><input type="submit" name="post_edit" value="Save Changes!" /> &nbsp;| <a href="{{ target_url }}">Cancel</a> | <a href="{{ return_to_view }}">Back to Page View</a></p>
 </form>
 <div id="divmainqsddatetimestamp">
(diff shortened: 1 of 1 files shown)


benjaminjkraft:
Whoa, magic!  I'm going to merge this manually so I can squash it.

benjaminjkraft:
Merged as 9691330e.

benjaminjkraft:
Also, since it may not be clear what the actual change was from the commit message:
screenshot

benjaminjkraft:
Oh, @mgersh just pointed out to me that probably you meant to make both changes that were in this branch.  (One had been reverted.)  I've merged the other one as 2174cdbe.  It now looks like this:
screenshot of QSD editor with formatting bar and markdown

wnavarre:
Wait so the pull request was affected when I did that revert? O.o I didn't expect that. (So is a pull request essentially a pointer to my branch, and whatever I do to my branch mutates the pull request accordingly? I was sort of imagining that a pull request was a 'diff' or a 'patch'). 

Anyway, I reverted that because Favyen said it would be better to assign it to a class that would give us monospace. This solution isn't that bad though because for example the class monospace (in main.less) is kind of bad because it specifies Courier New but I prefer if it's generic monospace so it uses whatever monospace the user prefers.

mgersh:
Yeah, pull requests are branches, and if you push to the branch you'll change the pull request. See for example the giant multi-person mess on #955.

Your second change alone (the class="wide markdown" one) doesn't give us monospace, it gives us what's shown in the first screenshot, which is a toolbar thing that wasn't there before. Both changes together give us both that and monospace.

benjaminjkraft:
> Anyway, I reverted that because @fbastani said it would be better to assign it to a class that would give us monospace.

Yeah, that would be ideal; if you want to make another pull request to fix it, that would be cool.  But we've already got plenty of inline styles, and they're not entirely bad, so I'm fine with it in this case.  You can definitely make another style that uses the default monospace font if you prefer, but if you're going to change it globally you should see what else already uses it, and if we are likely to care.

PR #1820, opened 2015-10-15
Rewrite text to not say "QSD" or "quasi-static"

luac:
QSD is not a descriptive name. It is a bad bad name. Can we get rid of it?


luac:
I also had some partial progress on renaming models and variables, but then I decided that was too much work, at least for now.

esp/esp/qsd/models.py
```diff
@@ -200,6 +200,9 @@ def program_from_url(url):
     def get_absolute_url(self):
         return "/"+self.url+".html"
 
+    class Meta:
+        verbose_name = 'Editable'
```
benjaminjkraft: We may want to look over the LU and MIT ESP wikis to make sure that there aren't references to "Quasi Static Datas on the admin panel".

benjaminjkraft:
I agree; merged.

PR #1821, opened 2015-10-15
Use sudo -H with npm so that it sets $HOME properly

luac:
Since npm tries to read and write $HOME/.npm, there can be issues
running sudo npm when root does not have permissions on $HOME/.npm
(such as if $HOME is in AFS). Fix this by using sudo -H, which sets
$HOME to root's home directory instead.

### DIFF
--- a/esp/packages_base_manual_install.sh
+++ b/esp/packages_base_manual_install.sh
@@ -7,4 +7,4 @@ sudo apt-get install -y python-software-properties
 sudo add-apt-repository -y ppa:chris-lea/node.js
 sudo apt-get update
 sudo apt-get install -y nodejs
-sudo npm install -g less@1.3.1
+sudo -H npm install -g less@1.3.1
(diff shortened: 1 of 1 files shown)


benjaminjkraft:
Won't this cause the user to be unable to see the packages they've installed without root, since the user probably can't see `/root/.npm`?  Since `npm` default install locations are really inconsistent by `npm` version and user config, we may want to do `npm install -g ... || sudo npm install -g ...` or something.  (Or we could actually check what `npm prefix` is, and install as root iff the current user can't read it, but that's harder.)

luac:
Hmm, okay. I was going off of the assumption that `npm prefix` was `/usr` or something, which was the case when I encountered the issue that prompted this. I also assume the intent is to install lessc in a place that is visible from everyone's path, so that you don't have to run update_deps as the same user as the web server.

luac:
Looks like `npm install -g --prefix /usr` works?

benjaminjkraft:
I'd prefer if we did *less* installation of global packages and use of `sudo` than before, but I guess that ship has sailed, so yeah, that should be fine.

luac:
I was about to make this change, but then I hesitated because it might break in a confusing way in case this is run on a platform where /usr isn't a thing. I don't know if we care about those.

benjaminjkraft:
I would think any platform with `sudo` would definitely have `/usr`, so I'd say we don't care.

PR #1830, opened 2015-10-20
Rewrite program cap logic

benjaminjkraft:
This change rewrites the program cap logic, to support new features requested
by Stanford.  In particular, rather than using `Program.isFull()` or
`Program.program_size_max`, all callers should now use
`Program.user_can_join(user)`, which will correctly determine if a user is
already in the program (in which case they can always register for it), and if
not, whether the cap is full.  Module views can use `@meets_cap` which will
display an error page if the user can't join the program.  (The cap can also be
overridden by giving them the "OverrideFull" permission.  Previously this was
implied by "Student/All"; it is no longer.)

In addition, programs wishing to set a per-grade program cap (this was
Stanford's request) may set the program_size_by_grade tag, which is documented
in `Program.user_can_join` and overrides `program.program_size_max`.

This also makes the cap apply to most program modules.  Previously users would
only encounter it when they got to the point of trying to add a class; this is
a bad user experience at best, and meant it was impossible to use a cap for the
lottery.  Now the program cap is checked before you do anything.  If we later
find it necessary to make this more configurable, we can do so at that point.

One thing where this doesn't update the logic is in the lottery script itself.
This should be fine for now, since the cap will prevent more students from
signing up than should be allowed.  If necessary we can add the cap to the
lottery, but that would be difficult.

This also fixes a minor bug: previously the lunch constraints module would not
act correctly if no program cap was set.  (It would set the lunch class's size
to zero.)  Now it will do something reasonable.

I'll be putting this on Stanford's site for them to test sometime in the next
few days.


benjaminjkraft:
Note: I specifically excluded a few module view from this.  I excluded the cancel class/program ones since they should be fine, and if something goes wrong it would be nice if people could still *leave* the program.  I excluded financial aid, because if somehow someone gets to that point and gets a program full error, I don't want it to be misinterpreted as a "we have no more finaid".  I don't feel super-strongly about those decisions, though.

esp/esp/program/controllers/lottery.py
```diff
@@ -290,6 +290,11 @@ def fill_section(self, si, priority=False, rank=10):
         
         #   Compute number of spaces - exit if section or program is already full.  Otherwise, set num_spaces to the number of students we can add without overfilling the section or program.
         num_spaces = self.section_capacities[si] - numpy.sum(self.student_sections[:, si])
+        # TODO(benkraft): This doesn't support program_size_by_grade.  For
+        # Stanford's use case, they don't need it -- they want the cap to
+        # apply to entering the lottery too.  But we should either decide that
+        # that's the Right Way and remove the program size logic from here
+        # entirely, or implement it here too.
```
mgersh: I agree that it's weird to have different kinds of support for program caps in different places, but I think I disagree that it's worse than the alternatives. One thing you could do is leave all the functionality as it is here, but have the lottery script print a warning if program_size_by_grade is set, since it's always run manually.

esp/esp/program/models/__init__.py
```diff
@@ -536,26 +536,121 @@ def volunteers_union(self, QObject = False):
...
+
+        The tag's value, if set, should be a json object.  The keys should be
+        grades (as strings, e.g. "9") or inclusive ranges of grades (separated
+        by hyphens, e.g.  "7-9"), and the values should be caps for that range.
+        Each cap will be checked, so having overlapping caps is possible,
+        although it may hurt performance.
+        """
+        # TODO(benkraft): maybe this shouldn't be a Tag.  For now it basically
+        # has to be because I don't want to deal with migrations until after
+        # we're on django 1.8.
```
mgersh: Once we can add migrations again, yeah, I'd rather have it not be a tag, because I just dislike tags in general and there are other good places to put this kind of thing. The two reasonable places to put it are as a new field on Program (makes sense for consistency with program_size_max, but ew, Program is too cluttered already) or in the SCRMI (inconsistent but this is what SCRMI exists for).

benjaminjkraft: The other problem is that it's more than just a number -- currently as JSON.  I'm not a huge fan of storing JSON -- we have a DB for structured data, so let's use it.  But making a model for this seems kind of overkill.

esp/esp/program/controllers/lottery.py
benjaminjkraft: I'll add a warning for now, and update the comment a bit.  As currently implemented, the fact that the cap (if it exists) applies to entering the lottery is hard-coded, so this logic is useless, but it's possible that that could change in the future if there is demand to make that configurable.  My guess is there won't be -- I think not many programs use the cap as-is, and the fact that until now having "All student deadlines" open would cause it to be ignored supports that theory.

mgersh: It's not useless in the case where you have no program cap while students are entering the lottery and then set a program cap right before running the lottery script, which is what we experimented with (although didn't end up actually using, so this is maybe not the strongest argument) for Summer HSSP.

benjaminjkraft: Right, duh.  In any case, I'm not going to add it now, so I'll just update the comment.

esp/esp/program/controllers/lottery.py
```diff
@@ -68,6 +69,16 @@ def __init__(self, program, **kwargs):
...
+
+        if Tag.getProgramTag('program_size_by_grade', self.program):
+            # TODO(benkraft): Consider implementing this.  Stanford's use case
+            # as of Fall 2015 (for which program_size_by_grade was written)
+            # doesn't need it, but we might want to implement it anyway, or
+            # remove the program size logic from the lottery entirely.
+            print ("WARNING: The lottery doesn't support the "
+                   "program_size_by_grade Tag yet.  It will run based on "
+                   "`Program.program_size_max`, which is %s." %
+                   self.program.program_size_max)
```
jmoldow: Shouldn't we ignore `Program.program_size_max` if `program_size_by_grade` is set? Since the Tag takes precedence over the the model value, if admins accidentally set `Program.program_size_max` (e.g. during program creation), but then later set the tag, then they will run through all of lottery studentreg as if `Program.program_size_max` was unset. But then suddenly we'll get to the lottery and it'll act as if it's set. If `Program.program_size_max` is accidentally smaller than the sum of the capacities in the Tag, then this will lead to students getting kicked out.

And/or we can print the `program_max_size`, the Tag, and the total number of students registered, and prompt the user if they want to continue with the `program_max_size`, ignore it, or quit to fix things up.

Although this isn't a problem with your code review, this made me realize that we're not handling OverrideFull correctly here. It is valid to have more than `program_max_size` students, but the lottery will strictly cap it there. Should we fix this now, or separately?

esp/esp/program/controllers/lunch_constraints.py
```diff
@@ -127,7 +127,12 @@ def get_lunch_subject(self, day):
             new_subject.category = category
             new_subject.class_info = 'Enjoy a break for lunch with your friends!  Please register for at least one lunch period on each day of the program.'
             new_subject.class_size_min = 0
-            new_subject.class_size_max = self.program.program_size_max
+            # If the program doesn't have a max size, we unfortunately still
+            # need one here.  Set a really big one.
+            if self.program.program_size_max:
+                new_subject.class_size_max = self.program.program_size_max
+            else:
+                new_subject.class_size_max = 10**6
```
jmoldow: This can be one line: ``new_subject.class_size_max = self.program.program_size_max or 10**6``.

esp/esp/program/controllers/lottery.py
benjaminjkraft: Fair point, will fix.  I don't think it's worth printing a bunch of information; the user should either realize "this is an issue, let me fix it" and investigate themselves, or realize "I know, it's fine" and continue.  Prompting the user is something the lottery script doesn't currently do and I'd like to keep it that way without a stronger reason to the contrary.

And yes, we're not handling OverrideFull correctly.  I'm not going to fix that now.  My inclination is that the eventual approach should be to remove program caps from here, because I think letting students mark classes in the lottery and *then* picking which ones get into the program is a poor approach that we need not support (in addition to being one that no chapters currently use).

esp/esp/program/controllers/lunch_constraints.py
benjaminjkraft: Good point, will fix.

esp/esp/program/models/__init__.py
```diff
@@ -536,26 +536,121 @@ def volunteers_union(self, QObject = False):
...
-        students_dict = self.students(QObjects = True)
-        if students_dict.has_key('classreg'):
-            students_count = ESPUser.objects.filter(students_dict['classreg']).distinct().count()
-        else:
-            students_count = ESPUser.objects.filter(record__event="reg_confirmed",record__program=self).distinct().count()
+        Returns a dict with tuples of valid grades as keys, and caps as values.
+        """
+        size_tag = Tag.getProgramTag("program_size_by_grade", self)
+        size_dict = {}
+        for k, v in json.loads(size_tag).iteritems():
```
jmoldow: ``Tag.getProgramTag(...)`` defaults to `None`, ``json.loads(None)`` raises. Use ``Tag.getProgramTag(..., default={})``.

esp/esp/program/models/__init__.py
```diff
@@ -536,26 +536,121 @@ def volunteers_union(self, QObject = False):
...
+        This may be very slow if the user is not in the program; unfortunately
+        there's not much we can do about it.
+        """
+        # Check these first because computing the number of students in the
+        # program is slow and uncacheable.
+        if user.canRegToFullProgram(self):
+            return True
+        if self._student_is_in_program(user):
+            return True
+        return self._students_in_program() < self.program_size_max
```
jmoldow: These five lines could be one ``or`` expression.

esp/esp/program/models/__init__.py
```diff
@@ -536,26 +536,121 @@ def volunteers_union(self, QObject = False):
...
+            if '-' in k:
+                low, high = map(int, k.split('-'))
+                size_dict[tuple(xrange(low, high + 1))] = v
+            else:
+                size_dict[(int(k),)] = v
+        return size_dict
+    _grade_caps.depend_on_model('tagdict.Tag')
 
-        isfull = ( students_count >= self.program_size_max )
+    def _user_can_join_at_all(self, user):
```
jmoldow: The caller has these lines:

```python
elif self.program_size_max:
    return self._user_can_join_at_all(user)
else:
    return True
```

But if we were to start calling ``_user_can_join_at_all`` from somewhere else, or if we refactored things, we could accidentally call ``_user_can_join_at_all()`` without doing the ``if self.program_size_max`` check. And this function would produce the incorrect result; it'd always return `False` (or raise an exception) if it reaches the last line.

Can we instead move the ``if self.program_size_max`` check out of the caller, and into the top of this function?

esp/esp/program/models/__init__.py
benjaminjkraft: Yes, they could, and it would be a lot harder to understand IMO.

esp/esp/program/models/__init__.py
benjaminjkraft: This function only gets called if we already know the tag exists.

benjaminjkraft: I decided to add this anyway, but as an `or {}` after the fact, so that it also handles the case of the tag being empty.

esp/esp/program/models/__init__.py
benjaminjkraft: I find this way clearer.  There are 3 main cases based on how the program is configured; `user_can_join` dispatches based on those, and each of those then does the user-specific logic.  This also makes it easy to see at a glance that not setting a `program_size_max` will not only not set a max, but will be more performant, since it won't have to do any of the checks.

It's a private method, and the only public method that calls it now has tests.  All of these private methods may produce incorrect results if called in the wrong circumstance; I'm happy to document that more clearly but I don't think it's a problem, and I think readability wins over a concern about a potential refactor we have no plans to do.

benjaminjkraft:
(Merged main again, I think this is ready to go @luac or @jmoldow.)

PR #1831, opened 2015-10-22
Add script to bulk-create users

benjaminjkraft:
Stanford wants us to create bulk user accounts for outreach schools.  This is a
modification of the script proposed in f0aacab (#1638).


mgersh:
The accounts that were created using this had trouble registering for a program, see the email thread with Stanford. Right now I've put https://github.com/learning-unlimited/ESP-Website/commit/5f1ad9d15d9de65edd8fb00f7aceedb528599e12 on their site as a temporary fix. I haven't looked at whether anything wants to be changed in this script or whether we should fix it on the RegistrationProfile end of things.

esp/useful_scripts/create_users_for_program.py
```diff
@@ -0,0 +1,97 @@
...
+        return Group.objects.get(name=group)
+    else:
+        raise AssertionError('{} is not a Group or Group name'.format(unicode(group)))
+
+
+def create_user_with_profile(username, password, program, groups):
+    print "creating", username
+    user = ESPUser(User.objects.create_user(username=username, password=password))
+    user.groups.add(*groups)
+    profile = RegistrationProfile.objects.create(user=user, program=program)
```
mgersh: Is there a reason why it's useful to create a profile at all here? Maybe this wasn't the case when this was originally written, but now it means that if when the student first logs in they fill out their profile from myesp, then this mostly blank one shows up in getLastForProgram(), and they have grade 0 for the program. Seems like this could be fixed by just not creating one here in the first place, unless I'm missing something. @jmoldow

benjaminjkraft: That sounds reasonable to me, but it would be nice to get confirmation from @jmoldow since he originally wrote it and I think has a better idea than I do of exactly what Stanford wants here.

btidor:
Warning: `ESPUser` casts are deprecated in #1862.

benjaminjkraft:
Yup.  #1862 makes `ESPUser`'s manager inherit from `UserManager`, so we can just make that an `ESPUser.objects.create_user`.

benjaminjkraft:
Ok, I updated for #1862, and also removed the `RegistrationProfile` bit.  @jmoldow does that seem reasonable to you?

PR #1834, opened 2015-10-25
Fab+Vagrant relocation

btidor:
Right now, our Fabric and Vagrant configurations are set up so that `fab` and `vagrant` can only be run from specific directories. It would be nice if these commands could be run from anywhere in the source tree. I believe this is the way we're "supposed" to do it: it simplifies some of our configuration (see 707023e) and I think new webmins will find it easier to understand and will run into fewer errors when they're getting started. However, this does require some migration for existing setups.

### Vagrant
How it works now:
* The Vagrantfile is located at `$PROJECT_ROOT/vagrant/Vagrantfile`
* All Vagrant commands must be run from `$PROJECT_ROOT/vagrant/`

The idiomatic way to do it:
* The Vagrantfile is located at `$PROJECT_ROOT/Vagrantfile`
* Vagrant commands can be run from anywhere in the source tree (just like `git`) -- see "Lookup Path" [in the docs](https://docs.vagrantup.com/v2/vagrantfile/index.html). Note that we could also achieve this behavior without a migration by asking users to set a specific environment variable (although we'd need to find a way to persist that, and we might need to modify the fabfile in addition).

The migration:
* In order to keep using their existing VM, developers will need to run:
```
cd $PROJECT_ROOT
mv vagrant/.vagrant/ .
```

### Fabric
How it works now:
* The fabfile is located at `$PROJECT_ROOT/esp/fabfile.py`
* Fabric commands could in theory be run from anywhere under `$PROJECT_ROOT/esp/`, except that the fabfile doesn't use absolute paths in certain places -- this is fixed in e09eccf

It might be better if:
* The fabfile is located at `$PROJECT_ROOT/fabfile.py`
* Fabric commands can be run from anywhere in the source tree (just like `git` and `vagrant`) -- this is one less edge case for developers to remember. It also seems to be idiomatic -- see "Fabfile discovery" [in the docs](http://docs.fabfile.org/en/1.10/usage/fabfiles.html).

The migration:
* (none)

### Other
* The directory `vagrant/boxes/` appears to contain the configuration we used to generate the original VirtualBox image (now hosted on S3). The scripts aren't used in development or in production, but we should definitely keep them in the repository.
* The directory `esp/config_templates/` holds configuration file templates used by Fabric when setting up a new VM.

I thought it would be nice to consolidate these into a machine/deployment-related directory, `deploy/`. Let me know if that was a bad idea.


luac:
I fixed a minor documentation thing; other than that, I tested this and it works fine for me.

mgersh:
Oh hey, I was in the middle of testing this at the same time. Found one minor issue, but it's a 1.8 issue, not an issue with this, so I'll throw the fix for it in a separate PR.

luac:
This clobbers #1794 kind of unfortunately, but I'm okay with that for now.

PR #1838, opened 2015-10-27
Dev setup and infrastructure cleanup

mgersh:
Various small infrastructure things, including some dev setup bugs that have been around for a while and some remaining 1.8 issues.


luac:
I left a few comments. I also added a separate dev setup related thing that I had locally.

PR #1819, opened 2015-10-14
AJAX scheduler: option to hide unapproved classes.

uakfdotb:
This mimics the text and filtering strategy from the old AJAX scheduler.

PR #1839, opened 2015-10-28
Ajax spike

luac:
This has been used for the past couple of MIT schedulings. If we resolve the merge conflicts, it's probably good to merge.


mgersh:
Note that one of the reasons why we can't merge it right now is because there's a South migration in here and we'll need to make it a Django migration instead.

luac:
It's also failing tests due to something with custom forms, probably a similar issue as the one in the 1.8 branch, though it might go away once we update things here.

luac:
Also, should merge #1819 first.

luac:
Note to self for resolving merge conflicts: something weird happened at bf0c34ff8bca745b12bcc7aa4326e14dd9eae4b5, if you manually try to merge its two parents you get a bunch of merge conflicts, it's not clear that the resolution we want is what was recorded there, and it deserves a closer look.

luac:
I merged main. It looks like it's broken though, probably migrations and/or some other 1.8 thing.

mgersh:
Now the tests are failing because testdata.json is based on very outdated versions of our models (we've changed a lot of things in the last year). I don't know how it was originally generated or what the best way to fix it is.

mgersh:
After thinking about it: we could migrate the initial data by checking out that commit, loading the data, and  then checking out a commit with up-to-date migrations and migrating. However, I'm not sure this is a good idea. It means we'll need to go through the process of migrating the test data every time we change any of the models. Seems like we're probably better off just creating the necessary data in Python during test setup. There's not as much as it looks like, most of that file is South migration history.

mgersh:
And now it's just failing on something about Firefox. Do we even know if this is supposed to work on Travis?

benjaminjkraft:
If it's trying to run selenium tests, I doubt those would work on Travis.

On Wed, Oct 28, 2015 at 3:58 PM, Miriam Gershenson <notifications@github.com

mgersh:
http://docs.travis-ci.com/user/gui-and-headless-browsers/ has instructions for getting it running. It's not working on dev servers either right now, though.

benjaminjkraft:
Ok, I looked at the tests, and I'm pretty convinced the selenium tests in this branch are useless and we can throw them out.  I'll push a commit to do that in a few minutes.  The branch does also have JS tests, which you can run by going to `/javascript_tests`.  They currently pass but I have no idea how much functionality they exercise.  After I remove the selenium tests I'll take a look at how hard it would be to run those automatically.

More details on the selenium tests: I had to upgrade selenium to get it to start at all.  I also had to make sure we inherit from `TransactionTestCase`, not `TestCase` or any of its subclasses, because otherwise the Selenium browser won't see any changes made python-side in the test.  Then I had to fix some bad assumptions that the tests were making (roughly, that they were running with all the same context as `AJAXSchedulingModuleTest` with the methods available there).  After that, one test passed, some of the tests failed in `self.scheduleClass`, and the remainder failed in Selenium assertions or methods.

mgersh:
I reviewed the python-side, non-test-related parts of this, and left comments on various commits. Progress! Or something.

luac:
Resolved merge conflicts and addressed comments, other than the race condition one.

luac:
Okay, so because I was sufficiently annoyed at the prospect of this branch accruing more merge conflicts, I just did a once-over all the JS.

It looks pretty okay to me overall (except for the thing where I now keep thinking this would be better in React, thanks @benjaminjkraft). I think the biggest things are that it would be useful to have an overview doc describing the code organization, and we should figure out how to automatically run the JS tests if they aren't already. Beyond that, there are minor things that can be fixed if someone is enthusiastic about them, but I would be in favor of merging at that point and punting any further issues to new tickets.

Here are a bunch of notes I took while reading it, some of which describe minor issues or places where documentation would help; I don't feel like processing them now, but I'll dump them here in case I don't get to it soon:

* APIClient.js
  * corresponds fairly straightforwardly to the Python endpoints
* Cell.js: an object representing a cell in the table. there is also a cell rendered for each section listed in the directory.
  * what is the ghostSection stuff? (answer: looks like it's the hover effect when a section is selected and you mouse over an available cell)
  * why is .css() called directly to update the cell colors? (answer: because the color is determined based on the section number digits?)
  * tooltip: why not in sections instead? tooltips seemed laggy at Spark scheduling; could we cache/precompute them?
  * typo: Sectinos
* CellColors.js
  * typo: occuppied
  * spacing issue on "section.emailcode [4]"
* ChangelogFetcher.js
* Directory.js
  * each Directory has a number of TableRows
  * Directory.render() -> remove all child elements, create new TableRows, call TableRow.render(), append results as child elements.
  * each TableRow has a Cell that functions as a handle for scheduling that section
  * typo: direcotry
  * this would be better with React, but oh well
  * re-renders on schedule-changed (actually quite expensive, I imagine)
  * is TableRow.hide/unHide() even used?
* Helpers.js
  * some utility functions: helpers_add_timeslots_order, helpers_hex_string_to_color, helpersIntersection
  * should be consistent about snake_case vs camelCase
  * should helpers_add_timeslots_order just be in the Timeslots class, since it's only used once there?
  * similarly, should hex_string_to_color just be in CellColors.js?
  * if we can also get rid of helpersIntersection, we can get rid of this file. (lodash seems to have one?)
* Main.js
  * entry point
  * fetching ajax_schedule_last_changed is asynchronous with setting up the rest of the scheduler; is this okay? why isn't the ChangelogFetcher responsible for this?
* Matrix.js
  * typo: secionInfoPanel
  * the Matrix has many Cells
  * has a pointer to the Sections, Timeslots, and the table of rooms
  * highlightTimeslots has some messy logic for applying the "teacher-available-not-first-cell" highlighting, but it's not unbearably bad
  * why is the tooltip applied to the whole matrix rather than the room cells?
* MessagePanel.js
  * typo: pannel
* Scheduler.js
  * has:
    * a Timeslots
    * a Sections
    * a MessagePanel
    * a SectionInfoPanel
    * a SectionCommentDialog
    * a Matrix
    * a Directory
    * a ChangelogFetcher
  * feature report: keyboard shortcuts! these should be documented.
  * this thing's init does a bunch of data denormalization, initializes all the sub-things, and sets up all the change handlers
  * most of the change handlers call methods on Sections
  * for some reason Scheduler.render() calls ChangelogFetcher.pollForChanges? seems more like init()'s job.
  * also this class doesn't actually have an init(), it just does all the setup directly in its constructor. maybe should be consistent.
  * spacing issue on mouseenter line
* SectionCommentDialog.js
* SectionInfoPanel.js
  * has a pointer to the sectionCommentDialog which is toggled by its "comment" button.
* Sections.js
  * encapsulates the section data, as well as UI state including the selected section, filter, and search term
  * typo: scheule assignments
  * has a pointer to the matrix when bindMatrix is called, 
  * setComment isn't reverted if the API call fails? why? could this lead to inconsistencies where a user locked a section locally and it appears locked to them, but not to others?
* Timeslots.js
  * encapsulates the timeslots data
* spec: contains unit tests for each of the classes and methods. actually looks fairly good, we should figure out how to run them automatically if they aren't already.
* why did this branch touch commpanel.js? student_reg.js?

uakfdotb:
> why did this branch touch commpanel.js? student_reg.js?

jquery-ui was upgraded, which broke some things in those modules that needed to be fixed.

mgersh:
For reference, here's the email @corinnherrick sent about the keyboard shortcuts:

> ESC = unselect currently selected class
DEL = unschedule selected class
F1 = switch to classes tab in right sidebar
F2 = switch to filters tab in right sidebar
/ = go to search tab (right now you have to press F1 first if you're in the filters tab but I'll fix that now.)

luac:
I apologize for introducing more changes, but I just happened to look at some of the server-side code and found some things I could not resist cleaning up.

Could someone review the last four commits? They're pretty small and hopefully shouldn't change functionality.

btidor:
Looks good to me!

btidor:
After talking to @luac, it sounds like the best thing to do is to merge this now and make issues for the remaining improvements.

PR #1844, opened 2015-11-01
Fix off-by-one error in fill_low_priorities logic

luac:
In af05527f6ca5c12046321451f4e4c5b4c5cef55f, in a couple of places,
`range(self.priority_limit+2 if self.grade_range_exceptions else self.priority_limit+1)`
is replaced by `range(self.effective_priority_limit)`, whereas it seems
like it should have been `range(self.effective_priority_limit+1)`.


benjaminjkraft:
:

benjaminjkraft:
(Whoops, misclicked.)  LGTM, thanks.

PR #1846, opened 2015-11-02
Remove a bunch of unused code

benjaminjkraft:
Mostly generated using https://pypi.python.org/pypi/vulture and checked by hand
with `git grep`.


esp/esp/program/modules/base.py
```diff
@@ -389,9 +365,6 @@ def isCompleted(self):
     def prepare(self, context):
         return context
 
-    def getNavBars(self):
```
benjaminjkraft: This appears to have been a way modules could do something magical with navbars.  In any case, despite a number of modules implementing the method, it doesn't appear to be used.

esp/esp/program/modules/base.py
```diff
@@ -483,35 +456,8 @@ def update_props(props):
...
-
-        Returns the modified context.
-
-        This is a stub, to be overridden by subclasses.
-        """
-        return context
-
-        class Meta:
-            app_label = 'modules'
+    class Meta:
```
benjaminjkraft: This was previously in under a method and thus effectively didn't exist.  I think we do want to have it, but it is potentially a slight behavior change and I'm not sure if that's correct.

esp/esp/users/models/__init__.py
```diff
@@ -551,12 +551,6 @@ def addAvailableTime(self, program, timeslot, role=None):
         new_availability, created = UserAvailability.objects.get_or_create(user=self, event=timeslot, role=role)
         new_availability.save()
         
-    def convertAvailability(self):
-        resources = Resource.objects.filter(user=self)
-        for res in resources:
-            self.addAvailableTime(Program.objects.all()[0], res.event)
-        resources.delete()
```
benjaminjkraft: Appears to have been intended for use to convert from the old availibility system.  Presumably this is no longer necessary; it certainly has no callers.

esp/esp/program/modules/base.py
mgersh: Yeah, looks like it was intended to automatically add entries to navbars when the module was enabled? That's not even something we want anymore.

esp/esp/program/modules/base.py
mgersh: Whoops, my bad. Based on the poking around I did before I tried and failed to add this, and the fact that tests are passing, I don't expect any issues.

mgersh:
One question: have you tried running this on ajax-spike and made sure they're all unused there too? A few of these look like someone writing a new scheduler might have found and decided to use them. Maybe unlikely, but I'd rather catch it now than later.

Otherwise, LGTM.

benjaminjkraft:
Good catch.  I did some manual grepping and it looks fine; tests are running on the merged branch `ajax-dead-code`.

benjaminjkraft:
(They passed: https://travis-ci.org/learning-unlimited/ESP-Website/builds/88942278.)

PR #1850, opened 2015-11-03
Move ArgCache out of the repository

luac:
I spent some time over the weekend factoring ArgCache out into a separate repository. It now lives here: https://github.com/luac/django-argcache

This pull request removes the implementation of ArgCache from this repository and changes everything to depend on the new repository instead.

There's a small question of whether/how we want to continue maintaining ArgCache, as an organization. The ESP website is currently its only user, and probably will be for the foreseeable future, so any drive for changes is probably going to come from here. It might also be more convenient to have a learning-unlimited fork of it in case people would otherwise have to wait for me to accept pull requests, especially if I become less involved in ESP website development later. Other than that, I would be happy to maintain it for now. Anyone else have strong opinions on this?


luac:
Oops, didn't mean to press submit just yet.

mgersh:
Huh, cool.

Thoughts about the organizational side of things: It looks like people other than you haven't been changing argcache too often, so having it in a separate repository won't cause frequent issues, and other people's changes have all been either nonurgent cleanup, ESP-specific things, things that could have been added somewhere in another app temporarily if necessary, or maintenance for Django version changes. So that makes me think that as long as you keep generally paying attention, it's unlikely to cause a problem.

My one concern is that if argcache is off somewhere else separated from the code we normally see, new developers are less likely to decide to get more familiar with it, and there won't be anyone to take over maintenance if you stop wanting to do it. On the other hand, plenty of other components also need people who understand them and are less stable than argcache, and we'll never have enough people to pay attention to all the components anyway, so this probably isn't worth worrying about. Future developers can always fork it later if there's a problem later.

esp/esp/cache/apps.py
```diff
@@ -1,14 +1,10 @@
...
+class ESPCacheConfig(ArgCacheConfig):
     """ Makes sure all caches are inserted. Replacement for cache_loader. """
 
-    name = 'esp.cache'
-
     def ready(self):
-        import sys
-        from django.conf import settings
-        from esp.cache.registry import _finalize_caches, _lock_caches
+        from argcache.registry import _finalize_caches, _lock_caches
```
benjaminjkraft: Is there a reason these can't be toplevel imports?  I'm assuming the later ones have to be, but we may as well minimize them.

esp/esp/cache/tests.py
```diff
@@ -1,494 +0,0 @@
...
-            arts1e = reporter1.articles_with_headline('Breaking News')
-            self.assertEqual(arts1d,arts1e)
-        with self.assertNumQueries(0):
-            arts2e = reporter1.articles_with_headline('Article II')
-            self.assertEqual(arts2d,arts2e)
-        with self.assertNumQueries(1):
-            arts3e = reporter3.articles_with_headline('Article II')
-            self.assertNotEqual(arts3d,arts3e)
-
-    @unittest.skip("Known issue, see Github #866")
```
benjaminjkraft: You may want to update this comment in the new one.

benjaminjkraft:
Whoa, so excite!  Thanks for doing this.  Just a minor comment; everything else looks good as far as I can tell.

I'm fine with the organizational things either way; if it doesn't work it should be easy enough to fork it back to learning-unlimited.

esp/esp/cache/__init__.py
```diff
@@ -1,40 +0,0 @@
...
-  527 Franklin St, Cambridge, MA 02139
-  Phone: 617-379-0178
-  Email: web-team@learningu.org
-"""
-
-# Convenience imports
-from esp.cache.function import cache_function, cache_function_for
-from esp.cache.key_set import wildcard
-
-default_app_config = 'esp.cache.apps.CacheConfig'
```
mgersh: Note for the new one: this was actually on my list of things to clean up, since apparently default_app_config is discouraged for new apps, which I missed the first time. The recommended thing is to tell users the app config name to put in their INSTALLED_APPS.

esp/esp/cache/apps.py
```diff
@@ -1,14 +1,10 @@
-from django.apps import AppConfig
+from argcache.apps import ArgCacheConfig
 
-class CacheConfig(AppConfig):
+class ESPCacheConfig(ArgCacheConfig):
```
mgersh: This whole file can just get replaced by uses of the caches.py thing you added in the new one, right?

luac: Yes. I can do that, I was just lazy / wanted to see the tests pass first.

esp/esp/cache/apps.py
luac: No, and I put these to the toplevel in the new non-ESP-specific version of this file (https://github.com/luac/django-argcache/blob/master/src/apps.py).

luac:
Removed the custom cache config in favor of utils/caches.py.

benjaminjkraft:
This looks good to me now.  Since it's a fairly significant change, I'll leave it open for another few days in case anyone has more thoughts or wants to review.

luac:
You're also welcome to review https://github.com/luac/django-argcache/compare/b817eaecdb99c9d66887c5f834a855a0dddaf9f6...master

I tried to make separate commits for copying code straight over and for changing it, so it should be possible to tell where the changes are.

benjaminjkraft:
Random thought, fine if the answer is no: do you want to move `CacheFlushTestCase` into argcache?

benjaminjkraft:
And thanks -- I did review it and on a quick look it looks fine to me.  There are a few minor comments both here and there pointing to things it would be nice to address, but that's all I saw.

luac:
I'm not sure I see the value in moving `CacheFlushTestCase` over, though I might change my mind later.

benjaminjkraft:
:100: Hearing no objections, I think this is good to go!

PR #1851, opened 2015-11-03
Update the increase_caps script

luac:
The main functional change here is to allow the user to retry errors
when they happen.


luac:
Untested, because I didn't want to actually send a bunch of emails and see if I got an SMTP error, but hopefully it works.

luac:
Actually, I just tested the retry logic by replacing the send_mail function with "if random.random() < 0.1: 1/0", and it seems to work.

PR #1856, opened 2015-11-08
Add migration for change of QSD to Editable

benjaminjkraft:
Apparently we were supposed to have a migration for the change of the model's
`verbose_name` in d336fba2, even though it doesn't affect the DB:
https://docs.djangoproject.com/en/1.8/ref/migration-operations/#altermodeloptions
This commit adds that migration.


luac:
How did you find this out?

benjaminjkraft:
I did a `./manage.py makemigrations` to autogenerate migrations for another thing, and it generated this and the one in #1857 as well.

PR #1859, opened 2015-11-09
Remove a number of unused tags

benjaminjkraft:
I did some queries to figure out which tags are used by which sites.  This removes a number of tags that were not set anywhere or had no effect anymore.  I don't think I got all of them; this is just the ones I found before I got bored.

It is probably easiest to review this pull request commitwise.


benjaminjkraft:
Note to self: when this gets merged, make sure to update http://wiki.learningu.org/Customize_behavior_with_Tags.

PR #1862, opened 2015-11-09
Refactor some code around creation of ESPUsers

### LINKED ISSUE #1519: ESPUser inherits from AnonymousUser
This seems silly and likely to cause us to do stupid things.  See #1290 for an example.  That said, changing it might subtly break things.

mgersh:
The immediate reason for this was removing the code which can cast
a Django User or AnonymousUser to an ESPUser, which caused a nasty
mess recently. Making a new auth backend which returns an ESPUser
instead of a User was simple enough. However, AnonymousUser means
there are still some issues.

Previously, ESPUser inherited from both User and AnonymousUser. We're
not sure if inheriting from AnonymousUser was doing anything, but it
was there. Either one would get cast to ESPUser, which would add
a property indicating whether it was anonymous. This means that we're
used to having all of ESPUser's methods available even for anonymous
users, and we don't know to what extent we're using that. It's a
pretty weird way of doing things, though.

So instead, I've put all the methods in a base class, and created
our own version of AnonymousUser. ESPUser and AnonymousESPUser now
inherit from both BaseESPUser and the corresponding Django class.
Any methods which override User will have to go in ESPUser directly,
but so far we've only been doing that to handle anonymity, which is
no longer needed.

Fixes #1519 as part of a bigger set of changes.


esp/esp/utils/auth_backend.py
```diff
@@ -0,0 +1,25 @@
+from django.contrib.auth.backends import ModelBackend
+
+from esp.users.models import ESPUser
+
+class ESPAuthBackend(ModelBackend):
+    """ Subclass of Django's ModelBackend that returns an ESPUser
+    instead of a User. Adapted from
+    http://stackoverflow.com/questions/10682414/django-user-proxy-model-from-request
+    and from django/contrib/auth/backends.py as of 1.8.5 """
```
benjaminjkraft: Might want a comment somewhere explaining why we need both this and the custom middleware.  (I'm assuming it's so that this part can get the custom user, and that one can get the custom anonymous user?)

esp/esp/users/models/__init__.py
```diff
@@ -127,62 +122,23 @@ def save(self, *args, **kwargs):
...
-
-        self.other_user = False
+    other_user = False
 
+    def __init__(self, *args, **kwargs):
+        # This is last in ESPUser's method resolution order, and
+        # AnonymousUser doesn't do anything in its __init__,
+        # so there's no need for a super() call unless you're changing
+        # inheritance structure of this, ESPUser, or AnonymousESPUser,
+        # or if Django has changed something.
```
benjaminjkraft: Is there a reason *not* to have the `super()` call?  It seems like it would be safer to include, in case django changes something and we don't notice.

mgersh: It would be a bit weird, because if I'm just including it for safety it's unclear whether the correct thing to write is
`super(BaseESPUser, self).__init__(*args, **kwargs)`
or
`super(BaseESPUser, self).__init__()`

The first version involves trying to pass arguments to `AnonymousUser.__init__()`, which will break if there's ever anything actually there, and that feels ugly. The second version seems about as vulnerable to breaking as just leaving it out.

That's not actually the original reason, though, the original reason is just that I got really excited about having solved the multiple inheritance puzzle.

esp/esp/utils/auth_backend.py
mgersh: Sure, will do.

esp/esp/users/models/__init__.py
benjaminjkraft: Hehe ok, sounds fine to me.

luac: I feel like either of those might break in a more obvious way than leaving it out, which might be better?

luac: Also, it would have to be the second one in order to work right now, since `AnonymousUser` defines `__init__(self)`.

mgersh:
That was a lot of merge conflicts, mostly with #1927. Done now.

btidor:
Sorry :frowning:

PR #1864, opened 2015-11-10
Remove -qq from travis build commands

benjaminjkraft:
Since Travis automatically collapses output, there's no reason to silence it,
and it sometimes causes our builds to time out.


jmoldow:
:+1:

PR #1866, opened 2015-11-11
Add a custom management command to create superusers properly

btidor:
Right now, running `manage.py createsuperuser` (which is done when starting from an empty database in development) results in an account that isn't a member of `Administrators`. The first problem this creates is that the user can't use the theme editor, which is the first step of setting up an empty site.

It's really hard to override an existing `manage.py` command, so we create our own: `createsuperuseresp`.

### DIFF
--- a/esp/esp/utils/management/commands/createsuperuseresp.py
+++ b/esp/esp/utils/management/commands/createsuperuseresp.py
@@ -1,0 +1,19 @@
+""" Overridden version of createsuperuser """
+
+from django.contrib.auth.management.commands.createsuperuser \
+    import Command as BaseSuperUserCommand
+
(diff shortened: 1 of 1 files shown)


mgersh:
If we're going to do this, we should make dev setup call it instead of calling the normal createsuperuser. I'd also recommend using make_user_admin() instead, and possibly also fixing users.models.install(), which is what used to do this but isn't working anymore (and in fact can make onsite accidentally an admin if you're not careful) because the order in which things happen has changed. Or we can do that last part in a separate PR if you'd like. Thanks for pointing out the issue!

btidor:
Closing in favor of #1869

PR #1868, opened 2015-11-12
Rewrite fabfile.py

btidor:
Our fabfile is big and complicated and doesn't always do the right thing. I've been trying to clean it up, but the easiest solution ended up being restructuring it from scratch. This version cuts out some of the messy logic by consolidating it into a small number of procedural tasks: `fab setup`, which performs one-time setup of a new VM; `fab emptydb` and `fab loaddb`, which configure the database; and `fab refresh`, which deals with Python dependencies, Django migrations and other tasks that need to be run more often.

As part of this refactor, there are some new features and improvements:
* Better prompts and error messages
* Consistent use of absolute paths, which makes the system less fragile and ensures that commands will work no matter where in the source tree `fab` is run
* Symlinks on Windows &mdash; if we create `images` and `styles` as directory symbolic links instead of directory junctions, Virtualbox will display them as Unix symlinks in the VM. We need Administrator rights to create directory symbolic links, so I added a helper script that makes this easier.
* Support for loading compressed Postgres dumps
* Support for loading database dumps from a URL (e.g. directly from Telekinesis)
* Automatic username detection &mdash; when loading a dump, there's no need to specify the name of the chapter/Postgres user to use
* An improved `refresh` command (formerly `post_db_load`, and `update` before that) that upgrades and downgrades Python requirements, efficiently cleans *.pycs, etc.

There are some features I didn't transfer over to the new fabfile. If people are still using these features, please let me know:
* Apache+mod_wsgi is no longer installed by default, since it doesn't seem useful for most developers. I also removed the port forwarding rules and the fab commands for interacting with Apache.
* Encrypted and compressed database dumps (other than ones encrypted or compressed by Postgres itself) are no longer automatically decrypted and decompressed: this is left up to the user.

I've tested this fabfile on Windows, Mac and Linux. Besides some trouble running `manage.py migrate` on the MIT ESP dumps (probably a Django 1.8 issue, the old fabfile has the same problem), it seems to work well. Unfortunately, it's not easy to migrate an existing VM to use this fabfile, so the easiest thing for existing developers to do is to destroy and recreate their VM. If this is a major issue, please let me know.


docs/dev/vagrant.rst
```diff
@@ -51,32 +53,28 @@ Note that you will not be able to see the VM, since it runs in a "headless" mode
...
-If you would like to load a database dump to a system that has already been set up, you may do so with the "load_db_dump" task (which overwrites the existing database on the dev server): ::
+    fab emptydb
 
-    fab load_db_dump:dbuser=chaptername,dbfile=/path/to/chaptername.sql.gz.gpg
+These commands can also be used on a system that has already been set up to bring your database up to date. They will overwrite the existing database on your dev server.
 
 Now you can run the dev server: ::
 
-    fab run_devserver
+    fab runserver
```
mgersh: Yaaaay, thank you, run_devserver was a confusing command name.

docs/dev/vagrant.rst
```diff
@@ -105,26 +94,19 @@ Once you have everything set up, normal usage of your vagrant dev server should
...
 
 To run your dev server: ::
 
-    fab run_devserver
+    fab runserver
 
 Other useful command examples: ::
 
     fab manage:shell_plus
-    fab manage:'migrate program'
```
mgersh: The point of having this in the examples was to demonstrate use of quotes. I'm inclined to leave it in, unless you can think of something more useful to replace it with.

fabfile.py
```diff
@@ -46,258 +68,276 @@
...
 
-def create_settings():
-    context = {
-        'db_user': 'ludev',
-        'db_name': 'devsite_django',
-        'db_password': gen_password(8),
-        'secret_key': gen_password(64),
-    }
+@task
+def psql(cmd=None, *args):
```
mgersh: Ooh, shiny.

fabfile.py
```diff
@@ -46,258 +68,276 @@
...
-def update_deps():
-    sudo('%s/esp/update_deps.sh --virtualenv=%s' % (REMOTE_PROJECT_DIR, REMOTE_VIRTUALENV_DIR))
+    """
+    Run a manage.py command in the remote host.
+    """
+    ensure_environment()
+
+    if cmd in ["shell", "shell_plus"]:
+        # cd() doesn't work with open_shell
+        open_shell("(cd " + env.rbase + "esp && python manage.py " + cmd + "); exit")
```
mgersh: Woah, is this a fab thing we didn't know about? :smiley:

mgersh:
It might take me a while to get around to testing this and looking at it in more detail, but I'm excited!

fabfile.py
```diff
@@ -46,258 +68,276 @@
...
-
-@task
-def update_deps():
-    sudo('%s/esp/update_deps.sh --virtualenv=%s' % (REMOTE_PROJECT_DIR, REMOTE_VIRTUALENV_DIR))
+    """
+    Run a manage.py command in the remote host.
+    """
+    ensure_environment()
+
+    if cmd in ["shell", "shell_plus"]:
```
mgersh: I think this wants to be `if "shell" in cmd`, or something else that correctly handles running shell_plus with arguments.

docs/dev/vagrant.rst
btidor: How about: `fab psql:"SELECT * FROM pg_stat_activity"`?

fabfile.py
btidor: [open_shell](http://docs.fabfile.org/en/1.10/api/core/operations.html#fabric.operations.open_shell): it's supposed to provide a "better interactive experience" by not printing the hostname on every line, and doing some terminal-related stuff I'm unclear on.

docs/dev/vagrant.rst
mgersh: Sure, that's a good one. I probably wouldn't run it on my dev server in real life, but I won't complain about more people knowing it exists, and it points out the existence of the `psql

fabfile.py
```diff
@@ -46,258 +68,281 @@
...
-    link_media()
+def gen_password(length):
+    return "".join([random.choice(string.letters + string.digits) for i in range(length)])
 
 @task
-def run_devserver():
-    """ Run Django dev server on port 8000. """
-    with esp_env():
-        sudo('python manage.py runserver 0.0.0.0:8000')
+def refresh():
```
mgersh: This doesn't fully replicate ./manage.py update, and it shouldn't have to--I think we should either continue recommending 'fab manage:update' for that, or have this call ./manage.py update, to avoid code duplication.

It looks like you've copied some of this from my post_db_load() in the old version. That was intended specifically to be what needs to happen after loading a db dump, not at any other time. I think it's still worth having update code specifically for that purpose, which isn't trying to also be a general purpose update task, since what you need after loading a db dump and what you need after pulling new code will always be fairly different.

fabfile.py
```diff
@@ -46,258 +68,281 @@
...
+        # Clean up *.pyc files that contain stale code. In practice, the only
+        # files that cause problems are orphaned *.pyc's where the corresponding
+        # *.py file has been deleted (git leaves the *.pyc in place). Other than
+        # this, *.pyc files cannot be stale because git updates the lastmod time
+        # when changing a file, so a freshly checked out *.py file will always
+        # be newer than its *.pyc, triggering a recompile. This is an optimization
+        # over 'manage.py clean_pyc', which deletes all *pyc's in the source
+        # tree.. Note that this command is run in the Vagrant VM, since the host
+        # may not have find installed.
+        run("""find . -name '*.pyc' -exec bash -c 'test ! -f "${1%c}"' -- {} \; -delete""")
```
mgersh: Maybe instead of putting this here, look into replacing the clean_pyc call in manage.py update with it?

fabfile.py
```diff
@@ -46,258 +68,281 @@
...
+
+        # Run Django migrations
+        manage("migrate")
+
+        # Recompile theme (run twice, to work around bug)
+        manage("recompile_theme")
```
mgersh: The specific bug this was intended to work around: after you load a database dump with a customized theme, the first recompile_theme call will fail, and the second will succeed. This won't work without the warn_only=True from the old version, or something that has a similar effect. It's not necessary when just changing code.

fabfile.py
```diff
@@ -46,258 +68,281 @@
...
+
+      - upgrades and downgrades Python packages
+      - removes orphaned *.pyc files
+      - runs Django database migrations
+    """
+    ensure_environment()
+
+    with cd(env.rbase):
+        # Update the virtualenv with correct packages
+        run("pip install --upgrade -r " + env.rbase + "esp/requirements.txt")
```
mgersh: I like that this command does things relevant to both update_deps and manage.py update. Is there a reason why it does Python dependencies but not Ubuntu packages?

docs/dev/vagrant.rst
```diff
@@ -105,26 +94,20 @@ Once you have everything set up, normal usage of your vagrant dev server should
...
-Functionality that is lacking
------------------------------
-
-This is a TODO list for the developers:
+One last command! When your devserver gets out of date, this command will update the dependencies, run migrations, and generally make things work again: ::
 
-* Support deploying to other targets (other than Vagrant VMs) - could be useful for deployment
-* Make things more customizable
-* Reduce number of setup steps
+    fab refresh
```
mgersh: See below comments about whether this is actually what we want.

fabfile.py
```diff
@@ -46,258 +68,281 @@
...
+    else:
+        with cd(env.rbase + "esp/public/media"):
+            with settings(warn_only=True):
+                run("ln -s -T default_images images")
+                run("ln -s -T default_styles styles")
+
+    sudo("touch /fab-setup-done")
+
+@runs_once
+def ensure_environment():
```
mgersh: As long as we have this anyway, it would be cool if it also checks that you've either run loaddb or emptydb. Right now, when you go straight from setup to runserver, you get a confusing error. (I checked because I was curious.)

mgersh:
I tried this, and it generally worked. Definitely better than the old one! I ran into a weird memcached error towards the end of setup, but I think it was probably either a blip or caused by something weird I did on purpose. I don't feel like waiting for the whole setup process again to see if it still happens.

I left some comments, mostly about the refresh task, otherwise looks good. You might also want to draft an email to send out once this is merged, since everyone will need to know to reinstall their dev server and relearn the commands.

:star2:

fabfile.py
btidor: Done!

mgersh:
Looks great! I like the addition of `vagrant ssh` for `shell_plus`.

Question before we actually merge: how do you want to deal with the backwards incompatibility issue? Immediately merging main into mit-prod and backporting to the stable release is one option, if you want to do that. Or people might just need to keep two dev servers around for a while, like I've been doing for the Django version change. We'll also need to make sure everyone who doesn't read their Github email knows they have to reinstall.

PR #1869, opened 2015-11-14
On granting an account is_superuser, add to Administrators group

btidor:
Another attempt at #1866

### DIFF
--- a/esp/esp/users/models/__init__.py
+++ b/esp/esp/users/models/__init__.py
@@ -998,6 +998,16 @@ class ESPUser(User, BaseESPUser):
         proxy = True
         verbose_name = 'ESP User'
 
+    def makeAdmin(self):
+        """
+        Make the user an Adminstrator and a Django superuser.
+        """
+        # Django auth
+        self.is_staff = True
+        self.is_superuser = True
+        self.makeRole("Administrator")
+        self.save()
+
 class AnonymousESPUser(BaseESPUser, AnonymousUser):
     pass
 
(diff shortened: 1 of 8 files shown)


esp/esp/users/models/__init__.py
```diff
@@ -1038,6 +1038,13 @@ def getRankInClass(student, subject, default=10):
             rank = default
         return rank
 
+@dispatch.receiver(signals.post_save, sender=User,
+                     dispatch_uid='make_admin_save')
+def make_admin_save(sender, instance, **kwargs):
+    if instance.is_superuser:
+        from esp.users.views.make_admin import make_user_admin
+        make_user_admin(ESPUser(instance))
```
mgersh: I like this approach, but this way of writing it won't work once #1862 is merged.

btidor: Could you explain the issue?

mgersh: Ah, sure. The major change in that pull request is removing the ability to cast from User to ESPUser, so doing ESPUser(instance) won't work anymore. You'll have to either get an ESPUser some other way, or make make_user_admin() accept a User (and leave a comment so it stays that way).

Mentioning this for the record since I missed it at first: I'm pretty sure this will only catch saves of User objects, not saves of the equivalent ESPUsers. I don't have a problem with that, since it does the thing you were really trying to do, but it might be a good idea to comment it, and maybe also mention why this receiver is useful.

benjaminjkraft:
Ok.  Not crazy about the potential for circularity, but I guess it's fine.

PR #1870, opened 2015-11-15
Update the migration check script for new migrate --list format.

luac:


esp/useful_scripts/migration_check.sh
```diff
@@ -3,7 +3,7 @@
 # Pipe the output of "manage.py migrate --list" to this script.
 # It will say whether there are unapplied django-south migrations.
 
-MIGRATIONS="`grep -v '(\*)' | grep -1 '( )'`"
```
luac: I'm not sure what -1 was supposed to do. I can't find it in grep's manpage.

benjaminjkraft: Looks like it's a synonym for `-C 1`.

luac: Ah. We might want that then, or at least `-B 1`, since most of the lines themselves just say "[ ] 0001_initial".

Issue #1872, opened 2015-11-15
Move mailgate and dbmail_cron to use script_setup


benjaminjkraft:
They do a bunch of manual path-munging, where they could just use script_setup.  Presumably this is just because they predate script_setup, but we'll need to make sure it doesn't break anything to switch.

btidor:
Just FYI, script_setup was designed for our useful scripts, so it sets up the environment in a way that might not be appropriate for "real" code. In particular, it magically imports everything in the way that shell_plus does.

Other good options could be to migrate these scripts to manage.py commands, or to refactor script_setup into a core part that sets up the environment like a Django script, and an extra part that sets up the environment like shell_plus.

PR #1873, opened 2015-11-15
Fix mail system for 1.8

mgersh:
I'm not sure if this is the correct fix, but it seems like people can send mail again, and I want to put this here so we don't lose track of it. See also #1872.


benjaminjkraft:
Seems fine to me.

PR #1876, opened 2015-11-16
Prevent people from generating many student schedules as PNG

### LINKED ISSUE #1874: student schedules should not let you generate PNG for multiple students
We support a bunch of formats, including PNG.  We're only set up to print a single-page schedule to PNG (we just print the first page if it's multiple pages), but if you try anyway, it will try to `convert` the whole thing to PNG and just return the first one, which will eat up all the server's memory if the PDF it's trying to convert is huge.  So we just shouldn't let you do that.

benjaminjkraft:
Generating PNG schedules for a lot of students will cause
`convert` to use a huge amount of memory and make the server sad.
It also doesn't work, since we just return the first page of the
PNG anyway.  So don't let people do that.  This has caused issues on both
diogenes and esp.mit.edu recently.  Fixes #1874.

### DIFF
--- a/esp/esp/program/modules/handlers/programprintables.py
+++ b/esp/esp/program/modules/handlers/programprintables.py
@@ -900,7 +900,7 @@ Volunteer schedule for %s:
             else:
                 file_type = 'pdf'
 
-        if len(student) > 1 and file_type == 'png':
+        if len(students) > 1 and file_type == 'png':
             # Generating PNG schedules for a lot of students will cause
             # `convert` to use a huge amount of memory and make the server sad.
             # It also doesn't work, since we just return the first page of the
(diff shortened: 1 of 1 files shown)


esp/esp/program/modules/handlers/programprintables.py
```diff
@@ -890,6 +890,24 @@ def get_student_schedules(request, students, prog, extra='', onsite=False):
...
+            file_type = extra.strip()
+        elif 'img_format' in request.GET:
+            file_type = request.GET['img_format']
+        else:
+            if onsite:
+                file_type = 'png'
+            else:
+                file_type = 'pdf'
+
+        if len(student) > 1 and file_type == 'png':
```
jmoldow: Typo: should be ``students`` (this caused tests to fail).

jmoldow:
Other than typo, looks good to me.

esp/esp/program/modules/handlers/programprintables.py
benjaminjkraft: Whoops, thanks, fixed.

PR #1877, opened 2015-11-16
Use subprocess.call() for gen_latex() type 'log'

jmoldow:
When using `gen_latex()` to generate a latex log file, don't use
`subprocess.check_call()`. Instead, use `subprocess.call()`. The
command might fail, but we still want to read the log file.

### DIFF
--- a/esp/esp/utils/latex.py
+++ b/esp/esp/utils/latex.py
@@ -36,7 +36,7 @@ Learning Unlimited, Inc.
 
 import os.path
 import os
-from subprocess import check_call, STDOUT
+import subprocess
 from random import random
 from functools import partial
 import hashlib
@@ -73,7 +73,7 @@ def render_to_latex(filepath, context_dict=None, filetype='pdf'):
 
     return gen_latex(rendered_source, filetype)
 
-def gen_latex(texcode, type='pdf', remove_files=False, stdout=_devnull_sentinel, stderr=STDOUT):
+def gen_latex(texcode, type='pdf', remove_files=False, stdout=_devnull_sentinel, stderr=subprocess.STDOUT):
     """Generate the latex code.
 
     :param texcode:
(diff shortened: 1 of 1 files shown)


esp/esp/web/util/latex.py
```diff
@@ -36,6 +36,7 @@
 
 import os.path
 import os
+import subprocess
 from subprocess import check_call, STDOUT
```
btidor: This pair of imports is sort of ugly. Can we pick one or the other to use?

esp/esp/web/util/latex.py
```diff
@@ -149,7 +150,11 @@ def _gen_latex(texcode, stdout, stderr, type='pdf', remove_files=False):
 
     elif type == 'log':
         mime = 'text/plain'
-        call(['latex'] + latex_options + ['%s.tex' % file_base])
+        # Use `subprocess.call()`, not `subprocess.check_call()`, because the
+        # command doesn't need to succeed in order to read the generated .log
+        # file. In fact, for the use case of this type (trying to debug why
+        # compilation isn't working), it is likely that the command will fail.
+        subprocess.call(['latex'] + latex_options + ['%s.tex' % file_base], cwd=TEX_TEMP, stdout=stdout, stderr=stderr)
```
btidor: This gets pretty confusing in combination with the other definition of `call`. I might make our `call` into a function that can either use `subprocess.check_call` or `subprocess.call` based on a flag, and maybe rename it to something else, too.

btidor:
Other than that, looks great!

esp/esp/web/util/latex.py
benjaminjkraft: It looks like it's just a `partial` defined above, so it might be better to just rename it to `check_call` and then make a similar `call` wrapper.

jmoldow: Went with @benjaminjkraft 's suggestion.

esp/esp/utils/latex.py
```diff
@@ -139,29 +139,35 @@ def _gen_latex(texcode, stdout, stderr, type='pdf', remove_files=False):
     latex_options = ['-interaction', 'nonstopmode', '-halt-on-error']
 
     # All command calls will use the same values for the cwd, stdout, and
-    # stderr arguments, so we define a partially-applied callable call()
-    # that makes it easier to call check_call() with these values.
-    call = partial(check_call, cwd=TEX_TEMP, stdout=stdout, stderr=stderr)
+    # stderr arguments, so we define a partially-applied callable check_call()
+    # that makes it easier to call subprocess.check_call() with these values
+    # (and similarly for check()).
```
benjaminjkraft: I think you mean `call()`.

benjaminjkraft:
Minor typo in a comment, otherwise LGTM

esp/esp/utils/latex.py
jmoldow: Fixed.

PR #1885, opened 2015-11-19
Fix #1649: hide multi-block classes starting in the past

### LINKED ISSUE #1649: Class changes: "hide past timeblocks" should hide multi-hour classes
Requested at MIT Spark: "class changes: if you hide past blocks it should hide classes that started in the past, even if they continue into the present"

luac:
When hiding past blocks, hide classes that started in the past, even
if they continue into the present.

### DIFF
--- a/esp/public/media/scripts/onsite/ajax_status.js
+++ b/esp/public/media/scripts/onsite/ajax_status.js
@@ -730,9 +730,9 @@ function render_table(display_mode, student_id)
             //  Hide the class if it started in the past (and we're not showing past timeblocks)
             if (settings.hide_past_time_blocks && section.timeslots.length > 1)
             {
-                for (var i in section.timeslots)
+                for (var j in section.timeslots)
                 {
-                    var sec_ts_id = section.timeslots[i];
+                    var sec_ts_id = section.timeslots[j];
                     var startTimeMillis = data.timeslots[sec_ts_id].startTimeMillis;
                     //excludes timeslots that have a start time 20 minutes prior to the current time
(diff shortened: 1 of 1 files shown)


esp/public/media/scripts/onsite/ajax_status.js
```diff
@@ -726,7 +726,22 @@ function render_table(display_mode, student_id)
             new_div.append($j("<span/>").addClass("room").html(section.rooms));
             //  TODO: make this snap to the right reliably
             new_div.append($j("<span/>").addClass("studentcounts").attr("id", "studentcounts_" + section.id).html(section.num_students_checked_in.toString() + "/" + section.num_students_enrolled + "/" + section.capacity));
-            
+
+            //  Hide the class if it started in the past (and we're not showing past timeblocks)
+            if (settings.hide_past_time_blocks && section.timeslots.length > 1)
```
benjaminjkraft: Should this not go up above with the other check?  I don't entirely understand the code here but it seems like this is unnecessarily inside a loop.

esp/public/media/scripts/onsite/ajax_status.js
```diff
@@ -726,7 +726,22 @@ function render_table(display_mode, student_id)
...
+            {
+                for (var i in section.timeslots)
```
benjaminjkraft: If this is staying here, it's a bit confusing that this `i` is distinct from the one in the parent loop.

esp/public/media/scripts/onsite/ajax_status.js
luac: Yeah, "rule of three".

There might be a smarter way of doing this, but I didn't think of one at the time.

PR #1888, opened 2015-11-19
Move esp.web.util into esp.utils

benjaminjkraft:
For historical reasons, we have two general-utils directories, and that's silly.  This pull request moves what was formerly in `esp.web.util` to `esp.utils`, and renames the imports.  We may also want to move some of the template tags in `esp.web` to `esp.utils`, but I left that for another pull request since it might be nice to simultaneously remove `esp.utils.globaltags` (see discussion on #1779).


mgersh:
LGTM except for a stray reference in documentation, which I just fixed.

benjaminjkraft:
Thanks!

PR #1898, opened 2015-11-22
Add default email template

milescalabresi:
Adds editable default HTML template for emails sent through the comm panel. The Travis build failed, but I believe this is (now, after a few fixes) because of the test. The code works when I tested by hand, and the Travis build appears to test the message body in a way that assumes the old code.
Of course, I could be wrong, so someone who knows Travis better than I do should verify, and perhaps update the build if this change seems like something worth implementing


benjaminjkraft:
Yep, it looks like the test is failing because the exact test changed.  So you can just update the test to use the correct message body instead of the one it has.

esp/esp/program/modules/handlers/commmodule.py
```diff
@@ -109,6 +110,11 @@ def commprev(self, request, tl, one, two, module, extra, prog):
         esp_firstuser = ESPUser(firstuser)
         contextdict = {'user'   : ActionHandler(esp_firstuser, esp_firstuser),
                        'program': ActionHandler(self.program, esp_firstuser) }
+
+        htmlbody = unicode(loader.get_template('email/default_email_html.txt').render(DjangoContext({'msgbdy': body,
```
benjaminjkraft: I think the more standard thing is to use `template.render_to_string()` rather than `unicode()`, and in that case I don't think you need to call the `Context` instance yourself, you just pass a dict.

esp/esp/program/modules/handlers/commmodule.py
```diff
@@ -175,7 +181,10 @@ def commfinal(self, request, tl, one, two, module, extra, prog):
                                                       sendto_fn_name  = sendto_fn_name,
                                                       sender     = fromemail,
                                                       creator    = request.user,
-                                                      msgtext    = body,
+                                                      msgtext = unicode(loader.get_template('email/default_email_html.txt').render(DjangoContext(
```
benjaminjkraft: Likewise here.  Also, nit: can this either be on one line or be wrapped to a reasonable length?

esp/templates/email/default_email_html.txt
```diff
@@ -0,0 +1,5 @@
+<html>
```
benjaminjkraft: I think this is redundant -- we tell users to include their own `<html>` if they want HTML.

benjaminjkraft: Also, this file should either be called `default_email.html`, if it's html, `default_email.txt`, if it's text, or `default_email` if it could be either.

esp/esp/program/modules/handlers/commmodule.py
benjaminjkraft: Also, I think you want `htmlbody` instead of `body` here, otherwise lines 105-108 are meaningless.  Also also, passing the user and program in here is unnecessary.

esp/esp/program/modules/handlers/commmodule.py
```diff
@@ -175,7 +181,10 @@ def commfinal(self, request, tl, one, two, module, extra, prog):
...
+                                                                                                                                       {'msgbdy': body,
```
benjaminjkraft: Here, conversely, you don't need to pass `msgbdy`.

esp/templates/program/modules/commmodule/step2.html
```diff
@@ -38,7 +38,8 @@
 Please write your email/message now. Note that you can use <tt>{{ parameter }}</tt> to enter a parameter
 into the text of the message&mdash;the parameters are listed below.
 
-<p><b>Tip:</b> If you would like to send your message in HTML format, wrap your entire message body in &lt;html&gt; and &lt;/html&gt; tags.</p>
+<p><b>Tip:</b> The default {{ settings.ORGANIZATION_SHORT_NAME }} template (which you can modify <a href="/admin/utils/templateoverride/">here</a>) is automatically wrapped around your message below.
+Don't forget to write the body in HTML, but no need for &lt;html&gt; and &lt;/html&gt; tags.</p>
```
benjaminjkraft: We should still allow writing plain-text emails, because it's easier and if you don't need any formatting it's better.  (Ideally we may want to switch to Markdown for emails, because that's both easier and can be used to make pretty things.)

esp/templates/email/default_email_html.txt
```diff
@@ -0,0 +1,5 @@
...
+{% autoescape off %}
+{{ msgbdy }}
+{% endautoescape %}
```
benjaminjkraft: I wonder if the default template should include an unsubscribe footer that links to the disable account page.

benjaminjkraft:
Ooh, cool!  I'm assuming that the point of this is to make it easier for chapters to add a standard header/footer to their message?

esp/templates/program/modules/commmodule/step2.html
```diff
@@ -38,7 +38,8 @@
 Please write your email/message now. Note that you can use <tt>{{ parameter }}</tt> to enter a parameter
 into the text of the message&mdash;the parameters are listed below.
 
-<p><b>Tip:</b> If you would like to send your message in HTML format, wrap your entire message body in &lt;html&gt; and &lt;/html&gt; tags.</p>
+<p><b>Tip:</b> The default {{ settings.ORGANIZATION_SHORT_NAME }} template (which you can modify <a href="/admin/utils/templateoverride/">here</a>) is automatically wrapped around your message below.
```
benjaminjkraft: You can actually probably link straight to `/admin/utils/templateoverride/?name=default_email_html.txt` or something to automagically do the search/filter for the right TO (if it exists).

milescalabresi:
Yes exactly! Neat idea, right? Yale Splash came up with the idea after people got tired of typing the "click here to unsubscribe" footer and wanted it automatically inserted. Now people can add, for instance, inline images from the filebrowser or make the emails more phone-friendly without having to copy-paste the same markup background text each time they use the comm panel.

As you can see, I executed this whole thing a little sloppily. I really appreciate your detailed comments, by the way! I will get to work on fixing it up. Regarding the <tt>loader.render_to_string</tt> command, I saw that, but when I tried to use it, it prematurely rendered the file so user-entered parameters would show up unrendered. Any idea on how to fix that without doing my hack?

esp/templates/program/modules/commmodule/step2.html
milescalabresi: Yeah, my thought was to include a default HTML email that did something basic (such as replacing newlines with paragraph tags, adding some kind of standard unsubscribe message, and maybe some phone-compatibility things) but benign so that it will automatically format it a bit for them. Most of the chapter board was very enthusiastic about this idea, and it seems nice to include some baseline level of professionalism (it's actually illegal to send mass emails like ours without some unsubscribe instructions, for instance, and unformatted plain text doesn't look very nice). Of course, programs could modify this file as they wish (that's the whole point), but having a default with some easy prettify-ing HTML in there seems like a nice way to add a little style to the emails. But of course I'm open to discussion on this point

esp/templates/program/modules/commmodule/step2.html
milescalabresi: That's a great idea! Thanks! I forgot you could search by name. I'll add that in once we settle on a good name for file, which will probably depend on what we decide for [this guy](https://github.com/learning-unlimited/ESP-Website/pull/1898/files#r45942372).

esp/templates/email/default_email_html.txt
milescalabresi: :+1: :+1: :+1: and maybe some phone-compatibility code? If we include unsubscribe footers (I have a good one of those that I can add) and other stuff, then does that tie us into making them use HTML or Markdown?

esp/templates/program/modules/commmodule/step2.html
benjaminjkraft: If we want to do that, we should actually do that, probably by converting markdown (perhaps optionally).  You can look at how we do it for QSD as a guide.  If we do that I'm fine not allowing plaintext, but as I understand it, as-is this will just eat the newlines and such, and I do think for short emails unformatted text looks fine.

Incidentally, I believe CAN-SPAM doesn't actually apply to our use case since we are non-commercial and all users have signed up for our mailing list by registering an account, although we certainly *should* include unsubscribe links.

esp/templates/email/default_email_html.txt
benjaminjkraft: If we're allowing both HTML and plain text, we probably actually want to have two templates; the plain-text one can just include the URL.

What do you mean by phone-compatibility code?

benjaminjkraft: Also, if we want, we could consider including a per-user unsubscribe link that doesn't require logging in, since making people log in to unsubscribe is really annoying.  It would require a bit more code so perhaps it's better to do that separately.

esp/templates/program/modules/commmodule/step2.html
benjaminjkraft: Also, while we're considering this, we may want to think about whether we should always be attaching a plain-text version of HTML emails, since there are still some email clients that will be sad if you don't.  If we do markdown and people don't include raw HTML tags, this is fairly easy; in the more general case it could be a lot trickier.  So we may not want to bother, or if we do, may want to do it in a separate pull request.

benjaminjkraft:
How were you calling `render_to_string`, and what was the output?

esp/templates/email/default_email_html.txt
milescalabresi: I mean code that makes emails display more nicely on phones (there are sometimes issues with rotating screens and font sizing that you can throw enough HTML at to make them vanish).

The per-user link would be nice if that's possible. We can also include an "email us if you want to unsubscribe" if the additional code wouldn't be worth the hassle.

benjaminjkraft: Huh, ok, cool.  If by "email us if you want to unsubscribe" you intend for a human to handle it, that's fine; we can't really do an automated thing since the from/reply-to could be an arbitrary address controlled by the chapter.

esp/templates/program/modules/commmodule/step2.html
milescalabresi: I think it would be nice to do the full-blown version. I'll take a look at the markdown conversion. Good call on including the plain-text version though. That seems worth doing. And maybe a line to the effect of "click here to see this email in your browser" if we can manage that without too much trouble?

I'm kind of tempted to force admins to write in something that looks nice. We could include an optional if-statement to catch emails under 50 words with no formatting tags (or whatever) that would just wrap them in <tt>&lt;p&gt;</tt> tags and add the default headers/footers and for emails over 50 words, require admins to use HTML. Thoughts on that?

esp/templates/email/default_email_html.txt
milescalabresi: Yes, sorry. I mean for humans to handle the request if logging in is annoying or they forgot their password.

benjaminjkraft: If we have a nice template, making a 1-click unsubscribe wouldn't be too hard -- there might even be a library that will do most of the work for us.  But in any case, we can do that after this is merged.

milescalabresi:
Oh, on second thought, I might have done that before I remembered to turn autoescaping off. I'll check again after fixing the other stuff. I'll probably open a new pull request, depending on what we decide on the HTML/plain text/Markdown question.

esp/templates/program/modules/commmodule/step2.html
benjaminjkraft: "Click here to see this email in your browser" is fairly reasonable; we'd just have to render the same template into a page.  The only problem is that we'd either have to always make people log in for it, or somehow make it user-specific without doing that, which is a little tricky to do correctly (i.e. without leaking user data).  But it's probably doable.

Markdown seems better (and probably easier) than trying to autowrap `<p>` tags ourselves, and should do that for us.

Incidentally, we'll want to make sure the release notes of whatever release this goes in include a mention of the fact that the defaults have changed, since some of these are breaking changes.

milescalabresi: Okay, cool. I don't know how to safely render emails without making them log in, but I can search around a bit.

Sounds good. I'm not super familiar with how Markdown works, but I'll check out the QSD-conversion code and see if I can figure out how to adapt it.

Good call on the release notes. At this point, I think I should open a new pull request after making some changes to this one.

To summarize, next steps are:
1. Look into making a more general code that will permit admins to enter plain or marked-up text from the comm panel and merge their message with the default HTML template.
2. Expand the current default template to include some "invisible" compatibility code (for phones, and maybe a "view in browser" link) as well as a generic unsubscribe footer and easy slots for admins to add/edit their own headers and footers.
3. Update step2.html (include a link directly to the template override) and release notes with new instructions for admins. This should probably include some warnings about what not to touch (autoescape tags and where not to mix HTML/Markdown)
4. Update Travis if necessary (depending on how steps 1 and 2 work out).
5. Fix style in the Python files (among other things, use <tt>render_to_string()</tt> or find out what went wrong last time).

benjaminjkraft: Basically we'd want to do it similar to an unsubscribe link: give each user a unique unguessable link, which the website can decode to get their username and therefore render the correct email for them.  I'm not sure what the best way to do that in practice is, but there may be some package out there that will do the crypto for you.

For release notes, if you want you can just add a blank release notes file now, and put your stuff in it, or you can do it when we actually get around to cutting the release.  It would probably be nice if we did more of the former, but since we haven't in the past it's fine if you want to do the latter.

Those steps sound good!

esp/esp/program/modules/handlers/commmodule.py
```diff
@@ -175,7 +181,10 @@ def commfinal(self, request, tl, one, two, module, extra, prog):
...
+                                                                                                                                        'user': request.user,
```
mgersh: This causes the template to render user variables in the template itself (not message body) with information about the sender. Not a problem with the default since there's no use of the user, but this is what's causing Yale's bug. Since the template is going to get rendered into one copy for all users, which then gets further transformed later into the individual copies, user information shouldn't be passed here at all.

milescalabresi: Is there a way to add user information into the template? It would be very useful to put, for instance, the username the email was sent to.

benjaminjkraft: Yeah, you can still put user information in the template; it will just get interpreted on the second render pass, when we add user variables.

esp/templates/email/default_email_html.txt
```diff
@@ -0,0 +1,5 @@
+<html>
+{% autoescape off %}
+{{ msgbdy }}
```
benjaminjkraft: If `msgbdy` might have HTML, you'll need to use `{{ msgbdy|safe }}` so django won't escape the HTML.  (When doing this you should make sure that only admins are passing `msgbdy`, or that it's escaped somewhere else; in this case the former is true.)

esp/esp/program/modules/handlers/commmodule.py
milescalabresi: Sorry, I don't quite understand what the difference is / what needs to be done in order to make sure the right thing happens instead of the variables being rendered in the template?

benjaminjkraft: This template should get rendered twice.  First, here, we render `msgbdy` into `default_email_html.txt` to get `msgtext`.  This happens once, and therefore should not include any user data, because we don't yet know what user the email will be sent to.  Then, later, below, we the output of that (`msgtext`) into the actual email, once for each user, injecting their information into the template.  Any variables in `msgtext` will get interpreted on the second pass.  When you include `request.user` the first time, it means that any user variables in `default_email_html.txt` get rendered according to the user *sending* the email, because that's who `request.user` is; instead you want to render them the second time with the email's recipient as the user.

I wrote too quickly above and was a bit incorrect: any template variables that you put in `default_email_html.txt` that should get rendered on the *second* pass, i.e., any user variables, need to have their template tags excaped, using `{% templatetag openvariable %}` instead of `{{` and `{% templatetag closevariable %}` instead of `}}`.  This way, the first rendering pass will convert those to `{{` and `}}` and then the second rendering pass will see them as normal variables, and interpolate the correct user's information.

benjaminjkraft:
@milescalabresi any progress on this?  Anything I or others can help out with?  If it would be helpful we can talk through some of the tricky bits on Skype sometime.

milescalabresi:
No, but thanks for the poke. I'll get on it this week and definitely take you up on the offer. Much appreciated.

harianbarasu:
Next steps are probably going to be:

1. Add radio buttons to step2.html giving users options for either plain text in email or HTML
2. Implement quick check to confirm plain text vs HTML (if we see <p> tags and they've selected HTML, yell at them)
3. If plain text is checked, run through simple regex parser to convert whitespace to <br> and add <p> tags

milescalabresi:
More specifics on the steps (thanks to Ben for the advice):
If the user selects neither HTML nor plaintext on the radio buttons, display a default Django forms error.

Step 2:
If they select plain text but then enter &lt;p&gt; or something, then use a jquery-ui (https://jqueryui.com/dialog/ for the dialog) that forces them to acknowledge "yes I know, do it anyway" so that the tags would get sanitized to &amp;lt;p&amp;gt; before the text is converted to HTML (and the whitespace is adjusted).

Details on Step 3:
if you choose plaintext, we want to replace &lt;, &gt;, and &amp; with their codes, double newlines with &lt;/p&gt; &lt;p&gt;, single newlines with &lt;br /&gt;, and then wrap the whole thing inside

[template stuff]
&lt;p&gt;
...body...
&lt;/p&gt;

(As I learned, &lt;html&gt; tags are not required in emails.)

Details on the template:
we would need to set the mimetype to text/html( double check?), but the standard Django libraries should be able to handle that for usand see http://www.inboxinspector.com/ for a preview.
Wikipedia says text/html; charset=UTF8 should do the trick, and Django should fill that in (see https://docs.djangoproject.com/en/1.9/topics/email/ and we can even include both HTML and text parts: https://docs.djangoproject.com/en/1.9/topics/email/#sending-alternative-content-types).

Added steps : consider having a "view this email in a browser" link, and see http://www.inboxinspector.com/ for a preview.

willgearty:
Hmm...did you mean to merge main?

milescalabresi:
I meant to rebase this on top of main b/c the fabfile from 2016 was outdated. I hope I didn't do more than just pulling in all of main; I certainly didn't mean to merge this into main (yet). Did I do something wrong?

willgearty:
Oh, I meant, "Did you mean to merge main into this branch?" Because now there's like 200+ new commits...

willgearty:
You probably should have just merged main into the branch, rather than rebasing, but I'm certainly not a git expert

willgearty:
https://www.atlassian.com/git/tutorials/merging-vs-rebasing

milescalabresi:
My mistake. Thanks for the link. I'll fix that later today.

benjaminjkraft:
It looks like you've got a bunch of changes from main into here, which is causing all the merge conflicts.  In general rebasing/squashing commits already in main is a bad idea: what I'd suggest is resetting the last 3 commits, merging main instead of squashing, and then cherry-picking or redoing the last 2 commits on top of that.

willgearty:
So...hate to be THAT guy, but I'm going to guess that #2666 will break this. But I do think #2666 is an important addition, so hopefully it won't be to hard to integrate these changes with that?

milescalabresi:
thanks for the heads-up. I'll look at it before I finish this branch

willgearty:
Any update on this? Do we think this will be ready for SR12?

milescalabresi:
haven't touched it in a while, sorry. I can try to make this the March priority, but I want to discuss the scope with the team given the impact of the GUI

willgearty:
Maybe something like https://github.com/learning-unlimited/ESP-Website/issues/2885 would be more appropriate then? We could have some array of buttons like they do in other mass-mailing services (e.g. Mailchimp) that would import/inject a template into the commpanel GUI. The most basic version could pull all of templates from a template folder (e.g. /program/commpanel/templates) and then just show the name of the template in the button. A more advanced version could actually render the templates within the buttons. This way, chapters could create any number of custom templates (as template overrides) within this folder, and each one would be selectable from the comm panel.
image

milescalabresi:
Closing in favor of #3817

PR #1902, opened 2015-11-22
Add attended number to studentreg big board

benjaminjkraft:


btidor:
For Splash 2014, this reports a number 4 higher than the dashboard. @mgersh suggests that it's because the big board counts deactivated accounts but the dashboard doesn't. The problem here probably isn't with the this pull request.

benjaminjkraft:
Yeah, that sounds plausible.  This is an inconsistency that pops up multiple places.  The thing is I'm not sure what the right thing to do is; it may depend on the context.

PR #1905, opened 2015-11-22
Relax Memcache validation when DEBUG=True

btidor:
In development, we validate that Memcache accesses are sane (i.e. that the value is not too large). The heuristic we use to do this is inexact, since pylibmc applies zlib compression by default, reducing the size of the value. Therefore, this check should probably only include the warning and not the hard limit.

### DIFF
--- a/esp/esp/utils/memcached_multikey.py
+++ b/esp/esp/utils/memcached_multikey.py
@@ -17,11 +17,7 @@ try:
 except:
     import pickle
 
-FAILFAST = getattr(settings, "DEBUG", True)
-
-# Is there any way to introspect this?
 CACHE_WARNING_SIZE = 1 * 1024**2
-CACHE_SIZE = 2 * 1024**2
 MAX_KEY_LENGTH = 250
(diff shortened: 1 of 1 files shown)


benjaminjkraft:
It would be nice if we had a better way to warn without failing, but whenever we do that we can change this to a real warning.

btidor:
Yep! It also turns out this warning isn't as important as I thought. Memcache will raise an error if the value is too big (the documentation is wrong, it doesn't auto-truncate), which means that a too-large value in production will manifest as a hard failure (`SERVER ERROR`) and not mysterious corruption. This makes me more comfortable that removing our hard limit (i.e., this PR) is a good long-term solution.

Right now, Memcache on the dev VM has a 1MB limit, which might cause issues when loading certain chapters' databases. We should probably raise the limit so that everything works, but we can also raise the limit on our servers to be a little higher than in development, which will help suppress issues that we don't catch right away. If we do that, we could even remove the warning entirely (I don't know how useful it is to other people) and just rely on failures to identify issues.

PR #1912, opened 2015-11-24
Fix broken use of TEMPLATE_DIRS in surveys

### LINKED ISSUE #1911: Can't see responses to numeric survey questions
I went to look at my class reviews for MIT Splash (https://esp.mit.edu/teach/Splash/2015/survey/review), and the page is incorrect. I can see responses to open ended questions for all 3 of my classes, but I can only see the responses to the numeric questions for one of the classes (EDIT: I can only see numeric response questions, for classes that have no responses to numeric questions).

It seems to be a bug in the code, not in the survey itself, because the results are messed up for old Splashes as well, even though I know they used to be correct (e.g. https://esp.mit.edu/teach/Splash/2013/survey/review).

It also seems to be a new bug, because I don't see this problem looking at Stanford Splash (which is on stable-release-6) (https://stanfordesp.org/teach/Splash/2014_Fall/survey/review).

mgersh:
Fixes #1911.

### DIFF
--- a/esp/esp/survey/templatetags/survey.py
+++ b/esp/esp/survey/templatetags/survey.py
@@ -159,7 +159,7 @@ def histogram(answer_list, format='html'):
     import hashlib
     file_base = hashlib.sha1(pickle.dumps(context)).hexdigest()
     file_name = os.path.join(tempfile.gettempdir(), file_base+'.eps')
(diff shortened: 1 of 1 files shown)


jmoldow:
:+1:

PR #1918, opened 2015-11-26
Remove gen_media app

benjaminjkraft:
It turns out that the whole `esp.gen_media` app was no longer used.  Some parts of it were removed in #1910.  The remaining part was some sort of facility to generate image-ified headings.  It was only used inside `{% block subsection %}` tags, which it turns out are overriding a nonexistent block, and have been since around 2010.  So I removed all of those, along with the presumably-once-related `{% block subsection_name %}` tags, which allowed me to remove the template tag library and thus the remainder of the app.


btidor:
:smiley_cat:

Issue #1789, opened 2015-10-07
User search and email capitalization


willgearty:
It appears that the user search only does *exact* email matches.
Would there be any way to allow for case differences?
Perhaps make both entirely lowercase and match them?
I've had a few cases where emails from the user list their email address with one capitalization and then their email address on the website is listed with different capitalization.

benjaminjkraft:
Partial duplicate of #1533; I'll add a mention of email addresses there as well since I assume they're easy to fix at the same time.

PR #1919, opened 2015-11-26
Remove confusing whitespace in commpanel code

wnavarre:
Fix a few minor style problems.


btidor:
Looks great! You seem to have branched off main too long ago, so this patch doesn't apply cleanly. Could I ask you to rebase and push an updated patch?

wnavarre:
Done.

mgersh:
Looks like you've got a weird git situation again.

In general, Github's workflow isn't designed to work with rebases, it plays much more nicely with merges. So even though it's uglier, we usually resolve merge conflicts by merging main into the pull request branch, and I recommend that as the solution we should normally be using and suggesting, unless you're confident that you can do a rebase without creating a mess. One good way of avoiding creating a mess is making a new branch instead of trying to push the rebase to the same branch (see #1744, #1789).

In this case, the best way to clean it up will be to close this pull request and make a new branch with just the two commits you want. Then you'll have to look at what the merge conflict was and figure out the correct resolution. In this case it's because #1888 moved `render_to_response` from `esp.web.util` to `esp.utils.web`, and you'll need to make sure the final version of the file is importing from the new location.

Issue #1913, opened 2015-11-24
Deprecation warning on Tag, apparently related to external argcache


mgersh:
#1850 introduced a "RemovedInDjango19Warning: Model class esp.tagdict.models.Tag doesn't declare an explicit app_label and either isn't in an application in INSTALLED_APPS or else was imported before its application was loaded. This will no longer be supported in Django 1.9."

PR #1923, opened 2015-11-27
Tagdict: add an app_label

btidor:
Fixes GH-1913.


mgersh:
The thing is, Tag isn't supposed to need an app_label. I'd be happier fixing this by finding where the import is that it's grumpy about. Unfortunately, Django isn't very good at useful error messages.

mgersh:
(I figured out how to get a useful error message, see #1925)

PR #1925, opened 2015-11-27
Move cf_cache import to inside ready()

mgersh:
This makes a deprecation warning go away.

For everyone's future reference, if you need to debug one of those, put 'raise Exception' at the top of the file with the model it's complaining about and now you have a traceback.


btidor:
Fixes #1913.

PR #1926, opened 2015-11-27
Fix some files not existing

mgersh:
flag-edit.js was moved, and the calendar thing hasn't existed in years, whatever it was.


benjaminjkraft:
:+1: will merge once tests pass

PR #1927, opened 2015-11-27
Remove trailing whitespace from *.py files

btidor:
Here's the command I used:
```
find . -type f -name '*.py' -not -iwholename '.git' -print0 | xargs -0 perl -pi -e 's/ +$//'
```
...courtesy of http://stackoverflow.com/a/149081.

I only skimmed this diff -- can anyone think of what could go wrong?


benjaminjkraft:
Ooh thanks you're the best!

Personally, I use https://github.com/facebook/codemod for these kinds of things, but that command looks fine to me.  The only problem I can think of is in triple-quoted strings, if for some reason we care about trailing whitespace.  Also, it looks like there are a bunch of files with no changes in the diff?  Also also, any idea why there's one extra removed line?

btidor:
Oh, cool, I didn't know about codemod. I'm not sure where the "no changes" message comes from. It might be a bug in Github: `git diff -R` clearly shows that there are changes. I have no idea where the -1 comes from, unfortunately.

benjaminjkraft:
Ah, looks like it removed a newline at the end of some file that had several.  Should be fine.  And I can't think of anywhere where this could cause a problem, so if you merge/rebase it onto latest main I can merge it real quick before we get any more conflicts.

benjaminjkraft:
Thanks!

PR #1928, opened 2015-11-27
Remove program check items

benjaminjkraft:
This is an old system that had a similar use to class flags, but was much
less powerful, and had fallen out of use since about 2008.  Since its
functionality is subsumed by class flags, let's get rid of it.

Also includes removal of some unused code I noticed while ripping it out.


mgersh:
Yay, deleting things! One more thing you can delete: there's a few lines of css in listclasses.html which looks like it was for these and is now unused.

benjaminjkraft:
Ooh, good catch, done.

PR #1934, opened 2015-11-27
Cache average survey responses with argcache

### LINKED ISSUE #1916: Missing number on teacher survey review
It says "Average for all classes: N/A" and I don't think it's supposed to.

mgersh:
They weren't getting invalidated, because someone wanted to cache them before argcache existed. Fixes #1916.


mgersh:
Note to self: this should get backported, but with `esp.cache` instead of `argcache`.

esp/esp/survey/models.py
```diff
@@ -289,24 +291,19 @@ def pretty_val(val):
 
         try:
             average_key = 'question_%d_avg' % self.id
```
benjaminjkraft: I think you can remove this line too.

PR #1941, opened 2015-11-28
Sentry linting

luac:
I went through the Sentry errors from the MIT site and fixed a bunch of them.

The bad(?) news is that a lot of them were just silly uninteresting edge cases that should've been user errors anyway and/or should've been caught by a linter. The good(?) news is that a lot of them involved awful code that should've been cleaned up anyway and/or will have to be if we eventually want to instate a linter.


benjaminjkraft:
A couple of minor things, otherwise looks good once tests pass.

luac:
Argh, that was an entirely predictable merge conflict. But anyway, resolved now.

benjaminjkraft:
Yeah, sorry, I had vaguely figured this would get merged first and I would get to resolve it.

btidor:
+++

PR #1942, opened 2015-11-28
Delete our JS error reporting

mgersh:
Because the Django 1.8 transition broke it, and everyone I told about that thought it was a good thing.


btidor:
++ LGTM (once tests pass)
If we want JS error reporting, we can use Sentry!

benjaminjkraft:
++ and ==btidor

PR #1943, opened 2015-11-28
Remove has_key

benjaminjkraft:
Welcome to the wonderful world of Python 2.3.  Apparently `in` is not only more
pythonic than `has_key`, it's also
[faster](http://stackoverflow.com/a/1323880/904174).  Generated mostly with
codemod, although in some cases I manually replaced it with `get`.

### DIFF
--- a/esp/esp/program/modules/handlers/teachereventsmodule.py
+++ b/esp/esp/program/modules/handlers/teachereventsmodule.py
@@ -181,7 +181,7 @@ class TeacherEventsModule(ProgramModuleObj):
                     # decide type
                     type = "training"
 
-                    if data.has_key('submit') and data['submit'] == "Add Interview":
+                    if 'submit' in data and data['submit'] == "Add Interview":
                         type = "interview"
 
                     form.save_timeslot(self.program, new_timeslot, type)
--- a/esp/esp/users/views/usersearch.py
+++ b/esp/esp/users/views/usersearch.py
@@ -66,7 +66,7 @@ def get_user_list(request, listDict2, extra=''):
                          'description': value['description']}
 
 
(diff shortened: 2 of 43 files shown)


esp/esp/program/modules/handlers/teachereventsmodule.py
```diff
@@ -181,7 +181,7 @@ def teacher_events(self, request, tl, one, two, module, extra, prog):
                     # decide type
                     type = "training"
 
-                    if data.has_key('submit') and data['submit'] == "Add Interview":
+                    if 'submit' in data and data['submit'] == "Add Interview":
```
mgersh: Here's one more that can be changed to a get()

esp/esp/users/views/usersearch.py
```diff
@@ -139,8 +135,7 @@ def get_user_list(request, listDict2, extra=''):
         keys = request.POST['keys'].split(',,')
 
         for key in keys:
-            if request.POST.has_key('operator_'+key) and \
-               request.POST['operator_'+key]         and \
+            if request.POST.get('operator_'+key) and \
```
mgersh: And this one actually wanted to be `in`

benjaminjkraft: Why?

mgersh: Oh, whoops, I misread the diff on this one, you're right.

I was continuing my thought from the other comment several files up the page! I dunno, it made sense in my head.

benjaminjkraft: Oh, I think github was being dumb and not showing me that one, whoops.

Issue #1722, opened 2015-07-03
Suggestion: set SESSION_EXPIRE_AT_BROWSER_CLOSE to False


mgersh:
I'm not sure why we have it set to True in the first place, and changing it might help the issues we have with people making multiple accounts and logging into the wrong one. Any objections? Anyone know why we did that to begin with? The default is False.

PR #1944, opened 2015-11-28
Fix a lot of 'undefined name' errors

benjaminjkraft:
Found using pyflakes, because there's no point in waiting until we see them in
Sentry.  Some of these were created recently, some were very old, some I didn't
bother checking.  Some were variables that had changed names, some were missing
imports.  Some were in unused code, some were in error-handling, some were
probably causing failures in things we might care about.  In any case, they're
(hopefully) gone now.


esp/esp/middleware/espauthmiddleware.py
```diff
@@ -31,6 +31,7 @@
   Phone: 617-379-0178
   Email: web-team@learningu.org
 """
+import datetime
```
benjaminjkraft: This is probably necessary if we want to turn off `settings.SESSION_EXPIRE_AT_BROWSER_CLOSE` (see #1722).

esp/esp/program/modules/handlers/onsitecheckinmodule.py
```diff
@@ -76,7 +76,7 @@ def create_record(self, event):
                                                      program=self.program)
         return created
 
-    def delete_record(self, extension):
+    def delete_record(self, event):
```
benjaminjkraft: I think the (non-barcode, non-rapid) checkin module has been failing if you *remove* any checkin steps since the datatree merge.

esp/esp/qsd/models.py
```diff
@@ -114,17 +114,6 @@ class QuasiStaticData(models.Model):
     def edit_id(self):
         return qsd_edit_id(self.url)
 
-    def get_file_id(self):
```
benjaminjkraft: The two sections in this file appear to be left over from e8dda3c75.

PR #1946, opened 2015-11-28
Add a lint test

benjaminjkraft:
Now that we don't have any undefined names (see #1944), let's keep it that
way. This adds a dependency on pyflakes and a test to check that it doesn't
find any undefined names.  We can add more similar tests as we see fit; it
should be fairly trivial to add more pyflakes-based ones and not that hard
to use other linting libraries.

Depends on #1944.


btidor:
Would it be better to put this directly in `travis.yml`? We wouldn't have to shell out from python, and it might make the build fail faster on lint errors.

benjaminjkraft:
Hmm, it's a possibility.  I would actually put a script somewhere and run that, so we can run it manually too, which would also be much faster.  I'll think some more and maybe change this.

benjaminjkraft:
Closing this in favor of a shell script.

PR #1947, opened 2015-11-28
Accounting cleanup

btidor:
1. Removes a lot of unused methods, including ones related to "executing" transfers
2. Adds a mechanism for Transfer objects representing fees (program admission, t-shirt, lunch, etc.) to be linked with Transfer objects representing payments upon being paid for

### DIFF
--- a/esp/esp/accounting/controllers.py
+++ b/esp/esp/accounting/controllers.py
@@ -290,7 +290,7 @@ class IndividualAccountingController(ProgramAccountingController):
         super(IndividualAccountingController, self).__init__(program, *args, **kwargs)
         self.user = user
 
-    def add_required_transfers(self):
+    def ensure_required_transfers(self):
         """ Function to ensure there are transfers for this user corresponding
             to required line item types, e.g. program admission """
 
@@ -462,7 +462,7 @@ class IndividualAccountingController(ProgramAccountingController):
         FinancialAidGrant.objects.filter(request__user=self.user, request__program=self.program).delete()
 
     def amount_requested(self):
-        self.add_required_transfers()
+        self.ensure_required_transfers()
         program_account = self.default_program_account()
 
         #   Compute sum of all transfers into program that are for this user
--- a/esp/esp/accounting/controllers.py
+++ b/esp/esp/accounting/controllers.py
@@ -257,27 +257,6 @@ class ProgramAccountingController(BaseAccountingController):
     def all_accounts(self):
         return Account.objects.filter(program=self.program)
 
-    def execute_pending_transfers(self, users):
-        """ "Close the books" for this program, with the selected users.
-            Typically this will be all students that attended the program. """
-
-        #   Finalize financial aid for these users
-        for grant in FinancialAidGrant.objects.filter(request__program=self.program, request__user__in=users):
-            grant.finalize()
-
-        if self.program.sibling_discount:
-            #   Execute sibling discounts for these users
-            for splashinfo in SplashInfo.objects.filter(program=self.program, student__in=users):
-                splashinfo.execute_sibling_discount()
-
-        #   Execute transfers for these users
-        self.execute_transfers(Transfer.objects.filter(user__in=users, line_item__program=self.program))
-
-    def remove_pending_transfers(self, users):
-        """ Clear financial records for these users; if they didn't show up,
-            we shouldn't be expecting their money.  """
-        self.all_transfers().filter(user__in=users, executed=False).delete()
-
     def payments_summary(self):
         """ Return a tuple with the number and total dollar amount of payments
             that have been made so far. """
@@ -519,12 +498,6 @@ class IndividualAccountingController(ProgramAccountingController):
         amt_sibling = self.amount_siblingdiscount()
         return amt_request - self.amount_finaid(amt_request, amt_sibling) - amt_sibling - self.amount_paid()
 
-    def clear_payments(self):
-        #   Remove all payments listed for this user at this program
-        line_item_type = self.default_payments_lineitemtype()
-        target_account = self.default_source_account()
-        Transfer.objects.filter(source=None, destination=target_account, user=self.user, line_item=line_item_type).delete()
-
     def submit_payment(self, amount, transaction_id=''):
         #   Create a transfer representing a user's payment for this program
         line_item_type = self.default_payments_lineitemtype()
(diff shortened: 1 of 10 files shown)


mgersh:
Can confirm that it's safe to delete Transfer.executed, I checked on the non-MIT sites a while ago when I was thinking about doing something like this and it's similarly unused on those.

esp/esp/accounting/controllers.py
```diff
@@ -290,28 +232,37 @@ def __init__(self, program, user, *args, **kwargs):
...
-        executed_item_types = executed_items.values_list('line_item__id', flat=True)
-
-        #   Create transfers for required line item types that have not already been executed
-        for lit in line_items:
-            if lit.id not in executed_item_types:
-                result.append(Transfer.objects.create(source=source_account, destination=program_account, user=self.user, line_item=lit, amount_dec=lit.amount_dec))
-        return result
+        required_line_items = self.get_lineitemtypes(required_only=True)
+
+        existing_transfers_by_li = {t.line_item.id: t for t in Transfer.objects.filter(
```
mgersh: Writing `t.line_item_id` rather than `t.line_item.id` avoids a query per transfer.

esp/esp/accounting/controllers.py
```diff
@@ -290,28 +232,37 @@ def __init__(self, program, user, *args, **kwargs):
...
+            elif transfer.paid_in:
+                # A Transfer for this Line Item Type already exists
+                # *and* has already been paid. It's too late, so do
+                # nothing.
+                pass
+            else:
+                # Adjust the amount of the Transfer to match the new
+                # value.
+                transfer.amount_dec = item.amount_dec
+                transfer.save()
```
mgersh: It might be worthwhile to only save if there's actually been a change.

esp/esp/accounting/controllers.py
```diff
@@ -519,24 +470,48 @@ def amount_due(self):
...
+        # find all of the Transfers representing the items that were paid for
+        # and add a link back to the payment.
+
+        # Find the paid transfers by examining Transfers in order of creation
+        # until they sum to the given amount.
+        total = 0
+        target = payment.get_amount()
+        paid_transfers = []
+
+        for transfer in self.get_transfers():
```
mgersh: `get_transfers()` returns a queryset, so you can add additional filters on it. Doing the `paid_in` check here is simple, `is_purchasable` is more complicated* but still doable with Q objects if you want to try that, although it might not be an advantage.

*There are no transfers in any site's database where `line_item` is null. If you want to add a migration to make that an official requirement, it could simplify some things.

esp/esp/accounting/controllers.py
```diff
@@ -519,24 +470,48 @@ def amount_due(self):
...
+            if not (transfer.line_item and transfer.line_item.is_purchasable) \
+               or transfer.paid_in:
+                # Filter out payment-related Transfers and already-paid Transfers
+                continue
+
+            total += transfer.get_amount()
+            paid_transfers.append(transfer)
+            if total >= target:
```
mgersh: It makes sense to me to have either this optimization or the correctness check right below it, but not both. I'd say keep the correctness check. Or you could change the comparison here to a `>`.

esp/esp/accounting/controllers.py
```diff
@@ -519,24 +470,48 @@ def amount_due(self):
...
+                break
+
+        if total != target:
+            raise ValueError("Transfers do not sum to target: %.2f" % target)
+
+        # Link the paid transfers back to the payment.
+        for transfer in paid_transfers:
```
mgersh: You can do this in the above loop instead of keeping track of a list. It's safe because we set `ATOMIC_REQUESTS=True`, which means that requests run in a transaction which will get rolled back if there's an exception.

esp/esp/accounting/models.py
```diff
@@ -164,20 +170,18 @@ class Account(models.Model):
...
 
     @property
     def balance(self):
-        result = self.balance_dec
-        if Transfer.objects.filter(source=self, executed=False).exists():
-            result -= Transfer.objects.filter(source=self, executed=False).aggregate(Sum('amount_dec'))['amount_dec__sum']
-        if Transfer.objects.filter(destination=self, executed=False).exists():
-            result += Transfer.objects.filter(destination=self, executed=False).aggregate(Sum('amount_dec'))['amount_dec__sum']
+        if Transfer.objects.filter(source=self).exists():
+            result -= Transfer.objects.filter(source=self).aggregate(Sum('amount_dec'))['amount_dec__sum']
```
mgersh: This errors, because `result` doesn't have a starting value.

mgersh:
Nice! This way of keeping track of things definitely makes more sense. I recommend playing around with the debug toolbar's SQL panel and/or `shell_plus --print-sql` if you haven't already, since this is a part of the codebase that has caused performance issues, but this is definitely an improvement so it's up to you how much more to fine-tune it right now.

esp/esp/accounting/controllers.py
btidor: Thanks!

esp/esp/accounting/controllers.py
btidor: Done.

esp/esp/accounting/controllers.py
btidor: Okay. I made `line_item` a required field and replaced this logic with a database-level filter. Since this is the only use the `is_purchasable` property, I removed it as well.

esp/esp/accounting/controllers.py
btidor: I think both of these are actually needed for correctness. If we change this check, we're requiring that all transactions sum to the target amount, when what we really want to test is that a prefix of transactions add up.

esp/esp/accounting/controllers.py
btidor: Okay. I wrapped this function in `@transaction.atomic` to make sure that the caller can't mess things up by suppressing the exception.

esp/esp/accounting/models.py
btidor: Oops.

btidor:
I made these changes and took a look at the SQL for `ensure_required_transfers` and `submit_payment`. As far as I can tell, they're both pretty fast.

esp/esp/accounting/controllers.py
mgersh: Oh, okay, makes sense.

mgersh:
:+1:

PR #1948, opened 2015-11-28
Add a lint test

benjaminjkraft:
Now that we don't have any undefined names (see #1944), let's keep it that way.
This adds a dependency on flake8 and script (`useful_scripts/run_lint.sh`) to
run a few `flake8` checks; travis should automatically run the script before
running tests.

We can add more checks as we clean up the codebase so that they pass.  (For
example, it looks like we still have a few trailing whitespace issues, so I
didn't turn those checks on.)  It should be trivial to add more `flake8`
checks; adding other linters (e.g.  `eslint` for our JS) shouldn't be too hard
either.

If you want to automatically run the linter before every commit, just make a
symlink from `.git/hooks/pre-commit` to the lint script.  It's a bit slower
than I'd like, unfortunately, so it might be good to have a git-hook version
that only lints changed files or something, but I'll leave that for another
day.

Replaces #1946; this version is a shell-script that travis runs, which means travis will fail faster in the case of lint errors, and makes the thing fast enough that one can plausibly run it as a commit hook.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,5 +18,6 @@ before_script:
   - psql -c "CREATE DATABASE test_django OWNER testuser;" -U postgres
 script: 
   - cd esp
+  - useful_scripts/run_lint.sh
   - ./manage.py collectstatic --noinput -v 0
   - ./manage.py test
--- a/esp/requirements.txt
+++ b/esp/requirements.txt
@@ -37,6 +37,7 @@ Pygments==2.0.2
 twilio==3.6.5
 shortuuid==0.4.2
 raven
+flake8==2.5.0
 
 # We are using the 2014-03-13 version of the Stripe API, which is v1.12.2.
 stripe==1.12.2
--- a/esp/useful_scripts/run_lint.sh
+++ b/esp/useful_scripts/run_lint.sh
@@ -1,0 +1,24 @@
+#!/bin/bash -e
+
+BASE_DIR="$(dirname "$(dirname "$(readlink -e "$0")")")" # /esp
+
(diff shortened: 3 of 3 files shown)


btidor:
+1 (once tests pass)

jmoldow:
Rather than running it in the same job as the unit tests, can we run it in a separate job? That way Travis will report separate results for linting and unit tests.

Our Travis CI file could have

```
matrix:
  include:
    - env: ENV="flake8"
    - env: ENV="test"
```

And then we could have the script invoke either unit tests or linting, depending on the value. Or, if we were okay with introducing another dependency, we could use tox to do that for us.

Thoughts on this?

benjaminjkraft:
I don't have a problem with doing that.  The advantage and disadvantage of this is that it won't even bother running tests if lint fails.  In any case, I'd propose to do that in a separate PR, since it's a more significant change to our travis setup.  Ideally the lint check wouldn't bother setting up the DB, installing all the packages ever, and so on, so that it could be even faster.

jmoldow:
Sure, in that case this looks fine to me.

I'm not familiar with flake8, but I've used pylint a bunch. For pylint tests, you would still need to install all of the Python pip packages, because there is a pylint test that will fail if any of the imports fail. But you could skip the apt packages, and as you said, all of the DB setup, etc.

benjaminjkraft:
I think flake8 (and pyflakes which it uses) don't require imports to be installed, but we could check.

btidor:
Cool! I'll open a new issue for the feature.

PR #1953, opened 2015-11-29
Travis: run linter on a separate machine

### LINKED ISSUE #1951: Run linters in a separate Travis build
From @jmoldow on #1948:

> Rather than running [the linter] in the same job as the unit tests, can we run it in a separate job? That way Travis will report separate results for linting and unit tests.
>
> Our Travis CI file could have
>  
>     matrix:
>       include:
>         - env: ENV="flake8"
>         - env: ENV="test"
> And then we could have the script invoke either unit tests or linting, depending on the value. Or, if we were okay with introducing another dependency, we could use tox to do that for us.
> 
> Thoughts on this?

btidor:
Fixes #1951, and also makes the linter portable.

### DIFF
--- a/deploy/lint
+++ b/deploy/lint
@@ -1,6 +1,11 @@
 #!/bin/bash -e
 
-BASE_DIR="$(dirname "$(dirname "$(readlink -e "$0")")")" # /esp
+# http://stackoverflow.com/a/4774063
+pushd "$(dirname "$0")" > /dev/null
+SCRIPTDIR="$(pwd -P)"
+popd > /dev/null
+
+BASEDIR="$(dirname "$SCRIPTDIR")" # /esp
 
 echo "Running lint checks..."
 
@@ -19,6 +24,6 @@ echo "Running lint checks..."
 #   W191: indentation contains tabs
 #   W291: whitespace at end of line
 #   W293: whitespace on blank line
-flake8 --ignore "E,F,W" --select "F82,F831,W601,W603,W604" $BASE_DIR
+flake8 --ignore "E,F,W" --select "F82,F831,W601,W603,W604" "$BASEDIR"
 
-echo "done."
+echo "...success!"
--- a/deploy/travis/before_script
+++ b/deploy/travis/before_script
@@ -1,0 +1,15 @@
+#!/bin/bash
+set -euf -o pipefail
+
+if [ "$TRAVIS_JOB" = "test" ]; then
+    cp esp/esp/local_settings.py.travis esp/esp/local_settings.py
+    ln -s `pwd`/esp/public/media/default_images esp/public/media/images
+    ln -s `pwd`/esp/public/media/default_styles esp/public/media/styles
+    psql -c "CREATE ROLE testuser PASSWORD 'testpassword' LOGIN CREATEDB;" -U postgres
+    psql -c "CREATE DATABASE test_django OWNER testuser;" -U postgres
+elif [ "$TRAVIS_JOB" = "lint" ]; then
+    :
+else
+    echo "Unknown TRAVIS_JOB: $TRAVIS_JOB"
+    exit 1
+fi
--- a/deploy/travis/install
+++ b/deploy/travis/install
@@ -1,0 +1,13 @@
+#!/bin/bash
+set -euf -o pipefail
+
+if [ "$TRAVIS_JOB" = "test" ]; then
+    sudo apt-get install -y $(cat esp/packages_base.txt | grep -v ^memcached | grep -v ^postgres)
+    esp/packages_base_manual_install.sh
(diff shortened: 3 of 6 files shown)


deploy/lint
```diff
@@ -1,8 +1,13 @@
 #!/bin/bash -e
 
-BASE_DIR="$(dirname "$(dirname "$(readlink -e "$0")")")" # /esp
+# http://stackoverflow.com/a/4774063
+pushd `dirname $0` > /dev/null
+SCRIPTPATH=`pwd -P`
```
benjaminjkraft: This should probably be called `SCRIPTDIR`

deploy/lint
```diff
@@ -1,8 +1,13 @@
...
+popd > /dev/null
 
-echo "Running lint checks..."
+BASE_DIR="$(dirname "$SCRIPTPATH")" # /esp
+
+echo -n "Running lint checks..."
```
benjaminjkraft: `-n` is going to make the first line of  `flake8` output look weird, that's why I didn't do it.

btidor: Oh, okay

deploy/travis/before_script
```diff
@@ -0,0 +1,15 @@
...
+    cp esp/esp/local_settings.py.travis esp/esp/local_settings.py
+    ln -s `pwd`/esp/public/media/default_images esp/public/media/images
+    ln -s `pwd`/esp/public/media/default_styles esp/public/media/styles
+    psql -c "CREATE ROLE testuser PASSWORD 'testpassword' LOGIN CREATEDB;" -U postgres
+    psql -c "CREATE DATABASE test_django OWNER testuser;" -U postgres
+elif [ "$TRAVIS_JOB" = "lint" ]; then
+    :
+else
+    echo "Unknown TRAVIS_JOB: $TRAVIS_JOB"
+    exit -1
```
benjaminjkraft: is `exit -1` valid?  I thought exit codes had to be in `[0, 128)` for normal exits, and `[128, 256)` (or perhaps `[-128, 0)` for signals.

deploy/travis/install
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+set -euf -o pipefail
+
+if [ "$TRAVIS_JOB" = "test" ]; then
+    sudo apt-get install -y $(cat esp/packages_base.txt | grep -v ^memcached | grep -v ^postgres)
```
benjaminjkraft: Do you know if we actually need this?

deploy/travis/before_script
btidor: Right, I probably meant `exit 1`.

deploy/travis/install
btidor: Nope, this is what was in `.travis.yml`.

deploy/travis/before_script
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+set -euf -o pipefail
```
benjaminjkraft: I wasn't aware of `pipefail`, cool.

deploy/travis/install
benjaminjkraft: Oh, whoops, I totally misread and thought this was for lint.

benjaminjkraft:
Do you know if there's a way to make the jobs show up separately in github?  I know django does that (look at any of their pull requests) and it's quite convenient.

btidor:
I don't believe so. See travis-ci issue 5035.

benjaminjkraft:
Ah, yeah, it looks like django does this by running their own CI server.  Never mind!

deploy/lint
```diff
@@ -1,6 +1,11 @@
 #!/bin/bash -e
 
-BASE_DIR="$(dirname "$(dirname "$(readlink -e "$0")")")" # /esp
+# http://stackoverflow.com/a/4774063
+pushd `dirname $0` > /dev/null
```
benjaminjkraft: I think you want quotes around the backticks here and on the following line.  (Although if anybody has spaces in their dev server path, they're probably already screwed.)

mgersh: People do have spaces in their dev server path, actually, we've dealt with this before.

deploy/travis/before_script
dehnert: I'm gonna bet that somebody's been reading https://sipb.mit.edu/doc/safe-shell/...

deploy/lint
dehnert: I believe to support spaces you need 
```
pushd "`dirname "$0"`"
```
-- around the `$0` so `dirname` uses the right thing, and around the backticks so the `pushd` does. (Though, if you're using `bash` already, you might consider `$()` instead of backticks.)

benjaminjkraft:
Cool!  Going to wait for tests to pass, but this looks good.

Issue #1901, opened 2015-11-22
Assert that every program module view has an auth decorator of some sort


luac:
or has some explicit no_auth decorator, for things we intend to make public.

This could be an automated test / part of #1900, or it could be an assert that triggers upon module load.

PR #1949, opened 2015-11-28
Basic Cybersource cleanup

btidor:
Simplify our Cybersource flow by consolidating from two pages to one, and removing some unneeded text.

PR #1954, opened 2015-11-29
Adds test that all program module views have some auth decorator

emannes:
This is an attempt to resolve issue #1901. 

I tried to add an explicit auth decorator (e.g., `needs_student`, `needs_admin`, etc.) to every program module view. I also wrote a test that checks that every such view has a decorator. Most views were given a new `no_auth` decorator.

I ran into problems with the `stats`, `sections`, `message_requests`, and `class_info` views in the JSONDataModule, however. If I put a `needs_admin` decorator below the cache decorator on the `stats` view, I get the error

> esp.middleware.esperrormiddleware.ESPError_Log: ArgCache does not support varargs.

If I put it above the cache decorator, the `needs_admin` decorator gets rid of the `cached_function` attribute of `stats`, which causes an error on jsondatamodule.py:804.

What's the best workaround? If I modify the `needs_whatever` decorators so that they preserve the cached_function attribute, will that preserve both caching and the auth check? Or should I be doing something else?


emannes:
After conferring with @mgersh, I added appropriate decorators and changed the instances of `something.cached_function` to `something.method.cached_function` as needed. I also filed issue #1955 so that `class_info` can safely not require auth.

esp/esp/program/modules/handlers/creditcardmodule_cybersource.py
```diff
@@ -71,13 +71,13 @@ def studentDesc(self):
...
     @main_call
-    @usercheck_usetl
+    @needs_student
     @meets_deadline('/Payment')
     def startpay_cybersource(self, request, tl, one, two, module, extra, prog):
         return render_to_response(self.baseDir() + 'cardstart.html', request, {})
 
     @aux_call
-    @usercheck_usetl
+    @needs_student
```
benjaminjkraft: You may want to not change these now to avoid conflicting with #1949 if not necessary.  @btidor?

esp/esp/program/modules/handlers/jsondatamodule.py
```diff
@@ -468,6 +475,7 @@ def class_info(self, request, tl, one, two, module, extra, prog):
     class_info.cached_function.depend_on_model(ClassSection)
 
     @aux_call
+    @no_auth
```
benjaminjkraft: In other cases the decorator went below the `@json_response()`.  Is there a deliberate difference?  Does it matter?

esp/esp/program/modules/handlers/onsiteprintschedules.py
```diff
@@ -56,7 +56,7 @@ def module_properties(cls):
             }
 
     @main_call
-#    @needs_onsite
+    @needs_onsite
```
benjaminjkraft: Did you test that printing schedules still works?  It wouldn't surprise me if we turned this off because the script wasn't *actually* doing login correctly.  (Although perhaps fixing this issue is more important than keeping `poll_schedules.sh` working, in which case just make sure to make an issue if you break it.)

esp/esp/program/modules/tests/auth.py
```diff
@@ -0,0 +1,54 @@
...
+        """Test that all views of all program modules have some sort of auth decorator,
+        e.g., @needs_admin, @needs_student, @needs_account, @no_auth"""
+
+        # self.program has all possible modules
+        modules = self.program.getModules()
+        for module in modules:
+            view_names = module.get_all_views()
+            for view_name in view_names:
+                view = getattr(module, view_name)
+                assert hasattr(view, 'has_auth_check') and view.has_auth_check, \
```
benjaminjkraft: If you want you can do `getattr(view, 'has_auth_check', None)`.  Also, I think `self.assert_` is preferred because it lets `unittest` correctly classify it as an assertion failure instead of an error and print a nicer message.  (Or something like that, I just know it's preferred.)

benjaminjkraft:
Whoo thanks!  The fact that this added several non-`@no_auth` decorators perhaps speaks to the value of doing this.

esp/esp/program/modules/handlers/creditcardmodule_cybersource.py
btidor: I'm fine resolving the merge conflict. `@needs_student` is probably cleaner.

esp/esp/program/modules/handlers/jsondatamodule.py
emannes: I think it doesn't matter because the definition of `@json_response` uses `functools.wraps`. I'll change it for consistency.

benjaminjkraft: Oh, right, I forgot that `wraps` copies `__dict__` in addition to `__name__` and so on.  In that case, consistency would be nice but I don't care that much.

esp/esp/program/modules/handlers/onsiteprintschedules.py
emannes: It didn't work, but not because of @needs_onsite but because latex.py doesn't support ps output, just pdf, log, tex, svg, and png. I pushed a fix.

emannes:
Okay. @benjaminjkraft, I think I've addressed all of your comments--do you want to check it over again ?

esp/esp/program/modules/handlers/onsiteprintschedules.py
benjaminjkraft: Oh, weird, looks like we have two copies of the print schedules script floating around -- the other one was fixed in 4368beb7b.  In any case, we can clean that up later.

benjaminjkraft:
This somehow seems to have a merge conflict still.  Try merging/rebasing it on main?  (Also, while you're at it, if you want you can remove the last commit since it'll be irrelevant after #1979.)

benjaminjkraft:
(I just did the merge.)  Thanks!

PR #1964, opened 2015-12-05
Fix auth for the reg profile module

mgersh:
Because this is what `usercheck_usetl` is for. However, `usercheck_usetl` was also kind of a mess so I fixed it to use the same templates as everything else and be more readable.


mgersh:
This should fix the issue which is the reason for the local changes we didn't merge in #1750, so we can get rid of those next release.

jmoldow:
:+1:

esp/esp/program/modules/handlers/regprofilemodule.py
```diff
@@ -95,14 +95,6 @@ def profile(self, request, tl, one, two, module, extra, prog, check_role=True):
         else:
```
benjaminjkraft: The above code is no longer wrong, but it's kind of weird.  Since we're using `@usercheck_usetl` it would be nice to just remove the check of the user's types in here, and always use the `tl`, since they now have to have the matching type already.  (Also, at the moment, this will work for `tl`s other than `teach` and `learn` if and only if the user is neither a teacher nor a student, which is weird.  Not sure if that will just end up with it erroring later anyway, though.)

PR #1969, opened 2015-12-10
Fabfile fixes based on user feedback

btidor:


benjaminjkraft:
Looks fine to me, but I haven't set up fabric yet, so probably someone who has should at least glance at this.

fabfile.py
```diff
@@ -89,7 +92,7 @@ def setup():
     print "***** Please choose a passphrase and type it at the three prompts."
```
mgersh: I haven't tried it myself, but based on your commit message this should say "two prompts" now?

btidor: Right, yes.

mgersh:
Cool, looks good. Since it's a security fix, want to email out about it?

PR #1971, opened 2015-12-10
Fix path-construction in linter

benjaminjkraft:
It broke when moved.  (Edit: to be more specific, it started trying to lint `env`, which it really shouldn't.)

### DIFF
--- a/deploy/lint
+++ b/deploy/lint
@@ -5,7 +5,7 @@ pushd "$(dirname "$0")" > /dev/null
 SCRIPTDIR="$(pwd -P)"
 popd > /dev/null
 
-BASEDIR="$(dirname "$SCRIPTDIR")" # /esp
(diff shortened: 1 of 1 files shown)


btidor:
Don't we also want to lint `fabfile.py` and stuff in `deploy`? The old way gets the whole repository.

benjaminjkraft:
If we want that we should specify it explicitly/separately.  If you're running this locally, as is it will try to lint `env`, which is not a plan.  (It does so on Travis too, but I guess `flake8` passes `flake8`, and that's the only thing we install.)

benjaminjkraft:
(Updated to do that.)

benjaminjkraft:
(Talked to @btidor off-thread, going to use excludes instead of linting only certain directories.)

btidor:
+1

benjaminjkraft:
Merging per @btidor's approval now that tests have passed.

Issue #1237, opened 2014-06-29
Improve logging setup


benjaminjkraft:
This is a pretty broad issue, but I'd like to improve our logging setup, because I think it sucks.  Reasons include: we're not using anything remotely standard, it's unclear how to log something that shouldn't show the user an ESPError page, but where we should know that something is wrong, serverlog is flooded with crap from the JS error reporter so no one actually looks at it, we can't really log anything outside the main django request-response flow, and so on.

Things I think we should do:
- [x] Use the [django](https://docs.djangoproject.com/en/1.6/topics/logging/)/[python](https://docs.python.org/2.7/library/logging.html) builtin logging functionality, and make all the logging go through that.
- [ ] Make ESPError a normal exception class which is raised, and then have the error middleware just log things using the normal logging module.  Or maybe have two classes, one for things which are a user error (e.g. deadlines) which raises a 4xx and logs an INFO and one for things which are a server error (e.g. a failed assertion) which raises a 500 and logs an ERROR.  Or maybe kill it entirely, have the middleware just catch any error, and raise actual customized or standard exception classes.
- [ ] Pull the JS error reporter into this, and set its logging to WARNING or INFO.
- [ ] Have one email list that gets all the logs, and another that gets only ERROR/CRITICAL so that maybe people will actually send those to their email.  It's nice to have everything off the server somewhere, but getting all logs is a bit much.
- [x] Actually log the errors on disk somewhere.  Right now they're _only_ in email.
- [x] Make the debug server print the errors usefully, and not even attempt to send them by email.  (Right now it does, and a lot of us have the emails set to print to the console on dev, which is useless.
- [x] Convert all the various print statements and such in the code (of which there are a terrifying number) to log appropriately, even if we don't want to show the user an error.

Thoughts?

luac:
Leaving some links here that I had in my personal notes, presumably to look at for client/JS logging:

https://github.com/RossiAndrea/django-statistics-and-errors
https://github.com/sorensen/django-client-errors

PR #1976, opened 2015-12-12
Program module extension cleanup

benjaminjkraft:
This is a cleanup of the program module extension system, completing some of the work started in #1232.

The major change (11b691b) is that module extension attributes are no longer accessible from the program module.  #1232 added a deprecation warning for this type of access, but unfortunately our setup doesn't seem to log those anywhere (see #1237).  So I decided to be bold and/or stupid and just grep for all possible fields and attributes that would be an issue and clean them up; luckily only a few were actually issues.

Removing this allowed a number of simplificiations.  The biggest one is a schema change: module extensions now point at programs, rather than at modules, and therefore are accessed like any other related model.  (The migration to do this involved a bit of cleanup; see the commit message and comments in 6a0cd37.)  They still have a slight tie to their former modules: if the module is added to a program, the corresponding module extension will be added too (see f5b1fd8).  But this is a very weak tie, and if we wanted we could easily sever it completely by creating all possible module extensions on program creation.

See individual commits for more details.  I tried to do things such that each commit is a valid, non-breaking change in and of itself, to make it easier to check correctness and do bisects; it may correspondingly be easier to review it commitwise.

### DIFF
--- a/esp/esp/application/models.py
+++ b/esp/esp/application/models.py
@@ -204,7 +204,7 @@ class StudentClassApp(models.Model):
 class FormstackStudentProgramAppManager(models.Manager):
     def create_from_submission(self, submission, settings):
         """ Takes a FormstackSubmission and creates an app from it. """
-        program = settings.module.program
+        program = settings.program
         data_dict = { int(entry['field']): entry['value']
                       for entry in submission.data() }
 
--- a/esp/esp/application/models.py
+++ b/esp/esp/application/models.py
@@ -5,7 +5,6 @@ from django.dispatch import receiver
 from argcache import cache_function
 from esp.users.models import ESPUser
 from esp.program.models import Program, ClassSubject
-from esp.program.modules.base import ProgramModuleObj
 from esp.formstack.api import Formstack
 from esp.formstack.objects import FormstackForm, FormstackSubmission
 from esp.formstack.signals import formstack_post_signal
@@ -16,7 +15,6 @@ class FormstackAppSettings(models.Model):
     module.
     """
 
-    module = models.ForeignKey(ProgramModuleObj)
     program = models.OneToOneField(Program)
 
     # formstack settings
@@ -44,6 +42,12 @@ include the content of a field, use {{field.12345}} where 12345 is the
 field id.""")
 
     app_is_open = models.BooleanField(default=False, verbose_name="Application is currently open")
+
+    @property
+    def module(self):
+        """Deprecated; you probably shouldn't need this."""
+        # TODO(benkraft): remove.
+        return self.program.getModule('FormstackAppModule')
 
     def formstack(self):
         """
--- a/esp/esp/application/models.py
+++ b/esp/esp/application/models.py
@@ -268,7 +268,7 @@ class FormstackStudentProgramAppManager(models.Manager):
         """ Get apps for a particular program from the Formstack API. """
 
         # get submissions from the API
-        settings = program.getModuleExtension('FormstackAppSettings')
+        settings = program.formstackappsettings
         submissions = settings.form().submissions(use_cache=False)
 
         # parse submitted data and make model instances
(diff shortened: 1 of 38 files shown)


luac:
Cool!

esp/esp/application/models.py
```diff
@@ -44,6 +43,12 @@ class FormstackAppSettings(models.Model):
 
     app_is_open = models.BooleanField(default=False, verbose_name="Application is currently open")
 
+    @property
+    def module(self):
+        """Deprecated; you probably shouldn't need this."""
+        # TODO(benkraft): remove.
+        self.program.getModule('FormstackAppModule')
```
luac: Looks like this is missing a "return" (and same for the others).

mgersh:
Nice! I'll miss the great get_program() comment too, but it hasn't actually made sense for a while now. And I highly approve of explaining the history of module_ext.py in a comment.

mgersh:
Some thoughts on future design decisions: I don't actually like the idea of completely cutting the ties between module extensions and modules. A lot of program settings information is closely related to a specific module, including settings that aren't currently stored that way (I'm thinking of various Tags and a few of the fields on Program), and I think it's a good way to organize settings and we should be doing more of that. And if we want to do this for a lot of modules, it'll be more convenient if we have some infrastructure to support it.

luac:
It's still possible to add more module extensions and have modules ensure that the ones they need are created. Is there more integration needed than that?

luac:
(One thing that might be nice is some way to enforce that a module doesn't use module extensions that it doesn't depend on, so that we don't get a whole bunch of things in StudentClassRegModuleInfo that aren't directly related to StudentClassRegModule, but I think that general class of problems will take more thought to fix.)

mgersh:
Yeah, that was really a response to the comment at the top of module_ext.py which suggests getting rid of the connection between modules and the settings models entirely. I'm being vague because I'm not sure what I mean, and this definitely doesn't matter for any of the code changes here.

esp/esp/application/models.py
benjaminjkraft: Whoops, yeah.  Will fix.

benjaminjkraft:
Yeah, I was going back and forth on this.  The fact that there are lots of users of the SCRMI outside the SCRM is a bit ugly, but on the other hand in the new system it would be easy enough to just have multiple modules require the SCRMI -- just hook up multiple signals.  I'll update the comment to make it a bit more equivocal about what Should be done.

esp/esp/program/modules/apps.py
```diff
@@ -2,3 +2,9 @@
 
 class ModulesConfig(InstallConfig):
     name = 'esp.program.modules'
+
+    def ready(self):
+        super(ModulesConfig, self).ready()
+        # TODO(benkraft): add a thing to InstallConfig that imports a signals
```
mgersh: Not all of our apps use InstallConfig, that one is specifically for apps that have an install(). If we wanted to refactor to have all our apps use some subclass of AppConfig that runs install() if it exists and imports .signals if it exists, that's a thing we could do.

benjaminjkraft: Oh, whoops.  Does that mean I should just be inheriting from `AppConfig` for `esp.application`?

mgersh: Yeah. I was going to say "huh, why did that not error?" but then I realized it's because you didn't set anything to use that AppConfig. You also need to tell Django to use that one instead of the default one by putting it in django_settings.py.

benjaminjkraft: Whoops, I had it in my head to do that, too, and then forgot.  Fixed both issues.

mgersh:
Is this ready to merge? I think it's ready to merge, but the rebases make it hard to check that everything we were talking about has been resolved. (And doing bisects on this wouldn't work anyway, the migrations don't run backwards.)

benjaminjkraft:
Oh, whoops, only one of them needs a backwards migration, but I didn't bother writing it.  Hopefully it won't be relevant, or if it is I can just write it later.  Anyway, yeah, this is ready to merge as far as I'm concerned.

PR #1981, opened 2015-12-19
Run linter-style checks for future Django support

mgersh:
This is intended to make gradually adding support for future Django versions easier, and hopefully make the next upgrade easier when it eventually happens. It creates a separate Travis job which runs a custom script to check for adding uses of features that won't work anymore and might not have test coverage but are greppable.

Right now there's a check in the script for every feature that has a backwards-incompatible change in 1.9, or is deprecated and will be removed in 1.9 or 1.10, as long as it's anywhere remotely close to greppable. Many of them are commented out. Some of those are because we're still using the feature, some of them are false positives. Some of the false positives are because I didn't feel like spending much time thinking about any individual check and wrote a lot of them in obviously sloppy ways, so at this point those checks are mostly just placeholders for better checks. Some of them will just be hard to filter out with this style of check.

The idea is that once we've fixed a check, we uncomment it, which allows us to avoid adding any more uses, and when we eventually upgrade we won't have to recheck all of these. It also serves as a to-do list.

A lot of these checks can be deleted if we ever implement logging for deprecation warnings.

### DIFF
--- a/deploy/future
+++ b/deploy/future
@@ -1,0 +1,172 @@
+#!/bin/bash
+# Don't run this with -e, grep exits with 1 when it finds no results
+
+function check_py () {
+# This is a bit janky. If you know how to get the output
+# to go on separate lines, please tell me.
+    results=`git --no-pager grep $1|grep "\.py"|grep -v "future-ok"`
+    if [ "$results" ]; then
+	echo $results
+	EXIT=1
+    fi
+}
+
+function check_html () {
+    results=`git --no-pager grep $1|grep "\.html:"|grep -v "future-ok"`
+    if [ "$results" ]; then
+	echo $results
+	EXIT=1
+    fi
+}
+
+EXIT=0
+
+# http://stackoverflow.com/a/4774063
+pushd "$(dirname "$0")" > /dev/null
+SCRIPTDIR="$(pwd -P)"
+popd > /dev/null
+
+BASEDIR="$(dirname "$SCRIPTDIR")"
+cd $BASEDIR
+
+echo "Running future Django lint checks..."
+
+# Uncomment checks here once you've deleted all the uses from the codebase,
+# or checked that we never used the feature in the first place,
+# or checked that future changes won't break your use of this feature.
+# Add more checks once a new version of Django is released,
+# and delete them once we've upgraded main to a new version and
+# cleared out old pull requests.
+# Only push to future-django-support if what you're doing won't work
+# on our current Django version. Otherwise, make a pull request to
+# main and add the check as part of your pull request.
+
+# A lot of these are impossible or annoying to write without some
+# false positives, and some ungreppable things are left out entirely.
+# Many of them should be considered warnings rather than errors: check
+# that a future behavior change won't break your code.
+# Add "#future-ok" to the end of a line if you've checked,
+# or rewrite the check to avoid the false positives.
+
+# Django 1.9 backwards incompatibility checks
+#check_py is_usable
+#check_py "\.add(" # warning
+check_py get_request_repr
+check_py add_to_builtins
+check_py simple_tag # warning
+check_py page_range
+check_py REASON_PHRASES
+check_py STATUS_CODE_TEXT
+check_py ValuesQuerySet
+#check_py ValuesListQuerySet
+#check_py get_context_data
+check_py SimpleTestCase # see also 1.10
+check_py app_name
+check_py total_ordering
+#check_html url # warning
+check_py enable_comments
+check_py setup_databases # warning
+check_py teardown_databases # warning
+check_py models.fields.related
+
+# Deprecated, removed in 1.9
+# (if we had logging for deprecation warnings, we wouldn't need these)
+# a lot of the commented out checks are just false positives
+check_py dictconfig
+check_py importlib
+check_py tzinfo
+#check_py utils.unittest
+#check_py syncdb # also catches pre_syncdb and post_syncdb
+check_py IPAddressField
+check_py handle_app
+check_py RequestSite
+check_py get_current_site
+check_py runfcgi
+check_py SortedDict
+check_py declared_fieldsets
+check_py django.contrib.admin.util
+check_py django.contrib.gis.db.backends.util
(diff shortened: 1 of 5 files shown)


deploy/future
```diff
@@ -0,0 +1,172 @@
+#!/bin/bash
+# Don't run this with -e, grep exits with 1 when it finds no results
+
+function check_py () {
+# This is a bit janky. If you know how to get the output
+# to go on separate lines, please tell me.
+    results=`git --no-pager grep $1|grep "\.py"|grep -v "future-ok"`
```
benjaminjkraft: I'd use some variation on the more standard `# noqa` instead, just for consistency (see e.g. [flake8](https://pypi.python.org/pypi/flake8)).

It looks like you can just do `git grep $1 -- '*.py'` instead of grepping the output to get the extension.  And you shouldn't need `--no-pager`, I think it should autodetect that if the output is not connected to a terminal.

deploy/future
```diff
@@ -0,0 +1,172 @@
...
+    if [ "$results" ]; then
+	echo $results
```
benjaminjkraft: I believe if you do ```results="`git grep ...`"``` above, then `echo "$results"` here, you'll get the newlines.  (Bash quoting is really annoying >.< .)

deploy/future
```diff
@@ -0,0 +1,172 @@
...
+#check_py Storage
+#check_py success_url
+check_py GeoQuerySet
+#check_py ContentType
+check_py allow_migrate
+#check_html cycle
+check_py Signer
+
+if [ $EXIT = 0 ]; then
+    echo "success!"
```
benjaminjkraft: If you want you can steal the code from `deploy/lint` to make this print on the same line if there was no output.

deploy/future
benjaminjkraft: Although, actually, instead of capturing the output, can you just do `git grep ... || EXIT=1` at the toplevel, and skip the reprinting and conditional entirely?  (This might also fix the `-e` issue.)

deploy/future
```diff
@@ -0,0 +1,172 @@
...
+
+echo "Running future Django lint checks..."
+
+# Uncomment checks here once you've deleted all the uses from the codebase,
+# or checked that we never used the feature in the first place,
+# or checked that future changes won't break your use of this feature.
+# Add more checks once a new version of Django is released,
+# and delete them once we've upgraded main to a new version and
+# cleared out old pull requests.
+# Only push to future-django-support if what you're doing won't work
```
benjaminjkraft: What does this mean?  What is "future-django-support"?

deploy/future
```diff
@@ -0,0 +1,172 @@
...
+# A lot of these are impossible or annoying to write without some
+# false positives, and some ungreppable things are left out entirely.
+# Many of them should be considered warnings rather than errors: check
+# that a future behavior change won't break your code.
+# Add "#future-ok" to the end of a line if you've checked,
+# or rewrite the check to avoid the false positives.
+
+# Django 1.9 backwards incompatibility checks
+#check_py is_usable
+#check_py "\.add(" # warning
```
benjaminjkraft: Turning on this check probably won't ever work, due to `set.add()`.

benjaminjkraft:
Cool!  I have two contradictory thoughts, both of which you're welcome to ignore, but I figured I'd put them out there.

The first one is that it might be nice to put this in the lint script, or to make a script that calls them both, so that we can use it as a git hook.  But maybe it's better to just defer this to be done separately (along with automagic setup for the git hook in the dev setup scripts), since it's easy enough to just have one script call both.

The second one is that it might be worth trying to put this in a separate library, because it might be useful to others.  We'd need support for turning the checks on and off though.  A flake8 plugin might be a reasonable strategy if we wanted to do that.  But it's probably more work than it's worth.

deploy/future
mgersh: I intentionally used something different from what flake8 uses because they're testing different things. I'm not too happy with the name I came up with, though--do you have a better idea? Or do you think I should just not worry about the collision?

The `--no-pager` was necessary when I added it. I can check again, since I might have changed something relevant, but I think it probably needs to be there.

deploy/future
mgersh: That means I forgot to clean up the comment after I discovered that a thing I was originally trying to do doesn't work.

deploy/future
mgersh: Yeah, and it's not the only one like that in here. In this case, the list of results isn't too long to look at by hand, so I figure it's more useful to leave this in here as a reminder. Some of the others will need to be rewritten if they're going to be at all useful, though.

mgersh:
Thanks for the help with bash!

I'd be happy moving this into the main lint script. There was originally a reason for them to be separate, but then I changed some things and there isn't anymore. I don't think I want to try to make this usable for other projects, it's too inherently messy.

deploy/future
benjaminjkraft: I wouldn't worry about the collision?  But if you want to keep it, I think `# nofuture` is what I would expect, if it is a bit amusing.

luac:
Only tangentially related to the content of the pull request, but while we're at it, could we (instead|also) have Travis run the tests under the latest released Django version as well?

mgersh:
So, I wanted to do that. The way I tried implementing it was starting a branch for future Django support (since there are some things we won't be able to change in main beyond just the Django version), and having Travis merge that branch into the current branch before running tests. It didn't work because Travis only does a shallow git clone, so you can't merge in other branches, and as far as I can tell there's no way around this.

mgersh:
I've actually tried running the tests locally though. The first thing it blocks on is some reversion stuff, which we could easily fix in main now, thanks for reminding me to actually push that change. The second thing it blocks on is argcache.

luac:
Seems like we ought to be able to make a setting that flips any changes we want to make for future Django, though I don't know what those changes are and if it would be too messy to do like that.

luac:
Oh, thanks for reminding me too, I should fix argcache.

mgersh:
Closing because most of what this does is now done better by having actual logging.

Issue #1982, opened 2015-12-20
Email processing fails when a duplicate recipient contains unicode


mgersh:
This will cause the entire email to fail, not just for that recipient. See https://github.com/learning-unlimited/ESP-Website/blob/stable-release-7/esp/esp/dbmail/models.py#L349.

benjaminjkraft:
I can't get to reproduce this -- when I try to reproduce in the abstract, `'%s' % u'föo'` works just fine, and when I try to actually repro it gives a unicode error on `/commfinal`, but then works just fine when I actually try to do `MessageRequest.process()`.  So this will require some more digging to figure out what was going on.

Dhirenderchoudhary:
Hi @hwatheod @willgearty  I’d like to take ownership of this issue. If this issues is unresolved Could you please assign it to me? I’ll start working on it and share updates . Thanks.

Dhirenderchoudhary:
@willgearty as I tried to fix the issue I found the issue is already solved and after researching according to me:  Issue  is already fixed in PR #3255 This issue can be closed

willgearty:
Thanks @Dhirenderchoudhary 👍

PR #1986, opened 2015-12-21
Set up logging

benjaminjkraft:
This sets us up to use django/python logging facilities.  In particular, we'll now log to a file in addition to serverlog, and error logs that don't raise an error will also go to serverlog and sentry (if configured).  This should make it a lot easier to do things like deprecations, since we'll be able to actually see whether any live site raised a deprecation warning, rather than having to guess based on whether anyone noticed one on a dev server; it will also mean that if we want to make sure we notice if something goes wrong, but still do something reasonable, we can log an error without 500ing.  I also cleaned up our error handling middleware while I was at it.

~~This still needs a bit more testing on a prod-like setup, but I figured I'd let people take a look in the meantime.~~

This is a start on some of the various things suggested in #1237.


luac:
What's the rationale for having scripts behave differently? It seems like it makes things more complicated.

luac:
Ah, I guess you need it so that scripts log to the console even with DEBUG=False, and if you have that you might as well tailor it a bit more anyway.

mgersh:
It seems like we really have three main cases:
- Code running unattended in prod. This includes the main site, but also includes `dbmail_cron` and `mailgate`. I don't have complaints about the settings for this one, but I'd say that the mail system stuff should be changed to go in this category. If it breaks, we want to get emails.
- The dev server equivalent of the above. No complaints about this one.
- People running scripts by hand or messing around in the shell. Would this log things like somebody making a typo in the shell? That seems likely to make the log full of things that look like problems but that we have no reason to ever care about. Maybe scripts should go to a separate log file? Also, how does this change what I see in the console/shell, if at all?

benjaminjkraft:
Hmm, good point @mgersh, `dbmail_cron` and `mailgate` should probably act like prod servers rather than scripts.  (Their stdout does already go to serverlog, but I doubt anyone ever reads it, so putting errors there (as we currently do) is not the most useful.  I think `dbmail_cron` at least already manually sends an email if it fails in certain ways, but that's also not the most ideal.)

For the shell, I agree it's a bit weird -- as-is it will indeed log errors in the shell.  It's easy enough to turn that off; I guess it probably isn't that useful.  Or we could put it in a separate file, although that's another set of local config to maintain.  I do want to make sure `manage,py` commands still get logged, though, since failures there are no more likely to be user error than failures in prod scripts, and errors are only a little more likely to be seen by the person running them (since many put out a bunch of spam); maybe those should log to a file but scripts shouldn't.

It will somewhat change what you see in the shell, in that things that were previously `print` statements that are now `logger.info` will be prefixed by `INFO:`.  (We can change that easily, but I think it's more useful than spammy, because they might be `ERROR:` instead of `INFO:`.)  You can also still `print` as normal, it will just be discouraged (forbidden?) in code not intended to be run in the shell.

esp/esp/dbmail/models.py
```diff
@@ -345,7 +348,7 @@ def process(self, debug=False):
                 # all recipients the first time.
                 newtxt, created = TextOfEmail.objects.get_or_create(**newtxt)
                 if not created:
-                    if debug: print 'Skipped duplicate creation of message to %s for message request %d: %s' % (send_to, self.id, self.subject)
+                    logger.warning('Skipped duplicate creation of message to %s for message request %d: %s', send_to, self.id, self.subject)
```
mgersh: Will this still error if there's unicode in the user's name?

benjaminjkraft: It looks like it prints a unicode error, but continues on, which is also not the most ideal.  In any case I can just make the actual fix while I'm at it.

benjaminjkraft: (Going to punt that since it seems like something more complicated is going on, see #1982.)

benjaminjkraft:
Updated the code and did some more testing, this is ready to go as far as I'm concerned.  I believe I've addressed all the comments except the ones I said I wouldn't.

mgersh:
:rocket:

PR #1987, opened 2015-12-21
Add mechanism for fabfile customization

mgersh:
So we can add our favorite commands without cluttering up the fabfile.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,7 @@ theme_compiled.css
 *_admin_thumbnail.gif
 *_admin_thumbnail.jpg
 *_admin_thumbnail.png
+local_fabfile.py
 
 # Mac
 .DS_Store
--- a/fabfile.py
+++ b/fabfile.py
@@ -376,3 +376,8 @@ def runserver():
     ensure_environment()
 
     manage("runserver 0.0.0.0:8000")
+
+try:
+    from local_fabfile import *
+except ImportError:
+    pass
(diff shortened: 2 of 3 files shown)


fabfile.py
```diff
@@ -130,6 +130,15 @@ def setup():
...
 
+    # Create a file for personal Fabric commands
+    # but only if it's not already there
+    if not files.exists(env.rbase + "fabfile_personal.py"):
+        files.upload_template(
+            join(env.lbase, "deploy", "config_templates", "fabfile_personal.py"),
+            env.rbase + "fabfile_personal.py",
+            backup=False
+        )
+
```
btidor: I'd rather not create a personal `fabfile` by default. I think it would be cleaner for people who don't want to use one.

.gitignore
```diff
@@ -27,6 +27,7 @@ theme_compiled.css
 *_admin_thumbnail.gif
 *_admin_thumbnail.jpg
 *_admin_thumbnail.png
+fabfile_personal.py
```
btidor: Not important, but could we call it `local_fabfile.py` to be in line with `local_settings.py`?

btidor:
A few minor comments, but looks good!

fabfile.py
mgersh: Sure. Since this makes it less discoverable, I added a note in the documentation about it, which probably should have been there in the first place anyway.

.gitignore
mgersh: Done.

btidor:
Awesome!

PR #1992, opened 2015-12-27
Bugfix: cannot loaddb using saved settings

btidor:
Error message:
```
/bin/bash: /mnt/encrypted/fabric/dbconfig: Permission denied
```

Cause: instead of `cat`-ing a file, we try to execute it. Oops!

I'm going to merge this to `main` and `mit-prod`, since those are the two branches I know are affected.


benjaminjkraft:
Also seems fine to me.

PR #1993, opened 2015-12-27
Fabfile: support username detection in more dump formats

btidor:
It turns out that the LU database dumps use the following format for their grant lines:
```
ALTER TABLE public.program_class OWNER TO umbc;
```

This patch makes the Postgres username detection in `fab loaddb` support this format.


benjaminjkraft:
BTW, I think we usually do reviews even for small changes like this.  (But it looks fine to me.)

PR #1995, opened 2015-12-28
Add list of all tags, and logging to maintain it

mgersh:
Also add warnings for nonstandard tag uses, like calling getTag with a program instead of getProgramTag or returning a boolean-looking value without using getBooleanTag. This involved some minor refactoring.

This is intended to be partial work towards eventually having a frontend for setting tags. I filled out some of the help text entries but got lazy and left most of them blank, since they aren't being used right now. The logging doesn't do anything yet, but once #1986 is merged it'll give us a list of which tags are being used in weird ways so we can clean them up.


esp/esp/tagdict/models.py
```diff
@@ -33,8 +38,25 @@ def __unicode__(self):
...
     def getTag(cls, key, target=None, default=None):
+        if key not in all_global_tags:
+            logger.warning("Tag %s not in list of global tags", key)
+        if target is not None:
+            logger.warning("getTag() called for key %s with specific target; consider using getProgramTag()",
+                           key)
+
+        result = cls._getTag(key, target=target)
+
+        if result and (result.lower() == "false" or result.lower() == "true"):
```
benjaminjkraft: perhaps also if `default` is either a boolean or one of these strings?

esp/esp/tagdict/models.py
```diff
@@ -75,8 +99,13 @@ def getProgramTag(cls, key, program=None, default=None, ):
     def getBooleanTag(cls, key, program=None, default=None):
         """ A variant of getProgramTag that returns boolean values.
             The default argument should also be boolean. """
-
-        tag_val = Tag.getProgramTag(key, program)
+        if key not in all_boolean_tags:
+            "getBooleanTag warning"
```
benjaminjkraft: Probably unintentional?

esp/esp/tagdict/__init__.py
```diff
@@ -0,0 +1,107 @@
...
+    'user_types': "",
+    'studentinfo_shirt_options': "",
+    'studentinfo_food_options': "",
+    'ask_about_duplicate_accounts': "",
+    'require_email_validation': "",
+}
+
+# Any tag used with Tag.getBooleanTag()
+all_boolean_tags = {
+    'increment_default_grade_levels': "Consider all students to be one grade level higher (useful for summer programs)",
```
benjaminjkraft: Some of these have docstrings here but not below.  It might be worth thinking about whether there's a good way to avoid duplication.

esp/esp/tagdict/models.py
```diff
@@ -57,15 +79,17 @@ def getProgramTag(cls, key, program=None, default=None, ):
         Given a key and program, return the corresponding value as string.
         If the program does not have the tag set, return the global value.
         """
+        if key not in all_program_tags:
+            logger.warning("Tag %s not in list of program tags", key)
```
benjaminjkraft: Should this also have a boolean warning?  (If so, you'll need to make sure it doesn't get tripped by `getBooleanTag` calling `getProgramTag`.)

benjaminjkraft:
Whoo ++ very exciting!

esp/esp/tagdict/models.py
mgersh: Sure, except there's at least one case where `default` is `True` (not the string) so I'd have to write it a bit differently. That's also a thing I should try to catch, actually.

esp/esp/tagdict/models.py
mgersh: Yup.

esp/esp/tagdict/__init__.py
mgersh: Right now it's just this one, but there will be more after some cleanup happens. Maybe what I really want is to put the boolean tags in the other two lists with an indicator that they're boolean. I'll think more about this.

PR #2014, opened 2016-01-31
Prevent student name leakage

benjaminjkraft:
The teacher bio page's error when the user is not a student previously included
the user's name; this allowed figuring out a student's name from their
username.  Now it will just show the same error page as if the user didn't
exist, was inactive, or had their bio hidden.


mgersh:
:+1:

PR #2015, opened 2016-01-31
Deprecate and redirect old teacher bio URLs

benjaminjkraft:
The canonical URLs for bios are /teach/teachers/<username>/bio.html.  But
it turns out you can also access them at
/teach/teachers/<lastname>/<firstname><num>/bio.html and at both of those
URLs with /learn/ instead of /teach/.  This commit makes the deprecated
URLs permanently redirect to the correct ones; then sometime in the future
we can just remove the old ones and simplify the code a good bit.

This is on top of #2014 to avoid merge conflicts.

### DIFF
--- a/esp/esp/urls.py
+++ b/esp/esp/urls.py
@@ -81,12 +81,13 @@ urlpatterns += patterns('',
 
 urlpatterns += patterns('esp.web.views.bio',
 
-                        # bios
-                        (r'^(?P<tl>teach|learn)/teachers/(?P<last>[-A-Za-z0-9_ \.]+)/(?P<first>[-A-Za-z_ \.]+)(?P<usernum>[0-9]*)/bio\.html$', 'bio'),
+                        # bios (/learn URLs are deprecated)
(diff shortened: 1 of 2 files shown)


mgersh:
Huh, I didn't realize /teach/ was considered the canonical URL for these things. They actually make more sense to me under /learn/, because the point of having them is that you click on a name in the catalog and find out who the person teaching the class is. Thoughts?

benjaminjkraft:
I don't have a strong opinion about this.  /learn/teachers/... comes out a little weird for me, but I agree that in practice they are for students, not for teachers.

Since we have been using /teach/ as canonical, if we switch to /learn/ we should probably keep that redirecting for longer, but that's okay -- it's not much added code, whereas firstname/lastname is added code and added potential privacy leaks.

benjaminjkraft:
Anyone else have thoughts on which URL should be canonical?

benjaminjkraft:
@btidor said he has a slight preference for /teach/, but that in any case we should just make a decision.  @mgersh if you're okay with it I think I'll propose we stick to /teach/ since that's what we've been linking to.

btidor:
@mgersh said she's okay with it

PR #2016, opened 2016-01-31
Cleanup of tag warnings

mgersh:
I checked these for safety with existing chapter data, none of them should be a behavior change for anyone. This gets rid of all the warnings that show up in tests except for the ones about the 'test' tag, which I'm leaving there for now since it's testing behavior that's deprecated but might still be used. Any warnings we see after this and logging are merged will be in code that doesn't have test coverage.


benjaminjkraft:
:tada: Thank you, I have wanted this for so long!

Issue #1720, opened 2015-07-01
Race condition in ProgramModuleObj creation can create duplicates


mgersh:
See https://github.com/learning-unlimited/ESP-Website/blob/main/esp/esp/program/modules/base.py#L234. I noticed this in the code a while ago and figured that it was unlikely to come up in practice and easy enough to clean up if it does, but Yale just ran into an issue that could have been caused by this.

luac:
I've also seen duplicate ProgramModuleObjs appear before, this could be why.

imran-techvoyager:
Hi @mgersh, cc @willgearty I was reviewing #1720 and noticed that while #2017 added a `unique_together` constraint on `ProgramModuleObj`, the `getFromProgModule` logic still follows a `filter()` + manual `save()` pattern.

With the database constraint in place, this prevents silent duplicates, but concurrent requests can now surface as an `IntegrityError` instead of resolving cleanly. In other words, the race condition has been constrained at the DB layer but not eliminated at the ORM level.

Would you be open to replacing that block with `get_or_create(defaults=...)`? Combined with the existing constraint, that would make the creation path atomic and remove the now-unreachable `"Too many module objects!"` assertion branch.

I’d be happy to submit a focused PR refactoring just this method if that approach sounds good.

willgearty:
Thanks for tracking that down @imran-techvoyager  This sounds like a good approach, looking forward to the PR 👍

PR #2017, opened 2016-02-03
Add unique_together constraint on ProgramModuleObj.

luac:
It's silly that we keep having to deal with "Too many module objects!" errors caused by people misconfiguring their program modules.


luac:
Whoops, migration conflict. Will fix.

mgersh:
It's not even user error, it's a race condition, see #1720. But fixing it this way will at least cause the error to only happen once instead of continuing to happen until someone deletes the extra PMO. Will merge once tests pass.

PR #2032, opened 2016-02-14
Use EMAIL_HOST_SENDER instead of EMAIL_HOST in templates.

uakfdotb:
EMAIL_HOST is the hostname through which outgoing e-mail is routed by Django's SMTP mail backend. (From what I can tell,) EMAIL_HOST_SENDER is there so that we can have a different visible hostname; for example, the from address for availability e-mails is based on EMAIL_HOST_SENDER. This commit fixes the templates so that this setting is used instead of EMAIL_HOST whenever server e-mail addresses are displayed.

This is needed because e.g. MIT has a different mailserver hostname than the desired e-mail domain.


mgersh:
:+1:

PR #2033, opened 2016-02-15
Fix various warnings from tests and Sentry

mgersh:
Like I said...I compulsively fix warnings.


luac:
This creates more merge conflicts with #1839, could we possibly merge that first?

mgersh:
Is that in any way close to merging? If it is, that's great, I support. But last I heard, most of it hasn't been reviewed at all and nobody has said anything about taking that on.

luac:
I'm not sure what the merit is of having code that we dig up every half year in order to use it because we sort of have to, while segregating it from the main code base because we haven't reviewed it. Either we're comfortable using it, or we're not, and I don't like the current state of affairs where changes like this one and moving code around and fixing whitespace can cause huge merge conflicts and either I end up fixing them again in 6 months or someone less obsessive does it and the code slowly rots from bugs introduced by merging.

mgersh:
Yeah, I agree there's a problem, I just hadn't heard anyone proposing solving it by merging it now. Want to say something on that PR or in email or something?

PR #2035, opened 2016-02-17
Exit early from assign_room on room-does-not-exist

luac:
The function checks that rooms_to_assign is the same length as
meeting_times, but if it fails that check, it goes ahead and starts
assigning in the ones that exist anyway. (MIT ESP actually encountered
the case where there were duplicate room objects, so the rooms_to_assign
was presumably actually longer than meeting_times.)

It looks like there's an allow_partial option that controls whether or
not to keep going even if there are errors, so let's use that.


benjaminjkraft:
Seems fine.  We should probably clean up some of the unused options in this function, but that can be another day.

PR #2039, opened 2016-02-19
Cybersource settings

btidor:
Remove hard-coded Cybersource configuration and move it into Django settings. This will make it easier for other chapters to use the Cybersource module, and it will also ensure that if a chapter inadvertently enables the module, it won't default to sending payments through MIT.


uakfdotb:
I think we usually add settings to django_settings.py.

btidor:
Oh, thanks. I updated this to match how `STRIPE_CONFIG` is set up, but without the fancy per-program tag overrides.

PR #2043, opened 2016-02-21
Fix shebang lines

benjaminjkraft:
Several of our `#!` lines are `#!/usr/bin/python` which doesn't do the right
thing if you are in a virtualenv.  (In particular, most of the scripts will
auto-activate the virtualenv if you run them from outside of any virtualenv,
but if you run them from inside the virtualenv but with /usr/bin/python as your
interpreter, they may do the wrong thing.)  This changes them to use
`#!/usr/bin/env python` which is more universal.

### DIFF
--- a/esp/dbmail_cron.py
+++ b/esp/dbmail_cron.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python
 
 import sys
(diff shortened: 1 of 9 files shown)


jmoldow:
Looks good to me.

PR #2061, opened 2016-03-03
Prettify front page login box for logged in users

betaveros:
- center text
- put name in \<strong\>s
- add an 18px margin, matching the login form for users that aren't
  logged in

pretty-login-box

This is an utterly trivial change, but doesn't it look so much prettier?

(We can apparently correctly vertically center the text with some combination of `display: table` and `display: table-cell` and `vertical-align: middle`, but I didn't think it was worth it; the form that appears if you're not logged in uses 18px anyway.)


btidor:
Thank you!

PR #2062, opened 2016-03-03
Cybersource: simplify the payment page

btidor:
Entering invalid information (e.g. an incorrect state abbreviation) causes Cybersource to return with an ERROR code. Because we auto-fill the country as `US`, this makes it impossible for non-U.S. students to pay. The simplest way to solve this problem is to get rid of all the fields and have people enter this information directly into Cybersource.


benjaminjkraft:
Is it possible to autofill things in a way such that users can change it?

btidor:
Yes, but it seems like a lot of work to get it right. If we autofill a value that Cybersource thinks is improperly formatted, it will send users back with an error. I don't think it's worth it right now.

PR #2073, opened 2016-03-10
make username search case-insensitive (fixes #1533)

### LINKED ISSUE #1533: User search should not be case-sensitive on usernames
This has led to difficulties in finding users if they remember their username but not the capitalization.

betaveros:
surprisingly straightforward :D


benjaminjkraft:
LGTM once tests pass.

PR #2074, opened 2016-03-10
Create README

benjaminjkraft:
We should really have a README for people who end up on the repository page.  This one is pretty skeletal, but we can add to it as we go.


mgersh:
Want to delete esp/README now that you've added a better one?

benjaminjkraft:
Good call, done.

PR #2075, opened 2016-03-10
add diff view for TemplateOverrides (fixes #2072)

### LINKED ISSUE #2072: Tool for diffing template overrides
Even if we don't get around to doing something like #692, it would be nice to have a built-in way to display the diff of a template override against the original.  (Or, if that's too complicated, even a quick script to do it would have some value.)  This would make it easier to see what changes chapters have made, and whether they will cause any issues as we update the code.

betaveros:


luac:
It looks like the lint build is failing because ESPError wasn't imported.

Also, do we want to link this from somewhere? Seems like the admin panel would be ideal, but I'm not sure if there's a way to put a link there (maybe by defining `TemplateOverride.get_absolute_url` to be the diff view? seems like a bit of an abuse of that feature). Alternatively just adding something to the admin docs.

benjaminjkraft:
Cool, this looks great!  The only thing I would say it needs is a link somewhere.  I like doing `TemplateOverride.get_absolute_url()`, because then there will be a magic "view on site" link on the admin panel page for the TO.  But a link and a listing page somewhere else would also work.

betaveros:
(have not tested that --- hope it works oops)

benjaminjkraft:
:100: thanks!

Issue #353, opened 2012-04-21
Link to Account Management Page in Templates


jmoldow:
Sites using templates that aren't the MIT ESP template don't have a link to an account management page. At my SplashCon talk about the website, I realized that most chapters don't know that you can disable an account, probably because there are no pages on their sites that link to that feature.

We should create a page identical to http://esp.mit.edu/myesp/accountmanage.html on all of the sites (or, better yet, implement Issue #352), and then link to this page in all main.html template overrides.

PR #2085, opened 2016-03-22
Add a "Problems" section to vagrant.rst docs

hwatheod:
I recently setup a dev server and ran into a few issues, so I documented them in the vagrant.rst file.
Also fixed a broken link in the fruitsalad theme.

### DIFF
--- a/docs/dev/vagrant.rst
+++ b/docs/dev/vagrant.rst
@@ -1,8 +1,5 @@
 Vagrant based dev servers
 =========================
-Authors:
-   - Michael Price <price@learningu.org>
-   - Theodore Hwa <ted@learningu.org>
 
 .. contents:: :local:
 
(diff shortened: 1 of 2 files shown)


btidor:
Thanks! It looks like you pulled in a change to the fruitsalad theme, but feel free to merge once that's fixed.

mgersh:
Good idea on adding these notes to the docs. The memcached error happens on loaddb too, and pretty much everyone has been running into it.

More specifically about the fruitsalad change: it's linking to a QSD page intentionally, which will be a "broken link" if you're setting up a new site but isn't intended to stay that way. Sites using fruitsalad should create the page. See also #353.

Also, can we just delete the authors section from this file? We don't usually have them in individual files, and this one is out of date.

Issue #2089, opened 2016-03-27
Approving all sections from the status button doesn't work if class is already "Approved but hidden" 


hwatheod:
To reproduce:
- Mark a class as "Accepted but hidden" from the manage page for that class
- Go back to the dashboard, it will say "[Approved but hidden]" for the class
- On the dashboard, click "Status" for the class, then "Approve (all sections)"
The display on the dashboard will change to "[Approved]", but it didn't really get approved.  If you refresh the dashboard, it will say "[Approved but hidden]" again.  This isn't a caching issue, since it can be confirmed by looking at the class status directly from the db in shell_plus.

benjaminjkraft:
This is/was intended behavior, but I agree it's a bit confusing -- see discussion on #2090.

PR #2090, opened 2016-03-27
Make the approve/unreview/reject buttons on the status popup behave correctly in all cases

hwatheod:
This fixes issue #2089 by creating a new accept_all_sections() function in class_.py that doesn't do any special checks (unlike accept()), then calling that function from adminclass.py.  While I was at it, I refactored some similar code in class_.py.

### DIFF
--- a/esp/esp/program/models/class_.py
+++ b/esp/esp/program/models/class_.py
@@ -1731,19 +1731,30 @@ was approved! Please go to http://esp.mit.edu/teach/%s/class_status/%s to view y
             user = AnonymousUser()
         return True
 
+    def set_all_sections_to_status(self, status):
+        self.status = status
+        self.save()
+        for sec in self.sections.all():
+            sec.status = status
+            sec.save()
+
+    def accept_all_sections(self):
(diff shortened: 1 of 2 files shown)


benjaminjkraft:
I think this was intended behavior at some point -- the idea being that if some sections have been marked as cancelled/rejected, the right thing to do if the class is accepted is to mark the unreviewed sections as accepted, but leave the cancelled/rejected ones as-is.  It would be nice to keep that working as-is.  Some options I can think of to solve your problem (which is a case where I agree it's a bit confusing):
* Change all sections that have the same status as the class's old status
* Change all sections with status >= unreviewed
* Document the existing behavior better
* Have one button that does the current behavior and one that does the same thing to all sections

hwatheod:
I don't think that's the existing behavior.  In lines 159-161 of the existing code in adminclass.py, it sets all sections of the class to status 10 regardless of their current status.

benjaminjkraft:
Ah, I see, we already have inconsistent behavior between here and the manage page.  That seems silly to me, but in any case this seems like it is if anything an improvement.

PR #2092, opened 2016-03-29
getListDescriptions() shouldn't be cached

jmoldow:
This is because arbitrary program modules may add new lists at any time.

Program.students(), Program.teachers(), etc. are already not cached.


benjaminjkraft:
This is pretty cheap to recompute in any case.

PR #2093, opened 2016-03-29
Don't clear teacher interview or training signups when editing class availability (fix #2071)

### LINKED ISSUE #2071: Editing teacher availability clears teacher training or interview signup
Reported by Will Gearty from Stanford.

hwatheod:
Add a function clearAvailableClassTimes() which clears only Class Time Block events, not all events (for a particular teacher and program).  This fixes #2071.

It's possible that this should just be the default behavior for clearAvailableTimes() since usually we think of "availability" as the times when a teacher can teach, not the times for attending other events.  But I wasn't sure if that would break anything so I created a separate function.

### DIFF
--- a/esp/esp/program/modules/handlers/availabilitymodule.py
+++ b/esp/esp/program/modules/handlers/availabilitymodule.py
@@ -167,8 +167,8 @@ class AvailabilityModule(ProgramModuleObj):
             #   Process form
             post_vars = request.POST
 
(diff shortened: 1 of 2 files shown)


benjaminjkraft:
From a quick `git grep`, it looks like this is the only non-test caller of the function.  So I think you can just change it, rather than adding a new one.

benjaminjkraft:
:100:

PR #2095, opened 2016-04-02
add date of birth as an option to listgen

hwatheod:
Needed by Stanford to complete "working with minors" registration form requested by the university.


hwatheod:
This is already merged into stanford-sr7 since they needed it urgently

benjaminjkraft:
Looks good, thanks.

PR #2097, opened 2016-04-07
Add setting to prevent theme clobbering

mgersh:
Some sites are still on themes that aren't in generic templates. Ideally they
wouldn't be, but until we fix that, this will make it harder to accidentally
clobber one of these themes. It also makes manage.py update safe to use on these
sites, which should make releases easier.


benjaminjkraft:
Good idea!

mgersh:
I added this setting to local_settings.py for Chicago, Stanford, Berkeley, and UMBC, so it should work next time we do a release.

Issue #2103, opened 2016-04-10
Generate lunch classes without generating constraints


hwatheod:
The lunch constraints form currently generates both lunch classes and lunch scheduling constraints when you click submit.  Stanford doesn't want lunch scheduling constraints, but still needs to generate lunch classes.  

Right now the workaround is to use the lunch constraints form to generate both classes and constraints, then delete the constraints from the admin page.

hwatheod:
Since this is likely to be a consistent policy from one program to the next, it should be represented in a Tag.

PR #2105, opened 2016-04-11
modify lunch constraints module to allow generating lunch sections without generating constraints

hwatheod:
Added an additional checkbox (generate_constraints) to the lunch constraints module which indicates whether to really generate lunch scheduling constraints, or generate lunch sections only.  Fix for #2103 

Also modified the instructions on lunch_constraints.html accordingly.

### DIFF
--- a/esp/esp/program/modules/forms/admincore.py
+++ b/esp/esp/program/modules/forms/admincore.py
@@ -5,7 +5,6 @@ from django.utils.safestring import mark_safe
 from esp.cal.models import Event
 from esp.program.models import RegistrationType
 from esp.program.controllers.lunch_constraints import LunchConstraintGenerator
-from esp.tagdict.models import Tag
 
 def get_rt_choices():
     choices = [("All","All")]
@@ -33,12 +32,6 @@ class LunchConstraintsForm(forms.Form):
     def load_data(self):
         lunch_timeslots = Event.objects.filter(meeting_times__parent_class__parent_program=self.program, meeting_times__parent_class__category__category='Lunch').distinct()
         self.initial['timeslots'] = lunch_timeslots.values_list('id', flat=True)
-        should_generate_constraints = True
-        if Tag.getTag('no_lunch_constraints'):
-            should_generate_constraints = False
-        self.initial['generate_constraints'] = should_generate_constraints
-        self.initial['autocorrect'] = should_generate_constraints
-        self.initial['include_conditions'] = should_generate_constraints
 
     def save_data(self):
         timeslots = Event.objects.filter(id__in=self.cleaned_data['timeslots']).order_by('start')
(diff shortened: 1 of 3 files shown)


esp/esp/program/modules/forms/admincore.py
```diff
@@ -32,13 +33,21 @@ def __init__(self, program, *args, **kwargs):
...
+        if Tag.getTag('no_lunch_constraints'):
+            should_generate_constraints = False
+        self.initial['generate_constraints'] = should_generate_constraints
+        self.initial['autocorrect'] = should_generate_constraints
+        self.initial['include_conditions'] = should_generate_constraints
 
     def save_data(self):
         timeslots = Event.objects.filter(id__in=self.cleaned_data['timeslots']).order_by('start')
-        cg = LunchConstraintGenerator(self.program, timeslots, autocorrect=(self.cleaned_data['autocorrect'] is True), include_conditions=(self.cleaned_data['include_conditions'] is True))
+        cg = LunchConstraintGenerator(self.program, timeslots, generate_constraints=(self.cleaned_data['generate_constraints'] is True), autocorrect=(self.cleaned_data['autocorrect'] is True), include_conditions=(self.cleaned_data['include_conditions'] is True))
```
benjaminjkraft: I don't see why all of these need the `is True` given that they're `BooleanField`s which can only be `True`, `False`, or perhaps `None`.

esp/esp/program/modules/forms/admincore.py
```diff
@@ -32,13 +33,21 @@ def __init__(self, program, *args, **kwargs):
     def load_data(self):
         lunch_timeslots = Event.objects.filter(meeting_times__parent_class__parent_program=self.program, meeting_times__parent_class__category__category='Lunch').distinct()
         self.initial['timeslots'] = lunch_timeslots.values_list('id', flat=True)
+        should_generate_constraints = True
+        if Tag.getTag('no_lunch_constraints'):
```
benjaminjkraft: Wait, so is this tag just the default value for the form?  That seems unnecessary; can we just keep things simple?  (Also, if we're keeping it, it should be a `getBooleanTag`.

esp/esp/program/modules/forms/admincore.py
hwatheod: That's a good point.  I was just following the existing code.  I will clean it up.

esp/esp/program/modules/forms/admincore.py
hwatheod: Since lunch constraints are a policy that's likely to remain stable for a given chapter, I was trying to make it possible to set a default, but it's just one checkbox so you're right that it's probably unnecessary.  I will get rid of it.

esp/esp/program/modules/forms/admincore.py
benjaminjkraft: It shouldn't matter.  But that's okay, we can leave it as-is.

Issue #2106, opened 2016-04-12
Survey Review under Manage Pages Redirects to Teach Pages


willgearty:
When reviewing survey results from a recent program (e.g. https://www.stanfordesp.org/manage/Splash/2016_Spring/surveys/review), often it is helpful to be able to read all survey responses from a single survey respondent.
However, clicking on a survey response (which should produce this desired effect), takes the admin user to the corresponding `/teach` page (e.g. [here](https://www.stanfordesp.org/teach/Splash/2016_Spring/survey/review_single?4743)) instead of the `/manage` page (e.g. [here](https://www.stanfordesp.org/manage/Splash/2016_Spring/survey/review_single?4743)).
Since not all students are in the admin's class(es) or the admin may not have any classes, this results in an error.

I assume this effect is a result of the shared code between `/manage/Splash/2016_Spring/surveys/review` and `/teach/Splash/2016_Spring/surveys/review`?

Would it be possible to alter the code (either here or with a template override) to have separate code for these two pages?

hwatheod:
Fixed and merged into stanford-sr7 and main.

PR #2109, opened 2016-04-13
make review_single links on manage survey review page go to manage not teach

hwatheod:
Fix for #2106.
I will push into stanford-sr7 soon because they're reviewing survey responses now.


benjaminjkraft:
Seems fine (once tests pass).

PR #2111, opened 2016-04-14
Update programprintables.py

milescalabresi:
if the "class" is a compulsory event, grab start time via the appropriate attribute


benjaminjkraft:
We need a conditional here.  The issue is that if `classes[i]` is a class, hitting `classes[i].start_time` is much slower, because it requires a DB hit per class (per student).  Even on a small site, that's a lot.  So we need to only use `classes[i].start_time` if `classes[i]` is actually an event, and use `classes[i].start_time_prefetchable()` otherwise.

milescalabresi:
ahhh I did something wrong

benjaminjkraft:
Closing in favor of #2112.

PR #2112, opened 2016-04-15
Fix bug in schedules with compulsory events

mgersh:
See #2111

### DIFF
--- a/esp/esp/program/modules/handlers/programprintables.py
+++ b/esp/esp/program/modules/handlers/programprintables.py
@@ -959,7 +959,11 @@ Volunteer schedule for %s:
             for t in times_compulsory:
                 i = min_index
                 while i < len(classes):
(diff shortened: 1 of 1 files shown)


benjaminjkraft:
Thanks @mgersh!

hwatheod:
I made several bug fixes recently (see #2113 , #2093 ) which filter events by Class Time Blocks and now that I see this, I'm wondering if they should actually be filtering by ('Class Time Block', 'Compulsory').

What is the use case for compulsory events?  Should they be treated as equivalent to Class Time Blocks for most purposes?

mgersh:
They're for when a program wants every student to be in the same place at a certain time, and wants it to appear on their schedules. I believe it's only relevant to student schedules; definitely neither of the cases you linked needs to know about them.

PR #2113, opened 2016-04-15
fix scrolling class list so only class time block events are considered

### LINKED ISSUE #2110: Scrolling class list should consider only class time blocks
The scrolling class list should consider only class time blocks when displaying the list of open classes, not time blocks for other events such as volunteer time blocks.  Right now, if there is a volunteer time block that starts at e.g. 10:30 then it will display a blank list on the list of open classes at 10:30.

hwatheod:
Fixes #2110.  Also took the opportunity to add a comment on what timedelta(-1, 85200) does.


benjaminjkraft:
Makes sense, thanks.

PR #2116, opened 2016-04-28
Fix issue where QSD edit is not saved in some cases.

uakfdotb:
Specifically, if only title is edited and other fields stay the same, then
the edit would not be made. The code is cleaned up so the condition and copies
aren't duplicated.


benjaminjkraft:
Nice cleanup.

Issue #2104, opened 2016-04-10
Big Board shouldn't show popular classes if there is no lottery


benjaminjkraft:
Right now it shows a blank list for priority/1 classes and a list of classes with 0 stars; it should just hide those if they're all zero.  It could alternately show a list of classes with the most regs or something, but that's perhaps less meaningful.

PR #2120, opened 2016-05-14
hides popular classes when no classes

tankersleyk:
see #2104


benjaminjkraft:
Sweet, this looks great.  Thanks!

Issue #1292, opened 2014-08-29
Organize urls better


jmoldow:
Our system of dealing with urls is not as great as it should be. I would like to see us eventually get to a more organized url system, by:
- Using the `url()` function (https://docs.djangoproject.com/en/1.4/topics/http/urls/#django.conf.urls.url), in order to
- Use named urls (https://docs.djangoproject.com/en/1.4/topics/http/urls/#naming-url-patterns)
- Split central monolithic URLconf into app-specific URLconfs, which are included in the root URLconf `esp/esp/urls.py` (https://docs.djangoproject.com/en/1.4/topics/http/urls/#including-other-urlconfs)
- Stop hard-coding urls, and instead use named urls to call urlresolver utility functions (https://docs.djangoproject.com/en/1.4/topics/http/urls/#django-core-urlresolvers-utility-functions), especially `reverse()` and the `{% url %}` template tag.
- Define `get_absolute_url()` (https://docs.djangoproject.com/en/1.4/ref/models/instances/#get-absolute-url), using the `@permalink` decorator (https://docs.djangoproject.com/en/1.4/ref/models/instances/#django.db.models.permalink), on models with canonical urls. This won't apply to most models, but it applies to `ESPUser`, some of the various resource models, `customforms.Form`, `TeacherBio`, and QSD, and maybe others.
- Even in places where it doesn't make sense to define `get_absolute_url()`, it might still be possible to write helper methods that use the `@permalink` decorator to make it easy to get canonical urls without hardcoding them. I can see this being useful on `Program` and `ProgramModule`s.

With regards to named urls: with the current architecture, we won't be able to take 100% advantage of named urls. This is because a large number of our view functions are hidden behind the program module framework. As a nearly-good-enough workaround, when reversing urls for program module views, reverse `esp.web.views.main.program`, with the `module` argument set to the view name. However, the true fix would be to refactor the program module mechanism to somehow use URLconfs, which can then be defined with named urls.

imran-techvoyager:
Hi all @jmoldow @mgersh @willgearty I did a current-state audit of URLs across the codebase to see what work remains under #1292.

PR #2121 addressed the deprecated patterns() and string-view arguments, and partially split URLconfs. That part looks complete.

What still remains is primarily:

Many hardcoded URLs in Python views (e.g. HttpResponseRedirect with string formatting).

Very limited use of named URLs and reverse().

get_absolute_url() methods that build strings instead of using reverse().

[...]

PR #2121, opened 2016-05-31
Clean up our URL patterns

mgersh:
The main goal of this commit is to remove uses of patterns() and string
view arguments to url(), both of which are deprecated. While I was doing
that, I moved some patterns into per-app files and cleaned up the main
urls.py a bit. Progress towards, but not a complete fix for, #1292.

(I still compulsively fix warnings.)


benjaminjkraft:
Whoo, thanks!  Looks good to me, although I didn't check every line super-carefully :see_no_evil: .

Issue #2008, opened 2016-01-20
Link to program survey results from list of all modules


jerrywu64:
there should exist a link to /manage/Sploosh/3010/surveys/review (or if not this page, then some other relevant page) from the complete module list on /manage/.../main.

mgersh:
It's possible that I caused this (and a similar bug on the learn/teach side) when I added the survey/surveys aliases in #1688. I don't have time to check right now, but if you're working on this, look there. (The page it should be linking to is just /surveys, since that's the main_call, not /surveys/review.)

willgearty:
Don't you just need to make `isStep()` return `True` for the survey management module?

PR #2125, opened 2016-06-06
Program modules refactors

benjaminjkraft:
This branch makes several refactors to program modules, to simplify things, and prepare for bigger refactors in the future.

The biggest change is to disallow "double-modules": modules with a single python class with multiple main calls for different `tl` values.  I've split both existing double-modules, and made changes to disallow them in the future.  There's more cleanup to be done around this, but this is at least a start that should allow other refactors to proceed.  Since double-modules already showed up separately in the admin panel and other UIs, I think this should not be a user-visible change, but it's possible there will be minor differences.

I also did some more minor refactors: factoring out some duplicated functionality in module view decorators, removing some unused arguments, and so on.  It's probably easiest to review this commitwise, and to merge it without squashing to keep them separate for future `blame`ing.


benjaminjkraft:
I realized that this might actually require a migration for the existing PMs/PMOs, so I'll probably need to add that before we can merge.

mgersh:
I know it's not done yet, but what you have looks good. The test failures might just be the survey/surveys aliases (#2008).

willgearty:
Closed in favor of #3464.

PR #2126, opened 2016-06-11
Dead code cleanup: per-program templates and emailprefs

mgersh:
See individual commits.


esp/esp/program/modules/base.py
```diff
@@ -292,22 +292,15 @@ def getTemplate(self):
...
             except TemplateDoesNotExist:
-                try:
-                    get_template(base_template)
-                    if self.useTemplate():
-                        return base_template
-                except TemplateDoesNotExist:
-                    pass
+                pass
 
             return 'program/modules/%s/%s' % (self.__class__.__name__.lower(), self.module.inline_template)
```
benjaminjkraft: Isn't this the same as `base_template`?  Can't you just return that unconditionally now?

esp/esp/program/modules/handlers/adminclass.py
```diff
@@ -388,11 +388,12 @@ def coteachers(self, request, tl, one, two, module, extra, prog):
                 error = 'Error - You already added this teacher as a coteacher!'
 
             if error:
-                return render_to_response(self.baseDir()+'coteachers.html', request,{'class':cls,
-                                                                                                 'ajax':ajax,
-                                                                                                 'txtTeachers': txtTeachers,
-                                                                                                 'coteachers':  coteachers,
-                                                                                                 'error': error})
+                return render_to_response(self.baseDir()+'coteachers.html', request,
+                                          {'class':cls,
```
benjaminjkraft: If you feel like it you could clean up the spaces after the colons too while you're at it.

benjaminjkraft:
> and one from #1830.

Eek!  Sorry about that.

esp/esp/program/modules/base.py
mgersh: Ah, so it is. This code didn't really make sense before I changed it either...

esp/esp/program/modules/handlers/adminclass.py
mgersh: I could! And I did :)

esp/esp/program/modules/base.py
```diff
@@ -292,23 +292,6 @@ def getTemplate(self):
         if self.module.inline_template:
             baseDir = 'program/modules/'+self.__class__.__name__.lower()+'/'
```
benjaminjkraft: I think you can even remove these two lines too now.  (Maybe some day we will turn on unused vars in the linter and it will catch these things!)

benjaminjkraft:
:ballot_box_with_check: feel free to merge after that fix.

esp/esp/program/modules/base.py
mgersh: Oh duh whoops.

PR #2127, opened 2016-06-13
Delete login by birthday and school

mgersh:
They leak user data. I wrote a more detailed explanation in the changelog.


benjaminjkraft:
:+1: from me, but cc @milescalabresi in case he has thoughts on communicating this to chapters.

milescalabresi:
Definitely support this change. I'll email chapter reps and check with them in person at the next meeting to make sure it gets communicated to the whole team. Given that it's the summer and that these don't affect much in terms of what admins can do or have to do, I have no big concerns if it doesn't get spread to everyone immediately.

mgersh:
Note that this won't take effect until next time we do a stable release anyway. (Although we should probably do one soon.)

milescalabresi:
Yeah, I just noticed that. Thanks for pointing that out. Ben, did you want my thoughts on communicating the stable release or communicating this change specifically? In either case, I'm happy to take responsibility for making sure the relevant info gets out to all admins in a reasonable amount of time. Not sure if that was what you meant though.

mgersh:
I think the specific concern here is users getting confused that the thing they'd used to log in before has disappeared and emailing chapters. Hopefully there won't be too much of that. But just in case, it might be nice to include an explanation and suggested response in the release email when it happens.

benjaminjkraft:
I meant this change specifically, since it's removing functionality without a replacement, and I think some chapters may actually use this feature, as @mgersh said.  I think it'll be okay, just something to keep an eye out for.

milescalabresi:
Okay, I'll make sure to highlight this specifically slightly before or when the release comes out in that case. Let me know if there are other such things that come up.

PR #2131, opened 2016-07-09
Delete consistency checks

### LINKED ISSUE #2037: Consistency checks should become scheduling checks
Some of the consistency checks in esp/program/controllers/consistency.py also exist as scheduling checks, but not all. Ideally they would all be available from the scheduling checks page in a form that doesn't take 15 minutes to run for a large program.

mgersh:
Most of them were already scheduling checks, and I added a new
scheduling check for the one that wasn't. This will make it easier to
run them all at once. They won't show up on manage class pages anymore,
but in my experience admins usually ignore those or don't see them in
the first place. Fixes #2037.


benjaminjkraft:
:100: thanks for checking this carefully to make sure all the cases are still covered.

PR #2132, opened 2016-07-09
Delete ClassImplications

### LINKED ISSUE #2029: Remove ClassImplications
This was a prerequisite system that we had in the past.  It saw some use at MIT Spring HSSP 2008 and a very small amount of use at Chicago Splash 2008, but none since then.  The code has probably bitrotted by now, and if we want to make a prerequisite system we will probably need to rewrite it anyway, so we should just remove them.

mgersh:
It was last used in 2008. Fixes #2029.


benjaminjkraft:
Yay, code removal!

For the sake of the record, this seems to have been used by MIT ESP for HSSPark in 2008, and for a few classes for Chicago Splash 2008.

PR #2138, opened 2016-07-12
Fix teacher bio edit redirects

### LINKED ISSUE #2137: "Edit this biography" links don't work
They just redirect back to the bio. Looks probably caused by #2015. We should really fix this before it gets onto any sites since they're permanent redirects.

mgersh:
Three different bugs, introduced in two unrelated changes. That's
kind of impressive. Fixes #2137.


benjaminjkraft:
:bug: Eek! My bad and thanks for fixing.

PR #2148, opened 2016-08-04
add encoding to fix morph error

milescalabresi:
add Unicode encoding to ret_title to allow morphing (and un-morphing) into and back from students with accents or other special characters in their names


benjaminjkraft:
:100: thanks!

PR #2149, opened 2016-08-15
Add some miscellaneous useful scripts

benjaminjkraft:
Various things I've had sitting around.  See script docstrings for more
details.

### DIFF
--- a/esp/useful_scripts/get_chapter_classreg_stats.py
+++ b/esp/useful_scripts/get_chapter_classreg_stats.py
@@ -1,0 +1,35 @@
+"""Gets class registration data for all programs ever, by chapter.
+
+Best run with run_queries.sh.
+"""
+import csv
+import logging
+
+from script_setup import *
+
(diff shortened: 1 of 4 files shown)


mgersh:
It's useful_scripts, so if you want to just merge them and ignore this, I won't complain. That said, if you want the two that don't use script_setup to run outside of run_queries.sh, you'll need to call django.setup() at the top. And I'm curious what you meant by "needed to initialize PMOs"?

benjaminjkraft:
Honestly I'm not sure either is that useful outside of run_queries, but I guess it doesn't hurt to add it.

The PMOs thing is to make sure we've imported all the handler classes, so `__subclasses__` works.  Maybe we already do that somewhere?  I did it just to be safe.

benjaminjkraft:
I added `from script_setup import *` to both, so they can be used either way, and removed the import thing because apparently we do that already somewhere.

PR #2151, opened 2016-08-18
Fix travis config

mgersh:
--use-mirrors was deleted and everything was erroring. This fixes it by not using mirrors.


benjaminjkraft:
Thanks!

PR #2155, opened 2016-08-30
Add profile affiliation field

hwatheod:
This change is requested by Stanford, to add an affiliation field to the profile form.  The options are undergrad, grad, postdoc, other, none ("none" is not needed by Stanford but is added so that this change can be used at all schools).  There is an additional field to specify further, if "other" or "none" is selected.

I also removed several questions from the form that are made redundant by this change: from_here, is_graduate_student, school.  The fields are not removed from the database, because they are used elsewhere.  Instead, these fields are populated automatically based on the affiliation.

"Affiliation" was also added as a display column and filter on the teacher infos admin page, and also on the userview display.


hwatheod:
Failing tests are because I need to update the TeacherInfo tests - working on it

esp/esp/users/forms/user_profile.py
```diff
@@ -310,6 +310,14 @@ def clean(self):
 StudentInfoForm.base_fields['studentrep_expl'].widget.attrs['rows'] = 8
 StudentInfoForm.base_fields['studentrep_expl'].widget.attrs['cols'] = 45
 
+affiliationChoices = (
+    ('Pick', 'Pick one...'),
```
benjaminjkraft: This shouldn't be in choices, since it's not a valid choice; you can instead set `initial=` in the field, I think, and then django will do most of the work for you.

esp/esp/users/models/__init__.py
```diff
@@ -1354,6 +1362,7 @@ class TeacherInfo(models.Model, CustomFormsLinkModel):
 
     user = AjaxForeignKey(ESPUser, blank=True, null=True)
     graduation_year = models.CharField(max_length=4, blank=True, null=True)
+    affiliation = models.CharField(max_length=100, blank=True, null=True)
```
benjaminjkraft: You can actually put the choices here as well.

esp/esp/users/forms/user_profile.py
```diff
@@ -310,6 +310,14 @@ def clean(self):
 StudentInfoForm.base_fields['studentrep_expl'].widget.attrs['rows'] = 8
 StudentInfoForm.base_fields['studentrep_expl'].widget.attrs['cols'] = 45
 
+affiliationChoices = (
```
benjaminjkraft: Minor thing: this should be a constant (i.e. `ALL_CAPS`), and it probably makes more sense to put it in models.

esp/esp/users/models/__init__.py
```diff
@@ -1354,6 +1362,7 @@ class TeacherInfo(models.Model, CustomFormsLinkModel):
...
     from_here = models.NullBooleanField(null=True)
```
benjaminjkraft: Hmm, thoughts on removing the old fields and just making them `@property`s?  (This will require some more migrations.)

benjaminjkraft:
Nice, I think this is a good change for all sites!  Mainly just wondering if it makes sense to do some more consolidation on the models.

esp/esp/users/models/__init__.py
hwatheod: It doesn't work with the DropdownOtherField, because in case the user selects "Other", it stores the field as a colon-separated string "Other:Whatever" where "Whatever" is the text that the user typed into the other box.  So there aren't actually a finite number of choices to list.

esp/esp/users/forms/user_profile.py
hwatheod: Can't put it in models, because of the "Other" issue (see previous comment).  But I did change it to all caps, and also made the same change to several other choice lists.  (By the way, WhatToDoAfterHS doesn't seem to be used.)

esp/esp/users/forms/user_profile.py
hwatheod: I couldn't get initial= to work properly.  (I think DropdownOtherField would need to be modified to pass the keyword args properly.)  Instead, I followed the example of the other uses of DropdownOtherField , where the "Other" option comes first, so then we don't need the "Pick one" anymore.

benjaminjkraft:
Hmm, I think you may have rebased the wrong way here.  Gonna try some git magic and force-push to fix it up; note to self to revert to 1ff0cb7 if I mess up.

benjaminjkraft:
Ok, what happened wasn't exactly what I thought and I don't really understand why it confused github so much, but in any case it's fixed up.  You may want to double-check that things are as you expect.  I'll now take a look at actually reviewing.

(For reference, it looks like the issue was that you had a branch with tip bb0dd62 based on some old version of main, then rebased that to get 15f9944, then merged those two branches, so you had two copies of all the commits.  To fix it I just reset the branch to the latter, since the two seemed to have the same actual code.  In general if you rebase a branch you'll then have to force push it (or open a new PR); if in doubt it's usually safer to just not rebase, and we can always squash the commits at the end.)

esp/esp/users/forms/user_profile.py
```diff
@@ -109,7 +109,7 @@ def clean(self):
             raise forms.ValidationError("Please provide either a day phone or cell phone for your parent/guardian.")
         return self.cleaned_data
 
-HeardAboutESPChoices = (
+HEARD_ABOUT_ESP_CHOICES = (
```
benjaminjkraft: :-) thanks for cleaning these up

esp/esp/users/forms/user_profile.py
```diff
@@ -339,14 +344,22 @@ def clean(self):
...
-        school = cleaned_data.get('school')
-
-        if from_here == "False" and school == "":
-            msg = u'Please enter your affiliation if you are not from %s.' % settings.INSTITUTION_NAME
-            self.add_error('school', msg)
-
+        affiliation_field = self.fields['affiliation']
+        affiliation_with_data = affiliation_field.widget.decompress(cleaned_data.get('affiliation'))
+        affiliation = affiliation_with_data[0]
+        if affiliation == 'Pick' or affiliation == '':
```
benjaminjkraft: what's `Pick`?

esp/esp/users/forms/user_profile.py
```diff
@@ -339,14 +344,22 @@ def clean(self):
...
-        # If teacher is not from MIT, make sure they've filled in the next box
-        from_here = cleaned_data.get('from_here')
-        school = cleaned_data.get('school')
-
-        if from_here == "False" and school == "":
-            msg = u'Please enter your affiliation if you are not from %s.' % settings.INSTITUTION_NAME
-            self.add_error('school', msg)
-
+        affiliation_field = self.fields['affiliation']
+        affiliation_with_data = affiliation_field.widget.decompress(cleaned_data.get('affiliation'))
```
benjaminjkraft: if both of the elements are guaranteed to exist, even if empty string, you can do `affiliation, school = ...` which might be cleaner.

esp/esp/users/models/__init__.py
benjaminjkraft: Ah, got it.

Also, contrary to what the other things in this file do, the correct thing is actually `blank=True` but not `null=True`.  This will also require updating the migration.

hwatheod:
Sorry about the rebase mess-up, and thanks for fixing it.  The merge commit at the end of the rebase was unexpected.  I couldn't reproduce it when I tried rebase again on a dummy project, using git commands from a terminal.  I think the problem might have been caused by using PyCharm's "Resolve conflicts" option in the middle of a rebase (to resolve rebase merge conflicts), which then issued merge commands.

esp/esp/users/forms/user_profile.py
hwatheod: decompress guarantees both elements exist.  Fixed.

esp/esp/users/forms/user_profile.py
hwatheod: Oops, that shouldn't be there.  "Pick" was leftover from when one of the options was "Pick One...", but now I just moved Other to the top of the options.

esp/esp/program/modules/migrations/0014_lotteryfrontendmodule.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
```
benjaminjkraft: I don't think this should be here -- maybe it was missed elsewhere but I would have thought we'd have had it.

esp/esp/users/migrations/0009_auto_20170201_0057.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('users', '0008_teacherinfo_affiliation'),
+    ]
+
+    operations = [
+        migrations.AlterField(
```
benjaminjkraft: likewise not sure why this is here.  Maybe your migrations aren't up to date?  Or maybe this is a different problem?  In any case to be honest I'd just manually fix up the old migration to not have `null=True` and remove the new ones, although maybe that will do weird things to your local DB.

benjaminjkraft:
Looks great, thanks!

Issue #2130, opened 2016-07-03
Teacher Event Descriptions


willgearty:
Enhancement request:
It would be great if the teacher event form (/teacher_events) permitted custom descriptions.
While it's fairly easy to edit event descriptions at /admin/cal/event, this would make event creation much easier, especially for admins that aren't very acquainted with the /admin interface.
Thanks,
Will

willgearty:
Will need to alter the form (specifically adding a description field) and method [here](https://github.com/learning-unlimited/ESP-Website/blob/0bad4b86375a8c4f1db43cde696ccea8761769d6/esp/esp/program/modules/forms/teacherevents.py).
See also handlers [here](https://github.com/learning-unlimited/ESP-Website/blob/0bad4b86375a8c4f1db43cde696ccea8761769d6/esp/esp/program/modules/handlers/teachereventsmodule.py).

PR #2170, opened 2016-09-13
Added ability to specify description when creating teacher events.

willgearty:
Only changed two lines to add a new field in the teacher training/interview creation form and to then assign that new field as the event's description (rather than just copying the short_description).

(See #2130)


esp/esp/program/modules/forms/teacherevents.py
```diff
@@ -25,7 +26,7 @@ def save_timeslot(self, program, slot, type):
 
         slot.program = program
         slot.short_description = slot.start.strftime('%A, %B %d %Y %I:%M %p') + " to " + slot.end.strftime('%I:%M %p')
-        slot.description = slot.short_description
+        slot.description = self.cleaned_data['description']
```
benjaminjkraft: Looks like `description` is required in the model, so it would be nice to make it optional in the form (i.e. set `blank=True` above), and autofill it if not provided here.

willgearty:
Just tested the new code on my server and it seems to work as intended (with a blank description).

Issue #2119, opened 2016-05-04
Volunteer Schedule Printable


willgearty:
Would it be possible to implement a printable for volunteer schedules?
This would be especially important for teachers that are also volunteering (so we could provide a list of the volunteer hours in addition to their classes).

Furthermore, is there a way to cross-reference volunteer hours when scheduling classes (in AJAX) so we prevent teachers' classes from getting scheduled when the teacher has also signed up to volunteer?

willgearty:
See teacherschedules and getSchedule in [programprintables.py](https://github.com/learning-unlimited/ESP-Website/blob/5f5de2df825e42456333758eaaa2d653bcd6d7d1/esp/esp/program/modules/handlers/programprintables.py)
And maybe [volunteersignup.py](https://github.com/learning-unlimited/ESP-Website/blob/0bad4b86375a8c4f1db43cde696ccea8761769d6/esp/esp/program/modules/handlers/volunteersignup.py) and [volunteermanage.py](https://github.com/learning-unlimited/ESP-Website/blob/0bad4b86375a8c4f1db43cde696ccea8761769d6/esp/esp/program/modules/handlers/volunteermanage.py)
@benjaminjkraft suggested something like `VolunteerOffer.objects.filter(user=<user>, request__program=<program>)` to get each volunteer's offers.

PR #2171, opened 2016-09-13
Added volunteer schedule printable

willgearty:
Added capability to print all volunteer schedules (linked to from "All printables" page) and for volunteers to see their own schedules in a print-friendly layout (not currently linked to from anywhere).

(See #2119)

### DIFF
--- a/esp/esp/program/modules/handlers/volunteersignup.py
+++ b/esp/esp/program/modules/handlers/volunteersignup.py
@@ -107,7 +107,7 @@ class VolunteerSignup(ProgramModuleObj, CoreModule):
             base_dict[key] = 'Volunteers for shift "%s"' % req.timeslot.description
         return base_dict
 
-    def volunteerhandout(self, request, tl, one, two, module, extra, prog, template_file='volunteerschedules.html'):
+    def volunteerhandout(self, request, tl, one, two, module, extra, prog, template_file='volunteerschedule.html'):
         #   Use the template defined in ProgramPrintables
         from esp.program.modules.handlers import ProgramPrintables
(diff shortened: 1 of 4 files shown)


esp/esp/program/modules/handlers/volunteersignup.py
```diff
@@ -106,6 +108,35 @@ def volunteerDesc(self):
...
+            #sort the offers by timeslot
+            scheditems.sort(key=lambda item: item['offer'].request.timeslot.start)
+            context['scheditems'] = scheditems
+            return render_to_response(pmo.baseDir()+template_file, request, context)
+        else:
+            raise ESPError('No printables module resolved, so this document cannot be generated.  Consult the webmasters.', log=False)
+
+    @aux_call
+    @needs_account
+    def volunteerschedule(self, request, tl, one, two, module, extra, prog):
```
benjaminjkraft: You can just inline this -- since you only use it once, no need to have a separate wrapper.  Just hardcode the `template_file` in the `render_to_response` call.

willgearty:
Yeah, I wasn't sure how to avoid duplicating the code.
I've actually been thinking about combined teacher/volunteer schedules which could potentially overcome the duplication? Essentially, if a user is either a teacher or a volunteer, it would make a schedule for them, including both their classes (if they have any) and their volunteer shifts (again, if they have any). Are there any other places that this code is duplicated?

benjaminjkraft:
Hmm, that's an interesting idea.  It seems like we'd still want to have the ability to print separate schedules, though.  I think the real problem is that the entire printables system needs a bunch of refactoring.

Issue #2108, opened 2016-04-13
Volunteer Registration Deadlines


willgearty:
We can only have Stanford students volunteering for our program, so we need to check their names and email addresses before the program to make sure we don't have any ineligible volunteers showing up onsite.
Therefore, we'd like to be able to institute a deadline after which no more people can sign up to volunteer.
Is there a way to institute deadlines for volunteer registration/would it be possible to institute this?

willgearty:
See Permission class and methods [here](https://github.com/learning-unlimited/ESP-Website/blob/main/esp/esp/users/models/__init__.py).

PR #2172, opened 2016-09-16
Added and implemented volunteer deadlines

willgearty:
Added a basic volunteer deadline and one for volunteer signup (see #2108).
Implemented the Volunteer/Signup deadline for the volunteer signup page.

Note: Not sure if there is a better way to do this, but in order for teacher accounts to also volunteer, you must make separate deadlines for teachers and volunteers, each with the Volunteer/Signup deadline.


benjaminjkraft:
The code looks fine.  But this makes two behavior changes, one of which is sorta unavoidable and maybe okay, and the other of which you'll need to figure out a way around.

The first one is that this will close volunteer registration on all existing sites, when deployed.  I don't see a good way around that -- perhaps you could write a migration to add deadlines opening volunteer signup for all programs that have the module enabled?  But that may be overkill; it's probably good enough to just make sure this gets in the release notes (which you could, if you feel like it, do -- they're in `docs/admin/releases/09`) and then any chapters that it affects can make the update themselves.  Perhaps other on the web team have thoughts?

The second one is that this means that one now has to log in to use volunteer signup.  It's deliberately `@no_auth`, to allow people to sign up without making an account, which MIT ESP at least has found desirable to make the barrier to entry as small as possible.  So you'll need to figure out how to avoid forcing a login.  This both requires getting around the `if not_logged_in(request):` on `esp/esp/program/modules/base.py:517`, and also figuring out how to check the deadline even if we don't have a logged-in user.  I'm not sure what the best way to do that is -- probably you'll want to manually check if there's a deadline for the "Volunteer" role, and if so apply that to `AnonymousUser`s.

willgearty:
1. I think just including it in the release notes should suffice. The best way, in my opinion, would be to include it in the "Program Creation Form" as is done with the teacher and student deadlines, but not everyone uses the volunteer module.

benjaminjkraft:
> include it in the "Program Creation Form"
For programs going forward, I think we're fine -- we'll need to make sure to document it, but I'm not too worried about that.  The issue is more existing programs, which may already have signup open, and won't go back through the program creation process.

willgearty:
(cont.)
2. Hmm, I've always disliked the way that the volunteer module makes new incomplete accounts for users. I can see why this would be appealing, but afterwards, these accounts don't even have passwords set, making it frustrating for users to use the website afterwards. But if that is preferred to requiring accounts ahead of time, I suppose I can work around that. Both of those things sound a little more complicated than what I've done so far, so it might take me some time.

benjaminjkraft:
Yeah -- I wouldn't be opposed to having the option of requiring an account.  But MIT definitely likes the functionality to not require it, and it seems like a reasonable thing to have.

willgearty:
Is there a way to use decorators with arguments? That way one could specify whether the meets_deadline function should require a user to be logged in or not (which would then be passed on to _checkDeadline_helper).

benjaminjkraft:
Yep!  They already take one argument -- namely the deadline in question -- so you'd just need to add another (optional) one.  I think that's a fine way to do it -- by default require the user to be logged in, but on specific pages, not.

benjaminjkraft:
Looks like this also now needs a merge with master.

willgearty:
Probably won't get to this until after the quarter ends.

willgearty:
Also, I think it's just missing the new volunteer schedules code that I made in a different pull request.

willgearty:
The behavior of `meets_deadline` should be dependent on a tag or a property of the volunteer module (rather than static as it is now), but I'm at a loss of how to implement that. I tried implementing a method that checks for a tag, which is then called as a decorator argument (see comment), but that didn't work. Any thoughts @benjaminjkraft ?

esp/esp/program/modules/handlers/volunteersignup.py
```diff
@@ -51,8 +53,17 @@ def module_properties(cls):
...
+    def require_auth(self):
+        if Tag.getBooleanTag('volunteer_require_auth', self.program, default=False):
+            return True
+        else:
+            return False
+
     @main_call
     @no_auth
+    @meets_deadline("/Signup", False)
+    #Would prefer something like @meets_deadline("/Signup", self.require_auth()),
```
benjaminjkraft: Ah, I see -- yeah, that won't work as such.  Probably you want to, in `check_deadline`, rather than passing in an arg, have it check `moduleObj.require_auth()`, on line `554`.  It means you have to say the method name is fixed, but that's probably a good idea anyway.  (Then you'll want to define it on `ProgramModuleObj` to match current default behavior.

willgearty:
Provided all checks pass, I think this should be good to merge.

esp/esp/program/modules/handlers/volunteersignup.py
```diff
@@ -51,8 +53,15 @@ def module_properties(cls):
             "seq": 0,
             }
 
+    def require_auth(self):
+        if Tag.getBooleanTag('volunteer_require_auth', self.program, default=False):
```
benjaminjkraft: You can just do `return Tag.getBooleanTag(...)` and skip the if.

esp/esp/program/modules/base.py
```diff
@@ -510,16 +513,23 @@ def _checkGrade(moduleObj, request, tl, *args, **kwargs):
 # Just broke out this function to allow combined deadlines (see meets_any_deadline,
 # meets_all_deadlines functions below).  -Michael P, 6/23/2009
 def _checkDeadline_helper(method, extension, moduleObj, request, tl, *args, **kwargs):
-    if tl != 'learn' and tl != 'teach':
+    if tl != 'learn' and tl != 'teach' and tl != 'volunteer':
         return (True, None)
+    if "require_auth" in kwargs:
```
benjaminjkraft: You can just do `..., *args, require_auth=True, **kwargs` in the function definition, and avoid the manual defaulting here.  (But see below, I think you may want to remove it entirely.)

esp/esp/program/modules/base.py
```diff
@@ -548,7 +558,7 @@ def meets_deadline(extension=''):
     def meets_deadline(method):
         def _checkDeadline(moduleObj, request, tl, *args, **kwargs):
             errorpage = 'errors/program/deadline-%s.html' % tl
-            (canView, response) = _checkDeadline_helper(method, extension, moduleObj, request, tl, *args, **kwargs)
+            (canView, response) = _checkDeadline_helper(method, extension, moduleObj, request, tl, require_auth = moduleObj.require_auth(), *args, **kwargs)
```
benjaminjkraft: I'd actually even check `moduleObj.require_auth()` inside `_checkDeadline_helper`, so that you can make `meets_any_deadline` and friends work right for free and save passing it around a bit.

esp/esp/program/modules/base.py
```diff
@@ -510,16 +513,23 @@ def _checkGrade(moduleObj, request, tl, *args, **kwargs):
...
+    if "require_auth" in kwargs:
+        require_auth = kwargs["require_auth"]
+    else:
+        require_auth = True
     response = None
     canView = False
+    perm_name = {'learn':'Student','teach':'Teacher','volunteer':'Volunteer'}[tl]+extension
     if not_logged_in(request):
-        response = HttpResponseRedirect('%s?%s=%s' % (LOGIN_URL, REDIRECT_FIELD_NAME, quote(request.get_full_path())))
+        if not require_auth and Permission.valid_objects().filter(permission_type=perm_name, program=request.program, user__isnull=True).exists():
```
benjaminjkraft: (Per the below, you could just check `moduleObj.require_auth()` inline here instead of passing things around.

willgearty:
Any chance this could make it into SR09? Or at least pushed to our branch?

benjaminjkraft:
Looks good, thanks for all the fixes!  I think this is a nontrivial enough feature that it shouldn't go into SR9 at this point, but we can definitely put it on the Stanford site soon.

PR #2177, opened 2016-09-28
Fix compulsory timeblocks that are last on the schedule

### LINKED ISSUE #2176: Compulsory events don't appear on schedules if they're last in the day
@mgersh found the culprit, the while loop on https://github.com/learning-unlimited/ESP-Website/blob/main/esp/esp/program/modules/handlers/programprintables.py#L974 needs to have a case for if we don't find *any* class to insert before.

Reported by UNC.

benjaminjkraft:
Fixes #2176.  I cherry-picked this to UNC's site.

### DIFF
--- a/esp/esp/program/modules/handlers/programprintables.py
+++ b/esp/esp/program/modules/handlers/programprintables.py
@@ -982,6 +982,8 @@ Volunteer schedule for %s:
                         classes.insert(i, t)
                         break
                     i += 1
+                else:
+                    classes.append(t)
                 min_index = i
 
             # get payment information


esp/esp/program/modules/handlers/programprintables.py
```diff
@@ -982,6 +982,8 @@ def get_student_schedules(request, students, prog, extra='', onsite=False):
                         classes.insert(i, t)
                         break
                     i += 1
+                else:
```
mgersh: Huh, TIL.

PR #2184, opened 2016-10-05
Add buttons to rank classes in listing in phase 2 of reg

### LINKED ISSUE #1041: Two-phase student reg: click on class to rank
When choosing the top choices in step 2, it would be nice to be able to click on a button on the class's listing to make it your top, second, third etc. choice. Currently, you have to select choices through the dropdown menu while separately browsing through the listings to see their descriptions.

Relevant files:
esp/esp/program/modules/handlers/studentregtwophase.py
esp/templates/program/modules/studentregtwophase/rank_classes.html
esp/public/media/scripts/catalog-new.js

betaveros:
When choosing the top choices in step 2, adds buttons to the class's
listing to rank the class as each of the possible priorities.

Fixes #1041.
screen shot 2016-10-04 at 8 52 38 pm

### DIFF
--- a/esp/templates/program/modules/studentregtwophase/rank_classes.html
+++ b/esp/templates/program/modules/studentregtwophase/rank_classes.html
@@ -73,5 +73,9 @@ for every priority slot to ensure a full schedule.
     <i class="star-fill icon-2x icon-star" data-bind="visible: interested"></i>
   </a>
   <label>Click the star to mark this class as interested.</label>
+  <br />
+  {% for pri, pri_display in priorities %}
+  <button data-bind="click: function(data) { $parent.prioritySelection[{{pri}}-1](data.id); $j('#pri-{{pri}}').trigger('change.select2'); }">Set as {{pri_display}}</button>
+  {% endfor %}
 </div>
 {% endblock %}


esp/templates/program/modules/studentregtwophase/rank_classes.html
```diff
@@ -73,5 +73,9 @@
     <i class="star-fill icon-2x icon-star" data-bind="visible: interested"></i>
   </a>
   <label>Click the star to mark this class as interested.</label>
+  <br />
+  {% for pri, pri_display in priorities %}
+  <button data-bind="click: function(data) { $parent.prioritySelection[{{pri}}-1](data.id); $j('#pri-{{pri}}').trigger('change.select2'); }">Set as {{pri_display}}</button>
```
benjaminjkraft: I don't know knockout super well, but is it possible to not have this be a giant inline function?  Even just defining it somewhere else and referencing here seems nicer to me, if that's not too hard.

betaveros: I could definitely put it in `esp/public/media/scripts/catalog-new.js`. I didn't really want to, though, because that script is used in a couple templates and this function is really tightly coupled to this particular templates, having to trigger the custom `change.select2` event on our special `#pri-{{pri}}` element (which doesn't exist in other templates using that JS file) for it to sync back up with knockout.js updating the computed checkbox.

benjaminjkraft: Ok.  It seems a little ugly to be triggering a custom event related to the select dropdown, but maybe not worth fixing now ¯\_(ツ)_/¯ .

PR #2189, opened 2016-10-08
Add a route to dump teachers as JSON for scheduling

betaveros:


esp/esp/program/modules/handlers/jsondatamodule.py
```diff
@@ -894,6 +894,27 @@ def set_donation_amount(self, request, tl, one, two, module, extra, prog):
         data = {'amount_donation': amount_donation, 'amount_due': iac.amount_due()}
         return HttpResponse(json.dumps(data), content_type='application/json')
 
+    @aux_call
+    @json_response()
+    @needs_admin
+    def actually_teachers(self, request, tl, one, two, module, extra, prog):
```
mgersh: "teachers_for_autoscheduler()", maybe? Also can you add a comment that this is needed for the autoscheduler so we don't think it's dead code and delete it?

The reason why it got sad when you named it "teachers()" is because teachers() and students() on program modules are special. See https://github.com/learning-unlimited/ESP-Website/blob/main/esp/esp/program/modules/handlers/teacherclassregmodule.py#L108 for an example, and https://github.com/learning-unlimited/ESP-Website/blob/main/esp/esp/program/models/__init__.py#L279 for the thing that does the magic.

esp/esp/program/modules/handlers/jsondatamodule.py
```diff
@@ -894,6 +894,27 @@ def set_donation_amount(self, request, tl, one, two, module, extra, prog):
...
+            resources_for_user[resource['user_id']].append(resource['event_id'])
+
+        teacher_dicts = [
+            {   'uid': t.id,
+                'text': t.name(),
+                'availability': resources_for_user[t.id]
+            } for t in teachers ]
+
+        return {'teachers': teacher_dicts}
+    # TODO: cache
```
mgersh: I wouldn't worry about it if this is just for the autoscheduler.

mgersh:
:100:

PR #2190, opened 2016-10-11
New "Phase Zero" module

willgearty:
Implements a basic student lottery system. Creates a lottery record for each student which can then be altered by the student if they wish to enter the lottery as a group (e.g. for siblings).
To be added: script for running student lottery and assigning permissions to students accepted to the lottery, accessible through admin interface?
Optional addition: lottery results page

Requested by Stanford Splash for upcoming program (i.e. urgent).


willgearty:
Just realized we also want to implement a way to bypass this module (e.g. outreach groups that have guaranteed access) and access the lottery phase.

esp/esp/program/modules/handlers/studentregphasezero.py
```diff
@@ -60,6 +64,69 @@ def module_properties(cls):
...
+        # The lottery algorithm is run, with randomization and processing in order.
+        # If any one in the group doesn't get in (due to cap size), no one in that group gets in.
+
+        # not sure if we need to deep copy here; kept as shallow
+        groups = sibgroups.keys()
+        order = copy.copy(groups)
+        random.shuffle(order)
+
+        counts = {key:0 for key in grade_caps}
+        winners = Group(name=(str(prog) + ' Winners'))
```
willgearty: You should use "role" that is provided as a parameter to the lottery function.

esp/esp/program/modules/handlers/studentregphasezero.py
```diff
@@ -60,6 +64,69 @@ def module_properties(cls):
...
+        waitlist = Group(name=(str(prog) + ' Waitlist'))
```
willgearty: Hmm, since we don't plan to add any permissions to the waitlist group, I don't think this is necessary.

esp/esp/program/modules/handlers/studentregphasezero.py
```diff
@@ -36,9 +36,13 @@
 from esp.middleware.threadlocalrequest import get_current_request
 from esp.program.modules.base import ProgramModuleObj, main_call, aux_call, meets_deadline, needs_student, meets_grade, meets_cap, no_auth, needs_admin
 from esp.users.models import Record, ESPUser
-from esp.program.models import PhaseZeroRecord
+from esp.program.models import PhaseZeroRecord, getGrade
```
willgearty: This is already imported as part of ESPUser from esp.users.models

esp/esp/program/models/__init__.py
```diff
@@ -1865,6 +1865,22 @@ def __unicode__(self):
         else:
             return self.name
 
+class PhaseZeroRecord(models.Model):
+
+    class Meta:
```
willgearty: The entire class? Or models.Model?

esp/esp/program/models/__init__.py
```diff
@@ -1865,6 +1865,22 @@ def __unicode__(self):
...
+        app_label = 'program'
+
+    def __unicode__(self):
+        return self.lottery_number
+
+    user = AjaxForeignKey(ESPUser)
+    program = models.ForeignKey(Program, blank=True, null=True)
+    phase_zero_time = models.DateTimeField(default=timezone.now)
+    lottery_number = models.CharField(max_length=6)
```
willgearty: Originally we planned to have the lottery codes be hex values (hence the CharField), but then the plan changed. We'll move over to IntegerField.

esp/esp/program/models/__init__.py
mgersh: Yeah, I meant class Meta. This is one of those cases where the older code isn't a good example--it was needed on older versions of Django but isn't anymore, and we didn't bother deleting all the uses in this file when we updated.

willgearty:
@mgersh I think I've made fixes for all of your specific comments. With regards to your two higher-level suggestions:

1. This sounds nice, but is slightly different from what we were originally envisioning. Therefore, it will probably take a bit of time (especially since the lookup feature requires a familiarity with js and jquery, although I could probably just play around with similar code from other modules). We are a bit pressed for time, unfortunately; we are planning on having the student lottery opening next Thursday (Nov. 3), so I'm slightly concerned with making major changes to the module at this point. I'd be more than happy to keep working on this after we open the lottery, though. One extra thought: the ability to lookup any students/userids in the database could be a security concern, no?

2. Personally, I hate this workflow. I thinks it's extremely inelegant that you have to disable a particular module halfway through registration. Enze and I had envisioned applying the "OverridePhaseZero" permission to all "winners", then leave PhaseZero open so any "losers" receive a "sorry, you didn't get in" page. We think this will help mitigate any confusion as to whether a student got in or not (if we just disabled the module, the "losers" would just get a generic error page for the two-phase lottery). Furthermore, having the OverridePhaseZero permission allows us to have some students (particularly underserved groups) bypass the student lottery and enter the class lottery. This is particularly helpful for Stanford because we need extra time to go to the underserved schools and help them register.

Would love any further feedback/input on these two items, and thanks for your comments so far!

mgersh:
Not going to look at the details tonight, but to respond to the main thread:

1. Yup, it sounds like you're not going to have time to get this into its final finished state for this program. It happens. It'll probably be hard to do a migration from the current version of the models to a significantly different one, so you'd lose some data when updating, but sometimes that's the price you pay for early features.

Username or ID lookup that then reveals a name would indeed be a privacy issue, but the existence of a particular username or ID isn't. I wasn't envisioning anything fancy like the coteachers interface, just a way to have to track less data and make things a bit more user-friendly. I've seen a lot of sad confused student emails, and having some basic check that they've done things right would eliminate one source of those, even if we can't tell them very much.

2. Yeah, I agree that the way the two-phase lottery works is kind of an abuse of the program modules system. Unfortunately we don't have a better way of doing it right now. One thing you should know is that students who can't bypass phase zero will still be able to get to the two-phase lottery page if they can find a direct link. Required modules only block access to the core module, not other modules, even if those other modules come later in the sequence. It's possible to block this with correct use of deadlines, but I'm nervous about admins managing to set things up correctly, and if you don't disable the module you'll end up with a "phase zero" link in the modules list when you get to FCFS, which might confuse students. This is easier to change later than (1), though, and I do see your point that leaving the phase zero module there will show a nicer error, so it's up to you.

willgearty:
@mgersh We just finished testing this on our local server (hence the large number of recent small bug fixes), and everything seems to work (including all of the edge cases we can think of). Can you (and possible @benjaminjkraft and @hwatheod ) briefly check our code for any glaring problems, then push to stanford-sr8?

willgearty:
For some reason (that I don't understand), the new students() and studentDesc() methods correctly populate the commpanel and dashboard, but while the commpanel returns the proper number of users, the dashboard shows 0 users (under program vitals).

Since the manage page for Phase Zero currently shows a count for the number of entrants, this isn't terribly important, but I'm confused as to why it would work properly in the commpanel but not on the dashboard.

Thoughts?

willgearty:
@mgersh I think we're good to go!

willgearty:
Thoughts on including this in SR09? @mgersh @benjaminjkraft

willgearty:
Another thought: will we need to do anything special for the models migration? Since we used different models for the previous student lottery, do we need to delete those records or do anything else?

willgearty:
bump @mgersh

willgearty:
bump @mgersh

willgearty:
Hmm...not sure what happened with the migrations there (see checks)

willgearty:
Oh, I see. My migrations conflict with the migrations of #2313.
I'm not really sure how to resolve that?

hwatheod:
To resolve, you can re-number your migrations to be later than the other one.  I assume that by "conflict", you just mean the numbering, and not that there is an actual conflict in the migration actions, since the migrations in #2313 just deletes some dead fields that are no longer used.

willgearty:
@hwatheod 
Well, it's 0009 of program depending on 0010 of users.
But yeah, it's probably not ideal.

To avoid the conflict with #2313, do I just need to change the dates? Or the first numbers?

hwatheod:
You need to change the numbers in front.  The dates don't matter.  Django sorts the migration filenames as strings, so the numbers in front take precedence.

willgearty:
Okay, well I changed the number in front of one of them, and I'm still getting the [conflict error] (https://travis-ci.org/learning-unlimited/ESP-Website/jobs/207422261)

hwatheod:
Can you git pull the latest migrations from the main branch for program and users, and number yours relative to that?  It looks like the conflicts aren't just with #2313.

esp/esp/program/migrations
esp/esp/users/migrations

willgearty:
Okay, did a rebase, everything looks good now (perhaps not the most elegant solution, but it provided a nice new slate).
Thanks @hwatheod for the help!
Now, it would be great if this could get reviewed (@mgersh , @benjaminjkraft , @milescalabresi , @bhorowitz ?) before there are another 2 months of changes that cause the migrations to break.
Otherwise, we might just need to cherry pick this @hwatheod , since student registration opens in two weeks.

hwatheod:
Hmmm, my suggestion may not have been the best. Looks like we have the duplicate commits .  (A chunk of commits appear twice in the list above.)  Last time this happened to me too, when I somehow did a rebase incorrectly (see #2155).  It's getting late now so I'll look into this tomorrow.

hwatheod:
Basically I think this happened: http://stackoverflow.com/questions/9264314/git-commits-are-duplicated-in-the-same-branch-after-doing-a-rebase

(the second answer from the top, by Whymarrh, has a nice detailed explanation)

hwatheod:
Following the steps in that stackoverflow link above, I'm going to hard reset back to `9f7ff07` (the last commit after rebasing but before changes were pulled in from this upstream `phase-zero` branch), then force-push the result to this upstream `phase-zero` branch.

hwatheod:
The duplicate commits have been removed.

hwatheod:
I've locally cherry-picked this to stanford-sr8 and tested it, and it seems to be working.  I will push it to Stanford's site this upcoming week (not right now, since they're in the middle of scheduling).

I found a minor issue where an exception is thrown if you try to run the phase-zero lottery and the program_size_by_grade Tag isn't set.  This isn't urgent since Stanford is the only school currently using it and the Tag is set on the Stanford site.  I'll open a separate issue for this.

esp/esp/users/models/__init__.py
```diff
@@ -613,6 +613,9 @@ def getFirstClassTime(self, program):
                 return sections[0].meeting_times.order_by('start')[0]
     getFirstClassTime.depend_on_row('program.StudentRegistration', lambda reg: {'self': reg.user})
 
+    def canPassPhaseZero(self, program):
```
benjaminjkraft: can we call this `can_skip_phase_zero` instead?  "pass" could mean either override or just has already filled it out.  (Although honestly I'd just inline it -- no need to define a wrapper for this.  But that's a stylistic preference, either way is ok.)

esp/esp/program/modules/handlers/studentregphasezero.py
```diff
@@ -0,0 +1,220 @@
...
+        context = {}
+        context['program'] = prog
+        user = request.user
+        in_lottery = PhaseZeroRecord.objects.filter(user=user, program=prog).exists()
+
+        op = ''
+        join_error = False
+        if 'op' in request.POST:
+            op = request.POST['op']
+        if op == 'join':
```
benjaminjkraft: Since this has totally separate code, can it just be a separate `@aux_call`?  Then you can get rid of `op`, and simplify some of the deeply-nested if blocks here.

esp/esp/program/modules/handlers/studentregphasezero.py
```diff
@@ -0,0 +1,220 @@
...
+            if len(request.POST['student_selected'].strip()) == 0:
```
benjaminjkraft: Should do `request.POST.get('student_selected', '').strip() ...`, so that we'll give a nice error if the field isn't included, too.

esp/esp/program/modules/handlers/studentregphasezero.py
```diff
@@ -0,0 +1,220 @@
...
+                elif in_lottery and old_group==group:
+                    join_error = 'Error - You are already in this lottery group.'
+                elif num_users < 4:
+                    group.user.add(user)
+                    group.save()
+                    self.send_confirmation_email(user)
+                    if in_lottery:
+                        old_group.user.remove(user)
+                        old_group.save()
+                        if len(old_group.user.all()) == 0:
```
benjaminjkraft: `old_group.user.exists()` is faster -- see for example https://docs.djangoproject.com/en/1.10/topics/db/optimization/#use-queryset-exists

esp/esp/program/modules/handlers/studentregphasezero.py
```diff
@@ -0,0 +1,220 @@
...
+
+            context['join_error'] = join_error
+            if join_error and not in_lottery:
+                form = SubmitForm(program=prog)
+                context['form'] = form
+                return render_to_response('program/modules/studentregphasezero/submit.html', request, context)
+
+            else:
+                context['lottery_group'] = PhaseZeroRecord.objects.filter(user=user, program=prog)[0]
+                context['lottery_size'] = len(context['lottery_group'].user.all())
```
benjaminjkraft: similarly, use `....user.count()` here.

esp/esp/program/modules/handlers/studentregphasezero.py
```diff
@@ -0,0 +1,220 @@
...
+        join_error = False
+        if 'op' in request.POST:
+            op = request.POST['op']
+        if op == 'join':
+            if len(request.POST['student_selected'].strip()) == 0:
+                join_error = 'Error - You must select a student\'s username.'
+
+            else:
+                join_user = request.POST['student_selected']
+                group = PhaseZeroRecord.objects.filter(user=join_user, program=prog)[0]
```
benjaminjkraft: Probably good to do `.get(...)` instead of `.filter(...)[0]` -- that'll raise a nicer error if there are none, and not let it pass silently if there are two.

...come to think of it, you should definitely handle the case where you try to join a nonexistent group.

esp/esp/program/modules/handlers/studentregphasezero.py
```diff
@@ -0,0 +1,220 @@
...
+                if in_lottery:
+                    old_group = PhaseZeroRecord.objects.filter(user=user, program=prog)[0]
```
benjaminjkraft: (likewise with the `.get()` here)

esp/esp/program/modules/handlers/studentregphasezero.py
```diff
@@ -0,0 +1,220 @@
...
+                users = group.user.all()
+                num_users = len(users)
```
benjaminjkraft: Are you actually using `users` anywhere?  If not, just do `group.user.count()`, as below.

esp/esp/program/modules/handlers/studentregphasezero.py
```diff
@@ -0,0 +1,220 @@
...
+                    join_error = 'Error - This group already contains the maximum number of students.'
+
+            context['join_error'] = join_error
+            if join_error and not in_lottery:
+                form = SubmitForm(program=prog)
+                context['form'] = form
+                return render_to_response('program/modules/studentregphasezero/submit.html', request, context)
+
+            else:
+                context['lottery_group'] = PhaseZeroRecord.objects.filter(user=user, program=prog)[0]
```
benjaminjkraft: and again `.get()` here.

esp/esp/program/modules/handlers/studentregphasezero.py
```diff
@@ -0,0 +1,220 @@
...
+                context['lottery_size'] = len(context['lottery_group'].user.all())
+                return render_to_response('program/modules/studentregphasezero/confirmation.html', request, context)
+
+        elif Permission.user_has_perm(user, 'Student/Classes/PhaseZero', program=prog):
+            if in_lottery:
+                context['lottery_group'] = PhaseZeroRecord.objects.filter(user=user, program=prog)[0]
```
benjaminjkraft: (and again `.get()` here and `.count()` on the next line)

esp/esp/program/modules/handlers/studentregphasezero.py
```diff
@@ -0,0 +1,220 @@
...
+                context['lottery_size'] = len(context['lottery_group'].user.all())
+                return render_to_response('program/modules/studentregphasezero/confirmation.html', request, context)
+            else:
+                if request.method == 'POST':
+                    form = SubmitForm(request.POST, program=prog)
+                    if form.is_valid():
+                        form.save(user, prog)
+                    context['lottery_group'] = PhaseZeroRecord.objects.filter(user=user, program=prog)[0]
```
benjaminjkraft: (and again `.get()` and `.count()`)

esp/esp/program/modules/handlers/studentregphasezero.py
```diff
@@ -0,0 +1,220 @@
...
+
+    @aux_call
+    @needs_student
+    def studentlookup(self, request, tl, one, two, module, extra, prog, newclass = None):
+
+        # Search for students with names that start with search string
+        if not 'username' in request.GET or 'username' in request.POST:
+            return self.goToCore(tl)
+
+        return StudentRegPhaseZero.studentlookup_logic(request, tl, one, two, module, extra, prog, newclass)
```
benjaminjkraft: You can just inline this -- no need for a separate method.

esp/esp/program/modules/handlers/studentregphasezero.py
```diff
@@ -0,0 +1,220 @@
...
+
+    @staticmethod
+    def studentlookup_logic(request, tl, one, two, module, extra, prog, newclass = None):
+        limit = 10
+        from esp.web.views.json_utils import JsonResponse
```
benjaminjkraft: import should go at the top of the file unless there's a reason not to.  (This is a bit faster since it means the import won't happen in a request, and is generally the convention to make it easy to find all the imports.)

esp/esp/program/modules/handlers/studentregphasezero.py
```diff
@@ -0,0 +1,220 @@
...
+
+        q_phasezero = Q(phasezerorecord__program=prog)
```
benjaminjkraft: no need to create a Q object here, you can just do it inline in the filter.

esp/esp/program/modules/handlers/studentregphasezeromanage.py
```diff
@@ -0,0 +1,161 @@
...
+            'required': False,
+            'admin_title': 'Manage Student Registration Phase Zero',
+            'link_title': 'Manage Phase Zero',
+        }
+
+    def lottery(self, prog, role):
+        # Run lottery algorithm.
+        # Get grade caps
+        grade_caps_str = prog._grade_caps()
+        grade_caps = {int(key[0]):grade_caps_str[key] for key in grade_caps_str}
```
benjaminjkraft: why `key[0]`?

esp/esp/program/modules/handlers/studentregphasezeromanage.py
```diff
@@ -0,0 +1,161 @@
...
+            "module_type": "manage",
+            'required': False,
+            'admin_title': 'Manage Student Registration Phase Zero',
+            'link_title': 'Manage Phase Zero',
+        }
+
+    def lottery(self, prog, role):
+        # Run lottery algorithm.
+        # Get grade caps
+        grade_caps_str = prog._grade_caps()
```
benjaminjkraft: The underscore-prefixed method means it's private, and you aren't supposed to call it from outside of `Program` itself.  (This makes it easy to tell whether a function is intended for external use.)  In this case, you can probably just make it not private -- just change the name.

esp/esp/program/modules/handlers/studentregphasezeromanage.py
```diff
@@ -0,0 +1,161 @@
...
+        grade_caps = {int(key[0]):grade_caps_str[key] for key in grade_caps_str}
+
+        #Get lottery records and randomize them
+        records = PhaseZeroRecord.objects.filter(program=prog).order_by('?')
+
+        ###############################################################################
+        # The lottery algorithm is run, with randomization and processing in order.
+        # If any one in the group doesn't get in (due to cap size), no one in that group gets in.
+        counts = {key:0 for key in grade_caps}
+        winners = Group.objects.get_or_create(name=role)[0]
```
benjaminjkraft: clearer to do `winners, _ = ...`

esp/esp/program/modules/handlers/studentregphasezeromanage.py
```diff
@@ -0,0 +1,161 @@
...
+
+        for i in records:
+            sibs = i.user.all()
+            newcounts = copy.copy(counts)
+            for j in sibs:
+                newcounts[j.getGrade(prog)] += 1
+
+            cpass = True
```
benjaminjkraft: Probably easier and clearer to do `cpass = not any(newcounts[c] > grade_caps[c] for c in counts)`

esp/esp/program/modules/handlers/studentregphasezeromanage.py
```diff
@@ -0,0 +1,161 @@
...
+            for c in counts.keys():
+                if newcounts[c] > grade_caps[c]:
+                    cpass = False
+
+            if cpass:
+                for user in sibs:
```
benjaminjkraft: You can do `user.groups.add(*sibs)` and it'll add them all at once (and it'll be faster).

esp/esp/program/modules/handlers/studentregphasezeromanage.py
```diff
@@ -0,0 +1,161 @@
...
+
+        ###############################################################################
+        # The lottery algorithm is run, with randomization and processing in order.
+        # If any one in the group doesn't get in (due to cap size), no one in that group gets in.
+        counts = {key:0 for key in grade_caps}
+        winners = Group.objects.get_or_create(name=role)[0]
+
+        for i in records:
+            sibs = i.user.all()
+            newcounts = copy.copy(counts)
```
benjaminjkraft: No need to copy here, or below, since you're not using `counts` anywhere else.  You can just update it in-place.

esp/esp/program/modules/handlers/studentregphasezeromanage.py
```diff
@@ -0,0 +1,161 @@
...
+
+        #Run lottery if requested
+        if request.POST:
+            if Tag.getBooleanTag('student_lottery_run', prog, default=False):
+                context['error'] = "You've already run the student lottery!"
+            else:
+                role = request.POST['rolename']
+                context['role'] = role
+                if "confirm" in request.POST:
+                    success = False
```
benjaminjkraft: No need to initialize like this.

esp/esp/program/modules/handlers/studentregphasezeromanage.py
```diff
@@ -0,0 +1,161 @@
...
+                    success = self.lottery(prog, role)
```
benjaminjkraft: Can this ever not return `True`?  If you want to catch errors, better to do `try`/`except`.

willgearty:
Thanks, @benjaminjkraft. Glad to hear there aren't any big problems!
We'll take a look at all of this as soon as possible.

And hopefully we'll have some extra things to think about after opening and running student registration these next few weeks.

willgearty:
Fixed #2350.

esp/esp/program/modules/handlers/studentregphasezeromanage.py
willgearty: @Enze-Chen ?

esp/esp/program/modules/handlers/studentregphasezeromanage.py
willgearty: Wait, sibs are the users, and we need to add all of them to the "winners" group.

esp/esp/program/modules/handlers/studentregphasezero.py
willgearty: I'll look into this later.

esp/esp/program/modules/handlers/studentregphasezeromanage.py
benjaminjkraft: Oh, sorry, I see.  Then I think you can do `winner.users.add(*sibs)`, conversely.

esp/esp/program/modules/handlers/studentregphasezero.py
```diff
@@ -95,16 +93,15 @@ def studentregphasezero(self, request, tl, one, two, module, extra, prog):
...
             op = request.POST['op']
         if op == 'join':
-            if len(request.POST['student_selected'].strip()) == 0:
+            if len(request.POST.get('student_selected', '').strip()) == 0:
                 join_error = 'Error - You must select a student\'s username.'
 
             else:
                 join_user = request.POST['student_selected']
-                group = PhaseZeroRecord.objects.filter(user=join_user, program=prog)[0]
+                group = PhaseZeroRecord.objects.get(user=join_user, program=prog)
```
benjaminjkraft: I think you still need to handle the case where there is no such group -- e.g. by catching the `DoesNotExist` exception.

esp/esp/program/modules/handlers/studentregphasezero.py
```diff
@@ -170,15 +167,9 @@ def studentlookup(self, request, tl, one, two, module, extra, prog, newclass = N
...
-        return StudentRegPhaseZero.studentlookup_logic(request, tl, one, two, module, extra, prog, newclass)
-
-    @staticmethod
-    def studentlookup_logic(request, tl, one, two, module, extra, prog, newclass = None):
         limit = 10
-        from esp.web.views.json_utils import JsonResponse
 
-        q_phasezero = Q(phasezerorecord__program=prog)
-        queryset = ESPUser.objects.filter(q_phasezero).distinct()
+        queryset = ESPUser.objects.filter(Q(phasezerorecord__program=prog)).distinct()
```
benjaminjkraft: Sorry, should have been clearer -- you can just remove the `Q()` and do `filter(phasezerorecord__program=prog)`, it does the same thing.

esp/esp/program/modules/handlers/studentregphasezeromanage.py
benjaminjkraft: Oh, also, you can do  `{int(key): value for key, value in grade_caps_str.iteritems()}`, so you don't have to do the dict lookup again to grab the value.

esp/esp/program/modules/handlers/studentregphasezeromanage.py
```diff
@@ -71,23 +71,20 @@ def lottery(self, prog, role):
         # The lottery algorithm is run, with randomization and processing in order.
         # If any one in the group doesn't get in (due to cap size), no one in that group gets in.
         counts = {key:0 for key in grade_caps}
-        winners = Group.objects.get_or_create(name=role)[0]
+        winners, _ = Group.objects.get_or_create(name=role)
 
         for i in records:
             sibs = i.user.all()
-            newcounts = copy.copy(counts)
+            newcounts = counts
```
benjaminjkraft: Is there a reason to rename it here anymore?

esp/esp/program/modules/handlers/studentregphasezeromanage.py
```diff
@@ -71,23 +71,20 @@ def lottery(self, prog, role):
...
-            for c in counts.keys():
-                if newcounts[c] > grade_caps[c]:
-                    cpass = False
+            cpass = not any(newcounts[c] > grade_caps[c] for c in counts)
 
             if cpass:
                 for user in sibs:
                     user.groups.add(winners)
-                counts = copy.copy(newcounts)
+                counts = newcounts
```
benjaminjkraft: (again here, why rename?)

esp/esp/program/modules/handlers/studentregphasezeromanage.py
willgearty: Umm, @Enze-Chen ? And below as well.

willgearty:
Not really sure why this is failing...

esp/esp/program/modules/handlers/studentregphasezeromanage.py
willgearty: Oh, now I remember. We only want to update the counts if we accept the group. If we can't accept the whole group, we move on to the next group with the old counts.

esp/esp/program/modules/handlers/studentregphasezero.py
willgearty: Wait, but in the `studentlookup` method, we only return students that are already in groups, so they shouldn't be able to provide a user that isn't in the lottery, right?

esp/esp/program/modules/handlers/studentregphasezeromanage.py
benjaminjkraft: Ah! So in that case you do want to do the copy at the beginning (but no need at the end). Sorry for leading you astray, I didn't read carefully enough.

esp/esp/program/modules/handlers/studentregphasezero.py
benjaminjkraft: I guess.  I'd still say include it, because bugs and unexpected situations (e.g. somebody manually edited phase zero groups for some reason or other) happen, and a clearer error message is nicer.

benjaminjkraft:
Ah, yeah, the test failure is #2388.  As long as that's the only test failing you can ignore.  In any case I think there are 1 or 2 more tiny things but then this is good to go!

esp/esp/program/modules/handlers/studentregphasezeromanage.py
willgearty: Ah, figured it out. Apparently this is what we get back from `grade_caps():`
`{(7,): 1, (8,): 220, (9,): 325, (10,): 325, (11,): 200, (12,): 150}`
You'll notice the keys are actually tuples (perhaps there's a problem in the grade_caps function), so we need to call key[0] to get the grade value.

willgearty:
Okay, handled the case where a selected user is not in a lottery group and moved the joingroup code to a separate @aux_call, making the appropriate changes to the templates.

esp/esp/program/modules/handlers/studentregphasezeromanage.py
willgearty: Ah, it's winners.user_set.add(*sibs)

esp/esp/program/modules/handlers/studentregphasezeromanage.py
benjaminjkraft: Ah, ok!  Can you comment that?

esp/esp/program/modules/handlers/studentregphasezero.py
```diff
@@ -161,6 +122,57 @@ def studentregphasezero(self, request, tl, one, two, module, extra, prog):
...
+        
+        join_error = False
+        if len(request.POST.get('student_selected', '').strip()) == 0:
+            join_error = 'Error - You must select a student\'s username.'
+
+        else:
+            join_user = request.POST['student_selected']
+            try:
+                group = PhaseZeroRecord.objects.get(user=join_user, program=prog)
+            except DoesNotExist:
```
benjaminjkraft: I think you'll need to import this, or refer to it as `PhaseZeroRecord.DoesNotExist` (I think the latter works?)

willgearty: I'll go with that, though I'm not too sure how to test it...

willgearty: Looks like you are correct! https://docs.djangoproject.com/en/1.11/ref/models/querysets/#django.db.models.query.QuerySet.get
You can also just use a generic `ObjectDoesNotExist` exception, but that needs to be imported.

willgearty:
Alright, that should be everything. There's probably some extra features that could be incorporated and the templates could probably be a little prettier, but for now this looks good to me.

PR #2193, opened 2016-10-19
Add sort key for student reg ranking priorities

### LINKED ISSUE #2194: Priorities (`RegistrationType`s, "First Choice" etc.) in lottery schedule aren't sorted correctly
The code tries to sort them but it appears that it sorts by a Django model that doesn't know how to sort itself and you end up with random stuff. I think we need a `key` function to sort by the `RegistrationType` internal name?
screen shot 2016-10-19 at 12 09 07 am
screen shot 2016-10-19 at 12 07 53 am

ecnerwala:
Currently , student registration is sorted by `RegistrationType`, which doesn't define a sort order. As a result, the ranking is essentially random when listed on the student reg page . Explicitly sort it by `relationship.name`, which is of the form `"Priority/1"`, `"Priority/2"`, `"Priority/3"`.

Discussed with @betaveros, tested locally.

Fixes #2194

### DIFF
--- a/esp/esp/program/modules/handlers/studentregtwophase.py
+++ b/esp/esp/program/modules/handlers/studentregtwophase.py
@@ -142,7 +142,8 @@ class StudentRegTwoPhase(ProgramModuleObj):
 
             if timeslot.id in timeslot_dict:
                 priority_dict = timeslot_dict[timeslot.id]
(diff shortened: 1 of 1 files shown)


mgersh:
:+1:

PR #2195, opened 2016-10-19
Update fresh classroom script to be more general

betaveros:
Squashed commits actually by @jerrywu64


mgersh:
It's a useful_script so I didn't look at it too closely. Yay scheduling.

PR #2196, opened 2016-10-19
Make filtered_sections logic a bit more sane

betaveros:


benjaminjkraft:
I'm confused -- this seems to change the arguments to the function, but not any callers?

betaveros:
Oh yeah, I was generalizing it in hopes I could add a switch to the AJAX scheduler to list even classes that had been scheduled, but it became a big change and scheduling ended and I got lazy. I can remove it if you want.

benjaminjkraft:
Ah, gotcha.  Seems fine to me as-is.

PR #2202, opened 2016-10-29
Fix handling of classroom names with commas

### LINKED ISSUE #2201: Scheduler cannot handle commas in classroom names
Whenever one places a class into a room with a comma in the name, it momentarily shows up and then disappears with an error "Error: An error occurred saving the schedule change."

mgersh:
Fixes #2201. Needs backport to stable-release-9. (This is based on an old branch so it can be merged rather than cherry-picked.)


benjaminjkraft:
Looks like the test failure was spurious and they passed on the PR version, so I'm going to ignore it and merge.  May be worth cleaning up this code further so we don't end up with more of these bugs, but that's of course harder.

mgersh:
It wasn't spurious, it was because travis is broken in the commit I based this on. Looks like you didn't do the backport, so I'll do that now.

PR #2210, opened 2016-11-15
Hyperlink teacher list in manage class page

betaveros:
This makes it easier to do a variety of tasks involving finding teachers
based on what classes they're teaching

### DIFF
--- a/esp/esp/users/models/__init__.py
+++ b/esp/esp/users/models/__init__.py
@@ -214,6 +214,11 @@ class BaseESPUser(object):
     def name(self):
         return u'%s %s' % (self.first_name, self.last_name)
 
+    def nonblank_name(self):
+        name = self.name()
+        if name.strip() == '': name = self.username
+        return name
+
     def get_email_sendto_address_pair(self):
         """
         Returns the pair of data needed to send an email to the user.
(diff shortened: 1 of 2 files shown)


esp/esp/users/models/__init__.py
```diff
@@ -214,6 +214,11 @@ def ajax_str(self):
     def name(self):
         return u'%s %s' % (self.first_name, self.last_name)
 
+    def nonblank_name(self):
+        name = self.name()
+        if name.strip() == '': name = self.username
```
betaveros: I don't know how necessary this defensive logic really is; I copied it from `getTeacherNames`

esp/templates/program/modules/adminclass/manageclass_class_info.html
```diff
@@ -1,6 +1,7 @@
 <tr>
     <th class="smaller">Teachers:</th>
-    <td>{{ class.getTeacherNames|join:", " }}<br />
+	<td>{% for teacher in class.get_teachers %}<a href="/manage/userview?username={{ teacher.username|urlencode }}">{{ teacher.nonblank_name }}</a>{% if not forloop.last %}, {% endif %}{% endfor %}
```
betaveros: The correct way to do this definitely involves `{% url something %}`, but I really want these links for sub wrangling and the rest of the codebase uses this URL creation method *a lot* so I'm going to keep this until after Splash at least

benjaminjkraft: Yeah, we don't really use `{% url %}` unfortunately, so this is fine.

esp/esp/users/models/__init__.py
benjaminjkraft: I'm not sure it's necessary, but it doesn't really hurt.

Issue #2211, opened 2016-11-19
Cache the viewpay_cybersource page


luac:
It takes a long time to load, people mainly want to look at it on the day of the program, and most programs will have closed credit card payment by then so it shouldn't need to update for the most part.

benjaminjkraft:
Fixed by #2213

PR #2213, opened 2016-11-21
Cache viewpay_cybersource page

luac:
Note: I opted to cache each line individually rather than the entire
page because apparently the entire page is huge and hits up against
memcached's key data limits.

Ideally the IAC methods themselves would be cached, and with less
conservative cache dependencies, but this is the quick and dirty
solution.

(Also, glancing at the implementations of the methods, maybe this
whole view can be rewritten to do fewer, batched queries and then it
won't be so slow even without caching. But that's also for another
day.)


luac:
Addresses #2211.

benjaminjkraft:
Looks good to me, thanks!

PR #2214, opened 2016-11-21
Add links to manage/edit classes from userview

betaveros:
Adds (manage | edit) links to each class in the `userview_class_entry`
template when enabled, and enable them in userview.

screen shot 2016-11-21 at 4 30 52 pm

screen shot 2016-11-21 at 4 31 00 pm

### DIFF
--- a/esp/esp/program/models/class_.py
+++ b/esp/esp/program/models/class_.py
@@ -360,6 +360,9 @@ class ClassSection(models.Model):
     def get_absolute_url(self):
         return self.parent_class.get_absolute_url()
 
+    def get_edit_absolute_url(self):
+        return self.parent_class.get_edit_absolute_url()
+
     @cache_function
     def get_meeting_times(self):
         return self.meeting_times.all()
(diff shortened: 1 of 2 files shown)


esp/templates/users/userview.html
```diff
@@ -160,7 +160,7 @@
 <br />
 {% for class in classes %}
 {% ifchanged class.parent_program %}{% if not forloop.first %}</ul>{% endif %}{{ class.parent_program.niceName }}<br /><ul>{% endifchanged %}
-{% with show_class_details="True" %}
+{% with show_class_manage_links=True show_class_details=True %}
```
benjaminjkraft: Hmm, why bother with the with+conditional here?  It doesn't look like anything else uses the `userview_class_entry.html` template so we could just always show it to keep things simple.

betaveros:
Good point. I assumed because of the flags that we were using the template somewhere public-facing as well as for management, and didn't check.

benjaminjkraft:
Awesome, thanks for the quick fix!  Will merge once tests pass.

Issue #2217, opened 2016-11-22
Teacher checkin page is slow


betaveros:
Loading the class details asynchronously on demand would probably make a big difference. The resource queries hit the database a lot.

Even if we did that, though, it looks like there's a separate SQL query for every room and phone number because they go through different tables. I dug through the code a little bit and tried some things and couldn't find any way to eliminate this. Perhaps somebody who knows the backend or Django queries can figure something out?

betaveros:
I figured some things out, I guess. Not the room numbers, but I think I'd be satisfied with the speed of the page after these optimizations

benjaminjkraft:
Fixed in #2218.

betaveros:
I want #2219 to be merged before closing this, I think it's just as much of a slowdown (it makes a zillion queries for the resources of every class)...

PR #2218, opened 2016-11-22
Consolidate queries for missingteachers' phones

betaveros:
Speeds up the teacher checkin page as per #2217.


esp/esp/program/modules/handlers/teachercheckinmodule.py
```diff
@@ -149,13 +150,31 @@ def ajaxteachercheckin(self, request, tl, one, two, module, extra, prog):
...
+        #       return contact_info.phone_cell or contact_info.phone_day or default
+        #   return default
+
+        # Only Postgres supports the following fancy database operation! See
+        # http://stackoverflow.com/a/20129229/3243497 .
+
+        profiles = (RegistrationProfile.objects
+                .filter(user__in=users)
+                .order_by('user__id', '-last_ts')
+                .distinct('user__id')
```
benjaminjkraft: Neat trick!  I wonder if we will want to use this elsewhere where we do crazy things to fetch lots of reg profile info.

PR #2219, opened 2016-11-22
Load class details on demand in teacher checkin

betaveros:
On the teacher checkin page, load class details asynchronously only on
demand to speed up the page, as per #2217.


betaveros:
(rebased to resolve merge conflicts, and added more comments)

benjaminjkraft:
Nice!  This all looks right to me.

PR #2222, opened 2016-11-22
Rewrite dashboard in more robust jQuery, plus UI tweaks

betaveros:
Construct the HTML through jQuery instead of splicing things into HTML. Link teachers to their userviews. Drop a few unnecessary `<span>`s.

### DIFF
--- a/esp/public/media/scripts/program/modules/adminclass.js
+++ b/esp/public/media/scripts/program/modules/adminclass.js
@@ -187,105 +187,119 @@ function fillClasses(data)
     $j("#classes_anchor").html('');
 
     // Now loop through and render each class row
-    for (var i in classes)
-    {
-	var cls = classes[i];
-	$j("#classes_anchor").append(createClassRow(cls));
+    for (var i in classes) {
+        var cls = classes[i];
+        $j("#classes_anchor").append(createClassRow(cls));
     }
-    
+
     //  Save the data for later if we need it
     classes_global = classes;
     sections_global = sections;
 }
 
-function createClassRow(clsObj)
-{
-    var template = " \
-    <tr id='clsid-{{ cls.id }}-row'> \
-    <td class='clsleft classname'> \
-      <span title='{{ cls.title }}'> \
-        <strong>{{ cls.emailcode }}.</strong> \
-        <span class='{{ title_css_class }}'>{{ cls.title }} \
-        <strong>[{{ cls_status }}]</strong></span> \
+function createClassTitleTd(clsObj, shortTitle) {
+    var status_details = getStatusDetails(clsObj.status);
+    var title_css_class = status_details.classes.join(" ");
+    var cls_status = status_details['text'];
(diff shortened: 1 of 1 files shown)


esp/public/media/scripts/program/modules/adminclass.js
```diff
@@ -187,105 +187,119 @@ function fillClasses(data)
...
+}
+function createStatusButtonTd(clsObj, shortTitle) {
+    return $j("<td class='clsmiddle rapid_approval_button'> \
+      <a href='#' title='Set the status of " + shortTitle + "' class='abutton' style='white-space: nowrap;' onclick='show_approve_class_popup(" + clsObj.id + "); return false;'>Status</a> \
+    </td>");
+}
+
+function createClassRow(clsObj)
+{
+    var shortTitle = clsObj.title;
```
benjaminjkraft: It seems cleaner to me to just put this on `clsObj`, if we can -- so we don't have to pass down a separate parameter in some cases.  But I could go either way.

benjaminjkraft: Or, perhaps better, break it out into a function of `clsObj` that each of the `...Td` functions can use, like `getStatusDetails`.

esp/public/media/scripts/program/modules/adminclass.js
```diff
@@ -187,105 +187,119 @@ function fillClasses(data)
...
-    <tr id='clsid-{{ cls.id }}-row'> \
-    <td class='clsleft classname'> \
-      <span title='{{ cls.title }}'> \
-        <strong>{{ cls.emailcode }}.</strong> \
-        <span class='{{ title_css_class }}'>{{ cls.title }} \
-        <strong>[{{ cls_status }}]</strong></span> \
+function createClassTitleTd(clsObj, shortTitle) {
+    var status_details = getStatusDetails(clsObj.status);
+    var title_css_class = status_details.classes.join(" ");
+    var cls_status = status_details['text'];
```
benjaminjkraft: nit: I'd just inline this, as `status_details.text`.

esp/public/media/scripts/program/modules/adminclass.js
```diff
@@ -187,105 +187,119 @@ function fillClasses(data)
...
+{
+    var shortTitle = clsObj.title;
+    if (shortTitle.length > 40) {
+        shortTitle = shortTitle.substring(0, 40);
+        shortTitle = shortTitle.concat("...");
+    }
 
-    // Turn the template into a jQuery node
-    $node = $j(template);
+    var tr = $j(document.createElement('tr'));
```
benjaminjkraft: Previously I think this was using the convention that variables starting with `$` are jQuery nodes.  I'm not super attached to that but if you feel like it it might be nice.

betaveros:
Bump?

benjaminjkraft:
:100: thanks for the fixes!

PR #2223, opened 2016-11-23
Added user group filter option to comm panel

willgearty:
Only required a few tweaks and additions. This implementation provides a drop-down menu to select a single user group to then filter the users by. Everything seems to work as it should, and it seems to play nicely with other filters.

The only little bug is that the new filter option isn't hidden along with the other filter options. However, once you collapse the option, it then behaves as it should. Not sure why...

### DIFF
--- a/esp/esp/users/controllers/usersearch.py
+++ b/esp/esp/users/controllers/usersearch.py
@@ -94,7 +94,7 @@ class UserSearchController(object):
                     group = criteria['group']
                     Q_include &= Q(registrationprofile__user__groups__name=group)
                     self.updated = True
-            
+
             for field in ['username','last_name','first_name', 'email']:
                 if criteria.get(field, '').strip():
                     #   Check that it's a valid regular expression
--- a/esp/esp/users/controllers/usersearch.py
+++ b/esp/esp/users/controllers/usersearch.py
@@ -89,11 +89,10 @@ class UserSearchController(object):
         else:
 
             ##  Select users based on all other criteria that was entered
-            if 'group' in criteria:
-                if criteria['group'] != "":
-                    group = criteria['group']
-                    Q_include &= Q(registrationprofile__user__groups__name=group)
-                    self.updated = True
+            if 'group' in criteria and criteria['group'] != "":
+                group = criteria['group']
+                Q_include &= Q(groups=group)
+                self.updated = True
 
             for field in ['username','last_name','first_name', 'email']:
                 if criteria.get(field, '').strip():
--- a/esp/esp/users/controllers/usersearch.py
+++ b/esp/esp/users/controllers/usersearch.py
@@ -40,6 +40,7 @@ from esp.dbmail.models import MessageRequest
 
 from django.db.models import Count
 from django.db.models.query import Q
+from django.contrib.auth.models import Group
 
 import collections
 import re
@@ -88,6 +89,12 @@ class UserSearchController(object):
         else:
 
             ##  Select users based on all other criteria that was entered
+            if 'group' in criteria:
+                if criteria['group'] != "":
+                    group = criteria['group']
+                    Q_include &= Q(registrationprofile__user__groups__name=group)
+                    self.updated = True
+            
             for field in ['username','last_name','first_name', 'email']:
                 if criteria.get(field, '').strip():
                     #   Check that it's a valid regular expression
@@ -357,6 +364,7 @@ class UserSearchController(object):
         if target_path is None:
             target_path = '/manage/%s/commpanel' % program.getUrlBase()
         context['action_path'] = target_path
+        context['groups'] = Group.objects.all()
 
         return context
 
--- a/esp/esp/users/controllers/usersearch.py
+++ b/esp/esp/users/controllers/usersearch.py
@@ -91,7 +91,8 @@ class UserSearchController(object):
             ##  Select users based on all other criteria that was entered
             if 'group' in criteria and criteria['group'] != "":
                 group = criteria['group']
-                Q_include &= Q(groups=group)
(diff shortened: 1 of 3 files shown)


esp/esp/users/controllers/usersearch.py
```diff
@@ -88,6 +89,12 @@ def query_from_criteria(self, user_type, criteria):
         else:
 
             ##  Select users based on all other criteria that was entered
+            if 'group' in criteria:
+                if criteria['group'] != "":
+                    group = criteria['group']
+                    Q_include &= Q(registrationprofile__user__groups__name=group)
```
benjaminjkraft: It should work to just do `Q(groups__name=group)`, `registrationprofile__user` should be unnecessary (and will make things slower in some cases).  (But see also below.)

esp/templates/users/usersearch/list_selector.html
```diff
@@ -73,6 +73,14 @@
             <div class="step_instructions step_highlight">Currently selected: <span id="filter_current_list"></span></div>
             <div id="filter_accordion">
             
+                <h4><a href="#">User Group</a></h4>
+                <div><select id="group" name="group">
+                    {% for group in groups %}
+                    <option value="{{ group.name }}" id="{{ group.name }}">{{ group.name }}</option>
```
benjaminjkraft: If you put `group.id` as the `value`, then you can just do `groups` instead of `groups__name` above, and it'll be more robust to naming collisions and possibly faster.  You can still use `group.name` for the dropdown so it'll be the same for users.  Also, I wouldn't bother adding an `id` to the `<option>` unless you for some reason need one; if you do, you should certainly make it more specific (e.g. `user-search-group-{{group.name}}`) since `id`s should be unique in the entire page.

esp/templates/users/usersearch/list_selector.html
```diff
@@ -73,6 +73,14 @@
...
+                    {% endfor %}
+                    <option value="" id="" selected hidden></option>
```
benjaminjkraft: Definitely shouldn't have an empty ID!

esp/templates/users/usersearch/list_selector.html
```diff
@@ -73,6 +73,14 @@
             <div class="step_instructions step_highlight">Currently selected: <span id="filter_current_list"></span></div>
             <div id="filter_accordion">
             
+                <h4><a href="#">User Group</a></h4>
```
benjaminjkraft: Let's put this down at the end, or at least further down -- I think it will be confusing to add something new at the top.  I'd put it either below email or at the end.

esp/templates/users/usersearch/list_selector.html
```diff
@@ -191,6 +199,14 @@
             <div class="step_instructions">(For any criteria you don't need, just leave them blank.)</div>
             <div id="combo_filter_accordion">
             
+                <h4><a href="#">User Group</a></h4>
```
benjaminjkraft: (all the same comments apply to this one)

benjaminjkraft:
Ah, and to make the accordion and such work right it looks like you may need to update a few things in `esp/public/media/scripts/commpanel.js`, if nothing else the list of fields in `clear_filters`.

willgearty:
So, I made all of the edits that were mentioned in your comments, and the code should theoretically work, but it doesn't.
Only the user group that corresponds to the user type being used in the comm panel can be selected (which then doesn't make any changes to the user list). Using any other user group in the dropdown list returns a list of 0 users.
I've been trying to debug it for nearly two hours now with no success. Any thoughts?

willgearty:
Hmm...okay further testing with a shell produced the following results. Group 1 is "Student" and Group 12 is "Splash Fall 2016 Winners". There are 7 accounts that are students, 4 of which are also winners.

`In [51]: ESPUser.objects.filter(groups__name="Student",groups=1).all()`
`Out[51]: [<ESPUser: dmalfoy>, <ESPUser: hgranger>, <ESPUser: gweasley>, <ESPUser: hpotter>, <ESPUser: fweasley>, <ESPUser: rweasley>, <ESPUser: willgearty>]`

`In [52]: ESPUser.objects.filter(groups__name="Student",groups=12).all()`
`Out[52]: []`

`In [53]: ESPUser.objects.filter(groups=1,groups=12).all()`
`File "<ipython-input-53-dc6d3a960242>", line 1 ESPUser.objects.filter(groups=1,groups=12).all() SyntaxError: keyword argument repeated`

`In [54]: ESPUser.objects.filter(groups=1,groups__name="Splash Fall 2016 Winners").all()`
`Out[54]: []`

`In [55]: ESPUser.objects.filter(groups=1,groups__name="Student").all()`
`Out[55]: [<ESPUser: dmalfoy>, <ESPUser: hgranger>, <ESPUser: gweasley>, <ESPUser: hpotter>, <ESPUser: fweasley>, <ESPUser: rweasley>, <ESPUser: willgearty>]`

`In [56]: ESPUser.objects.filter(groups=12,groups__name="Splash Fall 2016 Winners").all()`
`Out[56]: [<ESPUser: dmalfoy>, <ESPUser: gweasley>, <ESPUser: hpotter>, <ESPUser: fweasley>]`

My takeaway: it gets upset when you try to filter by two different groups?
But then why did it work when I was using registration profiles and group names before?

esp/templates/users/usersearch/list_selector.html
```diff
@@ -72,7 +72,7 @@
             <div class="step_instructions">(For any criteria you don't need, just leave them blank.)</div>
             <div class="step_instructions step_highlight">Currently selected: <span id="filter_current_list"></span></div>
             <div id="filter_accordion">
-            
+
```
benjaminjkraft: nit: Can you remove the whitespace change here?

benjaminjkraft:
Looks good, will merge once tests pass!

PR #2224, opened 2016-11-24
Uniformly style #battlescreen input.button/a.abutton

betaveros:
Combine their CSS rules and add line-height, cursor, box-shadow, and
transition rules to override other rules affecting only inputs, so they
look the same.


benjaminjkraft:
Yessssss thank you

PR #2024, opened 2016-02-06
Add bigpicture (Berkeley) theme (WIP, rebase of #1490)

benjaminjkraft:
See #1490 for existing discussion and screenshots.  Fixes towards release are ongoing.

PR #2227, opened 2016-11-24
Add "bigpicture" theme (formerly "berkeley-alpha"; rebase of #2024/#1490)

benjaminjkraft:
This pull request creates a new theme, which I've called "bigpicture", based on the custom theme created by @szhu and used by Berkeley.  See #1490 and #2024 for past discussion.  It should now be ready for review; I'd like to get it in the next stable release since I think it will be a much better theme for small chapters, and certainly looks more modern.

Here's what it looks like live on Berkeley's site:
screenshot from 2016-11-24 12-57-43

And here's what it looks like with little configuration on my dev server:
screenshot from 2016-11-24 12-57-57


benjaminjkraft:
A few notes to self for when we deploy this to berkeley's site:
- need to move media to their new locations
- need to add FB opengraph tags to some template override, probably `elements/html` (see also #2237)
- need to set up navbar to look as it currently does

esp/esp/utils/widgets.py
```diff
@@ -330,15 +331,311 @@ class NavStructureWidget(forms.Widget):
...
+        template = Template(self.template_text % {
+            'add_link_body': self.add_link_body})
         return template.render(Context(context))
 
     def value_from_datadict(self, data, files, name):
         result = json.loads(data[name])
         return result
+
+
+_ICONS = [  # copied from http://getbootstrap.com/components/
```
mgersh: Do we really need to copy the entire list here? I'm guessing the answer is yes because you wouldn't have done it otherwise, but ew.

esp/esp/themes/theme_data/bigpicture/templates/main.html
```diff
@@ -0,0 +1,91 @@
...
+
+{% block main %}
+
+<div class="container">
+  <div class="row">
+    {% block sidebar %}
+    <div id="sidebar" class="span3">
+      <ul class="nav nav-list sidebar">
+        {% include "sidebar/nav.html" %}
+        {% include "sidebar/programs.html" %}
```
mgersh: This doesn't appear to exist?

esp/esp/utils/widgets.py
```diff
@@ -219,6 +219,8 @@ def render(self, name, value, attrs=None, choices=()):
         return mark_safe(output)
 
 class NavStructureWidget(forms.Widget):
+    # TODO(benkraft): Convert this to an actual static script so we don't have
```
mgersh: I have no clue what's going on in this file. Have you checked that it doesn't break the other themes' forms? If so, I think I'm willing to not worry about it too much, since it was already like this.

esp/esp/themes/theme_data/bigpicture/templates/contact.html
```diff
@@ -0,0 +1,61 @@
+{% extends "main.html" %}
+{# TODO(benkraft): Many of these changes should go in the non-theme version #}
```
benjaminjkraft: Really it should just get merged into the main `contact.html`, I just need to check it doesn't break any other themes.  I'll just do that now as a separate PR.

esp/esp/utils/widgets.py
benjaminjkraft: Yeah, if we want to give them the option of those icons.  I could put them in a separate file but this whole thing needs a refactor as discussed so I just did the easy thing.

esp/esp/utils/widgets.py
benjaminjkraft: Yes, will double-check but I think fruitsalad is the only other one using this widget and I tested that.

esp/esp/utils/widgets.py
benjaminjkraft: Oh hmm.  Actually it probably makes sense to just make that switch now so we don't have to worry about breaking things.  Will do.

benjaminjkraft: Done, was pretty easy.

mgersh:
:art: :tada:

PR #2228, opened 2016-11-24
Consolidate render_qsd and render_qsd_inline

betaveros:
The two templates are more similar than different. Also I guess
qsd_default_hidden was pointless in the first place.


esp/templates/inclusion/qsd/render_qsd.html
```diff
@@ -1,12 +1,22 @@
...
+        {% if qsdrec.id %}
+            This is editable text.  Click here to edit the text; click outside the box to save changes.
+        {% else %}
+            This is a placeholder for editable text that has not yet been added.  Click here to edit the text; click outside the box to save changes.
+        {% endif %}
+    </div>
+    <div class="hidden" id="inline_edit_{{ qsdrec.edit_id }}">
+        <textarea rows="8" cols="80" id="qsd_content_{{ qsdrec.edit_id }}"
+                  onblur="qsd_inline_upload('{{ qsdrec.url }}', '{{ qsdrec.edit_id }}');"
+                  class="qsd_editor {% if not inline %}qsd_fullsize{% endif %} qsd_bits hidden {% if inline and not qsdrec.id %}qsd_halfsize{% endif %}"
```
benjaminjkraft: This is a little ugly, I almost wonder if we want to pass down a list of classes from the tag.  Or maybe this just means we need to set things up to not need so many classes.  Anyway, can be cleaned up at some future date if you don't feel like figuring it out now.

esp/templates/inclusion/qsd/render_qsd.html
```diff
@@ -1,12 +1,22 @@
...
+                  name="qsd_content">{% autoescape on %}{{ qsdrec.content }}{% endautoescape %}</textarea>
+        {# Avoid extra whitespace between <textarea> and </textarea>, or it'll
```
benjaminjkraft: I think `{#` is supposed to be only for single-line commments, and you need `{% comment %} {% endcomment %}` for multi-line.

esp/templates/inclusion/qsd/render_qsd.html
betaveros: Yeah, I don't know what half the classes do, but there are more changes I wanted to make to the QSD (mostly changing the blur-to-submit behavior, which makes it really easy to accidentally submit changes) and I can clean this up once I get more familiar from that.

benjaminjkraft: Yeah sounds good.

esp/templates/inclusion/qsd/render_qsd.html
```diff
@@ -1,12 +1,24 @@
...
+        {% else %}
+            This is a placeholder for editable text that has not yet been added.  Click here to edit the text; click outside the box to save changes.
+        {% endif %}
+    </div>
+    <div class="hidden" id="inline_edit_{{ qsdrec.edit_id }}">
+        <textarea rows="8" cols="80" id="qsd_content_{{ qsdrec.edit_id }}"
+                  onblur="qsd_inline_upload('{{ qsdrec.url }}', '{{ qsdrec.edit_id }}');"
+                  class="qsd_editor {% if not inline %}qsd_fullsize{% endif %} qsd_bits hidden {% if inline and not qsdrec.id %}qsd_halfsize{% endif %}"
+                  name="qsd_content">{% autoescape on %}{{ qsdrec.content }}{% endautoescape %}</textarea>
+        {% comment %}
```
benjaminjkraft: Thanks for adding this as a comment btw :-)

PR #2230, opened 2016-11-24
Fix buttons

betaveros:
Add Bootstrap styles to a variety of input elements that lack classes
and end up styled like a text field. Weaken the .fancybutton CSS rule to
style more such buttons. Fix other miscellaneous UI things.

### DIFF
--- a/esp/esp/themes/theme_data/circles/templates/main.html
+++ b/esp/esp/themes/theme_data/circles/templates/main.html
@@ -102,7 +102,7 @@
 <div id="usersearchform">
 <h3>Search for User</h3>
 <p>
-<form id="user_search_form" name="user_search_form" method="get" action="/manage/usersearch">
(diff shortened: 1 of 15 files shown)


betaveros:
These are just some of the buttons I noticed, really. I kind of want to replace all the `.fancybutton` and `.button` classes with Bootstrap classes, but it'll take some work to avoid or override partially conflicting rules with more specificity that style `input`s in elements with an ID and so on, and I'm not sure if this will be a problem for other branches' styles.

betaveros:
(rebased and Bootstrap-3-proofed with `btn-default`)

benjaminjkraft:
Sure!  I certainly don't mind moving more things over -- I think those are just cruft from before we had bootstrap -- but as you say it's a bunch of work so doing it incrementally is probably more feasible.

PR #2232, opened 2016-11-25
Fix line graph on bigboard for programs not using medical form

### LINKED ISSUE #2208: No Line Graph on BigBoard
I was looking through the big board [handler](https://github.com/learning-unlimited/ESP-Website/blob/main/esp/esp/program/modules/handlers/bigboardmodule.py) and [template](https://github.com/learning-unlimited/ESP-Website/blob/main/esp/templates/program/modules/bigboardmodule/bigboard.html), and noticed that there is supposed to be a line graph with registrations over time.
However, on our site's big board, we don't see a line graph. We have the large counts, the load averages, and the most popular classes, but no line graph.
(We're currently in the class lottery phase, but we don't have a line graph for any of the big boards for our previous programs either)

There's no console errors, so my best guess is that the `times_classes` method isn't working correctly? But really, I have no idea.

benjaminjkraft:
The line graph at the top of the studentreg bigboard shows both studentregs and
medical forms for programs that use them.  But it was getting hidden if either
series had no data.  Now it'll only get hidden if both have no data, and we'll
just remove individual series if they have no data.  Fixes #2208.

### DIFF
--- a/esp/esp/program/modules/handlers/bigboardmodule.py
+++ b/esp/esp/program/modules/handlers/bigboardmodule.py
@@ -64,11 +64,12 @@ class BigBoardModule(ProgramModuleObj):
             ("completed the medical form", self.times_medical(prog)),
             ("signed up for classes", self.times_classes(prog)),
         ]
+        timess = [(desc, times) for desc, times in timess if len(times) >= 25]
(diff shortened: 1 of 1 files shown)


jerrywu64:
lgtm

PR #2234, opened 2016-11-25
Add current program cookie and class search module to admin bar

betaveros:
- Add an argument to classsearch that searches for "title containing"
  the query.
- On pages for managing classes, add a "Make current class" button and
  path to store the class id in a cookie.
- Add a Class Search module to the Admin Toolbar that searches the class
  specified by this cookie if one is found.
- Style the admin bar buttons.

### DIFF
--- a/esp/esp/program/models/__init__.py
+++ b/esp/esp/program/models/__init__.py
@@ -54,7 +54,7 @@ from django.db.models.signals import m2m_changed
 from django.dispatch import receiver
 from django.utils import timezone
 
-from argcache import cache_function, wildcard
+from argcache import cache_function, cache_function_for, wildcard
 from esp.cal.models import Event
 from esp.customforms.linkfields import CustomFormsLinkModel
 from esp.db.fields import AjaxForeignKey
@@ -787,13 +787,58 @@ class Program(models.Model, CustomFormsLinkModel):
                 result.append(ts_day)
         return result
 
+    def datetime_range(self):
+        slots = self.getTimeSlots()
+        if slots:
+            return (min(slots).start, max(slots).end)
+        return None
+
+    # @staticmethod --- applied below after the depend_on_model call
+    @cache_function_for(60*60*24)
+    def current_program():
+        """ Guess the "current program", which is the first of the following programs that exists:
+
+        - the shortest program such that the current time is between the
+          start of its first timeslot and the end of its last timeslot
+        - the program in the future (<100 years) that will start the soonest
+        - the program in the past that ended the most recently
+        """
+        now = datetime.now()
+        far_future = now + timedelta(days=36500)
+        def currentness_penalty(program):
+            # The lower the return value (lexicographically), the more
+            # current a program is.
+            if "test" in program.name.lower():
+                return (9001, None)
+
+            datetime_range = program.datetime_range()
+            if datetime_range is None:
+                return (1337, None)
+            start, end = datetime_range
+            if start <= now <= end:
+                # most current: a program running now.
+                # tiebreak by shortest
+                return (0, (end - start))
+            elif now <= start <= far_future:
+                # second most current: program coming up in <100 years
+                # tiebreak by soonest
+                return (1, start)
+            else:
+                # past programs; tiebreak by latest
+                return (2, now - end)
+        programs = Program.objects.all()
+        if programs:
+            return min(programs, key=currentness_penalty)
+        return None
+    current_program.depend_on_model('cal.Event')
+    current_program = staticmethod(current_program)
+
     def date_range(self):
         """ Returns string range from earliest timeslot to latest timeslot, or NoneType if no timeslots set """
-        dates = self.getTimeSlots()
+        datetime_range = self.datetime_range()
 
-        if dates:
-            d1 = min(dates).start
-            d2 = max(dates).end
+        if datetime_range:
+            d1, d2 = datetime_range
             if d1.year == d2.year:
                 if d1.month == d2.month:
                     if d1.day == d2.day:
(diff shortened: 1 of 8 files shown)


betaveros:
screen shot 2016-11-25 at 1 17 39 pm

---

screen shot 2016-11-25 at 1 20 11 pm

---

screen shot 2016-11-25 at 1 18 36 pm

betaveros:
screen shot 2016-11-25 at 3 03 25 pm
screen shot 2016-11-25 at 3 03 33 pm

esp/esp/program/modules/handlers/classsearchmodule.py
```diff
@@ -103,8 +103,14 @@ def classsearch(self, request, tl, one, two, module, extra, prog):
             'program': self.program,
             'query': None,
         }
+        namequery = request.GET.get('namequery')
+        decoded = None
         if data is not None:
             decoded = json.loads(data)
+        elif namequery: # only search if not None and not ""
+            decoded = {'filter': 'title', 'negated': False, 'values': [namequery]}
```
benjaminjkraft: Hmm, is there a reason not to do this client-side?  It seems like that leaves things more flexible if we want to change exactly what the query is later, or add another similar thing with a slightly different query.  Also just seems cleaner that way.

betaveros: The reason is mostly that I'm lazy — the class search page proper goes through some contortions to turn what you enter into the form into a JSON query structure behind-the scenes, and pressing Enter in one of the text fields doesn't work (although I guess this probably isn't too hard to fix). And this is just a drop-dead simple form, there are fewer ways for it to break.

betaveros:
I agree the cookie is not a very friendly approach. I thought about having a default or a drop-down — choosing the program with the largest ID would probably not be a bad heuristic, but it feels like it would be really annoying when it failed. I'm a bit apprehensive about performing a complicated query about program dates or loading all programs into one massive dropdown on *every* page load, though maybe I'm prematurely optimizing. And there might be edge cases if a several-week program is interrupted briefly for a one-week program? What should the default program be then?

There's also the case that if you choose the program from a dropdown you have to dynamically choose the form action with JavaScript, although given the other concern maybe that's going to happen anyway. Or you could write a new endpoint that takes the program as one of the arguments along with the query.

betaveros:
Oh, I also thought it would be cool to have a global "current program" for some other purposes like saving one more click from the front page. If I'm visiting esp.mit.edu during Splash 2016, my first two clicks are usually "Manage" and "Splash 2016".

betaveros:
I guess there's also the option where I just append multiple Class Search modules to the Admin Toolbar based on multiple heuristics: the program of the page being viewed, the program from the user's cookie, maybe the program with the largest ID or all programs in the near future or something. This preserves drop-dead simplicity of forms and of queries. It's kinda ugly though.

luac:
Other places where having a notion of "current program" is relevant: the various shortcut buttons on usersearch pages

esp/esp/program/modules/handlers/classsearchmodule.py
benjaminjkraft: Ah, sure, ok.

benjaminjkraft:
Hmm, yeah.  I feel like in practice the userview thing usually works well enough as-is, and it's nicer than having to deal with a current program set.  This is especially true for chapters that aren't MIT where there is generally only one program that could possibly be active at a time (i.e. either there's exactly one in the future, or it's the most recent one).  I do see the perf issue although we could probably cache it all since it won't change very fast.

What if we do something like, the current program is the one for which you've most recently loaded a manage page?  I guess that might be a mess to do since you'd have to hook into program module routing.  And we might still want an override, although it seems to me that a dropdown is a more useful override than having to go to another page -- at that point I might as well just go to `/classsearch`.

Anyway, I guess I'm not necessarily opposed to doing it this way, it just feels like there might be a better way.

benjaminjkraft:
That's fair.  The thing with a site-wide setting is that if there's a single site-wide current program we should be able to determine that automatically.  I think I'm leaning towards having an automatic default, and then allowing setting it manually, it seems like there should be some reasonably easy set of logic that will do a good job there.  That way at large chapters people can figure out the setting if they want, and at small chapters we'll just do the right thing automatically.  We could also have a global default override, but that seems unnecessary, since we should mostly only need to override in cases where no single global default is correct.  I do still think allowing selecting it inline would be nice, but it can just be TODO.

betaveros:
Well, if there's reasonably easy and convincingly correct logic to pick the right current program, none of us have come up with it yet.

Do you think "largest ID" will do a good enough job? I just want to avoid weird date logic on every page load.

benjaminjkraft:
I had proposed "program with class dates soonest in the future, if any, or most recent program in the past if not".  I'd also add to that "without 'test' in the name", and limit "the future" to the next 100 years, to remove test programs.  The date logic can just get cached, no?

betaveros:
OK, sounds good. I don't know what exact queries I should use to get the class dates of a program or how caching works though. I'll need some help (or lots of time, which might need to wait until after the semester ends...)

esp/esp/program/models/__init__.py
```diff
@@ -787,13 +787,58 @@ def dates(self):
...
+            start, end = datetime_range
+            if start <= now <= end:
+                # most current: a program running now.
+                # tiebreak by shortest
+                return (0, (end - start))
+            elif now <= start <= far_future:
+                # second most current: program coming up in <100 years
+                # tiebreak by soonest
+                return (1, start)
+            else:
```
benjaminjkraft: This wants to be `elif start <= now` to exclude far future programs, or have another block up top similarly to `datetime_range is None`, right?

benjaminjkraft:
Cool, this sounds good!  Just a quick detail in the sorting.

benjaminjkraft:
Awesome, thanks!

betaveros:
Hmm, maybe we should display the "most current" program that has the class search module enabled?

(I should stop working on this until final projects are over)

PR #2235, opened 2016-11-25
Add some semantic styles to classsearch buttons

betaveros:


betaveros:
Colored buttons
screen shot 2016-11-25 at 1 18 36 pm

benjaminjkraft:
Ok!  I worry this is a little too busy but we can always fiddle with it more later.

PR #2239, opened 2016-11-27
Cleanup, refactors, and better error-handling for LaTeX generation

### LINKED ISSUE #1875: If schedule generation fails on CalledProcessError, return the log automatically
This would make it a lot easier to tell the difference between "LaTeX failed, your template override is wrong" and "there's a bug with schedule generation", both of which are quite common, and make it easier to debug the former.  You can already get the log by adding /log to the URL, but people don't always know to do that, and it means you have to run it twice to see it fail and then get the log.

benjaminjkraft:
This includes a bunch of refactoring and cleanup in our LaTeX utils, and in particular better error handling for a number of classes of errors, especially those which are often caused by a bad template override as opposed to a server error.  We'll now return more specific error text with details of what failed.  Fixes #1875.

### DIFF
--- a/esp/esp/utils/latex.py
+++ b/esp/esp/utils/latex.py
@@ -87,7 +87,8 @@ def render_to_latex(filepath, context_dict=None, filetype='pdf'):
 
     rendered_source = t.render(context_dict)
 
-    return gen_latex(rendered_source, filetype)
+    contents = gen_latex(rendered_source, filetype)
+    return HttpResponse(contents, content_type=FILE_MIME_TYPES[filetype])
 
 
 def gen_latex(texcode, type='pdf', stdout=_devnull_sentinel, stderr=subprocess.STDOUT):
@@ -117,9 +118,9 @@ def gen_latex(texcode, type='pdf', stdout=_devnull_sentinel, stderr=subprocess.S
     :type stderr:
         `int` or `file` or `None`
     :return:
-        The generated file.
+        The generated file contents.
     :rtype:
-        HttpResponse
+        `str`
     """
     with open(os.devnull, 'w') as devnull_file:
         # NOTE(jmoldow): `_devnull_sentinel` is private, and currently only the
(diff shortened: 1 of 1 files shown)


esp/esp/utils/latex.py
```diff
@@ -170,8 +170,17 @@ def _gen_latex(texcode, stdout, stderr, type='pdf'):
     elif retcode:
         # Otherwise, if there was an error, we want to exit now since things
         # didn't work.
-        # TODO(benkraft): Return part of the log file here.
-        raise ESPError("LaTeX failed; try looking at the log file.")
+        # TODO(benkraft): Try to extract the actual error out.
+        raise ESPError('LaTeX failed; try looking at the log file.  Here are '
+                       'the last 1000 characters of the log: %s'
+                       % tex_log[-1000:])
```
mgersh: Does this display okay? I haven't looked at it myself so maybe it's fine, but I'd be concerned about admins missing the message at the top and getting confused by the error spew.

esp/esp/utils/latex.py
```diff
@@ -170,8 +170,17 @@ def _gen_latex(texcode, stdout, stderr, type='pdf'):
...
+    elif 'No pages if output' in tex_log:
```
mgersh: Should this be 'No pages **of** output'?

benjaminjkraft: Eek, yes!

esp/esp/utils/latex.py
```diff
@@ -53,19 +53,25 @@
...
     
     This will run the latex interpreter and generate the necessary file type,
-    which must be one of those from FILE_TYPES.
+    which must be one of those from FILE_MIME_TYPES.
     """
-    if filetype not in FILE_TYPES:
+    if filetype not in FILE_MIME_TYPES:
         raise ESPError('Invalid type received for latex generation: %s should '
-                       'be one of %s' % (type, FILE_TYPES))
+                       'be one of %s' % (type, FILE_MIME_TYPES))
```
benjaminjkraft: Ah thanks.

esp/esp/utils/latex.py
benjaminjkraft: screenshot from 2016-12-11 11-31-56

I think I'm gonna call this good enough for now -- 1000 characters isn't that much spew, it's way better than it used to be, and short of a custom error page I don't see a great way to do better since I don't think we can do any real formatting from here.

esp/esp/utils/latex.py
```diff
@@ -71,7 +71,7 @@ def render_to_latex(filepath, context_dict=None, filetype='pdf'):
     """
     if filetype not in FILE_MIME_TYPES:
         raise ESPError('Invalid type received for latex generation: %s should '
-                       'be one of %s' % (type, FILE_MIME_TYPES))
+                       'be one of %s' % (filetype, ', '.join(FILE_MIME_TYPES)))
```
benjaminjkraft: Yep!  Using a dict as an iterator iterates over the keys.

PR #2241, opened 2016-11-28
Add documentation for create_users_for_schools function

hwatheod:
Added some docs for create_users_for_schools function in  useful_scripts/create_users_for_program.py

It wasn't obvious that you need to add new students to both the generic Student group as wel as whatever other new group you're creating.


mgersh:
:+1:

PR #2244, opened 2016-12-03
Add IDs to some commpanel elements so <label>s actually work

betaveros:


betaveros:
Wait whoaaa we have duplicate names/IDs? I had this commit lying around so I pushed it without thinking

benjaminjkraft:
Yeah -- they're in separate forms, but we do :cry:.  So you'll probably have to change them to do this.

esp/templates/users/usersearch/list_selector.html
```diff
@@ -149,16 +149,16 @@
             </select>
             <div class="sendto_fn_select Student">
                 <div>
-                    <input type="checkbox" name="student_sendto_self" value="1" checked="checked" />
-                    <label for="student_sendto_self">Email the students</label>
+                    <input type="checkbox" name="combo-student_sendto_self" id="combo-student_sendto_self" value="1" checked="checked" />
```
benjaminjkraft: Is the name change here intentional?  Won't that break the form?

willgearty:
What's the status on this PR @betaveros ?

willgearty:
This was also implemented as part of #3424.

PR #2245, opened 2016-12-04
Fix .less regexes: don't include variables_foo.less twice

betaveros:
The code tries to include all variables*.less first and then all other
files second, but the regex used to match the latter type of file name
isn't correct and inverting a regex is too damn hard to do or verify
anyway. Make the code do that.

### DIFF
--- a/esp/esp/themes/controllers.py
+++ b/esp/esp/themes/controllers.py
@@ -183,9 +183,17 @@ class ThemeController(object):
             result.append(os.path.join(themes_settings.less_dir, 'bootstrap.less'))
             result.append(os.path.join(themes_settings.less_dir, 'variables_custom.less'))
             result.append(os.path.join(themes_settings.less_dir, 'main.less'))
+
         #   Make sure variables.less is included first, before any other custom LESS code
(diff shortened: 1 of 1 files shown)


betaveros:
Oh, I just realized after clicking the PR button why this was written the way it is: the code still wants to only select *.less files the second time through. This is probably just wrong then (though right now it still works fine and actually better on MIT's website, as far as I can tell)

betaveros:
There are a bunch of ways to implement this... none of them are obviously correct to me 😕 . I could sort/rearrange the files outside of `list_filenames(..., r'\.less$')` myself, but then I'd be operating on the full path and it seems weird to match against it (what if one of the folders has "variables" in its name) or take it apart into parts again when the code had access to the filename earlier. I could add a regex argument with a more sensible interpretation to `list_filenames`, e.g. one that filters out filenames matching it, but the downsides of that were that it would take two filesystem walks and still have two regexes you'd have to keep in sync, and it doesn't look like `list_filenames` is going to be used *that* much anyway.

benjaminjkraft:
I'd just resplit the filenames -- I think having a readable and understandable API is more useful than saving a couple of string operations.

benjaminjkraft:
Cool, this looks plenty simple and correct.  Thanks!

PR #2247, opened 2016-12-05
Use Save/Cancel buttons to save editable text

betaveros:
instead of the old onblur behavior, which is easy to accidentally
trigger after you accidentally click into editable text

screen shot 2016-12-05 at 1 34 37 am

(That was easier than expected.)


jerrywu64:
lgtm

PR #2248, opened 2016-12-05
Update jQuery to 1.12.4, use cdnjs

### LINKED ISSUE #2221: jQuery version is old
1.7.2 is dated March 2012, so four and a half years, it looks like... jQuery even has major version 3 now...

betaveros:
Closes #2221.


benjaminjkraft:
Hmm, have you checked jquery changelogs to see what breaking changes we might need to worry about?  In the past we've often had to make several changes to upgrade jquery.

willgearty:
@betaveros Have you had a chance to follow up on this to check what breaking changes we might need to address? Our jQuery is horrendously out-of-date, and it would be great to upgrade.

willgearty:
OK, I went ahead and finished this PR up. Here are the additional changes I made:

1. Added our jquery version to the settings so we only need to change the number in a single place next time.
2. Same with jquery-ui (but that hasn't been updated since 2016, so I'm not expecting a new release any time soon).
3. Both are now always pulled from a remote CDN (not ours) and use the same version for every page (one exception, see #3210).
4. I used the migrate tool [here](https://github.com/jquery/jquery-migrate/tree/1.x-stable) to find any deprecated javascript in our codebase. I fixed any that I could find, and I left the tool migrate tool in so it will catch any remaining errors (until we want to upgrade to jquery 2 or 3). On dev servers, it will spit out a console warning if there are any problems. On the production server it will just quietly fix our code for us.
5. I updated our local version of jquery-grid (mostly used for the custom form responses page, I believe).
6. On two pages, we weren't importing common.js, which converts our version of jquery from "$" to "$j", so I fixed that.
7. A couple random fixes I ran into while testing various pages.

kkbrum (approved):
I tried to look through all the javascript heavy pages, particularly those with edits, and nothing looks broken to me! I did not look at the admissions dashboard since I've never really seen it in the past, but everything else looks good to me.

Issue #2236, opened 2016-11-25
Bootstrap version is old


betaveros:
v2.x is not supported any more. The current version is v3.3.7 but it's feature-frozen and there's even a v4 in alpha. Let's upgrade.

edit: This is probably a lot of work though. Just from the UI features nearest to my heart right now (i.e. buttons): `.btn` by itself does not style a button fully and you need to add `.btn-default`; `.btn-inverse` is gone, `.btn-large` and friends are now called `.btn-lg`.

betaveros:
OK I looked at this; it's really impractical given the number of layout classes and theme variable names that would be changed, and I don't really need it to do the things I'm interested in, so ignore me and carry on

PR #2250, opened 2016-12-05
Add v3.x bootstrap glyphicons, upgrade some

betaveros:
The new glyphicons are webfonts, so they're prettier and play nicer with
other CSS. There's also no name clashes since the new glyphicons class
names start with `glyphicon`.  I haven't upgraded or gotten rid of all
the old icons; they should still work.

Related to #2236, which I don't think will practically happen all at
once, but could happen partially when useful with piecemeal replacements
of this form.


betaveros:
I previously accidentally some words accidentally in the title

jerrywu64:
lgtm

PR #2251, opened 2016-12-05
Clean up expand_display.js and use legit &minus;

betaveros:
Unify the two different sources of state in the DOM.  Plus and hyphen
have different widths and makes the expansion/contraction ugly; use a
real minus sign instead for the expanded display.


benjaminjkraft:
Cool, this also looks more correct if things were somehow to get desynced.

PR #2252, opened 2016-12-09
Add buttons to undo checkins after checking in

betaveros:
On the way, combine it with the Ctrl-Z function and also rewrite the
jQuery to show/hide the input instead of swapping in/out an entirely new
`<td>`, as this makes it easier to not add messages repeatedly (previously
the "X is no longer checked in" message could be prepended multiple
times.)

### DIFF
--- a/esp/public/media/scripts/onsite/teacher_checkin.js
+++ b/esp/public/media/scripts/onsite/teacher_checkin.js
@@ -50,7 +50,7 @@ $j(function(){
     });
 
     function updateSelected(scroll){
-        var buttons = $j(".checkin");
+        var buttons = $j('.checkin:enabled');
         if(selected >= buttons.length)
             selected = buttons.length-1;
         if(selected < 0)
@@ -71,7 +71,7 @@ $j(function(){
     }
     updateSelected(false);
 
-    function checkIn(username, callback, undo){
+    function checkIn(username, callback, undo, errorCallback){
         refresh_csrf_cookie();
         var data = {teacher: username, csrfmiddlewaretoken: csrf_token()};
         if(undo)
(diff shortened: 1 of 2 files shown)


esp/public/media/scripts/onsite/teacher_checkin.js
```diff
@@ -108,28 +118,54 @@ $j(function(){
...
+            // search from the back because we assume more recent checkins
+            // are more likely to be undone
+            for (var i = checkins.length - 1; i >= 0; i--) {
+                if (checkins[i].username === username) {
+                    targetCheckin = checkins[i];
+                    break;
+                }
+            }
+        }
+        if (targetCheckin === undefined) return;
```
benjaminjkraft: Do we want to have some sort of error here?

esp/public/media/scripts/onsite/teacher_checkin.js
```diff
@@ -108,28 +118,54 @@ $j(function(){
...
     });
 
+    // Undo the check in of the user with the specified username, assuming
+    // that that user was not checked in when this page was loaded (because
+    // the DOM would look different depending on whether that were true). If
+    // no username specified, undo the last check in.
+    function undoLiveCheckIn(username) {
+        var targetCheckin;
+        if (username === undefined) {
+            if (checkins.length === 0) return false;
```
benjaminjkraft: Here and elsewhere, let's do a separate line (and braces) for the body to be explicit, it's just easier to read and harder to mess up that way.

esp/public/media/scripts/onsite/teacher_checkin.js
```diff
@@ -108,28 +118,54 @@ $j(function(){
...
+            targetCheckin = checkins[checkins.length - 1];
+        } else {
+            // search from the back because we assume more recent checkins
+            // are more likely to be undone
+            for (var i = checkins.length - 1; i >= 0; i--) {
+                if (checkins[i].username === username) {
```
benjaminjkraft: I wonder if checkins should be some sort of hash rather than an array, so we don't have to traverse it every time.  But that can be a TODO for now if you want.

esp/public/media/scripts/onsite/teacher_checkin.js
```diff
@@ -108,28 +118,54 @@ $j(function(){
...
+                    break;
+                }
+            }
+            $msg.html(response.message+"<br/>");
+            $td.children('.checkin').show().prop('disabled', false);
+            $td.prev().prop('class', 'not-checked-in');
+            selected = $j('.checkin:enabled').index($td.find('.checkin'));
+            updateSelected(true);
+        });
+        $msg.append(' Undoing...');
```
benjaminjkraft: Is the leading space intentional?

esp/public/media/scripts/onsite/teacher_checkin.js
betaveros: I think undoing any checkin that's not the last one is an extremely rare operation, but okay

esp/public/media/scripts/onsite/teacher_checkin.js
betaveros: Yeah, it'll go after the Undo button and we want some space to visually separate them.

benjaminjkraft:
Cool, thanks for the fixes!

PR #2257, opened 2016-12-11
Fix runserver warning

mgersh:
Now it should only print when you use runserver without extra arguments.


benjaminjkraft:
Thanks for the fix!

PR #2258, opened 2016-12-11
Link the logo to about/welcome.html instead of front page

betaveros:
The front page makes visual impact, but about/welcome.html is more
useful in practice (e.g. far more links).

### DIFF
--- a/esp/esp/themes/theme_data/fruitsalad/templates/main.html
+++ b/esp/esp/themes/theme_data/fruitsalad/templates/main.html
@@ -78,7 +78,7 @@
     <div id="page" class="{{ request.path|extract_theme }}">
     <div id="tab_sidebar">
       <ul id="menu">
-      <li id="tab_logo" class="tabcolor0" ><span class="accent">&nbsp;</span><a href="/"><img id="logo" alt="home" src="/media/images/theme/logo.png" /></a></li>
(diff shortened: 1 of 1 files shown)


benjaminjkraft:
I'm not sure I like this -- people have a strong expectation that the logo will go to the homepage, whether or not that's the most useful page for them to be on.

benjaminjkraft:
(Do others have thoughts on this?  If I'm outvoted, so be it -- this is just one person's opinion.)

mgersh:
I also pointed out in person that it won't work right anyway for sites that use fruitsalad without bubblesfront.

wishdasher:
Miriam brings up a good point. I personally expect it to go to the welcome
page, and I am still surprised sometimes when it doesn't.

On Sun, Dec 11, 2016 at 6:17 PM, Miriam Gershenson <notifications@github.com

betaveros:
I'm not sure what the typical user expects, but the current implementation (theoretically) tests `theme.front_page_style` which controls whether the theme uses bubblesfront

mgersh:
Ah, okay, I wasn't actually paying attention to the diff here, oops. In that case I agree that users would generally expect to end up at the homepage, but I'm not sure bubblesfront counts as the homepage. I think of it more as the page you see before you get the real homepage (I'd call it a "splash page" but that's kinda confusing in this context). But then it sends you to a different "real homepage" depending on which of learn/teach/logo you click, so...I don't know.

bhorowitz:
Since most programs don't upkeep the about/welcome.html currently, at this point this shouldn't be a system wide change. Closing for now.

PR #2260, opened 2016-12-11
Put search glyphicons in the adminbar search buttons

betaveros:
Requires a little css hacking to get the button heights to line up
because our Bootstrap is old, unfortunately.

screen shot 2016-12-11 at 3 44 52 pm


benjaminjkraft:
:mag: :+1:

PR #2261, opened 2016-12-12
Add button, path to preview editable text; update UI

betaveros:
yay glyphicons!

screen shot 2016-12-12 at 1 17 59 pm

---

screen shot 2016-12-12 at 1 18 24 pm

---

screen shot 2016-12-12 at 1 18 31 pm

### DIFF
--- a/esp/esp/qsd/views.py
+++ b/esp/esp/qsd/views.py
@@ -236,3 +236,14 @@ def ajax_qsd(request):
         result['url'] = qsd.url
 
     return HttpResponse(json.dumps(result))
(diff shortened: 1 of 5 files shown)


betaveros:
bump, we're starting to do things for Spark and HSSP

luac:
It seems a little weird that if you preview and then cancel, it still shows the previewed text rather than reverting to the original. But this is probably fine since you can reset it by reloading the page anyway.

luac:
Other than that, 👍

PR #2264, opened 2016-12-13
Add glyphicons to class search

betaveros:
screen shot 2016-12-12 at 10 15 18 pm


benjaminjkraft:
So pretty and even comes with all the right `aria-*` attributes :-)

PR #2265, opened 2016-12-13
Submit class search on Enter key in top-level text field

betaveros:
When the user hits <kbd>Enter</kbd> in a text field in the top level of
the class search form, submit the class search form.

Don't do this in lower-level children in case the user accidentally or
habitually hits Enter and interrupts the building of a complex query.

### DIFF
--- a/esp/public/media/scripts/query-builder.jsx
+++ b/esp/public/media/scripts/query-builder.jsx
@@ -155,7 +155,8 @@ var QueryBuilder = React.createClass({
       Find {this.props.spec.englishName}&hellip;
       <QueryNode ref="queryNode"
                  filters={this.allFilters()}
-                 filterNames={this.allFilterNames()} />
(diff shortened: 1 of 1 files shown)


esp/public/media/scripts/query-builder.jsx
```diff
@@ -190,6 +191,15 @@ var QueryNode = React.createClass({
     // A handler to call if the node is removed.  Leave out onRemove to not
     // allow removing the node (e.g. for the root).
     onRemove: React.PropTypes.func,
+
+    // Handler to call if user tries to submit the search form from here. For
```
benjaminjkraft: So to be clear, it's only if you have a single toplevel filter?

betaveros: Yup.

benjaminjkraft:
Sure, that seems like a good idea.

PR #2269, opened 2016-12-14
Restyle "add class" button

betaveros:
This looks way more like a button now

screen shot 2016-12-14 at 12 37 43 pm


benjaminjkraft:
Very loud indeed!

PR #2271, opened 2016-12-14
Restyle, add glyphicons to student reg submit buttons

betaveros:
Includes a few sort-of-backported patches of some disabled button style
rules, and a glyphicon in the info alert.

screen shot 2016-12-14 at 1 32 33 pm

screen shot 2016-12-14 at 1 32 01 pm

screen shot 2016-12-14 at 1 31 34 pm


benjaminjkraft:
Cool.  I don't know that the logic is what we actually want but it seems the same as before and more readable so that's still a win even if we want to change it later.

PR #2277, opened 2016-12-19
Recolor class search buttons about approval, group together

betaveros:
As noted in #2235 the button colors are a little noisy and kind of
overemphasize the approve/unreview/reject triad while they are pretty
rarely used compared to the other buttons, plus "reject" and "delete"
have very different degrees of danger; so restyle them with new
light-colored semantic styles and group them together.
screen shot 2016-12-19 at 12 25 40 am


benjaminjkraft:
Ok!  Might be worth trying to use these colors consistently as you bootstrapify other similar buttons.

PR #2278, opened 2016-12-19
Tear out some old button styles, replace with Bootstrap

betaveros:
Make the one about actually registering a class big and rewarding.

We don't need three different button styles in a single teacher
registration.


benjaminjkraft:
Seems like sharing some template code would be in order too!  But that can be later.

Issue #2173, opened 2016-09-20
Import volunteer time slots from previous program


willgearty:
This would be analogous to how you can import classroom/timeslot information from previous programs.

willgearty:
See resources_timeslot_import handler [here](https://github.com/learning-unlimited/ESP-Website/blob/main/esp/esp/program/modules/handlers/resourcemodule.py), forms [here](https://github.com/learning-unlimited/ESP-Website/blob/main/esp/esp/program/modules/forms/resources.py), and templates [here](https://github.com/learning-unlimited/ESP-Website/tree/main/esp/templates/program/modules/resourcemodule).

PR #2284, opened 2016-12-28
Volunteer shift import

willgearty:
Allows for the import of volunteer shifts from previous programs as is possible with class timeslots.

See #2173.


willgearty:
Would be nice to see this in SR09 or at least pushed to the Stanford branch.

esp/esp/program/modules/handlers/volunteermanage.py
```diff
@@ -73,7 +75,14 @@ def volunteering(self, request, tl, one, two, module, extra, prog):
             response['Content-Disposition'] = 'attachment; filename=volunteers.csv'
             return response
 
-        if 'op' in request.GET:
+        elif extra == "request_import":
```
benjaminjkraft: Can we just make `volunteer_import` an `aux_call` instead of using `extra`?  I know `csv` already does this but I think that's ugly too; this function is complicated enough when it's not trying to render several mostly-different pages.

esp/esp/program/modules/handlers/volunteermanage.py
```diff
@@ -91,10 +100,67 @@ def volunteering(self, request, tl, one, two, module, extra, prog):
...
+        if 'import_request_form' not in context:
+            context['import_request_form'] = VolunteerImportForm()
         context['requests'] = self.program.getVolunteerRequests()
         return render_to_response('program/modules/volunteermanage/main.html', request, context)
 
+    def volunteer_import(self, request, tl, one, two, module, extra, prog):
+        context = {}
+        response = None
+
+        import_mode = 'preview'
```
benjaminjkraft: If `import_mode` can only have two values, just make it a boolean, e.g. `has_confirmed = request.POST.get('import_confirm') == 'yes'`.

esp/esp/program/modules/handlers/volunteermanage.py
```diff
@@ -91,10 +100,67 @@ def volunteering(self, request, tl, one, two, module, extra, prog):
...
+                    priority = orig_timeslot.priority,
+                    start = orig_timeslot.start + time_delta,
+                    end   = orig_timeslot.end + time_delta,
+                )
+                #   Save the new timeslot and request only if it doesn't duplicate an existing one
+                if import_mode == 'save' and not Event.objects.filter(program=new_timeslot.program, start=new_timeslot.start, end=new_timeslot.end).exists():
+                    new_timeslot.save()
+                new_request = VolunteerRequest()
+                new_request.program = self.program
+                new_request.timeslot = new_timeslot
```
benjaminjkraft: If you didn't save the new timeslot, this won't work -- you need to get the existing one and use that here instead, since you'll end up throwing away the new one.  Same below in the `if`, and further below with `new_request`.  (But see above -- might make this moot.)

esp/esp/program/modules/handlers/volunteermanage.py
```diff
@@ -91,10 +100,67 @@ def volunteering(self, request, tl, one, two, module, extra, prog):
...
+                    program = self.program,
+                    event_type = orig_timeslot.event_type,
+                    short_description = orig_timeslot.short_description,
+                    description = orig_timeslot.description,
+                    priority = orig_timeslot.priority,
+                    start = orig_timeslot.start + time_delta,
+                    end   = orig_timeslot.end + time_delta,
+                )
+                #   Save the new timeslot and request only if it doesn't duplicate an existing one
+                if import_mode == 'save' and not Event.objects.filter(program=new_timeslot.program, start=new_timeslot.start, end=new_timeslot.end).exists():
```
benjaminjkraft: What's probably easier is to use `Event.objects.get_or_create(program=..., start=..., end=..., event_type=..., defaults={'short_description': ..., ...})`.  See the [docs](https://docs.djangoproject.com/en/1.10/ref/models/querysets/#django.db.models.query.QuerySet.get_or_create) for details.  This way you can also just take the return value of that and use it as the timeslot below, without having to worry about whether it was created or existing.  (Similarly for `new_request` below.)

esp/esp/program/modules/handlers/volunteermanage.py
willgearty: @benjaminjkraft How would you feel if I just got rid of the confirmation page (since the slots are already saved anyways, see my inline comment), too?

willgearty:
Haven't deleted the confirmation page template yet (just in case).
Otherwise, this should work fine now using get_or_create for the events and volunteerrequests.
@benjaminjkraft, What do you think?

esp/esp/program/modules/handlers/volunteermanage.py
```diff
@@ -61,6 +63,7 @@ def module_properties(cls):
     @needs_admin
     def volunteering(self, request, tl, one, two, module, extra, prog):
         context = {}
+        response = None
```
benjaminjkraft: you shouldn't need this since it doesn't get referred to anywhere outside the if block where it actually gets set.

esp/esp/program/modules/handlers/volunteermanage.py
```diff
@@ -91,10 +98,55 @@ def volunteering(self, request, tl, one, two, module, extra, prog):
...
+                context['import_error'] = "You can only import shifts from previous programs"
+            else:
+                #Figure out timeslot dates
+                prev_timeslots = []
+                prev_requests = past_program.getVolunteerRequests().order_by('timeslot__start')
+                for prev_request in prev_requests:
+                    prev_timeslots.append(prev_request.timeslot)
+                time_delta = start_date - prev_timeslots[0].start.date()
+                for i,orig_timeslot in enumerate(prev_timeslots):
+                    new_timeslot = Event.objects.get_or_create(
```
benjaminjkraft: nit: usually clearer to do `new_timeslot, _ = Event.objects.get_or_create(...)` rather than `new_timeslot = Event.object.get_or_create(...)[0]`.

esp/esp/program/modules/handlers/volunteermanage.py
```diff
@@ -91,10 +98,55 @@ def volunteering(self, request, tl, one, two, module, extra, prog):
...
+        response = None
+
+        import_form = VolunteerImportForm(request.POST)
+        if not import_form.is_valid():
+            context['import_request_form'] = import_form
+        else:
+            past_program = import_form.cleaned_data['program']
+            start_date = import_form.cleaned_data['start_date']
+            if past_program == prog:
+                context['import_error'] = "You can only import shifts from previous programs"
```
benjaminjkraft: Ideally, you might pass the current program into the form, and then you could make this a form validation error.  That would handle some of the error displaying.  But this is good enough for now.

willgearty:
Thanks for the comments; should be good to go now!

willgearty:
bump @benjaminjkraft

benjaminjkraft:
Looks good -- @hwatheod definitely feel free to merge to stanford's branch.

PR #2285, opened 2016-12-29
Make new site script prompt for release branch

mgersh:
We'll never remember to update this every time we do a release. Let's just ask.


benjaminjkraft:
Good idea.

PR #2302, opened 2017-01-30
Class styles

hwatheod:
Requested by Stanford: Add a field to the teacher registration form for "class style".  Also add corresponding changes to the JSON data module, two-phase catalog, and program dashboard "Status" pop-up box.  

Disabled by default; chapter admin must set a tag "class_style_choices" with the list of dropdown choices to enable.


esp/esp/program/modules/forms/teacherreg.py
```diff
@@ -60,6 +60,14 @@ class TeacherClassRegForm(FormWithRequiredCss):
         ("****", "**** - You should not expect to be able to understand most of this class.",),
     ]
 
+    # The following is a dummy list (because using None causes an error). To enable class styles, admins should set the
```
benjaminjkraft: Does using `[]` work?  That would be a little cleaner, if possible.  But if not this is fine.

hwatheod: Yes, [] works.  Changed.

PR #2306, opened 2017-01-31
CSV fresh classroom import script can now handle invalid rows in furnishings csv

jerrywu64:
Also I improved the text output a little bit


hwatheod:
Looks good to me.

PR #2310, opened 2017-02-03
How to add tags

milescalabresi:
A short explanation of the process for developers to create an use new tags that admins can later set. I welcome feedback, edits, and corrections!


benjaminjkraft:
I'd just put this as the start of a general file about tags, which you can call `tags.rst`.  (We do use extensions so github will display it all pretty.)

milescalabresi:
Oops, forgot about the extension. Fixed, and more prettified.

docs/dev/tags.rst
```diff
@@ -0,0 +1,27 @@
...
+to ``True``) or ``Tag.getTag`` (set it to ``False``). This formatting information is also in the comments of the file linked above.
+
+In order to make use of your new tag, find the file(s) where you want to use the tag's data, and make sure to include the
+line ``from esp.tagdict.models import Tag``.
+Use ``Tag.getTag('tag_name')`` (or the apprpriate variant, either ``getBooleanTag`` or ``getProgramTag``) to retrieve the value of the
+tag in each file.
+
+
+*Notes:*
+Due to the import line, it is not possible to use tag data in `models\__init__.py`.
```
benjaminjkraft: What?

docs/dev/tags.rst
```diff
@@ -0,0 +1,27 @@
+
+
+Basic process of adding and using a new tag:
+--------------------------------------------
+
+To create a *global tag* (one applies to all programs), add a new line to the global tag dictionary ``all_global_tags`` in the
+file https://github.com/learning-unlimited/ESP-Website/blob/main/esp/esp/tagdict/__init__.py.
+To create a *program-specific tag*, add the new tag's name to the program tag dictionary ``all_program_tags`` in the same file as above.
+Program tags are used with ``Tag.getProgramTag()`` or ``Tag.getBooleanTag()`` with a program argument in place of a boolean one in the
+first coordinate.
```
benjaminjkraft: I don't think the last part of this parses -- I'm not quite sure what you mean actually.

docs/dev/tags.rst
milescalabresi: I could be wrong, but it seems like you can't try to call `Tag.getTag('some_tag')` inside of `esp/esp/users/models/__init__.py` because `esp/esp/tagdict/__init__.py` needs to load models. Again, I may be misunderstanding the error I've been getting.

benjaminjkraft: Well, you can't call it at the toplevel of a models file or anything imported by one.  But you really shouldn't do that anyway since that would be evaluated when the process starts and not necessarily reloaded if the tag changes.  Inside a function should be fine.

docs/dev/tags.rst
milescalabresi: I thought it meant you can pass a program as the first argument to `getBooleanTag`, but I did a bad thing and copied blindly from [https://github.com/learning-unlimited/ESP-Website/blob/main/esp/esp/tagdict/\_\_init\_\_.py#L60](https://github.com/learning-unlimited/ESP-Website/blob/main/esp/esp/tagdict/__init__.py#L60) without giving it too much thought. If it doesn't actually make any sense, I'm perfectly happy to remove it from this commit (and possibly the other file as well?).

docs/dev/tags.rst
milescalabresi: Yeah, that makes much more sense. Will edit.

docs/dev/tags.rst
benjaminjkraft: I'd put it more like "Program tags can be used with `Tag.getProgramTag()` or by passing a `program` argument to `Tag.getBooleanTag()`.  It's not in place of anything -- it's an optional arg.

docs/dev/tags.rst
```diff
@@ -1,27 +1,25 @@
...
+Program tags can be used with ``Tag.getProgramTag('tag_name', program)`` or by passing a program argument to ``Tag.getBooleanTag()``.
 
 
 *Notes:*
-Due to the import line, it is not possible to use tag data in `models\__init__.py`.
-You may want to implement checks against the validity of the tag's data.
-It is especially helpful to send a helpful error message to either the admin and the server log. :)
 
-Please also add a description of your tag to the LU Wiki page on tags.
+* Due to the import line, it is not possible to retrieve tag data at the top-level of any ``models`` file or any other file that a ``models`` file imports.
```
mgersh: The new explanation still doesn't make sense to me, although the instruction itself is correct now. The reason for this error is that you [can't use the database until all models have finished loading](https://docs.djangoproject.com/en/1.8/ref/applications/#troubleshooting), and it would probably be a bad idea anyway, as Ben said. I think I'd express the underlying rule here as something like "Because tag data is stored in the database, it shouldn't (and in some cases can't) be accessed at import time."

docs/dev/tags.rst
```diff
@@ -1,27 +1,25 @@
...
+* You may want to implement checks against the validity of the tag's data.
+  It is especially helpful to send a descriptive error message to the admin when the tag is set.
```
mgersh: This would be helpful, but we don't currently have a good way to do it since tags are just set in the admin panel. I definitely wouldn't be opposed to fixing that, but since it doesn't exist yet, I don't want to tell people to do it. We can have helpful error messages when we try to use a tag and find that it's invalid, though, like what's done for quiz_form_id.

docs/dev/tags.rst
milescalabresi: Will do, thanks for the suggestion

docs/dev/tags.rst
milescalabresi: That's very fair. My limited coding instincts thought the next best thing would be to put some kind of check when the data called, but since we also don't have a standard error message or popup error available, I see what you're saying that it's asking a lot more, so I'll take it out.

PR #2311, opened 2017-02-03
Added attribute_value field to resources model

jerrywu64:
ResourceTypes can specify attributes_pickled, and ResourceRequests can specify desired_value, but Resources can't specify anything. This fixes that.

Nothing uses this right now, but I'd therefore highly doubt this will break anything, and we can make changes to other things to use this as needed, such as the AJAX Scheduler, or classroom import scripts, or the "autoscheduler" I'm working on right now.

### DIFF
--- a/esp/esp/resources/models.py
+++ b/esp/esp/resources/models.py
@@ -169,6 +169,7 @@ class Resource(models.Model):
     is_unique = models.BooleanField(default=False)
     user = AjaxForeignKey(ESPUser, null=True, blank=True)
     event = models.ForeignKey(Event)
+    attribute_value = models.TextField()
 
     def __unicode__(self):
         if self.user is not None:
--- a/esp/esp/resources/models.py
+++ b/esp/esp/resources/models.py
@@ -169,7 +169,7 @@ class Resource(models.Model):
     is_unique = models.BooleanField(default=False)
     user = AjaxForeignKey(ESPUser, null=True, blank=True)
     event = models.ForeignKey(Event)
-    attribute_value = models.TextField()
+    attribute_value = models.TextField(default="")
 
     def __unicode__(self):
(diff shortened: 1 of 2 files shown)


jerrywu64:
Huh, okay, migrations and tests seem to work on my computer; is there something else I'm supposed to do to make the tests not fail?

benjaminjkraft:
Looks like maybe you forgot to `git add` the migrations?

jerrywu64:
My git untracked files include:

	../../resources/migrations/0004_resource_attribute_value.py
	../../resources/migrations/0005_remove_resource_attribute_value.py
	../../resources/migrations/0006_resource_attribute_value.py
I assume everything else is irrelevant since they're not resources. What should I be adding, exactly? (or is there a way to redo my migrations so it's cleaner?)

benjaminjkraft:
You wan to add all of those migrations files.  Or possibly you can actually just (manually or automatically) squash them into a single file.  In general whenever you generate migrations you want to add the generated file along with the commit.

jerrywu64:
Cool, thanks.

esp/esp/resources/models.py
```diff
@@ -169,6 +169,7 @@ class Resource(models.Model):
     is_unique = models.BooleanField(default=False)
     user = AjaxForeignKey(ESPUser, null=True, blank=True)
     event = models.ForeignKey(Event)
+    attribute_value = models.TextField(default="")
```
mgersh: This [needs to have blank=True](https://docs.djangoproject.com/en/1.8/ref/models/fields/#blank) so people will be able to edit it from the admin panel or other forms. (Then you'll want to delete and regenerate the migration.)

It's also not clear from the code what it's for. I don't have bright ideas for better names, so maybe add a comment?

PR #2314, opened 2017-02-06
Fix K12School admin panel

### LINKED ISSUE #2307: K12 schools page raises server error
Happing on most (all?) sites but first reported by Oxford, the /admin/users/k12school/ won't load and instead displays a "whoops, the server messed up" error. Possible lead is that Django says the contact field is a nonetype.

mgersh:
Fixes #2307.


hwatheod:
Looks good to me.

PR #2315, opened 2017-02-06
Make manage/flushcache with no trailing slash work

### LINKED ISSUE #2175: /manage/flushcache (no trailing slash) loads but doesn't work
It sends you to /manage/ instead of back to the same page, which is typically a 404 and in any case doesn't flush the cache. I'm guessing this has something to do with the `<form action=".">`.

mgersh:
Fixes #2175.


hwatheod:
Looks good to me.

PR #2316, opened 2017-02-06
Add better error message for timeslot import from the same program

### LINKED ISSUE #2114: Show a reasonable error message when someone tries to import timeblocks from the current program
I've seen people try this multiple times. Right now it'll show a generic Whoops. Ideally it would be a form validation error that says something like "That's the current program! Try importing timeblocks from a past program that already has timeblocks."

mgersh:
Fixes #2114.


hwatheod:
Wouldn't it be better to exclude the current program from showing up in the drop down in the ClassroomImportForm?

benjaminjkraft:
Agreed but this seems fine in the meantime.

PR #2319, opened 2017-02-13
Add capacities, preregistered counts to Manage Class page

betaveros:
Add the capacity of sections as well as the number of people who
preregistered for each section in the first part of the Manage Class
view, with text descriptions so they are not confused.

Also break things into more lines and add a linebreak tag that went
missing.


hwatheod:
Looks good.  I remember being confused by the previous description.

PR #2321, opened 2017-02-14
Add a page that picks and displays random classes

betaveros:
Add a page that displays a random class with a button that dynamically
picks and displays another random class.


hwatheod:
Just curious, what is the purpose of this page?

benjaminjkraft:
¯\_(ツ)_/¯ seems fine/fun to me!

betaveros:
Oh, I never responded to this question — we sometimes link to this page in teacher recruitment materials and suggest teachers view some random classes for inspiration if they can't think of anything to teach.

PR #2323, opened 2017-02-15
Show attribute_value in scheduling application classroom tooltips.

uakfdotb:
In addition to displaying resource names, it will include the resource
attribute_value if set.

@werryju @jerrywu64


jerrywu64:
lgtm

PR #2325, opened 2017-02-18
Annotate request with permission info in checkDeadline, display it

### LINKED ISSUE #2324: Warn admins if deadlines would block the page they're viewing
As I understand it, the site allows admins to view all pages at all times, which is great for administering things but makes it dangerously hard for us to notice if we've misconfigured our deadlines and students or teachers can't see the page.

I mentioned this in #1700 but never elevated it to a full issue and totally forgot about it.

betaveros:
In the helper function for checking deadline satisfiability, if the user
is an administrator, annotate the request with information about which
permission or deadline was checked and what roles would have had
permission. Then display this information in the fruitsalad default
template.

To accommodate all this, also refactor deadline code to:

- add a new method for listing roles that have a particular permission,
  pulling out shared logic with user_has_perm
- eliminate duplication
- remove the unused meets_all_deadlines method.

Fixes #2324.

### DIFF
--- a/esp/esp/program/modules/base.py
+++ b/esp/esp/program/modules/base.py
@@ -512,27 +512,55 @@ def meets_grade(method):
 
 # Just broke out this function to allow combined deadlines (see meets_any_deadline,
 # meets_all_deadlines functions below).  -Michael P, 6/23/2009
+
+# Combined deadlines are annoying; meets_deadline has been absorbed into
+# meets_any_deadline and meets_all_deadlines has been deleted --bpchen,
+# 2017-02-17
 def _checkDeadline_helper(method, extension, moduleObj, request, tl, *args, **kwargs):
+    """
+    Given information about a request, return a pair of type (bool, None |
+    response), which indicates whether the user can view the requested page,
+    and an optional redirect if not.
+
+    If the user is an administrator, annotate the request with information
+    about what roles have permission to view the requested page.
+    """
     if tl != 'learn' and tl != 'teach' and tl != 'volunteer':
         return (True, None)
     response = None
     canView = False
     perm_name = {'learn':'Student','teach':'Teacher','volunteer':'Volunteer'}[tl]+extension
     if not_logged_in(request):
-        if not moduleObj.require_auth() and Permission.valid_objects().filter(permission_type=perm_name, program=request.program, user__isnull=True).exists():
+        if not moduleObj.require_auth() and Permission.null_user_has_perm(permission_type=perm_name, program=request.program):
             canView = True
         else:
             response = HttpResponseRedirect('%s?%s=%s' % (LOGIN_URL, REDIRECT_FIELD_NAME, quote(request.get_full_path())))
     else:
-        canView = request.user.updateOnsite(request)
+        user = request.user
+        program = request.program
+        canView = user.updateOnsite(request)
         if not canView:
-            canView = Permission.user_has_perm(request.user,
+            canView = Permission.user_has_perm(user,
                                                perm_name,
-                                               program=request.program)
+                                               program=program)
             #   For now, allow an exception if the user is of the wrong type
             #   This is because we are used to UserBits having a null user affecting everyone, regardless of user type.
-            if not canView and Permission.valid_objects().filter(permission_type=perm_name, program=request.program, user__isnull=True).exists():
+            if not canView and Permission.valid_objects().filter(permission_type=perm_name, program=program, user__isnull=True).exists():
                 canView = True
+
+            #   Give administrators additional information
+            if user.isAdministrator(program=program):
+                request.show_perm_info = True
+                if hasattr(request, 'perm_names'):
+                    request.perm_names.append(perm_name)
+                else:
+                    request.perm_names = [perm_name]
+
+                roles_with_perm = Permission.list_roles_with_perm(perm_name, program)
+                if hasattr(request, 'roles_with_perm'):
+                    request.roles_with_perm += roles_with_perm
+                else:
+                    request.roles_with_perm = roles_with_perm
 
     return (canView, response)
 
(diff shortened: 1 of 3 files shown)


esp/esp/program/modules/base.py
```diff
@@ -512,28 +512,56 @@ def _checkGrade(moduleObj, request, tl, *args, **kwargs):
 
 # Just broke out this function to allow combined deadlines (see meets_any_deadline,
 # meets_all_deadlines functions below).  -Michael P, 6/23/2009
+
+# Combined deadlines are annoying; meets_deadline has been absorbed into
+# meets_any_deadline and meets_all_deadlines has been deleted --bpchen,
+# 2017-02-17
```
benjaminjkraft: You can just get rid of these two comments, IMO.  The history is all in git if we need it, and I don't think it helps with understanding current code.

esp/esp/program/modules/base.py
```diff
@@ -512,28 +512,56 @@ def _checkGrade(moduleObj, request, tl, *args, **kwargs):
...
 def _checkDeadline_helper(method, extension, moduleObj, request, tl, *args, **kwargs):
+    """
```
benjaminjkraft: Thanks for the docstring :-D !  Per [PEP8/PEP257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings) multiline docstrings should have a one-line summary, then a blank line, then whatever else.  (Happy to make more specific suggestions about what to say, if helpful -- bug me on slack/zephyr/etc.)

esp/esp/program/modules/base.py
```diff
@@ -512,28 +512,56 @@ def _checkGrade(moduleObj, request, tl, *args, **kwargs):
...
             #   For now, allow an exception if the user is of the wrong type
             #   This is because we are used to UserBits having a null user affecting everyone, regardless of user type.
-            if not canView and Permission.valid_objects().filter(permission_type=perm_name, program=request.program, user__isnull=True).exists():
+            if not canView and Permission.valid_objects().filter(permission_type=perm_name, program=program, user__isnull=True).exists():
                 canView = True
 
+            #   Give administrators additional information
+            if user.isAdministrator(program=program):
+                request.show_perm_info = True
+                if hasattr(request, 'perm_names'):
```
benjaminjkraft: [`hasattr` has some weird edge cases](https://hynek.me/articles/hasattr/); normally I don't worry too much about them, but request objects in django are super duper magic so it could actually be an issue.  Better to use `getattr(obj, attr, None) is not None`.

esp/esp/users/models/__init__.py
```diff
@@ -2314,6 +2314,37 @@ class Meta:
         app_label = 'users'
 
     @classmethod
+    def null_user_has_perm(cls, permission_type, program):
+        return cls.valid_objects().filter(permission_type=permission_type,
+                program=program, user__isnull=True).exists()
+
+    @classmethod
+    def q_permissions_on_program(cls, perm_q, name, program=None, when=None, program_is_none_implies_all=False):
+        """
```
benjaminjkraft: (same for the docstring)

esp/esp/users/models/__init__.py
```diff
@@ -2364,18 +2395,34 @@ def user_has_perm(cls, user, name, program=None, when=None, program_is_none_impl
...
+        :type program:
+            `Program`
+        :return:
+            List of role names that would give the specified permission.
+        :rtype:
+            `list` of `str`
+        """
+
+        qrole = Q(user=None, role__isnull=False)
+        return [permission.role.name for permission in
```
benjaminjkraft: Probably slightly better to do `.values_list('role__name')` to avoid a second db fetch for the related lookup (and send less data back from the db to boot).

esp/esp/program/modules/base.py
```diff
@@ -547,59 +575,34 @@ def list_extensions(tl, extensions, andor=''):
     else:
         return 'the deadlines '+', '.join([nicetl+e for e in extensions[:-1]])+', '+andor+' '+nicetl+extensions[-1]+' were'
 
+def render_deadline_for_tl(tl, request, context):
+    errorpage = 'errors/program/deadline-%s.html' % tl
+    return render_to_response(errorpage, request, context)
+
 #   Return a decorator that returns a function calling the decorated function if
```
benjaminjkraft: These may as well become docstrings while you're at it.

esp/esp/themes/theme_data/fruitsalad/templates/main.html
```diff
@@ -209,17 +209,29 @@ <h1 class="title">{% block content_title %}&nbsp;{% endblock %}</h1>
...
       <div id="content">
+          {% if request.show_perm_info %}
+          <div class="alert alert-info">
+              <span class="glyphicon glyphicon-info-sign" aria-hidden="true"></span>
+              Permission/deadline controlling this page: {{ request.perm_names|join:", " }}
+              <br />
+              {% if request.roles_with_perm %}
+                  Roles with permission to view this page: {{ request.roles_with_perm|join:", " }}
+              {% else %}
+                  No roles with permission to view this page!
```
benjaminjkraft: super nit: I'd say "no roles have permission..." since that is clearer if you don't know what it might say if there were such roles.

esp/esp/themes/theme_data/fruitsalad/templates/main.html
```diff
@@ -209,17 +209,29 @@ <h1 class="title">{% block content_title %}&nbsp;{% endblock %}</h1>
 {% endblock %}
 
       <div id="content">
+          {% if request.show_perm_info %}
+          <div class="alert alert-info">
+              <span class="glyphicon glyphicon-info-sign" aria-hidden="true"></span>
+              Permission/deadline controlling this page: {{ request.perm_names|join:", " }}
```
benjaminjkraft: super nit: I'd say "Permission/deadline required to view this page" -- and you may as well link to the manage deadlines page, iff that's not too hard to do.

PR #2330, opened 2017-02-22
AJAX scheduler: Add gradient to classes with special scheduling needs

betaveros:
Add a gradient background to the corners of classes with special
scheduling needs, as determined by the tags.

### DIFF
--- a/esp/public/media/scripts/ajaxschedulingmodule/ESP/CellColors.js
+++ b/esp/public/media/scripts/ajaxschedulingmodule/ESP/CellColors.js
@@ -6,16 +6,32 @@
  * @method textColor(section)
  */
 function CellColors() {
+    this.specialColor = "rgb(255, 153, 0)";
     /**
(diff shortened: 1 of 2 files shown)


esp/public/media/scripts/ajaxschedulingmodule/ESP/CellColors.js
```diff
@@ -6,16 +6,32 @@
...
      *
      * @param section: The section to compute the background color of
      */
-    this.color = function(section){
-        return "#" +
-            section.emailcode[2] + "0" +
-            section.emailcode[3] + "0" +
-            section.emailcode [4] + "0";
+    this.color = function(section) {
+        return [17*(+section.emailcode[2]),
```
benjaminjkraft: If you're trying to keep the old logic, you want `16*...`, right?

benjaminjkraft:
:100:

PR #2333, opened 2017-02-22
Show module list on program manage directory by default.

uakfdotb:
Issue #357 added the current view with the nice looking sections and nice
looking buttons, but we still use the module list a lot.


benjaminjkraft:
Ok!  No point saving screen real estate at the bottom of the page anyway :-) .

PR #2334, opened 2017-02-22
Add more functions to medical bypass page

betaveros:
Add buttons to query the medical form submission status of the user and
to delete the medical bypass. Also clarify text and message passing.
screen shot 2017-02-21 at 9 48 12 pm


benjaminjkraft:
Seems fine to me!

PR #2344, opened 2017-03-08
Switch the dev VM image to 14.04

mgersh:
As a bonus, it now has our dependencies pre-installed. No more waiting
for all of texlive to download every time!


benjaminjkraft:
Thanks for doing this, and for adding the instructions!

PR #2345, opened 2017-03-08
Update pip before using it to install anything else

mgersh:
setuptools needs a new version of pip, but virtualenv packages a
very old version of pip. If we don't do this, we get strange errors.


benjaminjkraft:
¯\_(ツ)_/¯

PR #2349, opened 2017-03-11
Update automatic printing text for new Firefox version.

uakfdotb:
print_printer is now used instead of print.print_printer.


hwatheod:
Looks like print.print_color has also changed to print.print_in_color

benjaminjkraft:
Seems fine to me -- if someone wants to add the other one in a separate PR go for it.

PR #2356, opened 2017-03-18
Make sure new_site.sh sets up log files with the right permissions

benjaminjkraft:
This won't help the logrotate issues, but it'll at least prevent them on first setup.


mgersh:
:+1:, and I think I fixed the logrotate issues by fiddling with the config after last time it happened.

PR #2366, opened 2017-04-02
FCFS catalog fix

willgearty:
Fixes an unintended issue caused by #2317.
Need to update display property for within student registration (since it is now `display: none` by default).


hwatheod:
Looks good to me

Issue #347, opened 2012-04-17
Automate volunteer nametags


pricem:
From Stanford: There is now a volunteer management interface so the Web site knows who your volunteers are, but the nametag generator still expects you to enter volunteers yourself (with snarky subtitles).  So, it would be nice to integrate this with some kind of automatically generated list of volunteers.

willgearty:
Fixed by #2368.

PR #2368, opened 2017-04-04
Automatic nametags for any user group or volunteers

willgearty:
Adds a new option and dropdown menu to select any user group for nametag creation.

Requested by Stanford for use with custom student reg groups (so the nametags still have usernames on them).

### DIFF
--- a/esp/esp/program/modules/handlers/nametagmodule.py
+++ b/esp/esp/program/modules/handlers/nametagmodule.py
@@ -40,6 +40,9 @@ from esp.tagdict.models import Tag
 from esp.users.models import ESPUser
 from esp.utils.web import render_to_response
 
+from django.contrib.auth.models import Group
+from django.db.models.query import Q
+
 
 class NameTagModule(ProgramModuleObj):
     """ This module allows you to generate a bunch of IDs for everyone in the program. """
@@ -57,6 +60,7 @@ class NameTagModule(ProgramModuleObj):
     def selectidoptions(self, request, tl, one, two, module, extra, prog):
         """ Display a teacher eg page """
         context = {'module': self}
+        context['groups'] = Group.objects.all()
 
         return render_to_response(self.baseDir()+'selectoptions.html', request, context)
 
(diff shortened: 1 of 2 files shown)


willgearty:
The second commit fixes #347

esp/templates/program/modules/nametagmodule/selectoptions.html
```diff
@@ -47,36 +47,50 @@
...
+    <option value="" selected hidden></option>
+</select>
   <li><input type="radio" name="type" value="volunteers" id="volunt"
-       onclick="document.getElementById('volunteers').disabled = false;" />
+       onclick="document.getElementById('misc_info').disabled = true; document.getElementById('group').disabled = true;" />
       <label for="volunt">Volunteers</label>
   </li>
+  <li><input type="radio" name="type" value="misc" id="misc"
+       onclick="document.getElementById('misc_info').disabled = false; document.getElementById('group').disabled = true;" />
+      <label for="volunt">Miscellaneous (enter info below)</label>
```
benjaminjkraft: I think this will be a lot clearer -- I always forgot which one to pick :-) .

esp/esp/program/modules/handlers/nametagmodule.py
```diff
@@ -107,10 +111,45 @@ def generatetags(self, request, tl, one, two, module, extra, prog):
...
+            users_list.sort()
+
+            user_title = request.POST['blanktitle']
+            for user in users_list:
+                users.append({'title': user_title,
+                              'name' : '%s %s' % (user.first_name, user.last_name),
+                              'id'   : user.id,
+                              'username': user.username})
+
         elif idtype == 'volunteers':
```
benjaminjkraft: Can we duplicate a bit less of the logic between user types here?  You can just do the if-else block for generating the list of users (and title), and then computing the actual nametag data from there is almost the same in each case (except misc/blank).  I know we were already duplicating code, but since you're adding two more copies it's a good time to do a little refactoring.

esp/templates/program/modules/nametagmodule/selectoptions.html
```diff
@@ -47,36 +47,50 @@
...
   <li><input type="radio" name="type" value="teacher" id="teachers"
-       onclick="document.getElementById('volunteers').disabled = true;" />
+       onclick="document.getElementById('misc_info').disabled = true; document.getElementById('group').disabled = true;" />
       <label for="teachers">Teachers</label>
   </li>
+  <li><input type="radio" name="type" value="other" id="other"
+       onclick="document.getElementById('misc_info').disabled = true; document.getElementById('group').disabled = false;" />
+      <label for="teachers">Other Group</label>
+  </li>
+<select name="group" id="group" disabled>
```
benjaminjkraft: This should probably just go inside the `<li>`, perhaps to the right of "Other Group".

willgearty:
Yeah, nametags haven't been updated in a while, and I figured they should at least cover all of this different types of users.
Also, I've gotten really tired of handling all of the volunteer csv data and then plugging it into the nametags.
Also also, we ran into a problem last fall where the student nametags didn't have usernames (because we made "Miscellaneous" nametags for them), so this fixes that, too!

benjaminjkraft:
Great!

hwatheod:
In the stanford-sr8 branch of the code (but apparently not in main), the student section of the name tag code had a field "grade".  This got removed in the refactoring.

willgearty:
If we do want grade for students, I suppose I could be put in a check for whether a user is a student, then add a grade parameter.
However, it doesn't look like we even use the grade in our name tag template override (Stanford).

hwatheod:
Okay.  I will push this change as-is to stanford-sr8 then.

PR #2369, opened 2017-04-07
Volunteer stats

### LINKED ISSUE #627: Count total number of distinct volunteers
From Stanford: On the manage volunteer page (/manage/Splash/2013_Spring/volunteering) show the total number of distinct volunteers.  It's useful for things like knowing how many volunteer T-shirts we need to have.

Should we also put a "Volunteers" category in the dashboard?

willgearty:
Adds volunteer registration statistics to dashboard and volunteer page

Fixes #627 .

### DIFF
--- a/esp/esp/program/modules/handlers/jsondatamodule.py
+++ b/esp/esp/program/modules/handlers/jsondatamodule.py
@@ -746,7 +746,8 @@ teachers[key].filter(is_active = True).distinct().count()))
 
         volunteer_list = []
         volunteer_dict = prog.volunteers()
-        volunteer_list.append(("Volunteers who are signed up for at least one time slot", volunteer_dict['volunteer_all'].count()))
+        if 'volunteer_all' in volunteer_dict:
+            volunteer_list.append(("Volunteers who are signed up for at least one time slot", volunteer_dict['volunteer_all'].count()))
         vitals['volunteernum'] = volunteer_list
 
         timeslots = prog.getTimeSlots()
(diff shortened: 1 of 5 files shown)


willgearty:
Quick note: the dashboard stats change requires a cache flush.

esp/esp/program/modules/handlers/jsondatamodule.py
```diff
@@ -743,6 +743,12 @@ def stats(prog):
             else:
                 vitals['studentnum'].append((key, students[key].filter(is_active = True).distinct().count()))
 
+
+        volunteer_list = []
+        volunteer_dict = prog.volunteers()
+        volunteer_list.append(("Volunteers who are signed up for at least one time slot", volunteer_dict['volunteer_all'].count()))
```
benjaminjkraft: This probably needs to be wrapped in some sort of check if the module is enabled at all -- such as `if 'volunteer_all' in volunteer_dict`.

esp/public/media/scripts/program/modules/adminvitals.js
```diff
@@ -35,6 +35,18 @@ function fillStats(data)
 	}
     }
 
+    // Fill in the volunteer num data
```
benjaminjkraft: I'm less attached to it here as in #2368 since it's already a lot worse, but here, similarly, it would be nice while you're in the area to do a bit of cleanup and make it so we don't have to have 4 identical blocks of code for student/teacher/volunteer/class/hours data.  You could just define a function which takes an ID (e.g. `#stats_volunteers`) and a list of stats (e.g. `vitals.volunteernum`) and does all the work, and then just call it a bunch.

PR #2370, opened 2017-04-08
Fix error when an invalid barcode is submitted

benjaminjkraft:
In the case where a submitted ID is nonzero or invalid, we were catching the
error but handling it incorrectly.  In the case where this was not the first
ID, we'd re-use the previous one, thereby processing it twice and perhaps
causing a number in the results to be off by one.  In the case where this was
the first ID, we'd crash with an undefined name error.  Now we'll correctly
just continue to the next ID in all cases.


hwatheod:
Looks good

PR #2378, opened 2017-04-23
Hide class rooms in class changes grid (#2351)

### LINKED ISSUE #2351: Add option to class changes / on-site grid to hide classroom locations
Per Spark discussion, it is convenient at class changes to just show students the class changes screen for them to see what classes they can change into, but somewhat sketchy since the classroom locations are displayed in the grid, and we try not to release the locations to people who aren't in the class. We could fix this pretty easily by hiding the classroom locations or adding an option to hide them.

#2295 concerns a very nearby change and one might want to do them together.

ecnerwala:
Completely untested, as my dev server isn't up yet. Should just be copy-pasta from `show_class_titles`.

Fixes #2351

### DIFF
--- a/esp/public/media/scripts/onsite/ajax_status.js
+++ b/esp/public/media/scripts/onsite/ajax_status.js
@@ -17,6 +17,7 @@ var settings = {
     override_full: false,
     disable_grade_filter: false,
     show_class_titles: false,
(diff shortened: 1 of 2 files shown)


betaveros:
LGTM and works on dev server 👍

PR #2379, opened 2017-04-27
Delete Showdown

betaveros:
It's an alternate Markdown implementation we used in another era; it
hasn't been referenced since commit 0cbc07a33.


benjaminjkraft:
:+1: thanks!

PR #2380, opened 2017-04-30
Delete useless select_related for meeting_times

betaveros:
I think it doesn't work for ManyToMany fields; older Django ignored it
but now throws an error.

(I put this in prod to make class changes work some time ago and then forgot to make a pull request?)


benjaminjkraft:
Thanks!

PR #2382, opened 2017-04-30
Clamp num_spaces to >=0 in class changes controller

betaveros:
If students enroll in a class and then it's moved to a smaller
classroom, there can be negative space, so "more students than space",
and the code will try to randomly sample a negative number of students
to fill the space with, which naturally causes an exception. I think we
ran into this bug while running the class changes lottery for Spring
HSSP 2017.


benjaminjkraft:
Oops!

Issue #1401, opened 2014-10-17
Scheduler should be able to override constraints


benjaminjkraft:
Right now only the manageclass page can override any constraints (e.g. lunch, double-scheduling teachers).

willgearty:
OK, I think there's three things we would need to do to make this happen:

1. Add a checkbox to the filters/sections interface that enables/disables the override
2. When checking for the available timeslots for a class, check the status of that checkbox (or update some js variable when we check/uncheck it like with the filters) and if the override is enabled, just return all of the timeslots:
https://github.com/learning-unlimited/ESP-Website/blob/e5198eb53b618b85a2e5419c99fe2ee7ee1ec784/esp/public/media/scripts/ajaxschedulingmodule/ESP/Sections.js#L262
3. Do the same thing when submitting the schedule request to the server, so we tell the server whether or not to worry about constraints:
https://github.com/learning-unlimited/ESP-Website/blob/e5198eb53b618b85a2e5419c99fe2ee7ee1ec784/esp/public/media/scripts/ajaxschedulingmodule/ESP/ApiClient.js#L58
then use that to determine whether to run `cannotSchedule` here:
https://github.com/learning-unlimited/ESP-Website/blob/e5198eb53b618b85a2e5419c99fe2ee7ee1ec784/esp/esp/program/modules/handlers/ajaxschedulingmodule.py#L120

Issue #176, opened 2011-11-25
Variations on coteacher availability rules


pricem:
From Stanford:
For classes with coteachers, have something like an "use primary
teacher's availability only" button or something to avoid having to
email all the coteachers. Alternatively, somehow have a check on
coteachers availabilitys as they get added to a class to be sufficient
for the class and the overlap with the coteachers already signed up to
teach.

willgearty:
Resurrecting this issue since it is still relevant and still an issue.
Stanford has now run into a couple of instances where each teacher for a class has enough availability for the class but their combined availability is not enough for the class.

For example, a teacher teaching a single section of a class for 2 hours could be available all of Saturday.
They then add a co-teacher, who is available all of Sunday. The website doesn't flag this, since both teachers are available for enough time for the 2 hours of class.
However, when you go to schedule the class, there isn't actually any overlap between the two teachers' availabilities.

The best solution (imho), is to check **before** adding the co-teacher to the class whether there is enough overlap between all of the teachers.

As it is, all co-teachers are emailed when any changes are made to the class (including adding co-teachers). But even if an email was sent to the teacher as @jmoldow suggests, this wouldn't **ensure** all co-teachers have enough overlapping availability, merely encourage it.

PR #2391, opened 2017-05-27
AJAX scheduler: make teachers with no availability always available

### LINKED ISSUE #1005: AJAX scheduler should work if teacher availability module isn't enabled
The MIT Spring HSSP directors just tried to schedule and couldn't do it without teacher availability. We should add checks so that availability is ignored if the module isn't enabled.

betaveros:
If a teacher is not available for any timeslots, consider them available
for all timeslots, with a different striped yellow-green style to warn
about the fact that this was used. Fixes #1005 probably?

(If the availability module is enabled, scheduling classes onto such
fallback timeslots will fail on the server side and be reported to the
client. But I think this is fine since it moves us closer to #1401?)

### DIFF
--- a/esp/public/media/default_styles/scheduling.css
+++ b/esp/public/media/default_styles/scheduling.css
@@ -42,6 +42,14 @@ table,th,td
     opacity: .9;
 }
 
+.teacher-available-with-fallback-cell {
+    background: repeating-linear-gradient(
+                45deg,
+                #3c3, #3c3 10px, #ac3 10px, #ac3 20px
+    );
+    opacity: .9;
+}
+
 .teacher-available-not-first-cell {
     background: repeating-linear-gradient(
                 45deg,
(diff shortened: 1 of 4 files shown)


benjaminjkraft:
Hmm, why do this even if the availability module is enabled?  I would think we only want to do this if the availability module is not enabled (in which case we probably don't even need to have the warning).

betaveros:
It's been a while but I seem to recall that when the availability module is enabled, sometimes teachers add co-teachers to their classes who never fill out their availability, and when scheduling weekend happens we can only ignore the co-teachers' lack of availability.

benjaminjkraft:
Hmm, it seems to me like the far better thing is to just let us override that, no?  We shouldn't be letting it pass unnoticed, especially if we changed the timeslots for the program or something.  (Also, wait, I thought we prevented adding someone as a coteacher if they don't have enough compatible availability?)

willgearty:
I don't think it's possible for a teacher without availability to be added as a coteacher to a class. Furthermore, I think it would be better to fix how we handle teacher availability (esp. when teachers are added as coteachers), rather than trying to catch any messes after the fact (e.g. #2536, #176).

willgearty:
Closed in favor of #2859

Issue #1587, opened 2015-02-10
Make a teacherreg big board


benjaminjkraft:
Similar to the existing student one.  Possible stats:
- classes registered
- classes approved
- classes with X flag (probably not for all flags, maybe for all flags of which there are at least 10 or something)
- classes requesting X resource (maybe not for all resources)

We are now timestamping classes (#2013), so we can have a pretty graph.  (We may still want to backfill existing classes from email, which would also be useful for seeing if we have enough classes at some point during teacherreg.)

This might involve abstracting some of the studentreg big board functionality to save code duplication.

willgearty:
Any progress on this?

benjaminjkraft:
I haven't really been working on it; if someone else wants to grab the ticket feel free.

willgearty:
Fixed by #2396.

PR #2396, opened 2017-07-09
Teacher Big Board

willgearty:
This is my first stab at creating a teacher big board (#1587).
As of right now, it shows a graph of the following stats:

- Number of registered classes
- Number of approved classes
- Number of teachers teaching
- Number of registered hours of classes (class-hours and class-student-hours) (on separate y-axes)

It also displays the following static stats:

- Number of teachers registering in the last 10 minutes
- Number of teachers teaching a class
- Number of classes registered
- Number of classes approved
- Number of registered hours of classes (class-hours and class-student-hours)
- Number of teachers checked in today

Potential additions include:

- Separate graphs (or axes) for teacher numbers and class numbers (only necessary if numbers are different by multiple orders of magnitude)
- Number of sections registered (static or graph)
- Number of teachers that have filled out their availability (static or graph)
- Number of new vs. old teachers registered (graph?)
- ~~Number of registered hours of classes (graph) (class-hours or class-student-hours)~~
- Stats on number of classes with particular flags (static) (potentially with the ability to specify or select the flag(s) on the page)
- Stats on number of classes requesting particular resources (static)

All comments and suggestions welcome!


benjaminjkraft:
Haven't taken a look yet, but since I noticed -- this has a test failure beyond the known one, https://travis-ci.org/learning-unlimited/ESP-Website/jobs/251789443 tells you how to fix.

Also, including a screenshot in the pull request would be great!

esp/templates/program/modules/teacherbigboardmodule/teacherbigboard.html
```diff
@@ -0,0 +1,158 @@
+{% extends "elements/html" %}
+{% block fulltitle %}Teacher Registration Summary for {{program.niceName}}{% endblock %}
```
benjaminjkraft: copypasta

willgearty: Not sure I get the reference...are we not using blocks anymore or something?

willgearty:
There's not a lot of data (this is on an almost empty db), but here's an example of the current state:

image

Migrations are fixed, and I'll be looking into reducing the duplication at some point soon.

willgearty:
I think that should do it with merging the python and templates!

I'd really like to add the number of class-hours and/or number of class-student-hours, but I'm not really sure how to do that.
It'll also be a few orders of magnitude different from the other stats, so that could be a problem unless we use [multiple axes](https://www.highcharts.com/demo/combo-multi-axes).

willgearty:
Figured out a [messy] way to include class-hours.
In the current implementation, I've given class-student-hours a different y-axis (which only appears for the teacher big board), since they are usually ~1-2 orders of magnitude greater than the other stats.
I've also included static stats for class-hours and class-student-hours.
image
I feel like there's a way to get around semi-duplicating some of the code (`make_hours_data` is basically a modified version of `make_graph_data`+`chunk_data`), but I'm not sure how at this point.
I'm also slightly worried about how many lines there are on the graph for the teacherbigboard, but perhaps it will be better with actual data.

willgearty:
After looking at some real program data, I decided to add a third y-axis for class-hours.
image
I also moved the legend to the top-left, where it will hopefully be the least obtrusive.

esp/esp/program/modules/handlers/teacherbigboardmodule.py
```diff
@@ -0,0 +1,206 @@
...
+    def reg_classes(self, prog):
+        class_times = dict(ClassSubject.objects.filter(parent_program=prog
+        ).values_list('id').annotate(Min('timestamp')))
+        return sorted(class_times.itervalues())
+
+    @cache_function_for(105)
+    def app_classes(self, prog):
+        #all ClassSubjects that are approved (and have an approved section)
+        class_times = dict(ClassSubject.objects.filter(parent_program=prog, status__gt=0, sections__status__gt=0
+        ).values_list('id').annotate(Min('timestamp')))
```
benjaminjkraft: (if you do that you may need a `distinct()` here before the `values_list` due to the filter on `sections`

esp/esp/program/modules/handlers/teacherbigboardmodule.py
```diff
@@ -0,0 +1,206 @@
...
+        ).values_list('teachers').annotate(Min('timestamp')))
+        return sorted(teacher_times.itervalues())
+
+    @cache_function_for(105)
+    def get_hours(self, prog):
+        hours = ClassSubject.objects.filter(parent_program=prog
+        ).exclude(category__category__iexact="Lunch").values_list('id','duration','class_size_max'
+        ).annotate(Min('timestamp'))
+        sorted_hours = sorted(hours, key=operator.itemgetter(2))
+        return [hour[1:] for hour in sorted_hours]
```
benjaminjkraft: If you don't need `id`, no need to put it in the `values_list` solely in order to remove it!

esp/esp/program/modules/handlers/teacherbigboardmodule.py
```diff
@@ -0,0 +1,206 @@
...
+    @cache_function_for(105)
+    def static_hours(self, prog):
+        hours = ClassSubject.objects.filter(parent_program=prog
+        ).exclude(category__category__iexact="Lunch").values_list('id','duration','class_size_max'
+        )
+        hours = [[float(j[1])] + [float(j[1])*j[2]] for j in hours]
+        return [sum(j) for j in zip(*hours)]
+
+    @staticmethod
+    def make_hours_data(hourss, drop_beg = 0, drop_end = 0, cutoff = 0, delta=datetime.timedelta(0, 3600)):
```
benjaminjkraft: Hmm, why do you need a whole separate method here?  Can't you just pass a list of durations and a list of `class_size_max * duration`s to `make_graph_data`?  (In fact it seems like you already do much of that work in `static_hours`!  You could just return a line earlier and use the data one way for the graphs and the other way for the numbers.

esp/esp/program/modules/handlers/teacherbigboardmodule.py
```diff
@@ -0,0 +1,206 @@
...
+        ).annotate(Min('timestamp'))
+        sorted_hours = sorted(hours, key=operator.itemgetter(2))
+        return [hour[1:] for hour in sorted_hours]
+
+    @cache_function_for(105)
+    def static_hours(self, prog):
+        hours = ClassSubject.objects.filter(parent_program=prog
+        ).exclude(category__category__iexact="Lunch").values_list('id','duration','class_size_max'
+        )
+        hours = [[float(j[1])] + [float(j[1])*j[2]] for j in hours]
```
benjaminjkraft: This can just be `[[float(j[1]), ...] for j in hours]`, no need to add the two lists (since it's harder to read that way).

esp/esp/program/modules/handlers/teacherbigboardmodule.py
```diff
@@ -0,0 +1,206 @@
...
+        return ClassSubject.objects.filter(parent_program=prog).values_list('id').count()
+
+    @cache_function_for(105)
+    def num_class_app(self, prog):
+        return ClassSubject.objects.filter(parent_program=prog, status__gt=0, sections__status__gt=0).count()
+
+    @cache_function_for(105)
+    def reg_classes(self, prog):
+        class_times = dict(ClassSubject.objects.filter(parent_program=prog
+        ).values_list('id').annotate(Min('timestamp')))
```
willgearty: The way I understand it, if there are duplicates of whatever is in values_list, then only one of the duplicates will be returned.
So in this case, we want to get all classes by ID just in case there are two or more classes registered at the same time. (unrealistic, but still possible)

esp/esp/program/modules/handlers/teacherbigboardmodule.py
willgearty: The problem, at least that I saw, was that `make_graph_data` uses `chunk_times` which counts each class as 1, summing that for each hour, whereas the hours/student-hours for each class will be different. `make_hours_data` is essentially `make_graph_data` where `chunk_times` has been modified to account for that. I suppose a way around that would be to modify `chunk_times` and give it similar lists for the normal class data with 1's instead of hours or student-hours?

esp/esp/program/modules/handlers/teacherbigboardmodule.py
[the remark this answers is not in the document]
benjaminjkraft: Nope, it'll return all the dupes, unless you do `.distinct()` after the `values_list()`.  (You can always write a test to prove it ;-) .)

esp/esp/program/modules/handlers/teacherbigboardmodule.py
benjaminjkraft: Ah, I see.  yeah, I'd say do exactly that.

esp/esp/program/modules/handlers/teacherbigboardmodule.py
```diff
@@ -0,0 +1,163 @@
...
+        return ClassSubject.objects.filter(parent_program=prog, status__gt=0, sections__status__gt=0).count()
+
+    @cache_function_for(105)
+    def reg_classes(self, prog):
+        class_times = ClassSubject.objects.filter(parent_program=prog
+        ).values_list('timestamp', flat=True)
+        return sorted(class_times)
+
+    @cache_function_for(105)
+    def app_classes(self, prog):
```
benjaminjkraft: nit: can you call this `approved_classes`?  "app" can mean "approved" or "application".

esp/esp/program/modules/handlers/teacherbigboardmodule.py
```diff
@@ -0,0 +1,163 @@
...
+    def get_hours(self, prog):
+        hours = ClassSubject.objects.filter(parent_program=prog
+        ).exclude(category__category__iexact="Lunch").values_list('timestamp','duration','class_size_max')
+        sorted_hours = sorted(hours, key=operator.itemgetter(2))
+        class_hours = [(hour[1],hour[0]) for hour in sorted_hours]
+        student_hours = [(hour[1]*hour[2], hour[0]) for hour in sorted_hours]
+        return class_hours, student_hours
+
+    @cache_function_for(105)
+    def static_hours(self, prog):
```
benjaminjkraft: This does the same thing as `get_hours` that just extracts only some of the data, right?  Can it just be a wrapper around that that extracts the relevant data?

esp/templates/program/modules/bigboardmodule/bigboard.html
```diff
@@ -50,12 +52,29 @@
...
-          yAxis: {
+          yAxis: [{
             title: {text: null},
             min: 0,
             maxPadding: 0.01,
             endOnTick: false,
+          }
+          {% if hours_data %}
+          ,{
+            title: {text: 'Class-Hours'},
```
benjaminjkraft: Can these at least use `{{hours_data.0.description}}` or something like below?  It's a little weird to hardcode them here as the names of the second axis.  (It may also be better to call it `left_axis_data` and `right_axis_data` rather than `graph_data` and `hours_data`.)

esp/templates/program/modules/bigboardmodule/bigboard.html
```diff
@@ -68,6 +87,28 @@
                 data: {{series.data}},
               },
             {% endfor %}
+            {% if hours_data %}
+              {
```
benjaminjkraft: can you do a `for` here, so you don't have to write out the whole thing twice?

willgearty:
This is what the student big board looks like on my dev server:

image

But this is what it looks like on stanfordesp.org:

image

So it looks like it might be a problem with the theme I'm using (circles).

esp/esp/program/modules/handlers/teacherbigboardmodule.py
```diff
@@ -0,0 +1,163 @@
...
+        graph_data, start = BigBoardModule.make_graph_data(timess)
+
+        class_hours, student_hours = self.get_hours(prog)
+
+        hourss = [
+            ("number of registered class-hours", class_hours),
+            ("number of registered class-student-hours", student_hours),
+        ]
+
+        hours_data, _ = BigBoardModule.make_graph_data(hourss)
```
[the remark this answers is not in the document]
willgearty: Do you mean it *will* be the same? Because that's what I was assuming. I'm not sure how you'd have a teacher registered without any hours also registered, and those things should start at the same time.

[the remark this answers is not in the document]
benjaminjkraft: Ah, yeah, sorry, will be the same.

benjaminjkraft:
Great, thanks for the fixes!

PR #2397, opened 2017-07-09
Fix theme tests

### LINKED ISSUE #2388: esp.themes.tests.ThemesTest is failing on main

benjaminjkraft:
Adding the bigpicture theme caused a conflict between the tests for it and for
fruitsalad -- the two both set the same default files, which causes us to hit a
confirm page to clobber them in the test, and we don't know to click that, so
things don't work right.  This could also happen even before bigpicture, if
running tests locally on a dev server that had such a conflicting file.

One way to fix it would be to click past the confirm page. But that's a problem
since it may be clobbering real files, if someone is running locally!  So
instead, we now back up the offending files, and restore them at the end.  This
is super hacky -- we really shouldn't be modifying the user's files at all ever
if we can avoid it -- but it at least fixes the problem for now.

Fixes #2388.

### DIFF
--- a/esp/esp/themes/tests.py
+++ b/esp/esp/themes/tests.py
@@ -2,9 +2,10 @@
 Tests for the theme editor.
 """
 
-import re
 import os
 import random
+import re
+import shutil
 
 from django.conf import settings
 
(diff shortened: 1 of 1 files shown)


benjaminjkraft:
The diff here is kinda huge due to indent changes; here's `git diff -w`:
```diff
diff --git c/esp/esp/themes/tests.py w/esp/esp/themes/tests.py
index 3f09d8b..b77a93d 100755
--- c/esp/esp/themes/tests.py
+++ w/esp/esp/themes/tests.py
@@ -2,9 +2,10 @@
 Tests for the theme editor.
 """
 
-import re
 import os
 import random
+import re
+import shutil
 
 from django.conf import settings
 
@@ -77,12 +78,39 @@ def testSelector(self):
         response = self.client.get('/themes/select/')
         self.assertEqual(response.status_code, 200)
 
+        # Move the existing images dir out of the way, so we don't conflict
+        # with it (either what the user has there, or between themes).
+        # TODO(benkraft): Do the same for styles, although in practice
+        # conflicts there are less likely.
+        # TODO(benkraft): This is a super hacky way to do things!  Instead we
+        # should be doing all these tests in some sort of tmpdir to avoid
+        # touching anything of the user's.
+        images_dir = os.path.join(settings.MEDIA_ROOT, 'images', 'theme')
+        # Really we should use a tempdir, but on vagrant it may be on a
+        # different file system, which causes problems, so we do a hackier
+        # thing instead.
+        images_backup_dir = os.path.join(settings.MEDIA_ROOT, 'images',
+                                         'theme_backup_for_tests')
+        if os.path.exists(images_dir):
+            os.rename(images_dir, images_backup_dir)
+
+        try:
             #   Test each theme that is available.
             for theme_name in tc.get_theme_names():
                 #   Delete the theme_compiled.css file so we force a new one to be generated.
                 css_filename = os.path.join(settings.MEDIA_ROOT, 'styles', themes_settings.COMPILED_CSS_FILE)
                 if os.path.exists(css_filename):
                     os.remove(css_filename)
+                # Clobber any stray theme images dir, to avoid conflicts.
+                # Note that we've already backed up any one the user had
+                # created, above.
+                if os.path.exists(images_dir):
+                    shutil.rmtree(images_dir)
+
+                # Make sure there won't be any conflicts between this theme and
+                # existing files -- since they would cause harder-to-understand
+                # errors later on.
+                self.assertFalse(tc.check_local_modifications(theme_name))
 
                 #   POST to the theme selector with our choice of theme
                 response = self.client.post('/themes/select/', {'action': 'select', 'theme': theme_name})
@@ -126,6 +154,13 @@ def testSelector(self):
 
             self.client.logout()
 
+        finally:
+            # Restore the backed up images dir.
+            if os.path.exists(images_dir):
+                shutil.rmtree(images_dir)
+            if os.path.exists(images_backup_dir):
+                os.rename(images_backup_dir, images_dir)
+
     def testEditor(self):
         """ Check that the theme editor backend functionality is working. """
```

betaveros:
This looks reasonable to me. If I understand correctly that `check_local_modifications` is used as the condition elsewhere that triggers the problematic confirmation page, it looks like the same thing could also happen with JS files? If so, it might also be worth mentioning that as a TODO, but otherwise 👍

benjaminjkraft:
Ah, good point, I hadn't realized scripts were affected too.  Will add to the TODO.

Issue #2102, opened 2016-04-10
Userview behavior when profiles are missing for latest program


hwatheod:
If a teacher is teaching classes for a program but has no profile for the program, then the userview for the teacher will show blank (or maybe the latest program in which there was a profile) in the sections that depend on the program, e.g. the "teacher schedule (for program xxxx)" with a "print locally" section.  This means we can't print schedules for that teacher and program from the userview.

Admittedly, it takes a rare combination of events for this to happen. Stanford encountered it when we added an admin A who was otherwise not teaching any classes as a co-teacher onto a class which got cancelled but admin A was able to substitute-teach for.   I'm not sure exactly how it happened but probably this can happen if another admin edits A's availability and then morphs into the original teacher's account and adds admin A as a co-teacher to the class.  Thus admin A was added to a class in a program, but never created a profile for the program.  I was able to reproduce this scenario.

Stanford also encountered similar problems with students, which I don't have the details right now.

benjaminjkraft:
The best fix for this is probably #1880.

mgersh:
The more common reason why this happens (for both teachers and students) is when a user fills out their profile for the program and then later edits their profile from the myesp (non-program-specific) page.

benjaminjkraft:
Hopefully mitigated sufficeintly by #2399; if we still see this a lot feel free to reopen.

Issue #1880, opened 2015-11-18
Userview: "view profile for [dropdown] program"


luanthe:
There is no plan to make userview the One Page to Rule Them All, but if there were, it would be useful to be able to use the sidebar shortcuts for programs other than the one for which the user last updated their profile.

Possible implementation: a dropdown at the top of the page where you can select a program (that the user has a registration profile for), which will reload the page adding a GET parameter that tells it to use the student's last profile for that program.

betaveros:
Do we actually need/want the profile and the program shortcuts to come from the same program? After making the commit above referencing this issue I think those two things are pretty independent, so we could make the page display shortcuts for a different program but still display the same most–up-to-date profile information. Is there a common use case for viewing old profile information?

betaveros:
Also, if we want to fix the original rare case described in #2102 with this, we'll need to come up with a list of programs larger than the list of programs the user has a profile for. Might as well allow all programs?

PR #2399, opened 2017-07-11
Get last program with profile on userview

betaveros:
Also add some documentation, simplify the template, and add some if
guards and a link to the admin panel to search for registration
profiles. Fixes the most common cause of #2102 and possibly makes #1880
slightly easier since we now have the program and the profile
independently in the template.

### DIFF
--- a/esp/esp/program/models/__init__.py
+++ b/esp/esp/program/models/__init__.py
@@ -1301,12 +1301,19 @@ class RegistrationProfile(models.Model):
 
     @cache_function
     def getLastProfile(user):
+        """Return the user's most recent profile, or an empty profile.
+
+        Return the user's most recently written profile if one exists. If none
+        exist because the user has no profiles or is an AnonymousUser, create
+        and return (but don't save) an empty profile for the user.
+        """
         regProf = None
 
+        # check if this is an actual User, not an AnonymousUser
         if isinstance(user.id, (int, long)):
             try:
                 regProf = RegistrationProfile.objects.filter(user__exact=user).select_related().latest('last_ts')
-            except:
+            except RegistrationProfile.DoesNotExist:
                 pass
 
         if regProf != None:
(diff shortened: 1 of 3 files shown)


esp/esp/program/models/__init__.py
```diff
@@ -1301,8 +1301,16 @@ class Meta:
 
     @cache_function
     def getLastProfile(user):
+        """
```
benjaminjkraft: nit: docstring summary goes on the line with the open-quote

esp/esp/users/models/__init__.py
```diff
@@ -253,6 +253,11 @@ def getLastProfile(self):
         from esp.program.models import RegistrationProfile
         return RegistrationProfile.getLastProfile(self)
 
+    def get_last_program_with_profile(self):
+        # ditto
```
benjaminjkraft: nit: can you at least say what you're referencing?  it's easy for someone to add a new method in between and this will no longer make any sense.

esp/esp/program/models/__init__.py
```diff
@@ -1319,6 +1327,31 @@ def getLastProfile(user):
...
+        try:
+            return (
+                RegistrationProfile.objects
+                .filter(user__exact=user, program__isnull=False)
+                .select_related('program')
+                .latest('last_ts')
+                .program
+            )
+        except:
```
benjaminjkraft: this should catch a more specifc error (likely `RegistrationProfile.DoesNotExist`)

esp/esp/program/models/__init__.py
```diff
@@ -1319,6 +1327,31 @@ def getLastProfile(user):
...
+        program. If no such programs exist, return None. Used as an
+        approximation of the most recent program attended by the user, which is
+        also often a currently running program.
+        """
+        try:
+            return (
+                RegistrationProfile.objects
+                .filter(user__exact=user, program__isnull=False)
+                .select_related('program')
+                .latest('last_ts')
```
benjaminjkraft: Huh, TIL about `.latest`

Issue #1642, opened 2015-03-23
Easier way to check credit card payments


luac:
Requested at MIT Spark wrap-up

Issue #2394, opened 2017-07-09
Quickly list a user's transfers and financial aid


betaveros:
It would be helpful if, given a user, admins could quickly view all their transfers, so we can figure out whether they've paid online and whether they've received financial aid.

As a first approximation, links to the relevant Django administration pages with the appropriate filters might be enough, but there are two pages you want to look at, and the line items for transfers are a bit cryptic (if I'm understanding correctly, "Program admission" means you owe this much money for attending this program, and "Student payment" means you paid this online? (relevant: #2185)), so it would be nice to have our own page to list these things nicely for an admin.

Issue #1330, opened 2014-09-21
Defer creation of program charges until after enrollment


jmoldow:
There are transfers from receivable to the program account whenever you do anything with a program, even if you aren't in any classes. To make the account balances more accurate, we might want to only create new Transfers when a student enrolls in his first class and/or when a student has selected an extra cost and is already in at least one class, and might want to delete Transfers when a student drops all of their enrollments (and hasn't paid yet).

PR #2400, opened 2017-07-12
Add accounting module for showing user costs human-readably

betaveros:
The module gives a human-readable summary of costs, payments, financial
aid, and discounts involving a particular user on a particular program.

Part of #2394, but currently only accessible through URL-hacking; can be
easily added to userview after #2399 is resolved or similar.

(Note: Interacting with the IndividualAccountingController sometimes creates the default required transfers from a user to a program if it doesn't see any, just to be safe. I tried to guard against this because #1330 made it sound like it was undesirable, but poking around the database it appears that there are some programs for which every user already had a transfer created (at least, all out of a large random sample I tried), even if the user had graduated long before the program, so I'm not sure how worth it that was.)

### DIFF
--- a/esp/esp/accounting/controllers.py
+++ b/esp/esp/accounting/controllers.py
@@ -232,10 +232,34 @@ class ProgramAccountingController(BaseAccountingController):
         payments = Transfer.objects.filter(line_item=payment_li_type)
         return (payments.count(), payments.aggregate(total=Sum('amount_dec'))['total'])
 
+    def classify_transfer(self, transfer):
+        """Give a short human-readable description of a transfer.
+
+        Given a transfer, return a short human-readable phrase describing
+        the type of the transfer."""
+
+        line_item = transfer.line_item
+        if line_item == self.default_payments_lineitemtype():
+            return 'Payment'
+        elif line_item == self.default_finaid_lineitemtype():
+            return 'Financial aid'
+        elif line_item == self.default_siblingdiscount_lineitemtype():
+            return 'Sibling discount'
+        elif transfer.destination == self.default_program_account():
+            req_desc = "required" if line_item.required else "optional"
+            return "Cost ({})".format(req_desc)
+        else:
+            return 'Unrelated!?'
+
 class IndividualAccountingController(ProgramAccountingController):
     def __init__(self, program, user, *args, **kwargs):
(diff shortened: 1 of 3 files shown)


betaveros:
May fix #1642

esp/esp/accounting/controllers.py
```diff
@@ -232,11 +232,35 @@ def payments_summary(self):
...
+        line_item = transfer.line_item
+        if line_item == self.default_payments_lineitemtype():
+            return 'Payment'
+        elif line_item == self.default_finaid_lineitemtype():
+            return 'Financial aid'
+        elif line_item == self.default_siblingdiscount_lineitemtype():
+            return 'Sibling discount'
+        elif transfer.destination == self.default_program_account():
+            req_desc = "required" if line_item.required else "optional"
+            return "Cost ({})".format(req_desc)
```
benjaminjkraft: Does the line item not have a name that we could use here?  Or do I misunderstand the kind of description you want?

esp/esp/accounting/controllers.py
```diff
@@ -482,17 +506,19 @@ def grant_full_financial_aid(self):
     def revoke_financial_aid(self):
         FinancialAidGrant.objects.filter(request__user=self.user, request__program=self.program).delete()
 
-    def amount_requested(self):
-        self.ensure_required_transfers()
-        program_account = self.default_program_account()
+    def requested_transfers(self, ensure_required=True):
+        if ensure_required: self.ensure_required_transfers()
```
benjaminjkraft: nit: `if` body on a separate line

benjaminjkraft:
Oh, and probably a good idea to at least have a TODO to have a user search page as the `@main_call`; that way there will be something useful in the complete list of modules, even if many people end up using userview to get here.

esp/esp/accounting/controllers.py
betaveros: The entire line item is also printed in the table. This classification is supposed to supplement that, and to be short so that the columns align reasonably, because when I saw the the line item "Program admission for HSSP Summer 2016 ($40.00)" I didn't know if that meant whether the user needed to pay that or had paid that.

benjaminjkraft: Ah ok!

Issue #147, opened 2011-11-24
Text Message for Cancelled Classes


jmoldow:
The top of manageclass page recently got an addition of a set of boxes for canceling classes. You can type a message to students in the boxes, then click the "Cancel class" button, and all enrolled students would be e-mailed that message about their class cancellation. 

I suggest that we add a checkbox that says, "Check here to notify all students via text message". And perhaps an extra box, since the text message will probably need to be shorter than the e-mail. Either that, or the text message can be auto-generated (something like: "X1234 (hh:00-nn:00) has just been canceled. Come to class changes at any time to add a new class."), with the text configurable in a Tag. Or auto-populate the box with the default message, and allow someone to change it before sending.

This is related to Issue #29. However, I am suggesting being able to send text messages via the manageclass pages when a class is canceled, which is distinct from being able to send text messages via the Comm Panel at any time.

willgearty:
Fixed by #2404.

PR #2404, opened 2017-07-15
Adds option to text students when cancelling classes

willgearty:
As the title states, this is an attempt to add the option to text students when cancelling a class.
Unfortunately, at this time it's not working because it doesn't like importing a handler into a model.

### DIFF
--- a/esp/esp/program/models/class_.py
+++ b/esp/esp/program/models/class_.py
@@ -69,12 +69,11 @@ from esp.cal.models import Event
 from esp.dbmail.models import send_mail
 from esp.qsd.models import QuasiStaticData
 from esp.qsdmedia.models import Media
-from esp.users.models import ESPUser, Permission
+from esp.users.models import ESPUser, Permission, PersistentQueryFilter
 from esp.program.models import Program
 from esp.program.models import StudentRegistration, StudentSubjectInterest, RegistrationType
 from esp.program.models import ScheduleMap, ScheduleConstraint
 from esp.program.models import ArchiveClass
-from esp.program.modules.handlers.grouptextmodule import GroupTextModule
 from esp.resources.models         import Resource, ResourceRequest, ResourceAssignment, ResourceType
 from argcache                     import cache_function, wildcard
 from argcache.extras.derivedfield import DerivedField
@@ -963,6 +962,9 @@ class ClassSection(models.Model):
     enrolled_students = DerivedField(models.IntegerField, count_enrolled_students)(null=False, default=0)
 
     def cancel(self, email_students=True, include_lottery_students=False, text_students=False, explanation=None, unschedule=True):
+        # To avoid circular imports
+        from esp.program.modules.handlers.grouptextmodule import GroupTextModule
+
         if include_lottery_students:
             student_verbs = ['Enrolled', 'Interested', 'Priority/1']
         else:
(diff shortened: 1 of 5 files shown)


willgearty:
When this works, it will hopefully fix #147.

esp/esp/program/models/class_.py
```diff
@@ -1002,6 +1003,12 @@ def cancel(self, email_students=True, include_lottery_students=False, explanatio
                 else:
                     send_mail(ssi_email_title, msgtext, from_email, to_email)
 
+        if self.parent_program.hasModule('GroupTextModule') and text_students:
```
benjaminjkraft: Probably we should error or otherwise report back to the user if `text_students` is somehow `True`  but the module is not enabled!  (Alternately or in addition, we can just remove it from the form if the module isn't there?)

willgearty:
This works now, with the import now within the `cancel()` function.
The text_students form field is hidden if GroupTextModule is not enabled or if Twilio is not configured.
I was trying to implement some sort of error message, but it was tricky passing something from the cancel function back to the view and then to the template (or perhaps I just was doing something wrong).
However, since the field is now hidden, the only real condition that needs to be caught is if the class has no students in it. In this case, it skips the texting part within `cancel()` and continues with the rest of the function.

esp/esp/program/models/class_.py
```diff
@@ -1002,6 +1005,16 @@ def cancel(self, email_students=True, include_lottery_students=False, explanatio
...
 
+        if text_students and self.parent_program.hasModule('GroupTextModule') and GroupTextModule.is_configured():
+            if self.students(student_verbs).distinct().count() > 0:
+                msgtext = render_to_string('texts/class_cancellation.txt', context)
+                students_to_text = PersistentQueryFilter.create_from_Q(ESPUser, Q(id__in=[x.id for x in self.students(student_verbs)]))
+                send_log = GroupTextModule.sendMessages(students_to_text, msgtext)
+            else:
+                send_log = "No students to text"
+        else:
+            send_log = "GroupTextModule is not enabled or configured"
```
benjaminjkraft: Wait, does `send_log` go anywhere?  If you want, you can `logging.info` it, but you can also just not, if you're not going to display it back to the user, in which case you can just remove it!

esp/esp/program/models/class_.py
```diff
@@ -1002,6 +1005,11 @@ def cancel(self, email_students=True, email_teachers=True, include_lottery_stude
                 else:
                     send_mail(ssi_email_title, msgtext, from_email, to_email)
 
+        if text_students and self.parent_program.hasModule('GroupTextModule') and GroupTextModule.is_configured():
+            if self.students(student_verbs).distinct().count() > 0:
+                msgtext = render_to_string('texts/class_cancellation.txt', context)
+                students_to_text = PersistentQueryFilter.create_from_Q(ESPUser, Q(id__in=[x.id for x in self.students(student_verbs)]))
```
benjaminjkraft: I think you may have accidentally also the line that actually sends the messages!

willgearty: Whoops! Thanks for catching that!

esp/esp/program/models/class_.py
```diff
@@ -1002,6 +1005,12 @@ def cancel(self, email_students=True, email_teachers=True, include_lottery_stude
...
+                _ = GroupTextModule.sendMessages(students_to_text, msgtext)
```
benjaminjkraft: No need for `_ =`, you can just call the fn and not assign a variable.

benjaminjkraft:
Great, thanks!

Issue #439, opened 2012-11-05
Client-side JS grade-sorting?


shulinye:
tl;dr awesome HS student made a script that would filter by grade level.

Can we implement something similar that client-side hides classes that are out-of-grade-level?

Issue #65, opened 2011-11-23
Improve Student Catalog filtering mechanism


aseering:
Provide and improve sorting/filtering mechanism of student catalog.

This is only moderately easy, but it's not terrible, and it's a good valuable standalone project.

benjaminjkraft:
This issue is pretty vague, so I'm gonna say it's fixed by #2405.

PR #2405, opened 2017-07-16
Grade filter for student catalog

willgearty:
Adds a dropdown menu to select grade level, then filters catalog by selected grade level.

image


willgearty:
Fixes #439 and part of #65.

esp/esp/program/templatetags/intrange.py
```diff
@@ -0,0 +1,7 @@
+from django import template
+
+register = template.Library()
+
+@register.filter
+def intrange(min_val, max_val):
```
benjaminjkraft: Ew, this isn't really a filter... and having to do the calculation in the template is not great.  Is there a way to instead e.g. just have `class.grades` or something return the class's list of grades, and then we can iterate over that?

esp/templates/program/modules/studentclassregmodule/catalog.html
```diff
@@ -153,6 +167,19 @@ <h2 style="text-align: center;">
 hideClassesOutOfGradeRange(student_grade);
 </script>
 
+<script type="text/javascript">
+$j("#grade_filter").change(function(){
```
benjaminjkraft: The script right above this one seems to be doing approximately the same thing (except using the student's current grade).  Why do we need a new one? Is it just to allow looking at a different grade?  Can we remove the old one?  Can we make sure it still defaults to showing the current user's grade, if they are logged in?

esp/templates/program/modules/studentclassregmodule/catalog.html
```diff
@@ -153,6 +167,19 @@ <h2 style="text-align: center;">
...
+		var grade =  $j(this).find('option:selected').prop('value');
+    if(grade == "all"){
+        $j('.show_class').show();
+    }
+    else{
+    		$j('.show_class').hide();
```
benjaminjkraft: Can you fix the indentation here (not sure why it's wrong, perhaps a spaces/tabs issue?), and move `} else {` onto one line?  Having things inconsistently indented makes them really hard to read.

esp/templates/program/modules/studentclassregmodule/catalog.html
willgearty: So, I'm having a hard time figuring out where `student_grade` in the above script actually comes from?

willgearty: @benjaminjkraft ?

benjaminjkraft: Heh, I think it's a global thanks to the wonders of JavaScript.  From searching the codebase it looks like it's set here: https://github.com/learning-unlimited/ESP-Website/blob/9d40eedecac28ad895db5518f5b83afb68813a99/esp/templates/program/modules/studentclassregmodule/catalog.html#L21

willgearty:
How's this look, @benjaminjkraft ?

esp/templates/program/modules/studentclassregmodule/catalog.html
benjaminjkraft: Great, thanks for cleaning this up!  Happy to see a big pile of code go away :-) .

esp/templates/inclusion/program/class_catalog.html
```diff
@@ -1,7 +1,6 @@
 {% if show_class %}
   {% load class_render %}
-  {% load intrange %}
-  <div id="class_{{ class.id }}" class="{% for grade in class.grade_min|intrange:class.grade_max %}grade_{{ grade }} {% endfor %}show_class{% if errormsg %}class_error{% endif %}">
+  <div id="class_{{ class.id }}" class="{% for grade in class.grades %}grade_{{ grade }} {% endfor %}show_class{% if errormsg %}class_error{% endif %}">
```
benjaminjkraft: Great, much cleaner!

benjaminjkraft:
(I believe merging master would fix the "push" check; "pr" is what really matters anyway so I'm just gonna merge.)

PR #2408, opened 2017-07-17
Add class style to main catalog

### LINKED ISSUE #2407: Class styles only in lottery phase
The new class styles of #2302 should be included in the FCFS catalog and the main catalog.

willgearty:
Adds the class style implemented in #2302 to classes in the main catalog and FCFS phase.
(Only shown if a class has a class style, just like prereqs)
image

Fixes #2407.


benjaminjkraft:
Nice, thanks for the screenshot!

PR #2410, opened 2017-07-17
Send emails to teachers re: cancelled classes

### LINKED ISSUE #2360: Confirm class cancellation email to teachers
Clark Splash has requested a checkbox so that when an admin (or teacher) cancels a class, the website should email all teachers (including co-teachers) of the class. The logic is that if a teacher cancels their own class, the email can serve as a confirmation receipt, and if an admin cancels the class (if teacher emails in with an emergency or if too few students register) it can notify the teacher that the class has been cancelled. Ideally this checkbox would allow an optional message explaining why the class has been cancelled.

willgearty:
When a class/section is cancelled, the website now sends an email to each of the teachers of the class.
Fixes #2360.

### DIFF
--- a/esp/esp/program/models/class_.py
+++ b/esp/esp/program/models/class_.py
@@ -1018,7 +1018,7 @@ class ClassSection(models.Model):
         from_email = '%s at %s <%s>' % (self.parent_program.program_type, settings.INSTITUTION_NAME, self.parent_program.director_email)
         if email_ssis:
             email_content += '\n' + render_to_string('email/class_cancellation_body.txt', context)
-        teachers = self.get_teachers()
+        teachers = self.parent_class.get_teachers()
         for t in teachers:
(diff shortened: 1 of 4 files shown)


benjaminjkraft:
Hmm, I think I am in favor of having this be optional.  (I don't care what the default is.)  But if we're not going to make it optional, we definitely need to have text that says clearly that teachers will be emailed, especially since it's a behavior change!  We don't want an admin to be surprised that a teacher got an email (whereas it is generally fine to send admin-internal emails without warning).

willgearty:
How about making it optional, but make the option selected by default?

willgearty:
image

benjaminjkraft:
Fine with me!  In general I always hate to add more checkboxes and options, but this one seems both useful and simple enough to add.

PR #2411, opened 2017-07-17
Sort room schedules alphabetically

### LINKED ISSUE #1531: Room schedules should print alphabetically
Or in some other useful order, but I think alphabetically by room name would be both easy and good enough.  This would make sorting them to put them up a lot easier.

willgearty:
Fixes #1531.


benjaminjkraft:
Thanks!

PR #2414, opened 2017-07-21
Teacher reg option of grade ranges instead of mins and maxes

### LINKED ISSUE #51: Grade range options
Have teachers pick from a list of preconfigured grade ranges rather than pick a grade min and max

willgearty:
This adds the option to provide teachers with grade ranges to choose from when registering a class instead of requiring the teacher to specify a minimum and maximum grade level for their class.
Grade range options are set with the grade_ranges tag.
e.g. `[[7,9],[9,10],[9,12],[10,12],[11,12]]` gives five grade ranges: 7-9, 9-10, 9-12, 10-12, and 11-12)
When the form is submitted, the form parses the grade range and sets the minimum and maximum grade levels for the class.
Fixes #51.

image

### DIFF
--- a/esp/esp/program/modules/forms/teacherreg.py
+++ b/esp/esp/program/modules/forms/teacherreg.py
@@ -151,9 +151,9 @@ class TeacherClassRegForm(FormWithRequiredCss):
         # grade_min, grade_max: crmi.getClassGrades
         self.fields['grade_min'].choices = class_grades
         self.fields['grade_max'].choices = class_grades
-        grade_ranges = json.loads(Tag.getTag('grade_ranges'))
-        if grade_ranges:
-            self.fields['grade_range'].choices = [(range,`range[0]` + " - " + `range[1]`) for range in grade_ranges]
+        if Tag.getTag('grade_ranges'):
+            grade_ranges = json.loads(Tag.getTag('grade_ranges'))
+            self.fields['grade_range'].choices = [(range,str(range[0]) + " - " + str(range[1])) for range in grade_ranges]
             hide_field( self.fields['grade_min'] )
             hide_field( self.fields['grade_max'] )
         else:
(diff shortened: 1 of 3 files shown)


benjaminjkraft:
This functionality already exists -- see `StudentClassRegModuleInfo.allowable_class_size_ranges`!  I don't know if it works exactly the same way -- so pull requests to extend it are certainly welcome -- but we shouldn't have two ways of doing the same thing.

willgearty:
Wait, isn't that for class _size_ ranges? This is to specify particular _grade_ ranges that teachers can choose from instead of setting minimum and maximum grades.

esp/esp/program/modules/forms/teacherreg.py
```diff
@@ -277,6 +289,12 @@ def clean(self):
         if class_size_optimal == '':
             cleaned_data['class_size_optimal'] = None
 
+        # If using grade ranges instead of min and max, extract min and max from grade range.
+        if cleaned_data.get('grade_range'):
+            grade_range = json.loads(cleaned_data.get('grade_range'))
```
benjaminjkraft: nit: clearer to do `grade_min, grade_max = json.loads(...)` or even `cleaned_data['grade_min'], cleaned_data['grade_max'] = json.loads(...)`

esp/esp/program/modules/forms/teacherreg.py
```diff
@@ -66,6 +66,10 @@ class TeacherClassRegForm(FormWithRequiredCss):
     #     [["Lecture", "Lecture Style Class"], ["Seminar", "Seminar Style Class"]]
     style_choices = []
 
+    # To enable class styles, admins should set the Tag grade_ranges.
+    # e.g. [[7,9],[9,10],[9,12],[10,12],[11,12]] gives five grade ranges: 7-9, 9-10, 9-12, 10-12, and 11-12
+    grade_range_choices = []
```
benjaminjkraft: You can just inline this, and put the comment by the field.  (I guess you're following `style_choices` but I think that's a silly pattern.)

esp/esp/program/modules/forms/teacherreg.py
willgearty: So used to using R and always forget you can do this in python

benjaminjkraft:
Great!

PR #2416, opened 2017-08-06
ClassChangeController multi-request revamp

betaveros:
Let the class changes lottery accept multiple requests with priorities from each user, fix up the API and variable names, confirm email details with the user and allow them to customize it, and print more stats.

I haven't tested it yet because I don't have good data to test it on, so I don't think it should be merged yet (worst case we wait until Spring HSSP 2018 :P --- I am reasonably confident in my changes to the class change request form, which are completely independent), but it would be nice if somebody could look it over asynchronously.


willgearty:
I think merging main will fix the travis test error.

esp/esp/program/modules/handlers/classchangerequestmodule.py
```diff
@@ -46,7 +46,23 @@
...
+    return 1
+
+def extract_request_rank(self, user, section):
+    regs = StudentRegistration.valid_objects().filter(
+            user=user, section=section, relationship__name__startswith="Request")
+    if regs:
+        # there should be at most one; should we fail if there are more than one?
+        return extract_request_rank(regs[0])
+    else:
+        return 0
```
cjquines, line 65: this is... probably not right. but it's what's currently on `mit-prod`, so i'm just doing this for merge reasons

cjquines:
(when we merge this, it should be `Merge pull request`, and neither rebase nor squash)

willgearty:
We're mid testing and release right now, so I probably won't be able to take a look at this until November.

esp/esp/program/modules/handlers/classchangerequestmodule.py
willgearty: We should probably do what's right, not just whatever is on `mit-prod`

PR #2417, opened 2017-08-07
Specify program with dropdown in userview

### LINKED ISSUE #1880: Userview: "view profile for [dropdown] program"

betaveros:
Add a GET parameter to the userview that lets you specify a program to
list shortcuts for, and a <select> element on the page that chooses this
parameter. Also some other UI tweaks.

Fixes #1880.

### DIFF
--- a/esp/esp/program/views.py
+++ b/esp/esp/program/views.py
@@ -355,8 +355,16 @@ def userview(request):
     """ Render a template displaying all the information about the specified user """
     try:
         user = ESPUser.objects.get(username=request.GET['username'])
-    except:
+    except ESPUser.DoesNotExist:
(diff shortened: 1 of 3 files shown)


benjaminjkraft:
Looks good to me, thanks for fixing this and ~~helping make userview the One Page to Rule Them All~~ helping make userview a bit more useful.

PR #2418, opened 2017-08-17
Request class cancellation button for teachers

### LINKED ISSUE #577: Changing delete --> request delete for teachers
Currently, under some circumstances, teachers are not allowed to delete their class. Sometimes, they still want to. If they could, say, click a button that says "Request Delete", which emails us with a link that lets us delete, that could be useful.

willgearty:
Adds a button to the teacher registration page for each class enabling teachers to request a class cancellation.
Teachers provide a reason for cancellation and submit a form which sends an email to the admin email and sends emails to all teachers of the class (as confirmation).
The admin email has a link to the class manage page where the admin can then cancel the class, etc.

Fixes #577.

image

image

### DIFF
--- a/esp/templates/program/modules/teacherclassregmodule/listclasses.html
+++ b/esp/templates/program/modules/teacherclassregmodule/listclasses.html
@@ -1,5 +1,40 @@
+<style>
+.ui-dialog-titlebar-close {
+    padding:0 !important;
+}
+
+.ui-dialog-titlebar-close:after {
+    content: '';
+    width: 20px;
+    height: 20px;
+    display: inline-block;
+    background-image: url(/media/styles/jquery-ui/images/ui-icons_222222_256x240.png);
+    background-position: -96px -128px;
+    background-repeat: no-repeat;
+}
+</style>
+
 {% block xtrajs %}
 <script type="text/javascript">
+function requestCancel(cls) {
+    $j('#request_' + cls).dialog({
+        height: "auto",
+        width: 500,
+        resizable: false,
+        modal: true,
+        autoOpen: true,
+        closeText: "Close",
+        closeOnEscape: true,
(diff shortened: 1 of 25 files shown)


esp/templates/program/modules/teacherclassregmodule/listclasses.html
```diff
@@ -1,5 +1,40 @@
...
+    $j('#request_' + cls).dialog({
+        height: "auto",
+        width: 500,
+        resizable: false,
+        modal: true,
+        autoOpen: true,
+        closeText: "Close",
+        closeOnEscape: true,
+        title: "Class Cancellation Request",
+        open: function(event, ui) {
```
hwatheod: What does this open function do? It seems like other places in the code that pop-up dialog boxes don't have this parameter. Could you write a comment explaning how this case is different?

esp/templates/program/modules/teacherclassregmodule/listclasses.html
```diff
@@ -1,5 +1,40 @@
...
+        resizable: false,
+        modal: true,
+        autoOpen: true,
+        closeText: "Close",
+        closeOnEscape: true,
+        title: "Class Cancellation Request",
+        open: function(event, ui) { 
+            $j('.ui-widget-overlay').bind('click', function()
+            { 
+                $j(this).siblings('.ui-dialog').find('.ui-dialog-content').dialog('close');
```
hwatheod: What does this open function do? It seems like other places in the code that pop-up dialog boxes don't have this parameter. Could you write a comment explaning how this case is different?

willgearty: If I remember correctly, this allows for the user to escape the popup by clicking anywhere outside it (hence the `$j('.ui-widget-overlay')`.

Issue #181, opened 2011-11-25
Improve teacher availability form


pricem:
From Stanford:
- More of a design than functionality: time availability for teachers
  in a more calendar looking format rather than checkboxes

willgearty:
Fixed by #2419.

PR #2419, opened 2017-08-18
Style change for teacher availability form

### LINKED ISSUE #1258: Availability module involves checking too many boxes
It's kind of an annoying UI, and more prone to user error than something nicer. In my ideal world we'd have something that looks like whenisgood.

willgearty:
Moves to a more teacher-friendly availability form (similar to whenisgood) with the following features:
- Can change multiple blocks at once
- Prevents changing blocks that already have scheduled classes
- Can change entire blocks by clicking header
- No changes to backend

image

Current drawbacks:
- Does not combine multiple blocks on the same day (currently thinking about ways to fix this)
- Colors are fixed (this could be fixed by moving the styles to a styles file which can be altered by chapters)

Fixes #1258 and #181.

### DIFF
--- a/esp/templates/program/modules/availabilitymodule/availability_form.html
+++ b/esp/templates/program/modules/availabilitymodule/availability_form.html
@@ -8,6 +8,99 @@
 
 <link rel="stylesheet" href="/media/styles/forms.css" type="text/css" />
 
+<style>
+.wrap {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+
+.side {
+  width: 280px;
+}
+
+.summary {
+  display: block;
+  padding: 5px 5px;
+  font-size: 16px;
+  width: 100px;
+  margin: 5px 5px 5px 5px;
+  font-family: sans-serif;
+}
+
+.noselect {
+  -webkit-user-select: none;
+  -moz-user-select: none;
+  -ms-user-select: none;
+  user-select: none;
+}
+
+input,
+select {
+  padding: 2px;
+}
+
+td.group, table.group {
+  vertical-align: top !important;
+  border-spacing: 0px 2px !important;
+  padding: 0 !important;
+}
+
+#program_form table {
+  border-width: 0px;
+  border-collapse: separate;
+  border-spacing: 5px 1px;
+}
+
+#program_form th {
+  text-align: center;
+  font-weight: bold;
+  background: #EFEFEF;
+  padding: 3px;
+  height: 80px;
+  width: 75px;
+}
+
+.label {
+  text-align: right;
+  /* font-weight:bold; */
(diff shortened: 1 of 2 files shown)


esp/templates/program/modules/availabilitymodule/availability_form.html
```diff
@@ -54,31 +147,130 @@
...
+    }
+  }
+}
+
+//Sets colors of cells based on status of checkboxes upon loading page
+$j(document).ready(function() {
+  $j('#grid input:checked').each(function(i, e) {
+    if (this.disabled == true) {
+      document.getElementsByName($j(this).attr('value'))[0].className = "teaching";
+      document.getElementsByName($j(this).attr('value'))[0].title = "You are currently scheduled to teach during this timeblock";
```
hwatheod: "Timeblock" is jargon specific to our codebase and shouldn't be used in user-facing text. Replace it with "time".

esp/templates/program/modules/availabilitymodule/availability_form.html
```diff
@@ -54,31 +147,130 @@
...
+var down = false;
+
+//Records mousedown
+function md(td) {
+  down = true;
+  setting = !isSet(td);
+  mo(td);
+}
+
+//Enables hightlighting multiple cells
```
hwatheod: "hightlighting" -> "highlighting"

esp/templates/program/modules/availabilitymodule/availability_form.html
```diff
@@ -54,31 +147,130 @@
...
+  if (td.className != "teaching") {
+  	td.className = "canDo";
+  	checkbox.checked = true;
+  }
+}
+
+//Deactivate checkbox
+function off(td) {
+  var checkbox = $j('input[value='+td.getAttribute("name")+']')[0]
+  if (!checkbox.disabled) {
```
hwatheod: Since the on() function uses td.className in its condition (line 229), the code is clearer to read if this off() function also does. Can you change the condition to 

if (td.className == "canDo") {

hwatheod:
If you schedule one or more classes for a teacher, and then the teacher goes to the availability page and unsets all timeslots other than the times where they're scheduled to teach, the submission is refused with the error: "You submitted a schedule with no available times."   This is possibly a previously existing bug but it isn't the correct behavior.

willgearty:
I just tried on our website, and this does seem to be a pre-existing bug.
I'll look into it when I make the other changes for this.

willgearty:
That should do it (the view wasn't including the teaching times when calculating if the availability was empty).

benjaminjkraft:
Wow!  I'm late to the party but this is super cool, thanks.

PR #2421, opened 2017-08-19
Simplify coteacher interface

### LINKED ISSUE #729: Coteacher adding is confusing
You have to type and autocomplete the teacher's name, click "add teacher", _then_ click "save and continue".  This is confusing; some people try to just type in the name and directly click save, which won't in fact save it.  A JS warning would be useful.

willgearty:
Removes need to "Save and continue" after deleting or adding teachers.
Now, adding or deleting a teacher is saved automatically.
This only causes one downside: an email is sent every time a teacher is added or deleted.
Fixes #729.

### DIFF
--- a/esp/templates/program/modules/teacherclassregmodule/coteachers.html
+++ b/esp/templates/program/modules/teacherclassregmodule/coteachers.html
@@ -71,7 +71,7 @@ You currently have no coteachers for this class.
 <form action="{{request.path}}" method="post" name="addteacher" onsubmit="cleanTeacherSubmit($j(this))">
 <table align="center" width="500">
     <tr>
-        <th colspan="2">Add A Coteacher</th>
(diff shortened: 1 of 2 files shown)


esp/templates/program/modules/teacherclassregmodule/coteachers.html
```diff
@@ -74,7 +71,7 @@
 <form action="{{request.path}}" method="post" name="addteacher" onsubmit="cleanTeacherSubmit($j(this))">
 <table align="center" width="500">
     <tr>
-        <th colspan="2">Add More Coteachers</th>
+        <th colspan="2">Add A Coteacher</th>
```
hwatheod: Maybe it's just me, but this capitalization looks really weird (especially capitalizing "A").  I recommend to capitalize the first word only.

Issue #1166, opened 2014-03-21
TwoPhase Lottery Registration Should Have Confirm Button


jmoldow:
From Stanford: TwoPhase lottery registration submits preferences automatically, but this confuses students because they don't know that their preferences have been saved. There should be a "send me an email confirmation of my preferences" button, which tells them that their preferences have indeed been saved, shows their selected classes, and e-mails them. Or a view and print button that brings up a page displaying the exact preferences that we would see.

willgearty:
The red banner at the top of this page has made this perhaps more confusing:
![image.png](https://user-images.githubusercontent.com/7232514/236098821-9b40f95d-84de-4891-bce0-e063dddc4581.png)

Dhirenderchoudhary:
@willgearty love to work on this can i get this assigned ?

willgearty:
@Dhirenderchoudhary Thanks for your interest in contributing to our project! I've assigned you to this issue. Please make sure to check out our [instructions for getting started](https://github.com/learning-unlimited/ESP-Website/wiki#getting-started) and follow our [contribution rules](https://github.com/learning-unlimited/ESP-Website/wiki#rules). Let me know if you have any questions.

PR #2425, opened 2017-08-25
Progress bars for starred classes

### LINKED ISSUE #382: Progress bar per-timeblock in lottery interface
Have a progress bar in each timeslot in the lottery interface that shows how many interested classes a student has selected up to a maximum of n.  This will hopefully discourage students from listing only one or two classes and then being disappointed when they don't get many classes.

### LINKED ISSUE #1437: Show a "you're done!" in the two-phase lottery
There's no indication that you're done, and your submissions have been saved.  A notification of some sort might be useful, although when to put it could be a difficult question.

willgearty:
Adds a progress bar for each timeslot in two phase registration that reports the number of starred classes for that time slot out of the desired number of starred classes (use tag `num_stars`, default = 10).

Fixes #382.
Also maybe fixes #1437, although I'm in favor of also implementing #1166.

image

### DIFF
--- a/esp/esp/program/modules/handlers/studentregtwophase.py
+++ b/esp/esp/program/modules/handlers/studentregtwophase.py
@@ -46,6 +46,7 @@ from esp.middleware.threadlocalrequest import get_current_request
 from esp.program.models import ClassCategories, ClassSection, ClassSubject, RegistrationType, StudentRegistration, StudentSubjectInterest
 from esp.program.modules.base import ProgramModuleObj, main_call, aux_call, meets_deadline, needs_student, meets_grade, meets_cap, no_auth
 from esp.users.models import Record, ESPUser
+from esp.tagdict.models import Tag
 from esp.utils.web import render_to_response
 from esp.utils.query_utils import nest_Q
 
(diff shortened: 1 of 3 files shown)


willgearty:
Also, does anyone know if there is a way to change how many breaks the progress bar has?
Ideally, you would want it to have the same number of segments as the number of classes you want a student to star, otherwise it seems the default is 8 segments/7 breaks.

willgearty:
The bootstrap progress bar seems like a reasonable alternative for now. I'll implement that as soon as I get a chance.

And you're totally right, the priority classes seem out of place now. A second progress bar would be nice, although would we still include the "Priority/1", etc. text beneath it?

You're last idea (in parens) might be nice. Basically have a list of n items (where n is the number of priority classes for each slot) and then fill in the class if one is selected (with a little checkmark next to it) or put an alert symbol if a class has yet to be selected?

benjaminjkraft:
> Basically have a list of n items [...]

Yeah, that's the idea.  That way it's clear what they need to do -- the missing priority will be clearly called out.

willgearty:
Thanks for the great suggestions; here's what it looks like now:

image

My last thought: is 10 too many? Especially for small programs? Even though there's a tag to change it, I feel like maybe it's a little high for the default?

willgearty:
@benjaminjkraft Thoughts on this?

PR #2426, opened 2017-08-26
Catalog preview link more noticeable

### LINKED ISSUE #1626: Make catalog preview link more obvious in teacherreg
There are a bunch of buttons for other things, but it's a tiny blue-underlined "»", and it's not obvious that that's what it does.  I knew it existed, was looking for it, and still had trouble finding it.

willgearty:
Fixes #1626.

image


benjaminjkraft:
Good idea!

PR #2429, opened 2017-09-01
Small changes to text message profile form field

willgearty:
This changes the form field from a boolean field to a radio select field, allowing to have both `True` and `False` options, with one of the two options required to submit the form.
This fixes the potential problem of students submitting the form without reading/thinking about the field in the form (since the field was previously not required).
This also removes the field from the form if you are not a student or if it is not enabled by the `text_messages_to_students` tag.
Finally, I also added a little CSS to remove the bullet points that would otherwise be next to the radio buttons.

image

N.B. Since the option values are converted to strings, it required some coercing to save the value as a boolean when the form is submitted and some setup in the `__init__` of the form in case the user previously selected "No" (otherwise, we could get some very grumpy students who complain about having to select "No" over and over every time they have to update their profile).

### DIFF
--- a/esp/esp/users/forms/user_profile.py
+++ b/esp/esp/users/forms/user_profile.py
@@ -61,8 +61,6 @@ class UserContactForm(FormUnrestrictedOtherUser, FormWithTagInitialValues):
             del self.fields['phone_day']
         if not Tag.getBooleanTag('text_messages_to_students') or not self.user.isStudent():
             del self.fields['receive_txt_message']
-        if self.user.getLastProfile().contact_user and not self.user.getLastProfile().contact_user.receive_txt_message:
-            self.fields['receive_txt_message'].initial = 'False'
 
     def clean(self):
         super(UserContactForm, self).clean()
--- a/esp/esp/users/forms/user_profile.py
+++ b/esp/esp/users/forms/user_profile.py
@@ -61,7 +61,7 @@ class UserContactForm(FormUnrestrictedOtherUser, FormWithTagInitialValues):
             del self.fields['phone_day']
         if not Tag.getBooleanTag('text_messages_to_students') or not self.user.isStudent():
             del self.fields['receive_txt_message']
-        if self.user.getLastProfile() and not self.user.getLastProfile().contact_user.receive_txt_message:
+        if self.user.getLastProfile().contact_user and not self.user.getLastProfile().contact_user.receive_txt_message:
             self.fields['receive_txt_message'].initial = 'False'
 
     def clean(self):
--- a/esp/esp/users/forms/user_profile.py
+++ b/esp/esp/users/forms/user_profile.py
@@ -48,7 +48,7 @@ class UserContactForm(FormUnrestrictedOtherUser, FormWithTagInitialValues):
     e_mail = forms.EmailField()
     phone_day = USPhoneNumberField(required=False)
     phone_cell = USPhoneNumberField(required=False)
-    receive_txt_message = forms.BooleanField(required=False)
+    receive_txt_message = forms.TypedChoiceField(coerce=lambda x: x =='True', choices=((True, 'Yes'),(False, 'No')), widget=forms.RadioSelect)
     address_street = StrippedCharField(length=40, max_length=100)
     address_city = StrippedCharField(length=20, max_length=50)
     address_state = forms.ChoiceField(choices=zip(_states,_states), widget=forms.Select(attrs={'class': 'input-mini'}))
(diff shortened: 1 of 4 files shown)


esp/esp/users/forms/user_profile.py
```diff
@@ -59,8 +59,10 @@ def __init__(self, *args, **kwargs):
         super(UserContactForm, self).__init__(*args, **kwargs)
         if not Tag.getBooleanTag('request_student_phonenum', default=True):
             del self.fields['phone_day']
-        if not Tag.getBooleanTag('text_messages_to_students'):
+        if not Tag.getBooleanTag('text_messages_to_students') or not self.user.isStudent():
             del self.fields['receive_txt_message']
+        if self.user.getLastProfile().contact_user and not self.user.getLastProfile().contact_user.receive_txt_message:
```
benjaminjkraft: Hmm, I'm a little confused what you're trying to do here.  Don't we initialize the form to the values from the user's profile, if they have one?  Is this just something to do with TypedChoiceField being weird, and if so can we just fix up the existing `.initial` value rather than re-feching it from the old profile?

benjaminjkraft:
Ah, I see the problem!  I'm not totally sure if all the details are right here.  Also, I think you'll need to do a merge with main here.

esp/esp/users/forms/user_profile.py
willgearty: Yeah, it doesn't play very well with the boolean model. I'll look into fixing up the .initial value.

willgearty: So, after some digging, it looks like when the TypedChoiceField reads in the values from the user's profile, it reads it as either `True` or `None` (even when the value is actually `False`).
Therefore, this seems like the only way to initialize it correctly. I can probably clean it up a little bit by using `self.user.getLastProfile().text_reminder`?
Unless you have any other ideas?

willgearty: Or maybe I'm misunderstanding "fix up the existing `.initial` value"

willgearty: To further explain, there's 3 different cases we need to distinguish:

-None (there's no previous profile data)
-False (the user previously selected "No")
-True (the user previously selected "Yes")

For some reason (which I don't really understand), the form reads in the False case as `None`, which means we can't distinguish between the first two cases without any other information (like checking the last profile which is what I'm currently doing).

benjaminjkraft: Interesting.  I still don't think using `getLastProfile()` here is right; we should be using whatever data is passed into the form, which often will be from `getLastProfile()` but may be [`getLastForProgram()`](https://github.com/learning-unlimited/ESP-Website/blob/benkraft.new-site/esp/esp/web/views/myesp.py#L204) or some such.  Probably best is to use the initial data that has been passed into the form -- if `self.fields['receive_txt_message'].initial` doesn't have the right value, we should still be able to grab it from `kwargs.get('initial', {}).get('receive_txt_message')`.  Or we could have the caller (e.g. where I linked above) do the munging, before the form gets to it, although that's a bit uglier.  But I think leaving it as-is is going to be the source of some nasty edge case bugs when the profile in question isn't `getLastProfile()`.

benjaminjkraft: Ah, missed your two later comments.  By "fix up the existing `.initial` value", my hope is that even if the form doesn't display things right, we can still distinguish those three cases by looking at `self.fields['receive_txt_message'].initial` and fixing it.  But if not, i think the initial arg should also have the right data.

willgearty: Ah, I wasn't aware of kwargs.get('initial', {}), and I bet that will work. I'll look into that ASAP.

willgearty: So...that didn't actually work. It's still giving me `None` or `True` from the kwargs. Still investigating...

willgearty: The model is correct, because it's stored as `True`/`False` if the user has filled out the form and `None` if the user has never filled out a profile.
So somewhere the `False` is getting converted to `None` as it's passed to the form?

willgearty: Ahah, it appears that the receive_txt_message argument is only passed to kwargs when it is `True`. It isn't passed when it is `False`, which seems weird.
Hence why `kwargs.get('initial', {}).get('receive_txt_message')` returns `None`.

willgearty: After much digging, it looks like the [method](https://github.com/learning-unlimited/ESP-Website/blob/main/esp/esp/users/models/__init__.py#L1794) that supplies the initial values for the form skips any values that are `False`.
I assume it was intended to remove any missing values?

benjaminjkraft: Huh, interesting.  Yeah, I'd guess so.  Probably it should just be checking for `None` and `''`, but not for `False`, so you could explicitly check for those.  Just make sure you test a few different configurations of the form to make sure it doesn't break anything else.

willgearty:
Okay, it should be fixed. I'll do some checks when I get a chance.

willgearty:
Everything looks good to me. I've tried the form as a teacher, guardian, volunteer, and student, and the form hasn't broken and everything seems to populate properly.

esp/templates/users/profiles/usercontact.html
```diff
@@ -1,5 +1,11 @@
 <!-- User ContactInfo -->
 
+<style>
+#id_receive_txt_message {
```
benjaminjkraft: Can you put this in a proper css file?  I think forms.css is probably already included in the relevant pages.

benjaminjkraft:
Great, thanks!

PR #2432, opened 2017-09-03
Front end for resource type choices

### LINKED ISSUE #372: Allow editing resource type choices
We currently link to the admin interface for this, but ideally the ability to enter choices for each resource type would reside on the ResourceTypeForm used by the ResourceModule.

willgearty:
Fixes #372.

image


esp/esp/program/modules/forms/resources.py
```diff
@@ -61,9 +61,16 @@ def save_restype(self, program, res_type):
             res_type.program = program
         if self.cleaned_data['priority']:
             res_type.priority_default = self.cleaned_data['priority']
+        if choices and filter(None, choices):
+            res_type.choices = filter(None, choices)
+        else:
+            res_type.attributes_pickled = ResourceType._meta.get_field('attributes_pickled').default
```
benjaminjkraft: Do you need this?  I would think it would set the default automatically here.  (Also, if you do, may want to comment that setting `choices` and `attributes_pickled` is equivalent.)

hwatheod:
A bug I noticed:

If I create a resource, save it (by clicking "update resource"), then it shows up on the list of resource types underneath ("resource types for _____").  Then I click "edit" for that resource, check the "global" box (not sure what that checkbox means), and click "update resource" again, the resource I just added disappears from the list of resource types shown on the page.  It's not really deleted though; it still shows up in the admin interface.

esp/esp/program/modules/forms/resources.py
willgearty: Hmm...you may be correct. I'll look into it.

willgearty: After some testing, I recalled that this is in case you remove all of the choices from a pre-existing resource type. I've added a comment explaining that and your note as well.

willgearty:
@hwatheod Global refers to whether a resource type is associated with a particular program (not global) or not (global). The behavior that you are describing appears to be the intended behavior of the original code, given that the [default behavior](https://github.com/learning-unlimited/ESP-Website/blob/main/esp/esp/program/models/__init__.py#L864) for `getResourceTypes()` is to exclude global resources. So, when you set the resource type to global, it was no longer included in the resource types available in the form. I've modified the code such that it follows the logic of the teacher class reg module, in that it will show global resource types if the `allow_global_restypes` tag is set to `True`. Otherwise, the form will only show resources associated with the current program. Oddly, the code was set up to handle global resource types and the template already had logic to identify global resource types, so perhaps the `getResourceTypes()` default was changed at some point or something else changed?

willgearty:
I think this has been tested pretty exhaustively by me and @hwatheod, so it could theoretically make it into SR10. @benjaminjkraft, what do you think?

benjaminjkraft:
Hmm, since the resource system is pretty complicated, I'd kinda prefer to wait until we've had this on Stanford or MIT's site for a while before we push it to chapters.  I'm happy to push it to Stanford soon, though.

willgearty:
Certainly happy to test it on the Stanford site!

willgearty:
Now that we've set up a separate branch for SR10, I think this should be good to merge to main? @benjaminjkraft

benjaminjkraft:
Yep!  You may want to create a separate issue about the global resource types thing, it seems like a less than ideal UI.

benjaminjkraft:
Oh, and while it's fresh, could you go and add a line in the docs to #2540?

PR #2433, opened 2017-09-04
Grade range fix for open class reg

willgearty:
I forgot to test open class registration with the new grade ranges option (#2414).
Now it works.


hwatheod:
Looks good to me.

PR #2438, opened 2017-09-09
Link to accounting page from userview

### LINKED ISSUE #2394: Quickly list a user's transfers and financial aid

betaveros:
Fixes #2394.

### DIFF
--- a/esp/templates/users/userview.html
+++ b/esp/templates/users/userview.html
@@ -44,6 +44,9 @@
             Check availability for {{program.niceName}}</a>
     {% endif %}
     {% if profile.student_info %}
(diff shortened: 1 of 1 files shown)


jerrywu64:
lgtm

PR #2440, opened 2017-09-13
teacherreg: add hours reg'ed count

jakob223:
Add a feature where when you're registering classes it tells you something like:

"You've registered for 12 of the 14 hours of class you're available in this program."


benjaminjkraft:
Looks like tests are still failing -- I think `duration` is a decimal so you may need to cast to a float or something.

PR #2450, opened 2017-10-07
Upgrade jQuery UI to v1.12.1 and fix CSS

betaveros:
This jQuery UI version fixes a tooltip positioning bug that is annoying during AJAX scheduling.

Also fix the commpanel tab bar.


benjaminjkraft:
Great, thanks -- and sorry for the delay in reviewing?  Can you merge main to see if that fixes the tests?

PR #2451, opened 2017-10-07
Filter random classes by a tag

betaveros:
Add a JSON tag in which we can specify certain class names and program
name substrings to exclude from random class generation.


betaveros:
Resolves certain shortcomings of #2321

PR #2456, opened 2017-10-11
Remove grade range from walkins form

### LINKED ISSUE #2454: Editing open classes can cause grade range error
Open classes are automatically given the minimum and maximum grades of the program.
Therefore, _if this grade range is not an option for the grade ranges_ (#2414) in the class registration form, this will cause an error.
Since the min and max grades are only applied to the open class when it is created, this is only a problem if the teacher tries to edit the open class (or if they are copying an open class from a previous program).
This same error can also occur if the grade min/max of a normal class are changed through the manage page and the teacher then tries to edit the class.

image

willgearty:
Fixes #2454.


hwatheod:
Failing tests were because of Travis issue fixed by #2459.

PR #2458, opened 2017-10-11
Request Cancellation Button Deadline and Logic

### LINKED ISSUE #2447: Make "Cancel Class" Button Smaller/Disappear
The new "Cancel Class" button (#2418) is cool, but is sadly being overused by MIT Splash teachers :( Is there anyway to make the button smaller so it doesn't overshadow all of the other buttons on the page? Or can there be a deadline associated with it so that it disappears once their classes have been scheduled and they're committed to teaching?

@willgearty

willgearty:
This was originally to fix #2447, but then I realized a larger problem.
It appears there was once a working "Delete" button for classes but it doesn't seem to work anymore.
`<input class="button" type="submit" value="Delete" {% if cls.num_students or not module.deadline_met %}disabled {% endif %} />`
Since this was within the `class_teacher_list_row` inclusion template, rather than the `list_classes` template, the `module` variable was not passed, and so the logic always disabled the button.

Perhaps this originally worked for some reason or I'm interpreting all of this incorrectly. (N.B. Stanford uses a pretty different template for `list_classes.html`, so some of my knowledge/understanding may be a little off.)

Either way, I had originally kept this logic for #2418, thinking that whenever the Delete button was disabled, I would hide it and show the Request Cancellation button instead.
Of course, since the Delete button was always disabled, this new logic made it so that the Request Cancellation button was always shown instead of the delete button.

Anywho, as I was trying to implement a deadline for the Request Cancellation button as per #2447, I realized all of this, so I implemented new logic, thinking that people might want the delete button to actually work, but this probably doesn't actually fix #2447.
The current logic is this:

- If a class has not been scheduled and has no students enrolled, the Delete button is shown.
- If a class is scheduled or has students enrolled AND the deadline for requesting a cancellation is still open, the Request Cancellation button is shown.
- If a class is scheduled or has students enrolled AND the deadline is closed, neither button is shown.

The deadline is included within the normal Teacher deadlines, so it will always be open if "All teacher deadlines" is open. However, once that deadline is closed, a chapter can open "Classes/CancelReq" as  a separate deadline if they'd like to allow cancellation requests after teacher registration closes.

@mariolsen0112358, does this fix #2447? Would it be better if there was no delete button and the request cancellation is shown until reg closes, at which point it can be controlled by a deadline? Can you envision any reason why you would want to make the Cancellation Request button disappear _before_ teacher reg closes?

### DIFF
--- a/esp/esp/users/models/__init__.py
+++ b/esp/esp/users/models/__init__.py
@@ -2279,7 +2279,7 @@ class Permission(ExpirableModel):
             ("Teacher/Classes/All", "Classes/All"),
             ("Teacher/Classes/View", "Classes/View"),
             ("Teacher/Classes/Edit", "Classes/Edit"),
-            ("Teacher/Classes/CancelReq", "Classes/CancelReq"),
+            ("Teacher/Classes/CancelReq", "Request class cancellation"),
             ("Teacher/Classes/Create", "Create classes of all types"),
             ("Teacher/Classes/Create/Class", "Create standard classes"),
             ("Teacher/Classes/Create/OpenClass", "Create open classes"),
--- a/esp/esp/users/models/__init__.py
+++ b/esp/esp/users/models/__init__.py
@@ -2279,6 +2279,7 @@ class Permission(ExpirableModel):
             ("Teacher/Classes/All", "Classes/All"),
             ("Teacher/Classes/View", "Classes/View"),
             ("Teacher/Classes/Edit", "Classes/Edit"),
+            ("Teacher/Classes/CancelReq", "Classes/CancelReq"),
             ("Teacher/Classes/Create", "Create classes of all types"),
             ("Teacher/Classes/Create/Class", "Create standard classes"),
             ("Teacher/Classes/Create/OpenClass", "Create open classes"),
--- a/esp/templates/inclusion/program/class_teacher_list_row.html
+++ b/esp/templates/inclusion/program/class_teacher_list_row.html
@@ -1,5 +1,33 @@
 {# expects program, class, can_req_cancel, and crmi (ClassRegModuleInfo) #}
 
+<style>
+.clsmiddle {
+    vertical-align: middle;
+}
+
+.cancelbutton {
+    color: #000000;
+    font-family: Arial,helvetica,sans-serif;
+    font-size: 11px;
+    font-weight: bold;
+    border: #cccccc 1px solid;
+    background-color: #cccccc;
+    cursor: default;
+    padding: 2px 5px 2px 5px;
+    text-decoration: none;
+    margin: 0px;
+    line-height: normal;
+    cursor: pointer;
+    box-shadow: none;
+    transition: border linear 0.2s, box-shadow linear 0.2s;
+    border-radius: 0;
+}
+
+.teachbutton {
+    margin: 0 !important;
+}
+</style>
+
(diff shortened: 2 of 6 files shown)


mariolsen0112358:
Thanks for figuring this out! To answer your questions in order:

I think this definitely satisfies the deadline request in #2447 . I'm not sure if it also satisfies the other request (which I didn't emphasize as much) which was to make the button smaller/more of a size of the other buttons like "Modify Co-Teachers" or the "Delete" button. We felt like the relative size of the button was maybe giving teachers ideas by implying that cancelling their class is easier than say figuring out how to add co-teachers or edit the class description.

I think it would be better if there was no delete button: MIT ESP likes "request cancellation" more than "delete" because if they have to go through us, we can possibly talk them out of cancelling if it's due to a misunderstanding (for ex. they want to delete because they think someone else is already teaching their subject so they shouldn't). 

I think it'd be useful to have Cancellation Request disappear before teacher reg closes because we sometimes leave teacher reg open through scheduling just to make it easier on late registrants/teachers updating things, but it's annoying when teachers are cancelling while we're scheduling. So that's on us and our usage more than the feature itself, and if it's easier to simply have it come and go with teacher reg deadlines rather than add yet another deadline, we can live with it.

willgearty:
Great, thanks for the feedback @mariolsen0112358 !
I got rid of the delete button entirely and now the request cancellation button will appear whenever the CancelReq deadline is open.

Here's how it looks on my dev server:
image

mariolsen0112358:
Thanks for fixing it! Looks great to me :)

esp/templates/inclusion/program/class_teacher_list_row.html
```diff
@@ -1,4 +1,32 @@
-{# expects program, class, and crmi (ClassRegModuleInfo) #}
+{# expects program, class, can_req_cancel, and crmi (ClassRegModuleInfo) #}
+
+<style>
```
benjaminjkraft: This should really go in a proper CSS file, especially since this template gets rendered many times so putting it here means we're adding a `<style>` for every class.  Not sure offhand what file is best but I'm guessing the page already has something like `esp/public/media/styles/catalog.css` and you can just add it there.

esp/esp/users/models/__init__.py
```diff
@@ -2279,6 +2279,7 @@ class Permission(ExpirableModel):
             ("Teacher/Classes/All", "Classes/All"),
             ("Teacher/Classes/View", "Classes/View"),
             ("Teacher/Classes/Edit", "Classes/Edit"),
+            ("Teacher/Classes/CancelReq", "Classes/CancelReq"),
```
benjaminjkraft: The second one should be human-readable, e.g. "Request class cancellation".  Not sure why the few before it aren't.

esp/templates/inclusion/program/class_teacher_list_row.html
```diff
@@ -1,4 +1,32 @@
...
+    text-decoration: none;
+    margin: 0px;
+    line-height: normal;
+    cursor: pointer;
+    box-shadow: none;
+    transition: border linear 0.2s, box-shadow linear 0.2s;
+    border-radius: 0;
+}
+
+.teachbutton {
```
benjaminjkraft: What's this change?

willgearty: For some reason, the buttons had asymmetric margins, which made them not centered vertically.

PR #2459, opened 2017-10-13
don't upgrade libpq-dev because version 10 breaks our Travis tests, apparently

hwatheod:
All tests have been failing at the install step for about a week or so, and the culprit appears to be upgrading libpq-dev to version 10.  

Last good test: https://travis-ci.org/learning-unlimited/ESP-Website/jobs/283544097
First bad test: https://travis-ci.org/learning-unlimited/ESP-Website/jobs/283820126

Comparing the deploy/travis/install logs for both shows that the difference is the version number of libpq-dev and its dependency, libpq5.

Unfortunately, I cannot test this fix locally because on my dev box, apt-cache reports the latest version of libpq-dev as 9.3.9.

### DIFF
--- a/deploy/travis/install
+++ b/deploy/travis/install
@@ -2,7 +2,7 @@
 set -euf -o pipefail
 
 if [ "$TRAVIS_JOB" = "test" ]; then
-    sudo apt-get install -y $(cat esp/packages_base.txt | grep -v ^memcached | grep -v ^postgres)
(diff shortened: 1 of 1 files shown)


hwatheod:
Tests passed, so this appears to have fixed the problem.

benjaminjkraft:
Hmm, I'm surprised that we don't need to specify that we need postgres in some other way.  But hey, if it works, good enough for me.

Issue #2070, opened 2016-03-06
T-shirt costs included in financial aid


jerrywu64:
If you pre-order a t-shirt (which is an optional fee I think) and you receive 100% financial aid, you will also receive financial aid for the shirt, which is usually not what we want. Of course you could alternatively specify the dollar amount of financial aid to avoid this issue, but it seems like this would be easy to mess up on. Maybe add an option for line item types for whether they should be included as part of finaid?

betaveros:
It turns out there is already a `for_finaid` field on LineItemType, and some code somewhere has set it correctly for recent t-shirt line items on our site; we're just not using it...

betaveros:
Update: I guess I did not correctly understand what `for_finaid` meant to whoever put it there. Right now it's set to True on the `LineItemType`s for financial aid grants and sibling discounts, so you can't actually use it to distinguish between t-shirt orders and program admission fees proper. I'm not sure what it *means*, though, and whether it would break things or cause confusion if we partially repurposed it and set it to True on program admission fees, or whether we should just make a new field (ew migrations)

PR #2460, opened 2017-10-20
Give finaid only for line item types marked as such

betaveros:
This doesn't actually work because for_finaid doesn't mean what I
thought it does, but it shows the simple logic changes we need to make.


betaveros:
Forgot to say this will fix #2070

willgearty:
Any plans to make this actually work?

willgearty:
Any plans to make this actually work @betaveros ?

willgearty:
Closed in favor of #3465.

PR #2463, opened 2017-11-05
Add duration-from-now labels next to deadline datepickers

### LINKED ISSUE #1700: Deadlines page should warn you if it's the wrong year
Similarly, new program creation. We'd have to think a bit about how best to do this, but accidentally setting your deadline to open 11 months ago instead of 1 month in the future is a very common user error and it would be nice to prevent it.

betaveros:
On datepickers on the deadline page, add a little colorful Bootstrap
label with a glyphicon that indicates how far away the selected time is
from now, which will alert users about dates more than 120 days in the
past or future and otherwise indicate whether the date is in the past or
future.

Fixes #1700.

screen shot 2017-11-05 at 3 53 50 pm

### DIFF
--- a/esp/esp/utils/widgets.py
+++ b/esp/esp/utils/widgets.py
@@ -17,14 +17,9 @@ import time
 
 # DATETIMEWIDGET
 calEnable = u"""
+<span id="{id}-info"><span id="{id}-gi"></span> <span id="{id}-duration"></span></span>
 <script type="text/javascript">
-    $j("#%s").%s({
-        showOn: 'button',
-        buttonImage: '%simages/calbutton_tight.png',
-        buttonImageOnly: true,
-        dateFormat: '%s',
-        timeFormat: '%s'
-    });
+    setupDatepickerDurationLabel('{widget}', '{id}', '{media_url}', '{date_format}', '{time_format}');
 </script>"""
 
 class DateTimeWidget(forms.widgets.TextInput):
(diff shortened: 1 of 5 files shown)


esp/esp/utils/widgets.py
```diff
@@ -17,14 +17,9 @@
...
-        showOn: 'button',
-        buttonImage: '%simages/calbutton_tight.png',
-        buttonImageOnly: true,
-        dateFormat: '%s',
-        timeFormat: '%s'
-    });
+    setupDatepickerDurationLabel('{widget}', '{id}', '{media_url}', '{date_format}', '{time_format}');
 </script>"""
 
 class DateTimeWidget(forms.widgets.TextInput):
```
jerrywu64: These widgets appear in a bunch of places. I don't think there's anywhere where it's bad to have this display (if such a place would exist), but some pages (e.g. volunteering, accessible via manage -> volunteers) make it wrap weirdly.

esp/public/media/scripts/datepicker-duration.js
```diff
@@ -0,0 +1,42 @@
...
+        abshours.toFixed(1) + " hours"
+      ) : (
+        (abshours / 24).toFixed(1) + " days"
+      );
+      if (abshours > 24 * 120) {
+        $info.prop('class', 'label label-important');
+        $gi.prop('class', 'glyphicon glyphicon-alert');
+      } else {
+        if (hours < 0) {
+          $info.prop('class', 'label label-success');
```
jerrywu64: What's the rationale behind these colors, or is it arbitrary? (It's also a bit confusing, because I want to think "green" means "deadline open" but this isn't consistent because right now green in the open date means deadline open and green in the closed date means deadline closed. Not sure you can do anything about that though

willgearty:
@betaveros Any update on this? This is a great feature that would be great for the next stable release!

willgearty:
@betaveros ???

willgearty:
@betaveros ????

esp/public/media/scripts/datepicker-duration.js
betaveros: Fair. I got rid of the green/blue colors; now it's just blue if close and red if far away.

esp/esp/utils/widgets.py
betaveros: OK, I made the textboxes narrower so wrapping is less likely. It might not fix this in all cases, but should be good enough in most.

PR #2489, opened 2018-01-21
Fix Travis tests

### LINKED ISSUE #2488: Tests failing on main again
Seems to be due to travis config changes, possibly this one: https://blog.travis-ci.com/2017-12-12-new-trusty-images-q4-launch

benjaminjkraft:
Something is funky with Travis's postgres packaging; I used the fix
suggested here: https://github.com/travis-ci/travis-ci/issues/9011#issuecomment-355561967

Fixes #2488.


betaveros:
Did you choose 9.1 deliberately? It looks like PostgreSQL 9.x goes up to 9.6?

benjaminjkraft:
9.1 was on diogenes, so I figured if in doubt we ought to use that to be consistent.  We need to upgrade diogenes to 9.6 but that's separate.

PR #2490, opened 2018-01-22
CSS fix for student lottery progress bars

willgearty:
I realized that the original class (`bar`) I was using was theme specific, so I've fixed it to use the up-to-date bootstrap class (`progress-bar`).


willgearty:
Sorry about the extra commits. I realized adding the bootstrap css styled a lot more than just the progress bars. I've pulled out the css that is needed for the progress bars and included in the template.

PR #2498, opened 2018-02-05
Adds email code to class cancellation request email subject

### LINKED ISSUE #2443: All cancellation request emails have same subject line
The cancellation request emails that were implemented by #2418 all have the same subject line, and the different emails can be confused by gmail as the same thread which can be confusing for admins.

willgearty:
Fixes #2443.


benjaminjkraft:
I don't think this will actually work!  Gmail collapses threads that have the same subject except for a different prefix in brackets.  I'd suggest doing "[program name] Class cancellation request for X1235: Class Title" or some such.

willgearty:
Ah, good call @benjaminjkraft ! Should be good now!

PR #2517, opened 2018-03-10
Bump django version

### LINKED ISSUE #2513: Upgrade django to latest patch version
We should start looking at upgrading to 1.11 real soon, but before cutting the release we should update to latest 1.8.x which should be comparatively easy.

benjaminjkraft:
This contains some minor security fixes; nothing that looks particularly interesting but always a good idea to have.  The main change is actually on dev servers; I think our default settings should continue to work
fine.  See https://docs.djangoproject.com/en/1.11/releases/1.8.16/#dns-rebinding-vulnerability-when-debug-true for details.

Fixes #2513.

### DIFF
--- a/esp/requirements.txt
+++ b/esp/requirements.txt
@@ -1,5 +1,5 @@
 # Django
-Django==1.8.15
+Django==1.8.19
 
 # Django-specific packages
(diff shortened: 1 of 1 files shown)


willgearty:
Looks fine to me.

PR #2521, opened 2018-03-17
Explicitly order sections to make sure index() is stable

### LINKED ISSUE #2505: Section indices are nondeterministic
As far as I can tell, the index of ClassSection objects is determined by the order in which they appear in a reverse `all` query through the `AjaxForeignKey` with no sort key, and Django doesn't specify what order this is in. So a section code like `X12345s1` can refer to different ClassSections depending on how many times you restart the server or which shell you're using...?

benjaminjkraft:
I'm amazed this has ever worked right, but I guess in the past postgres
has mostly ordered consistently; apparently in whatever version MIT ESP
is using (9.6, I think) it doesn't.  Anyway, this is definitely safer.

Hopefully fixes #2505.

### DIFF
--- a/esp/esp/program/models/class_.py
+++ b/esp/esp/program/models/class_.py
@@ -1320,7 +1320,9 @@ class ClassSubject(models.Model, CustomFormsLinkModel):
 
     def get_sections(self):
         if not hasattr(self, "_sections") or self._sections is None:
-            self._sections = self.sections.all()
+            # We explicitly order by ID to make sure we get reproducible
+            # ordering for e.g. index().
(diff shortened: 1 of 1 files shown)


benjaminjkraft:
cc @betavers if you had a repro for #2505 do you want to test that this actually fixes it?

willgearty:
The code seems straight forward enough, and I can't imagine it having any negative effects on the website. Should we wait for @betaveros to try this out with his test case, or just merge?

hwatheod:
I agree with @willgearty , merging.

PR #2522, opened 2018-03-17
Fix error in lunch constraint generation

### LINKED ISSUE #2510: Lunch constraint generator doesn't set the right seqs
The lunch constraint generator generates a set of `ScheduleConstraint` based on the selections in the interface and the program's timeblocks; each consists of two `BooleanExpression`s, a condition and a requirement, which consist of an RPN-style stack of `BooleanToken`s, so for example `[ScheduleTestOccupied(timeblock_1), ScheduleTestOccupied(timeblock_2), '||']` would be true if the user has a class in `timeblock_1` or `timeblock_2`.  The ordering of the stack is stored in the `seq` field of the `BooleanToken`.

For some reason, the lunch constraint generator hasn't always been setting distinct `seq` objects on each of its `BooleanToken`s, such that their ordering is not well-defined.  In the past, they seem to somehow have always come out in the right order, I'm guessing because the database broke ties by `id`, which happened to be the intended ordering.  But for MIT ESP Spark 2018, possibly because of a postgres upgrade, they didn't, which caused an expression like the above to always evaluate to False, which as the requirement meant the constraint not satisfied when it should have been.

The fix will be to correctly set the `seq` on each token.  I'm not totally sure why it's not getting set correctly, but it seems that a simple OR like the above consistently results in all three `seq`s being 0, whereas really the OR should have a strictly higher `seq`.

benjaminjkraft:
This fixes #2510, a nasty bug in lunch constraint generation, which has
been around forever but formerly was masked by postgres being fairly
stable about ordering in practice.  In particular, the constraints were
generated with several duplicate `seq` values rather than each token
having a distinct `seq`.

The underlying issue was that `BooleanExpression.get_stack()` is cached,
and the cache wasn't being expired when new tokens were added, because
apparently `depend_on_row` doesn't work correctly for subclasses --
django only sends the signal for the exact model class being written,
not the subclasses, it seems.  This meant that sometimes when tokens
were added, they wouldn't see the preceding token, because they would
get a cached value of `get_stack()` generated before the previous token
was added.

The fix -- not the nicest but it should work -- is to just explicitly
depend on every subclass.  Luckily this seems to be the only cache that
depends on `BooleanToken`, so I didn't have to change it anywhere else.

I tested this by regenerating some lunch constraints on my dev server;
they now have distinct `seq`s for every token within each expression,
and the ordering looks correct to me.

### DIFF
--- a/esp/esp/program/models/__init__.py
+++ b/esp/esp/program/models/__init__.py
@@ -1633,7 +1633,13 @@ class BooleanExpression(models.Model):
     @cache_function
     def get_stack(self):
         return [s.subclass_instance() for s in self.booleantoken_set.all().order_by('seq')]
+    # Sadly, the way django signals work, we have to depend on every subclass
+    # of BooleanToken, not just BooleanToken itself.
     get_stack.depend_on_row('program.BooleanToken', lambda token: {'self': token.exp})
+    get_stack.depend_on_row('program.ScheduleTestTimeblock', lambda token: {'self': token.exp})
+    get_stack.depend_on_row('program.ScheduleTestOccupied', lambda token: {'self': token.exp})
+    get_stack.depend_on_row('program.ScheduleTestCategory', lambda token: {'self': token.exp})
+    get_stack.depend_on_row('program.ScheduleTestSectionList', lambda token: {'self': token.exp})
 
     def reset(self):
         self.booleantoken_set.all().delete()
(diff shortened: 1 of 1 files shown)


hwatheod:
Line 1572 in this file also has similar code.  Are these changes needed there too?

`subclass_instance.depend_on_row('program.BooleanToken', lambda bt: {'self': bt})`

benjaminjkraft:
Whoa, yeah, good catch.  I thought I had checked for other instances but I'll check more carefully!

PR #2523, opened 2018-03-17
Display better info

kkbrum:
Added a grade filter to the scheduler. This takes the minimum or maximum grade inputted and only displays sections that have that grade as the lower or upper bound.  For example, if you set the min to 10, you will get 10-11 and 10-12th grade classes, and if you also set the max to 12 you will get just 10-12th grade classes.

Added a column to show whether the user is activated or deactivated in the user search. This is helpful if one person has many accounts.

Added a select all button to the recipient checklist in the communications panel.

### DIFF
--- a/esp/public/media/scripts/ajaxschedulingmodule/ESP/Sections.js
+++ b/esp/public/media/scripts/ajaxschedulingmodule/ESP/Sections.js
@@ -24,7 +24,7 @@ function Sections(sections_data, section_details_data, teacher_data, scheduleAss
         classLengthMax: {active: false, el: $j("input#section-filter-length-max"), type: "number"},
         classCapacityMin: {active: false, el: $j("input#section-filter-capacity-min"), type: "number"},
         classCapacityMax: {active: false, el: $j("input#section-filter-capacity-max"), type: "number"},
-	    gradeMin: {active: false, el: $j("input#section-filter-grade-min"), type: "number"},
+        gradeMin: {active: false, el: $j("input#section-filter-grade-min"), type: "number"},
         gradeMax: {active: false, el: $j("input#section-filter-grade-max"), type: "number"},
         classTeacher: {active: false, el: $j("input#section-filter-teacher-text"), type: "string"},
         classHideUnapproved: {active: false, el: $j("input#section-filter-unapproved"), type: "boolean"},
--- a/esp/public/media/scripts/ajaxschedulingmodule/ESP/Sections.js
+++ b/esp/public/media/scripts/ajaxschedulingmodule/ESP/Sections.js
@@ -24,7 +24,7 @@ function Sections(sections_data, section_details_data, teacher_data, scheduleAss
         classLengthMax: {active: false, el: $j("input#section-filter-length-max"), type: "number"},
         classCapacityMin: {active: false, el: $j("input#section-filter-capacity-min"), type: "number"},
         classCapacityMax: {active: false, el: $j("input#section-filter-capacity-max"), type: "number"},
-		gradeMin: {active: false, el: $j("input#section-filter-grade-min"), type: "number"},
+	    gradeMin: {active: false, el: $j("input#section-filter-grade-min"), type: "number"},
         gradeMax: {active: false, el: $j("input#section-filter-grade-max"), type: "number"},
         classTeacher: {active: false, el: $j("input#section-filter-teacher-text"), type: "string"},
         classHideUnapproved: {active: false, el: $j("input#section-filter-unapproved"), type: "boolean"},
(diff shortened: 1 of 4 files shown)


esp/public/media/scripts/ajaxschedulingmodule/ESP/Sections.js
```diff
@@ -24,6 +24,8 @@ function Sections(sections_data, section_details_data, teacher_data, scheduleAss
         classLengthMax: {active: false, el: $j("input#section-filter-length-max"), type: "number"},
         classCapacityMin: {active: false, el: $j("input#section-filter-capacity-min"), type: "number"},
         classCapacityMax: {active: false, el: $j("input#section-filter-capacity-max"), type: "number"},
+	    gradeMin: {active: false, el: $j("input#section-filter-grade-min"), type: "number"},
```
benjaminjkraft: super minor: can you fix the indentation here, to align with the other lines?

esp/templates/users/userchecklist.html
```diff
@@ -23,10 +38,16 @@ <h1>Select users</h1>
...
+<td>
+<input type="checkbox" id="chk_new" onclick="checkAll('chk');">
+Select all users
+</tr>
+</td>
 {% for user in users %}
 <tr>
 <td>
-<input type="checkbox" name="userno{{user.id}}" value="{{user.id}}" />
+<input type="checkbox" id="chk" name="userno{{user.id}}" value="{{user.id}}" />
```
benjaminjkraft: Since this is inside a `for`, there will be a bunch of them, so we should use `class` instead of `id`, since having multiple of the same `id` can confuse browsers.

esp/templates/users/userview_search.html
```diff
@@ -29,7 +29,7 @@ <h1>Select User</h1>
 <p>Multiple users matched the criteria that you specified.  Please select one to view:</p>
 
 <table>
-<tr><th>Name</th><th>Username</th><th>E-mail</th><th>User Type</th></tr>
+<tr><th>Name</th><th>Username</th><th>E-mail</th><th>User Type</th><th>Activated?</th></tr>
```
benjaminjkraft: Cool, great idea!

esp/templates/users/userchecklist.html
```diff
@@ -14,6 +14,21 @@
 <br />
 <br />
 
+<script type="text/javascript">
+    function checkAll(checkId){
+        var inputs = document.getElementsByTagName("input");
+        for (var i = 0; i < inputs.length; i++) { 
+            if (inputs[i].type == "checkbox" && inputs[i].id == checkId) { 
+                if(inputs[i].checked == true) {
```
benjaminjkraft: Here, you can just do `if (inputs[i].checked) {` (and `} else {` below)!  (Aso, while you're at it, if you could fix the whitespace -- our style is spaces after the `if`, no spaces before the `;`.)

esp/templates/users/userchecklist.html
```diff
@@ -14,6 +14,21 @@
 <br />
 <br />
 
+<script type="text/javascript">
+    function checkAll(checkId){
+        var inputs = document.getElementsByTagName("input");
```
benjaminjkraft: Just realized -- I think it's faster and simpler to do e.g. `document.querySelectorAll('input[type="checkbox"].class')` -- no need to iterate over all the inputs in the page.  (Alternately, we have jquery, it's `$j` rather than `$` for historical reasons, so you could use that instead.)

esp/templates/users/userchecklist.html
```diff
@@ -14,6 +14,19 @@
 <br />
 <br />
 
+<script type="text/javascript">
+    function checkAll(checkId){
+        var inputs = document.querySelectorAll("checkId");
```
benjaminjkraft: I think you want `checkId` here -- the variable, not the string.  Also, I think you'll need to prefix this with a `.`, so you get things with the _class_ `checkId`, not the node type.  See [querySelectorAll docs here](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelectorAll).

kkbrum: It seems to function either way-- but I could have just been lucky before!

esp/templates/users/userchecklist.html
```diff
@@ -14,6 +14,19 @@
 <br />
 <br />
 
+<script type="text/javascript">
```
benjaminjkraft: I tested this out -- the behavior if you do "select all", then uncheck a few, then change your mind and remove "select all" is very confusing.  I'd suggest setting everything to the new state of the "select all" checkbox, rather than toggling all the other checkboxes.

esp/templates/users/userchecklist.html
```diff
@@ -17,10 +17,12 @@
...
-            if (chk_new.checked) {
+        if (chk_new.checked) {
+            for (var i = 0; i < inputs.length; i++) {  
                 inputs[i].checked = true;
-            } else {
+            }
+        } else {
+            for (i = 0; i < inputs.length; i++) { 
                 inputs[i].checked = false;
             }
```
willgearty: If you really want to be efficient, you could just have a single for loop where you set all inputs to the state of chk_new.checked, right?

esp/templates/users/userchecklist.html
benjaminjkraft: Great, much simpler and better behavior.  Thanks!

Issue #1194, opened 2014-04-08
Shirt statistics missing when fitted shirt option is removed


pricem:
The program dashboard shows the number of shirts in each size requested by teachers.  However, if you set the "teacherinfo_shirt_type_selection" Tag to False in order to remove the distinction between "plain" and "fitted" shirts, the dashboard shows all zeros.  This should be an easy fix, probably in the Program.getShirtInfo() function at esp/esp/program/models/**init**.py:1025.

Princeton ran into this bug two years in a row.

kkbrum:
Pull request #2524 fixes this (and the checks passed after fixing some whitespace)?  Is there something else that needs to be done here?

willgearty:
We don't close issues until pull requests get merged, just a workflow thing :)
And if you put "Fixes #XYZ" in the description of a pull request, the issue will actually get closed automatically when the PR gets merged.

PR #2524, opened 2018-03-18
Fixed teacher shirt statistics  (issue #1194)

kkbrum:
The teacher shirt statistics weren't working when the types were removed. I fixed it by tallying the shirt as a plain ('M') shirt if shirt types were turned off.

### DIFF
--- a/esp/esp/program/models/__init__.py
+++ b/esp/esp/program/models/__init__.py
@@ -1067,7 +1067,7 @@ class Program(models.Model, CustomFormsLinkModel):
         if 'class_approved' in teacher_dict:
             query = teacher_dict['class_approved']
             query = query.filter(registrationprofile__most_recent_profile=True)
-            if Tag.getTag('teacherinfo_shirt_type_selection') == 'False':
+            if Tag.getBooleanTag('teacherinfo_shirt_type_selection') == False:
                 query = query.values_list('registrationprofile__teacher_info__shirt_size')
                 query = query.annotate(people=Count('id', distinct=True))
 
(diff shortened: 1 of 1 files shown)


esp/esp/program/models/__init__.py
```diff
@@ -1067,18 +1067,28 @@ def getShirtInfo(self):
         if 'class_approved' in teacher_dict:
             query = teacher_dict['class_approved']
             query = query.filter(registrationprofile__most_recent_profile=True)
-            query = query.values_list('registrationprofile__teacher_info__shirt_type',
+            if Tag.getTag('teacherinfo_shirt_type_selection') == 'False':
```
benjaminjkraft: Can you use `Tag.getBooleanTag` here?  It will take care of parsing things as `False` or `True` in a slightly more permissive way.

esp/esp/program/models/__init__.py
```diff
@@ -1067,18 +1067,28 @@ def getShirtInfo(self):
...
+                query = query.values_list('registrationprofile__teacher_info__shirt_size')
+                query = query.annotate(people=Count('id', distinct=True))
+
+                for row in query:
+                    shirt_size, count = row
+                    shirt_count['M'][shirt_size] = count
+
+            else:
+                query = query.values_list('registrationprofile__teacher_info__shirt_type',
                                       'registrationprofile__teacher_info__shirt_size')
```
benjaminjkraft: minor nit: can you fix the indentation of this line?  It should line up to the string right above.  I'm not sure why the linter isn't complaining here!

esp/esp/program/models/__init__.py
```diff
@@ -1067,18 +1067,28 @@ def getShirtInfo(self):
         if 'class_approved' in teacher_dict:
             query = teacher_dict['class_approved']
             query = query.filter(registrationprofile__most_recent_profile=True)
-            query = query.values_list('registrationprofile__teacher_info__shirt_type',
-                                      'registrationprofile__teacher_info__shirt_size')
-            query = query.annotate(people=Count('id', distinct=True))
+            if Tag.getBooleanTag('teacherinfo_shirt_type_selection') == False:
```
benjaminjkraft: Since this is a boolean now, you can just do `if not Tag.getBooleanTag(...):`, no need to compare to `False`.

benjaminjkraft:
Thanks for the fix!

Issue #2479, opened 2017-12-18
Align "log in" and "sign up" buttons in fruitsalad


milescalabresi:
When logged out, the "log in" and the "sign up" buttons are not vertically aligned. I think this comes from the fact that "log in" is an input form while "sign up" is an anchor redirect, each of which have different styles
Example:
![image](https://user-images.githubusercontent.com/5506191/34087149-2a975d90-e36f-11e7-8bbe-90209a8c6997.png)

PR #2525, opened 2018-03-18
Aligned Log in and Sign up buttons

kkbrum:
Fixes issue #2479. Just added a style tag to the log in button removing the padding and thus align it with the sign up button.

Note that it seems template overrides are automatically created when the theme is set, so anyone who actually wants this change will have to fix this in their own template overrides on their sites (I think).

### DIFF
--- a/esp/esp/themes/theme_data/fruitsalad/less/main.less
+++ b/esp/esp/themes/theme_data/fruitsalad/less/main.less
@@ -487,12 +487,13 @@ position: absolute;
 top: 9px; right: 48px;
 height: 21px; width: 45px;
 font-size: 10.5px;
-padding-right: 15px;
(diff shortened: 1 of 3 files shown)


benjaminjkraft:
Thanks for taking a look!  Can we do this in the theme stylesheets instead, to keep the template simpler?  I'm looking at around esp/esp/themes/theme_data/fruitsalad/less/main.less:485.  I think it might be more maintainable to do it that way, since that's where the other login styling is.

BTW, for either template overrides or stylesheets, if you go to the themes, then "recompile theme", it should update all the template overrides and stylesheets.

benjaminjkraft:
Thanks so much!  I tested this out and it definitely looks better.

PR #2526, opened 2018-03-18
add ndmin parameter for corner case in import_assignments in lottery.py

hwatheod:
If there is only one student or section, then the import_assignements lottery.py fails because numpy.loadtxt generates an array with the wrong number of dimensions (one fewer than expected).  

This likely wouldn't occur in a real program, but I discovered it while testing the lottery module on stanford-sr8 with a small amount of data.

This is fixed by specifying the parameter ndmin to force numpy.loadtxt to produce an array with the right number of dimensions (see documentation here: https://docs.scipy.org/doc/numpy-1.14.0/reference/generated/numpy.loadtxt.html )


benjaminjkraft:
Hah, that's a funny edge case!  Thanks for fixing.

Issue #373, opened 2012-06-12
Formatting of schedules 


pricem:
We have the ability to include a personalized student/teacher/volunteer schedule in comm panel e-mails by using the smart text tags {{ program.student_schedule }} et al.  However, these schedules are rendered in a tabular plain text format that only looks good if rendered in a fixed-width font and the fields fit within the allotted number of spaces.  We would like an alternative rendering scheme (either rendering with HTML or making more clever use of plain text) so schedules look good in a wider variety of situations.

willgearty:
Stanford would also love to have this enhancement implemented. It is a valuable tool to be able to include these schedules in emails, but the formatting is currently atrocious and unchangeable.
Would there be a way to implement template overrides for each smart text tag so chapters can alter the default text and formatting?

PR #2528, opened 2018-03-20
Better email schedules (issue #373)

kkbrum:
Turned the in line email schedules into html instead of text to make them display in a table that doesn't get messed up.

Fixes issue #373.

### DIFF
--- a/esp/esp/program/modules/handlers/programprintables.py
+++ b/esp/esp/program/modules/handlers/programprintables.py
@@ -792,21 +792,23 @@ class ProgramPrintables(ProgramModuleObj):
 
             if schedule_type == u'Student':
                 classes = ProgramPrintables.get_student_classlist(program, user)
+                classes.sort()
             elif schedule_type == u'Teacher':
                 classes = ProgramPrintables.get_teacher_classlist(program, user)
+                classes.sort()
 
             schedule = """
-                        <p> %s schedule for %s: </p>
-                        """ % (schedule_type, user.name())
+                        <p> %s schedule for %s for %s: </p>
+                        """ % (schedule_type, user.name(), program.niceName())
             schedule += """
-                        <table cellspacing="0" cellpadding="10" border="1">
+                        <table cellspacing="0" cellpadding="10" border="1" width="100%">
                             <tr>
-                            <th width="100">Time</th>
-                            <th width="300">Class</th>
+                            <th width="20%">Time</th>
+                            <th width="60%">Class</th>
                             """
             if room_numbers:
                 schedule += """
-                            <th width="100">Room</th>
+                            <th width="20%">Room</th>
                             """
             schedule += """
                         </tr>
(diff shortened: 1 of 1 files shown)


esp/esp/program/modules/handlers/programprintables.py
```diff
@@ -788,43 +789,87 @@ def getSchedule(program, user, schedule_type=None, room_numbers=True):
...
-
- Time                   | Shift                                  \n""" % (user.name())
-            schedule += u'------------------------+----------------------------------------\n'
+            schedule = """
+                       <p> %s schedule for %s: </p>
+                       """ % (schedule_type, user.name())
+            schedule += """
+                        <table cellspacing="0" cellpadding="10" border="1">
+                            <tr>
+                            <th width="100">Time</th>
```
willgearty: You might want to use percentages here instead of pixels, otherwise the width of the email could be way too big, especially on mobile

willgearty: (And above for students/teachers)

esp/esp/program/modules/handlers/programprintables.py
```diff
@@ -788,43 +789,87 @@ def getSchedule(program, user, schedule_type=None, room_numbers=True):
...
 
-     Time                   | Class                                  \n""" % (schedule_type, user.name())
-            schedule += u'------------------------+---------------------------------------------------\n'
             if schedule_type == u'Student':
                 classes = ProgramPrintables.get_student_classlist(program, user)
             elif schedule_type == u'Teacher':
                 classes = ProgramPrintables.get_teacher_classlist(program, user)
+
+            schedule = """
+                        <p> %s schedule for %s: </p>
```
willgearty: I know it's not in the original version, but would it be valuable to specify the program the schedule is for as well?

willgearty:
Just a few comments, otherwise it looks great!

kkbrum:
Is there a way to actually send an email from my test environment? I tried to send myself my schedule to see how it would actually look on a phone and not just in the preview panel, but the email didn't send (it didn't even make it to email texts).

Anyways I hope it will at least look better than the old ones... it looks good in the preview panel and I changed the width to percentages so hopefully it'll be okay.

willgearty:
I've wondered this for quite some time...maybe @benjaminjkraft or @milescalabresi know?
Another way to check, although not really the same thing, would be to just resize your browser window while looking at the preview panel to see how it responds to smaller window sizes.

Anyways, I think what you've got here is leaps and bounds beyond what it used to be (I've actually had this issue on a to-do list for quite some time, and it's great to see it finally fixed!).

esp/esp/program/modules/handlers/programprintables.py
```diff
@@ -788,43 +789,89 @@ def getSchedule(program, user, schedule_type=None, room_numbers=True):
...
+                            <td> %s </td>
+                            <td> %s </td>
+                            </tr>
+                            """ % (str(shift.request.timeslot.pretty_time()), str(shift.request.timeslot.description))
+            schedule += """
+                        </indent>
+                    </table>
+                    """
+
+        return mark_safe(schedule)
```
hwatheod: This mark_safe looks problematic because no HTML sanitization (escaping special symbols like <>, etc.) has been done when building up the schedule variable.  (Unless we're already sanitizing all names, classes, etc. before storing in the database?)  So it's not really safe to use directly in HTML output.

Instead, the schedule should be built up with format_html:

https://docs.djangoproject.com/en/1.8/ref/utils/#django.utils.html.format_html

benjaminjkraft:
If it's not the default, you can [set things up](https://docs.djangoproject.com/en/1.8/topics/email/#console-backend) to print email to the dev server console, but that doesn't show how it will display in reality.  Sadly we don't really have a good way to send an actual email.

PR #2539, opened 2018-04-07
Functionality to scan barcodes with device's camera

### LINKED ISSUE #2506: Use device's camera to scan barcodes
Right now, you can request barcode scanners from LU, but it would also be great if a user could just use their phone to scan barcodes to check students in.

Could use a JS library like this one: https://serratus.github.io/quaggaJS/
Good example (with source) here: https://serratus.github.io/quaggaJS/v1.0.0-beta.1/examples/scan-to-input/

willgearty:
Fixes #2506.

Adds functionality to the barcode checkin page to allow a user to scan barcodes with their device's camera (specifically the rear-facing camera if there are two), which adds the user IDs to the text box as an external scanner would. The user can then submit these user IDs just as they normally would.

The javascript library ([quaggaJS](https://serratus.github.io/quaggaJS/)) has the potential to work with various types of barcodes, although I have found the most reliability with Code 39, Code 128, and other linear codes, hence these are the only implemented options at the moment. Furthermore, this library, as opposed to other libraries, is invariant to scale and rotation, allowing for more flexibility and efficiency.

This works on computers (PC, Mac, and presumably Linux), (recent) Android devices, and iOS 11. Full compatibility [here](https://caniuse.com/#feat=stream).

### DIFF
--- a/esp/templates/program/modules/onsitecheckinmodule/barcodecheckin.html
+++ b/esp/templates/program/modules/onsitecheckinmodule/barcodecheckin.html
@@ -73,73 +73,66 @@
 <h1>Barcode Checkin &mdash; For {{ program.niceName }}</h1>
 
 <div id='program_form'>
+    <p>
+    Welcome to barcode check-in for {{ program.niceName }}.  Please enter a list of user IDs.
+    </p>
 
-<p>
-Welcome to barcode check-in for {{ program.niceName }}.  Please enter a list of user IDs.
-</p>
+    {% if results %}
+    <p>
+    Checked in {{ results.new|length }} students.  {{ results.not_found|length }} students were not found, {{ results.existing|length }} students were already checked in, and {{ results.not_student|length }} IDs did not correspond to students.
+    </p>
+    {% if results.not_found %}
+    <p>
+    Students not found:
+    {% for id in results.not_found %}
+    {{ id }}{% if not forloop.last %},{% endif %}
+    {% endfor %}
+    </p>
+    {% endif %}
+    {% if results.not_student %}
+    <p>
+    IDs not corresponding to students:
+    {% for id in results.not_student %}
+    {{ id }}{% if not forloop.last %},{% endif %}
+    {% endfor %}
+    </p>
+    {% endif %}
+    {% comment %}
+    {% if results.existing %}
+    <p>
+    IDs already checked in:
+    {% for id in results.existing %}
+    {{ id }}{% if not forloop.last %},{% endif %}
+    {% endfor %}
+    </p>
+    {% endif %}
+    {% endcomment %}
+    {% endif %}
 
-{% if results %}
-<p>
-Checked in {{ results.new|length }} students.  {{ results.not_found|length }} students were not found, {{ results.existing|length }} students were already checked in, and {{ results.not_student|length }} IDs did not correspond to students.
-</p>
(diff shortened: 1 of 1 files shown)


esp/templates/program/modules/onsitecheckinmodule/barcodecheckin.html
```diff
@@ -57,17 +107,137 @@ <h1>Barcode Checkin &mdash; For {{ program.niceName }}</h1>
...
+        // format.className = "format";
+        // code.className = "code";
+
+        // li.appendChild(format);
+        // li.appendChild(code);
+
+        // format.appendChild(document.createTextNode(result.codeResult.format));
+        // code.appendChild(document.createTextNode(result.codeResult.code));
+
+        // results.insertBefore(li, results.firstChild);
```
hwatheod: Can this blob of commented-out code be removed? (lines 177-192)

esp/templates/program/modules/onsitecheckinmodule/barcodecheckin.html
```diff
@@ -57,17 +112,133 @@ <h1>Barcode Checkin &mdash; For {{ program.niceName }}</h1>
...
+<div class="overlay overlay--inline">
+    <div class="overlay__content viewport" id="interactive">
+        <div class="overlay__close">X</div>
+    </div>
+</div>
+
+<!--<br />
 {% load render_qsd %}
-{% render_inline_program_qsd program "onsite:status" %}
+{% render_inline_program_qsd program "onsite:status" %}-->
```
hwatheod: Remove these commented-out lines.

esp/templates/program/modules/onsitecheckinmodule/barcodecheckin.html
```diff
@@ -57,17 +112,133 @@ <h1>Barcode Checkin &mdash; For {{ program.niceName }}</h1>
...
+            });
+        },
+        activateScanner: function() {
+            var scanner = this.configureScanner('.overlay__content'),
+                onDetected = function (result) {
+                    this.addToResults(result);
+                }.bind(this),
+                stop = function() {
+                    Quagga.stop();
+                    console.log("Scanner stopped")
```
hwatheod: Should we get rid of all the console.log lines? Normally, I would say yes, but since this is functionality only used by admins, I could see an argument for keeping them.

willgearty: This was mostly for debugging, but I could also seeing it be valuable to admins in case they are having trouble. However, given that most use of this will be on mobile devices, and you can't see console logs on mobile, it probably isn't very helpful after all?

hwatheod:
I had an old "Draco Malfoy" student barcode sheet from a while back and it scanned it in correctly under Code 39.

I didn't have much luck getting it to scan any other barcodes I had lying around on other products, though (with any of the four formats implemented).  Sometimes it scanned in wrong numbers from these other products.  I guess they must be in other formats.

However, since it seems to work on the barcodes produced by our website, I am merging.

benjaminjkraft:
When you get a chance, can you make a quick PR to add Stable Release 11 docs for this?  You can add a section to docs/admin/releases/11/README.rst.

hwatheod:
Done here: #2556

PR #2547, opened 2018-04-09
Classroom resource choices

### LINKED ISSUE #2140: Resource choices as classroom furnishings
Classroom furnishings should include the option to select resource choices rather than entire resource types (if a resource type has a choice, it is most probable that classrooms do not have all of the choices as furnishings).

Currently, we need to add additional resource types after teacher registration closes in order to facilitate proper scheduling (manual). This process is extremely convoluted.

willgearty:
Replaces old classroom furnishings selection with formsets that allow for the user to specify a custom `attribute_value`. Loads and saves `attribute_value`s appropriately when creating new classrooms and editing old classrooms. Also adds prefixes to the formsets so as not to conflict with the resource types form.
Fixes #2140.

image


willgearty:
I got nothing on the Travis errors...

benjaminjkraft:
Travis was down, so you may need to rebuild things.

willgearty:
@jerrywu64 How does this look for a first pass at allowing admins to assign furnishing choices?

jerrywu64:
Oh yay, this is a nice feature, thanks! MIT has been circumventing the previous lack of a UI by importing most of the classrooms we use by script and using admin panel to manually fix the rest, but admin panel kind of sucks for this. I didn't really look over your code but conceptually (based on what you wrote in the opening post) sounds good.

willgearty:
I think I originally tried to do that, but I couldn't figure out how to do that. I'll give it another look.

willgearty:
Okay, now I remember why that would be tricky @kkbrum. Basically, the form doesn't know at initialization which furnishing you are going to pick, so it can't decide whether to have an open response field (CharField) or a dropdown field (ChoiceField). I'm not sure if there's a way around this to do specifically what you are looking for (although I agree that would be ideal).

One possible alternative would be to dynamically provide suggestions for the (option) field, like the autocomplete feature for adding coteachers, except it would provide the suggestions right from the start? And then when the form is submitted, it could check, if there is a preset list of choices for the furnishing, that the provided option matches one of those (and if not raise an error on the page).

willgearty:
An alternative strategy would be to always have a dropdown with an open response field available if you choose the "other" option in the dropdown? I could probably even make it such that the open response field is disabled unless the furnishing that was chosen has no specific options?

benjaminjkraft:
I think if you want, it is possible to set things up so you have both detail fields, but neither is required, and both are hidden by default -- and then in javascript you unhide the relevant one based on the dropdown.  That will definitely be more work than the textbox, of course, although probably less work than autocomplete.

willgearty:
I would also need to then change the options for the select based on the furnishing, right?

benjaminjkraft:
Oh, ok.  Yeah, you'd need to populate those in JS, too.  At that point maybe you just have the JS replace the field dynamically?

willgearty:
Phew! Thanks to @benjaminjkraft's insight I designed a way to dynamically alter and populate the option field with JS. When a new furnishing formset is added, the default is to have an open response field (`input`). As soon as the furnishing `select` field is changed by the user, the page then checks for the options associated with that furnishing and updates the option field accordingly (if there is more than one choice, make a `select` field with those possible choices; if there is only one choice--presumably "Don't care"-- make a `input` field).

If the form is set up to edit a classroom, the same process occurs for each furnishing, except it then populates the new `input` or `select` fields with the values that were previously in the `input` fields.

Since there's a high chance of people submitting the form with blank furnishings (not valid) or blank options (totally valid), I've added a little bit of code to account for that, filter out the blanks, and then remove any duplicate furnishing/option combinations.

I also realized that a lot of the original classroom creation code assumed that you could only have one instance of a resource type for a given classroom, but that is no longer a valid assumption, since you might want to have two of the same resource type with different choices. So I changed that, too.

@kkbrum Please let me know how this runs and if it seems intuitive/whether it fulfills your request.

image

kkbrum:
This looks beautiful and intuitive! It does what I would expect it to do.

I don't have experience with floating resources but those might want to have a choice field too? Right now they get a name and a type but no option. Might make it more consistent to add that.

@benjaminjkraft can you look at the code? It looks more or less legible and decent to me but it is also a lot of code that is over my head right now so maybe someone else should give it a look

esp/templates/program/modules/resourcemodule/classrooms.html
```diff
@@ -74,3 +97,71 @@ <h2>Step 3: Classrooms</h2>
...
+    var furnishing = $j(obj).val()
+    getChoices(furnishing, function(response) {
+        var choices = response.choices;
+        var num = $j(obj).attr('id').split("-")[1];
+        $j('#id_furnishings-' + num + '-choice').remove();
+        if (choices.length > 1) {
+            $j(obj).after('<select style="margin-left: 4px" id="id_furnishings-' + num + '-choice" name="furnishings-' + num + '-choice" type="text"></select>');
+            $j('#id_furnishings-' + num + '-choice').append('<option>(option)</option>');
+            for (i in choices) {
+                $j('#id_furnishings-' + num + '-choice').append('<option value=' + choices[i] + '>' + choices[i] + '</option>');
```
benjaminjkraft: Doing the string interpolation like this is a little questionable security-wise -- maybe it's not a problem since only admins can edit these, but better would be to either escape it properly, or use jquery's methods to populate it (e.g. I think you can do something like `...append($('<option>').value(choices[i]).text(choices[i]))`, although I don't recall the syntax offhand).  (You could also ideally do that for `num`, although it may be easier to just ensure it's an integer instead.)

esp/templates/program/modules/resourcemodule/classrooms.html
```diff
@@ -74,3 +97,71 @@ <h2>Step 3: Classrooms</h2>
 {% endifequal %}
 
 <br />
+
+<!--This branch has a fix for when the form is originally hidden-->
+<script src="https://rawgit.com/elo80ka/django-dynamic-formset/f5f68c818953ddf43ffa7cb2e9aa391c3d4dfb47/src/jquery.formset.js"></script>
+<!--jQuery.initialize plugin is created to help maintain dynamically created elements on the page-->
+<script src="https://rawgit.com/pie6k/jquery.initialize/master/jquery.initialize.js"></script>
+<script>
+<!--Formset settings-->
```
benjaminjkraft: This is JS, so you want `//` or `/*` comments -- I think these technically work too since the JS is directly in the HTML, but JS-style comments are much more readable/expected and will continue to work in a standalone script.

esp/templates/program/modules/resourcemodule/classrooms.html
```diff
@@ -74,3 +97,71 @@ <h2>Step 3: Classrooms</h2>
...
+    <!--Convert option field to dropdown if necessary-->
+    if ($j(this).attr('id').split("-")[2] == "furnishing") {
+        var num = $j(this).attr('id').split("-")[1];
+        <!--Make sure to preserve selected value-->
+        var value = $j('#id_furnishings-' + num + '-choice').val();
+        if (value) {
+            update_choices($j(this), value);
+        }
+    }
+}, { target: $j("#furnishing_formset")[0] });
```
benjaminjkraft: Why the `[0]`?  I think jquery doesn't actually require it here, although I could be wrong.

esp/templates/program/modules/resourcemodule/classrooms.html
```diff
@@ -74,3 +97,71 @@ <h2>Step 3: Classrooms</h2>
...
+});
+
+<!--Get choices for a furnishing via POST-->
+function getChoices(furnishing, callback){
+    var data = {furnishing: furnishing, csrfmiddlewaretoken: csrf_token()};
+    $j.post('/manage/{{ prog.url }}/ajaxfurnishingchoices', data, "json").success(callback);
+}
+
+<!--Update the option field to reflect whether there are specific choices (dropdown) or not (open response). If a value is supplied, set that value as selected afterwards.-->
+function update_choices(obj, value) {
```
benjaminjkraft: Can you give `obj` a more descriptive name here?  I think it's the furnishing dropdown, so maybe `furnishing_select_obj` or something?

esp/templates/program/modules/resourcemodule/classrooms.html
```diff
@@ -74,3 +97,71 @@ <h2>Step 3: Classrooms</h2>
...
+            $j(obj).after('<input style="margin-left: 4px" id="id_furnishings-' + num + '-choice" name="furnishings-' + num + '-choice" maxlength="50" placeholder="(option)" type="text"></select>');
+        }
+        if (value) {
+            $j('#id_furnishings-' + num + '-choice').val(value);
+        }
+    });
+}
+
+<!--Upon page load and dynamic creation of new selects-->
+$j.initialize("select", function() {
```
benjaminjkraft: Hmm, can you instead do this where you create the select?  It's a bit surprising to see it here, and potentially has unexpected consequences (say, if someone else writes other JS to add another unrelated select to the page in a similar fashion).

esp/templates/program/modules/resourcemodule/classrooms.html
willgearty: I think this had something to do with the function requiring a single object, not a jQuery query. I've changed it anyways such that it just uses JS to get the object.

esp/templates/program/modules/resourcemodule/classrooms.html
willgearty: Doh, that's what I get for going back and forth between JS and HTML.

esp/templates/program/modules/resourcemodule/classrooms.html
willgearty: Good point, it's also extremely easy to mess up because you need to keep track of the different quote types!

esp/templates/program/modules/resourcemodule/classrooms.html
```diff
@@ -74,3 +97,71 @@ <h2>Step 3: Classrooms</h2>
...
+    var data = {furnishing: furnishing, csrfmiddlewaretoken: csrf_token()};
+    $j.post('/manage/{{ prog.url }}/ajaxfurnishingchoices', data, "json").success(callback);
+}
+
+<!--Update the option field to reflect whether there are specific choices (dropdown) or not (open response). If a value is supplied, set that value as selected afterwards.-->
+function update_choices(obj, value) {
+    var furnishing = $j(obj).val()
+    getChoices(furnishing, function(response) {
+        var choices = response.choices;
+        var num = $j(obj).attr('id').split("-")[1];
```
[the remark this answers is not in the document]
willgearty: Yep, it will always be an integer that represents the index of the formset, thanks to the dynamic formset plugin (the ID of the furnishing field is always of the form `id_furnishings-[index]-furnishing`, while the ID of the choice field is always of the form `furnishings-[index]-choice`). Please do let me know if you know of a better way, though!

esp/templates/program/modules/resourcemodule/classrooms.html
willgearty: The selects are made by the dynamic formset plugin, so I can't really move this there. Also, the `target` option below this indicates to `initialize` that it should only look for `select` objects within the specified object (the div with the id `furnishing_formset`), so that should keep the scope limited.

willgearty:
OK, I think that addresses all of your comments @benjaminjkraft, thanks!

And in terms of floating resources @kkbrum, I suppose they could also use a little love (although it would be a bit simpler since it wouldn't require formsets).
However, I was kinda hoping to get this cleaned up and merged so I could push it to the Stanford site soon (I know I can cherry-pick, but trying to avoid that if possible from now on).
I'd be more than happy to work on the floating resource stuff in a separate PR though!

willgearty:
Great, thanks! I've put the floating resources request in an issue (#2615).

PR #2548, opened 2018-04-10
add optional argument for nonstandard db names

milescalabresi:
This is my first time writing bash, so please correct me


benjaminjkraft:
This is actually Make, not bash!  So that won't work.  You want to use make variables, e.g. https://www.gnu.org/software/make/manual/html_node/Reference.html#Reference

milescalabresi:
derp. I should have known that. So then wouldn't it suffice to just put SITE=dbname when we run it? no changes required?

benjaminjkraft:
Aha, you're right! That should work I think.

PR #2549, opened 2018-04-11
Fix HTML formatting / Unicode bug in sending e-mails.

uakfdotb:


uakfdotb:
I tested this stops the dbmail_cron from running into Unicode UnicodeEncodeError, but wasn't able to test whether the resulting e-mail actually looks sane. (It's possible that we have to set some encoding metadata in the e-mail itself, in addition to doing this.)

benjaminjkraft:
Thanks!  @willgearty do you want to give this a try to see if it works right?

willgearty:
Is this the same hotfix that you did for our site @uakfdotb? If so, the emails looked fine. If not, we'd be happy to test it out (with test emails, not official emails, of course), so could someone push this to our site?

uakfdotb:
It is not exactly the same. This one is a cleaner change. I pushed it to stanford lu website. thanks!

willgearty:
The test email request was processed successfully. In general, everything looks fine. However, when I specifically open it in my gmail, the schedule table has extra rows:

image

I was able to fix this by removing the `<indent>` tags that were added with #2528, so I'm going to include that fix in this pull request.

kkbrum:
Nope, I think I was referencing something that had them and it seemed to work fine in the preview. But I would very much like the table not to have extra rows when sent (I tested it from the Yale site once it got updated and was also deeply saddened by the wonky rows). Thanks!

benjaminjkraft:
Cool, thanks y'all!

Issue #2509, opened 2018-02-27
Makefile should pull the branch, rather than just checking it out


benjaminjkraft:
The makefile (e.g. for stable releases) doesn't actually do a pull, just a fetch, which means if the branch already exists, it won't update it.  This is only a little bad if you use it on a stable release, and a lot bad if you try to use it on main.

PR #2553, opened 2018-04-16
pull origin, not just fetch

milescalabresi:
this way, it will retrieve updates to existing branches


milescalabresi:
Fixes #2509

benjaminjkraft:
It might be better to checkout `origin/$NEWBRANCH` instead.  I'm not sure if this will do the right thing in every case if we have a local branch in some unknown state.

milescalabresi:
oh, right. Better now?

benjaminjkraft:
Sorry, I should have been clearer: if you're using `origin/$NEWBRANCH` instead, you can just do the `fetch`, not the `pull`, which is simpler and has less that can go wrong.

PR #2558, opened 2018-04-17
Fix ordering of two-phase priorities

### LINKED ISSUE #2552: Ranking order in step 2 of two-phase lottery does not match that of /rank_classes
It's unclear what's going on here, but it looks like the ordering of priorities that is used for the mainpage of the two-phase registration does not match the ordering of priorities on the /rank_classes pages. I didn't change the sorting in #2425, so perhaps this is a problem that has been around but was never identified?

It appears that the ordering on /rank_classes is the correct ordering, since it stays the same based on the students' actual rankings.

Reported by Stanford.

willgearty:
Fixes #2552.

I wasn't able to replicate the odd behavior reported in #2552 on my dev server, but I replaced the old sorting behavior with a much better way of parsing through the priorities (using dictionary keys), which seems to be better nonetheless. It also fixed another bug that was introduced with #2425 where the priorities would shift upwards (in priority number) if there were skipped priority levels (in other words, if a student had set Priority/1 and Priority/3, they would be displayed as Priority/1 and Priority/2). This is also fixed by the new implementation.

image

### DIFF
--- a/esp/esp/program/modules/handlers/studentregtwophase.py
+++ b/esp/esp/program/modules/handlers/studentregtwophase.py
@@ -102,7 +102,7 @@ class StudentRegTwoPhase(ProgramModuleObj):
         priority_regs = priority_regs.select_related(
             'relationship', 'section', 'section__parent_class')
         for student_reg in priority_regs:
-            rel = student_reg.relationship
+            rel = student_reg.relationship.name
             title = student_reg.section.parent_class.title
             sec = student_reg.section
             times = sec.meeting_times.all().order_by('start')
@@ -148,14 +148,12 @@ class StudentRegTwoPhase(ProgramModuleObj):
 
             if timeslot.id in timeslot_dict:
                 priority_dict = timeslot_dict[timeslot.id]
-                # (relationship, class_title) -> relationship.name
-                priority_list = sorted(priority_dict.items(), key=lambda item: item[0].name)
             else:
-                priority_list = []
+                priority_dict = {}
             temp_list = []
             for i in range(0, context['num_priority']):
-                if i < len(priority_list):
-                    temp_list.append(("Priority " + str(i + 1), priority_list[i][1]))
+                if "Priority/" + str(i + 1) in priority_dict:
+                    temp_list.append(("Priority " + str(i + 1), priority_dict["Priority/" + str(i + 1)]))
                 else:
                     temp_list.append(("Priority " + str(i + 1), ""))
             priority_list = temp_list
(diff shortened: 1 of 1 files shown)


esp/esp/program/modules/handlers/studentregtwophase.py
```diff
@@ -148,14 +148,12 @@ def studentreg2phase(self, request, tl, one, two, module, extra, prog):
 
             if timeslot.id in timeslot_dict:
                 priority_dict = timeslot_dict[timeslot.id]
-                # (relationship, class_title) -> relationship.name
-                priority_list = sorted(priority_dict.items(), key=lambda item: item[0].name)
             else:
-                priority_list = []
+                priority_dict = {}
             temp_list = []
             for i in range(0, context['num_priority']):
```
benjaminjkraft: What's probably even better, here, is to use `relationship.displayName` as the description.  I think doing your own custom stringification is really not ideal here; much better to just sort by the `.name` but use the `.displayName`.  You might need to refactor how this works a little, but I think it will make things a lot clearer.

willgearty: Oh, I wasn't even aware of `.displayName`. In fact, it's fairly odd that that wasn't used from the get-go, especially for the user interface (the /rank_classes still uses the Priority/X format, and the main page used that format until I changed it with #2425).

I'm not totally sure I understand what you're looking for though, since it sounds like you want to go back to the sorting method that seemed bugged? Also, I'm not sure how we'd be able to accommodate empty/absent priorities without looping through the priorities that _should_ exist (via stringification) rather than just the priorities that _do_ exist.

willgearty: bump @benjaminjkraft

benjaminjkraft: Yeah, I guess it's ok to specifically look for `Priority/1` and so on to generate the priorities.  But instead of manually stringifying `Priority 1` for the display name, just use `relationship.displayName`, which on some sites is set to something even nicer like `First Choice`.

willgearty: Oh, okay. And if they are missing a priority, just use the "Priority X" text? Even if it won't match the other `displayName`s?

benjaminjkraft: Oh, I see the issue.  I think what you can do is look up the `RegistrationType` with that name, even if it's not attached to any `StudentRegistration`s for this student.  It should be guaranteed to exist.

willgearty:
Should be fixed now, unless there's any other tweaks that should be made. Tested with default `registrationType` names and custom `registrationType` names.

benjaminjkraft:
Great, thanks!

PR #2563, opened 2018-04-22
Add logging to lottery_execute errors

benjaminjkraft:
They were previously swallowing the error without logging it.


benjaminjkraft:
This didn't turn out to be the problem on stanford's site -- it was apache giving the 500, not us -- but it still seems like a good idea.

willgearty:
Looks fine to me!

Issue #2573, opened 2018-05-06
Handle invalid phone number and other exceptions from Twilio more gracefully


hwatheod:
When texting a group of students/teachers, if Twilio encounters an invalid phone number, it throws an exception (TwilioRestException) which in some cases causes an Internal Server Error on our side. We should handle this more gracefully, probably just ignoring invalid phone numbers.

For example, for a class cancellation where we text all students in the class (phone number masked):

Request Method: 	POST
Request URL: 	https://stanfordesp.org/manage/Splash/2018_Spring/manageclass/6475?action=cancel_cls
Django Version: 	1.8.14
Exception Type: 	TwilioRestException
Exception Value: 	

HTTP ERROR 400: 21211: The 'To' number +1XXXXXXXXXX is not a valid phone number. 

This can cause the cancellation to not be processed for some students (presumably the ones that occurred in the list after the student with the invalid phone number).

willgearty:
Would it be best to check each number to determine whether or not it is valid beforehand? Or can we just handle the twilio exceptions better (or at all)?

benjaminjkraft:
We should just handle the exceptions better.  There's no way we will be as good at validating phone numbers as they are, and presumably we can get errors that aren't an invalid phone number.

PR #2574, opened 2018-05-06
when cancelling a class, remove the students first before emailing/texting

hwatheod:
When cancelling a class, remove the students first before emailing/texting.

Otherwise, if the email or text operations fail, then the students don't get removed.

This mitigates the impact of #2573.


hwatheod:
Ah, tests failed because once you've cleared the students, then you can't email them anymore... will fix it.

hwatheod:
never mind, this needs a different fix.

PR #2580, opened 2018-06-20
Add buttons to change status of all classes on class search page

### LINKED ISSUE #2537: Approve all with X flag
At least at Stanford, we often use flags to indicate which classes _should_ be scheduled before they actually get approved. However, afterwards, it would save so much time to just click a single button which would approve all of the classes with such a flag.

willgearty:
Adds 3 buttons (analogous to the three buttons that are listed for each class on the search page) which allows the user to change the status of all of the classes that are currently being shown on the class search page. Works by clicking all of the individual buttons via jQuery.

image

Fixes #2537.

### DIFF
--- a/esp/public/media/scripts/program/modules/flag-results-page.js
+++ b/esp/public/media/scripts/program/modules/flag-results-page.js
@@ -59,6 +59,18 @@ function hideAll () {
     $j(".flag-detail").hide();
 }
 
+function approveAll () {
+    $j("#program_form .btn-approve").click();
+}
+
+function unreviewAll () {
+    $j("#program_form .btn-unreview").click();
+}
+
+function rejectAll () {
+    $j("#program_form .btn-reject").click();
+}
+
 $j(document).ready(function () {
     $j(".flag-detail").hide();
     $j(".manage-approve-link").hide();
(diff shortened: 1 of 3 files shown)


willgearty:
Part of me worries that this makes it too easy to accidentally click the new buttons...

esp/public/media/scripts/program/modules/flag-results-page.js
```diff
@@ -59,6 +59,27 @@ function hideAll () {
     $j(".flag-detail").hide();
 }
 
+function approveAll () {
+    var r = confirm("Are you sure you'd like to approve all of these classes?");
+    if (r) {
+        $j("#program_form .btn-approve").click();
```
benjaminjkraft: This seems likely to be fragile: if someone e.g. changes the name of the button they might not think to check here.  Instead, it would be much better to loop through each of the classes and call `approve` above.  You can either stick a list of IDs in the template somewhere, or pull them from the HTML if necessary.

esp/public/media/scripts/program/modules/flag-results-page.js
```diff
@@ -59,6 +59,27 @@ function hideAll () {
     $j(".flag-detail").hide();
 }
 
+function approveAll () {
+    var r = confirm("Are you sure you'd like to approve all of these classes?");
```
benjaminjkraft: I'd allcaps "all" to make it more obvious, but otherwise I think this is enough.

benjaminjkraft:
This looks great!  One minor comment but feel free to merge after that.

willgearty:
@benjaminkraft, did you forget the comment?

Issue #2468, opened 2017-11-09
Manage page to edit volunteer offers


willgearty:
I'm thinking of something like the check teacher availability page, but for volunteer offers.
This would allow us to

1. Check the specific offers of a given volunteer (rather than searching through the manage volunteer page or using the admin pages)
2. Edit the offers of a given volunteer (particularly after the volunteer deadline)

Issue #2496, opened 2018-02-05
check_availability styling like teacher availability


willgearty:
See #2419.

benjaminjkraft:
We could probably still do it!  But it would be a little more complicated to add the class codes (or maybe a hover-box for class details).

willgearty:
Just like I did with volunteer availability? That would definitely work! I was just worried it might make potentially important information harder to find, but I think it would be fine.

willgearty:
The new edit_availability page (#2583) now covers most of the functionality of check_availability, except for maybe the searching feature and unscheduled classes feature (both of which could be incorporated). So at this point perhaps it makes more sense to just phase one of the two out and merge the remaining features into whichever we choose to keep?

willgearty:
Fixed by #2583.

PR #2583, opened 2018-06-21
Changed volunteer reg layout to match that of teacher availability

### LINKED ISSUE #2495: Volunteer styling like teacher availability form
See #2419.

willgearty:
Modified the layout of volunteer registration to look like that of the teacher availability module (a la whenisgood). Since the same css is used for both, I moved all of that to a separate file and then confirmed the styling still worked for both. As with the teacher availability module, the old form is still there, but hidden, and the new layout interfaces with it via js/jquery. Also, because the details in the form are now hidden, I've added a little popup on the left which provides those details when a user hovers over a particular timeslot (see example below).

Fixes #2495.

New Layout:
image

Old Layout:
image


esp/esp/program/modules/handlers/volunteersignup.py
```diff
@@ -81,13 +82,19 @@ def signup(self, request, tl, one, two, module, extra, prog):
         if hasattr(request.user, 'email'):
             form.load(request.user)
 
-        #   Override default appearance; template doesn't mind taking a string instead
-        context['form'] = form._html_output(
```
benjaminjkraft: can you delete the form's `_html_output` method now, or is that a django thing?

esp/templates/program/modules/volunteersignup/signup.html
```diff
@@ -41,10 +42,88 @@
         }
     });
 });
+
+var setting = false;
+var down = false;
+
+//Records mousedown
```
benjaminjkraft: Wait, is this stuff all still duplicated?  Can we share it too?

esp/templates/program/modules/volunteersignup/signup.html
```diff
@@ -73,24 +152,73 @@ <h3>Volunteer Shifts Cancelled</h3>
...
-    <td colspan="2"><span style="color: red;">Please correct the errors in the form.</span></td>
-</tr>
+<p><span style="color: red;">Please correct the errors in the form.</span></p>
 {% endif %}
-    {{ form }}
-<tr>
-    <td colspan="2" align="center"><input id="volunteer_form_submit" type="submit" value="Submit" class="btn btn-primary" /></td>
-</tr>
-</table>
+{{ form.name.label_tag }}
```
benjaminjkraft: Can you loop through the fields here instead of hardcoding?

esp/templates/program/modules/volunteersignup/signup.html
```diff
@@ -73,24 +152,73 @@ <h3>Volunteer Shifts Cancelled</h3>
...
+{{ form.name.errors }}
+{{ form.name }}
+{{ form.email.label_tag }}
+{{ form.email.errors }}
+{{ form.email }}
+{{ form.phone.label_tag }}
+{{ form.phone.errors }}
+{{ form.phone }}
+
+  <div class="wrap">
```
benjaminjkraft: can this be shared, too?  I'm guessing it's pretty similar to the teacher version so you might be able to share the template.

benjaminjkraft:
(I should say -- if the answer is "yes, but I'm not sure what way is best", you can either take a stab at it and go from there, or you can bug me and I can take a closer look and make suggestions.)

esp/templates/program/modules/volunteersignup/signup.html
willgearty: Hmm...maybe? Although if I loop through all fields it'll include the confirm field...

willgearty: Oh, this might work: https://stackoverflow.com/a/21698814/4660582

willgearty:
I think you are correct that most of the js and html can be shared, so I'll work on that at some point. There might be a little bit of js that is unique to each (e.g. hover details for volunteer availability), but that should be easy enough to work around.

esp/esp/program/modules/handlers/volunteersignup.py
willgearty: It's a django thing

willgearty:
OK, after modifying the code a bit, I was able to pull out a substantial bit of JS and HTML that was shared between the teacher availability and volunteer signup pages. This should make it much easier to use the same layout in the future (e.g. #2468 and #2496). @benjaminjkraft, please let me know if you see anything else that can be shared and if there's anything else you think could be added/changed.

willgearty:
I was working on another feature, but I figured it could just be included here since it is very relevant (and dependent on the original pull request). The new addition adds a page analogous to check/edit_availability for volunteer offers. This page can be reached by clicking on any volunteer name on `/volunteering`. By default, the form on `/check_volunteer` does not allow editing, but this can by changed by clicking the button at the top of the page.

Fixes #2468.

willgearty:
Now shows the section that a teacher is teaching when hovering over a teaching block.

image

willgearty:
Availability form now shows unscheduled classes:

image

willgearty:
Phew that ended up being a bigger project than I thought it would, but I'm pretty happy with the outcome (a unified availability experience for teachers, volunteers, and admins).
It's all yours @benjaminjkraft! (thanks in advance)

esp/templates/program/modules/availabilitymodule/availability_form.html
```diff
@@ -2,107 +2,50 @@
...
-  background: #EFEFEF;
-}
+{% block xtrajs %}
+<script type="text/javascript" src="/media/scripts/program/modules/availability.js"></script>
 
-.canDo {
-  background: #00FF00;
-}
+<script>
+$j(document).ready(function() {
```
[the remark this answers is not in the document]
willgearty: Most of the stuff in this ready is actually specific to the teacher availability form, and there are analogous and very different functions for the volunteer form. This is because of the pretty different setup of the two original forms (for the teacher form we want to extract the data from the checkboxes and for the volunteer form we want to extract the labels for the checkboxes). I suppose I could move both versions to the JS file and specify which version to use in the `init` call? Let me see if there's a way to make the two forms more similar...

willgearty:
OK, I found a way to make the two forms a lot more similar than they were before, so I was able to pull a bunch of the JS out of the template files. I also renamed a number of the JS functions to make them more descriptive and unique. I think all that's left to do based on Ben's initial comments is to write up some more documentation on how to use this in the future and to maybe use IIFE/class and export/import (which I have very little experience with in JS).

willgearty:
And there we go! Let me know if that covers everything @benjaminjkraft or if you can think of anything else!

esp/public/media/scripts/program/modules/availability.js
```diff
@@ -0,0 +1,145 @@
...
+
+    return {
+        mouseup_event: mouseup_event,
+        mousedown_event: mousedown_event,
+        mouseover_event: mouseover_event,
+        header_switch: header_switch,
+        toggle_edit: toggle_edit,
+        init: init
+    };
+}());
```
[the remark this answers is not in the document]
willgearty: I was just following [this article](https://medium.com/@vvkchandra/essential-javascript-mastering-immediately-invoked-function-expressions-67791338ddc6), but the [MDN web docs](https://developer.mozilla.org/en-US/docs/Glossary/IIFE) agree with you.

willgearty:
Example teacher page:

image

PR #2585, opened 2018-06-21
Fix for classrosters template

willgearty:
There was a view/template mismatch that resulted in a blank page, as reported by UCSD.
I fixed it, but I'm not sure this should even exist anymore for teachers with the implementation of rosters with `section_students` and `class_students` in `teacherclassregmodule` (although the /manage version of this is probably still useful).


benjaminjkraft:
Thanks Will!  As you say, hopefully we can get rid of this -- I said a bit more on the email thread.  But I'm happy to fix in the meantime.

PR #2586, opened 2018-06-22
Add dates to volunteer offers

### LINKED ISSUE #2465: Volunteer offers don't show day
In the admin pages, the volunteer offers only specify the time of the offer, not the day.
This can be confusing when a program has a multi-day event and the descriptions are the same for all of the days.

willgearty:
Volunteer requests are now separated by date on `/volunteering`. I changed the displayed time in the first column so that it wasn't redundant with the new headers. While I was at it, I fixed up some of the admin page displays so that they now include the dates of the events.

Fixes #2465.

image


benjaminjkraft:
Sounds good!

PR #2591, opened 2018-07-09
Add coteacher deadline

### LINKED ISSUE #2371: Coteacher Deadline
There's currently no coteacher deadline (it depends on Teacher/MainPage).
This could be problematic for programs that need to finalize their teacher lists by a certain deadline for training, printing, etc.

Requested by Stanford.

willgearty:
Changes the deadline dependency of the coteachers page from `/Teacher/MainPage` to `/Teacher/Classes/Coteachers` so the coteachers page can be closed independently.

Fixes #2371.

### DIFF
--- a/esp/esp/users/migrations/0015_auto_20180709_1328.py
+++ b/esp/esp/users/migrations/0015_auto_20180709_1328.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('users', '0014_auto_20180217_1243'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='permission',
+            name='permission_type',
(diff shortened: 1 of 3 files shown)


esp/esp/users/migrations/0015_auto_20180709_1328.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
kkbrum: For my understanding-- what is happening with these migration files?

willgearty: The migrations change the way the database is set up. They can add fields, remove fields, or change fields. In this case, this is modifying the potential options for the permission field.

willgearty: Any time you make a modification to the database models, you need to create a migration for it. `fab manage:makemigrations` makes the migration files for you, which you then need to `git add`. To then update the database of your own dev server, you do `fab manage:migrate`.

kkbrum: oh cool! thanks

Issue #1394, opened 2014-10-17
Tag a classroom


benjaminjkraft:
MIT often wants to make a note about a particular classroom that will be visible in the scheduler (e.g. "this is a computer cluster" or "this is where we are scheduling messy classes").  Currently the only way is to change the name, which is a bit of a hack.

willgearty:
Fixed by #2595, I'd say.

PR #2595, opened 2018-07-10
Hidden resource types

### LINKED ISSUE #2593: Add classroom resources that aren't shown in teacherreg
There are often classroom "resources" (loosely defined) that are important to keep track of, particularly for scheduling, but are not important for teacher registration (e.g. "Computer Cluster", "Walkins Only", etc.). It should be fairly easy to add a model to resource types as to whether they should be shown in teacher registration or not. Perhaps this would also be a good fix for #1394, since these resources would then be visible in the scheduler?

Requested by Stanford.

willgearty:
This adds an option to resource types to have them hidden during teacher registration, which allows for associating resources with classrooms that admins may not want teachers to see when creating or editing classes (for scheduling purposes, etc.). These resources are still shown on the resources page under a separate heading, and I've added annotations in the classrooms and floating resources forms so that it is obvious which resources are hidden.

Fixes #2593 and maybe #1394.

### DIFF
--- a/esp/esp/program/models/__init__.py
+++ b/esp/esp/program/models/__init__.py
@@ -867,14 +867,17 @@ class Program(models.Model, CustomFormsLinkModel):
             return None
 
     @cache_function
-    def getResourceTypes(self, include_classroom=False, include_global=None):
-        #   Show all resources pertaining to the program that aren't these two hidden ones.
+    def getResourceTypes(self, include_classroom=False, include_global=None, include_hidden=False):
+        #   Show all resources pertaining to the program that aren't these hidden ones.
         from esp.resources.models import ResourceType
 
-        if include_classroom:
+        if include_hidden:
             exclude_types = []
         else:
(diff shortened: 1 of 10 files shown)


esp/esp/program/models/__init__.py
```diff
@@ -867,14 +867,17 @@ def date_range(self):
             return None
 
     @cache_function
-    def getResourceTypes(self, include_classroom=False, include_global=None):
-        #   Show all resources pertaining to the program that aren't these two hidden ones.
+    def getResourceTypes(self, include_classroom=False, include_global=None, include_hidden=True):
+        #   Show all resources pertaining to the program that aren't these hidden ones.
```
kkbrum: This comment is no longer accurate due to the extra argument, right?

willgearty:
@kkbrum I think you need to apply the migrations to add the new hidden field to your database.

PR #2597, opened 2018-07-11
Category fixes in catalog

### LINKED ISSUE #2567: Ordering/presence of categories at top of catalog
Categories are listed at the top of the course catalog and are currently sorted by the IDs of the categories (I think?), but the ordering should either:

1. match the ordering of the categories in the catalog (if the catalog is sorted by category/symbol)
or
2. be hidden if the catalog is not sorted by category (or category symbol) (this is [actually possible](https://github.com/learning-unlimited/ESP-Website/blob/main/esp/esp/program/models/class_.py#L219) with the `catalog_sort_fields` tag)

willgearty:
This makes sure that the categories at the top of the catalog are sorted properly (matching the sorting in the catalog). It also removes the category headings within the catalog if the catalog is not primarily sorted by category.

Fixes #2567.

### DIFF
--- a/esp/esp/program/modules/handlers/studentclassregmodule.py
+++ b/esp/esp/program/modules/handlers/studentclassregmodule.py
@@ -453,12 +453,25 @@ class StudentClassRegModule(ProgramModuleObj):
 
         categories = {}
         for cls in classes:
-            categories[cls.category_id] = {'id':cls.category_id, 'category':cls.category_txt if hasattr(cls, 'category_txt') else cls.category.category}
+            categories[cls.category_id] = {'id':cls.category_id, 'category':cls.category_txt if hasattr(cls, 'category_txt') else cls.category.category, 'symbol':cls.category.symbol}
+
+        # Is the catalog sorted by category? If so, by which aspect of category?
+        # Default is to sort by category symbol
+        cat_sort = 'category__symbol'
+        program_sort_fields = Tag.getProgramTag('catalog_sort_fields', prog)
+        if program_sort_fields:
+            cat_sort = program_sort_fields.split(',')[0]
+
+        cat_sort_split = cat_sort.split('__')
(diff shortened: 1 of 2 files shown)


esp/esp/program/modules/handlers/studentclassregmodule.py
```diff
@@ -453,12 +453,25 @@ def catalog_render(self, request, tl, one, two, module, extra, prog, timeslot=No
...
+        # Is the catalog sorted by category? If so, by which aspect of category?
+        # Default is to sort by category symbol
+        cat_sort = 'category__symbol'
+        program_sort_fields = Tag.getProgramTag('catalog_sort_fields', prog)
+        if program_sort_fields:
+            cat_sort = program_sort_fields.split(',')[0]
+
+        cat_sort_split = cat_sort.split('__')
+        if cat_sort_split[0] == 'category' and cat_sort_split[1] in ['id', 'category', 'symbol']:
+            categories_sort = sorted(categories.values(), key = lambda cat: cat[cat_sort_split[1]])
```
kkbrum: Could we rename "cat_sort" to be "catalog_sort"? "Cat" could be an abbreviation for categories and the distinction between cat_sort and categories_sort is a lot more clear if the first is catalog_sort.

esp/esp/program/modules/handlers/studentclassregmodule.py
```diff
@@ -479,7 +492,7 @@ def catalog_render(self, request, tl, one, two, module, extra, prog, timeslot=No
 
         class_category_id = None
         for cls in classes:
-            if cls.category.id != class_category_id:
+            if cls.category.id != class_category_id and categories_sort:
```
benjaminjkraft: oh man does this finally fix the old bug where if you don't sort by category you get a bazillion section headings!?

willgearty: Yes, finally!

PR #2598, opened 2018-07-13
Import resources from a previous program

### LINKED ISSUE #2288: Allow for resources to be imported from previous programs

willgearty:
This is a first stab at importing resources from a previous program. The current implementation allows you to import furnishings along with classrooms. If particular resource types do not exist yet, they are automatically created based on the resource types from the previous program. There's still a few things that I think need to be done:

1. Have a way to import resource types independent of importing classrooms. (should be fairly easy)
2. I made a comment about this in the code, but the furnishings can't be associated with the classrooms until the classroom resources are saved, which means there needs to be a way to temporarily associate them with the classrooms for the preview page.
3. Code reduction (there's a bit of code duplication at the moment)

(Includes changes from #2595)
Fixes #2288 (eventually).

### DIFF
--- a/esp/esp/program/models/__init__.py
+++ b/esp/esp/program/models/__init__.py
@@ -867,14 +867,17 @@ class Program(models.Model, CustomFormsLinkModel):
             return None
 
     @cache_function
-    def getResourceTypes(self, include_classroom=False, include_global=None):
-        #   Show all resources pertaining to the program that aren't these two hidden ones.
+    def getResourceTypes(self, include_classroom=False, include_global=None, include_hidden=True):
+        #   Show all resources pertaining to the program that aren't these hidden ones.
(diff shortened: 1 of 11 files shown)


kkbrum:
I get the same error here as in the hidden_resources pull request

kkbrum:
Can you make the "add/modify a floating resource" box the same size as the "import a floating resource" box right above it? (The sizes match for the times, classrooms, resources, so its weird that this one doesn't.

image

kkbrum:
And I know this isn't what you were working on... but while you're changing the sizes of things... the timeslots boxes are a different size than the boxes for classrooms/resources/floating resources and it's bothering me... but I'll still approve the pull request if you dont change this one :)
image

PR #2602, opened 2018-07-30
Onsite backlinks

### LINKED ISSUE #2364: Rapid check-in and other onsite pages should have backlinks to /main
This is useful for switching between onsite pages.

willgearty:
This adds backlinks to `onsite/_/_/main` to the onsite pages. I'm pretty sure I got all of the onsite pages, but please let me know if I missed any. While I was at it, I changed the manage backlinks to buttons instead of bland text/hyperlinks and added a little extra breathing room above the buttons.

Fixes #2364.

### DIFF
--- a/esp/esp/program/modules/handlers/onsiteclassschedule.py
+++ b/esp/esp/program/modules/handlers/onsiteclassschedule.py
@@ -84,7 +84,7 @@ class OnsiteClassSchedule(ProgramModuleObj):
         """ Redirect to student registration, having morphed into the desired
         student. """
 
-        user, found = search_for_user(request, ESPUser.getAllOfType('Student', False))
+        user, found = search_for_user(request, ESPUser.getAllOfType('Student', False), add_to_context = {'tl': 'onsite'})
         if not found:
             return user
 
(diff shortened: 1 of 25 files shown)


willgearty:
Thanks!
Added the link to:

- List of open classes
- Unenroll students

Encountered the following issues/conundrums:

- Grid based class changes: this is intended to be the entire screen, so not really sure where to stick the link
- Scheduling and class changes: this uses a generic template and doesn't even pass any sort of context, so I don't think I can add the link only when it's being used for onsite
- View purchased items for a student: same as scheduling and class changes

benjaminjkraft:
Yeah I think it's fine to skip the grid.  You could also skip the ones where the user is morphed, since they have the button in the sidebar/menu, but if they're reasonable to add, no reason not to.

willgearty:
OK, I modified the link on the student registration page to be a button just like these other buttons. Were there any other pages you were thinking of @benjaminjkraft ?

benjaminjkraft:
Oh sorry I was just saying you could leave it as is if you want, but that's fine too!

PR #2603, opened 2018-07-30
Buttons to next and previous teacher checkin pages

### LINKED ISSUE #2293: Teacher check-in URL is strange
The URL is not friendly to URL hacking, e.g: "?start=1254 is 2pm""

willgearty:
I added buttons that go to the next and previous teacher checkin pages (hours or days, depending on what is being used) if they exist. For example, the first hour looks like this:

image

And the second hour looks like this:

image

Fixes #2293.


kkbrum:
When I press "next time slot" or "previous time slot" to checkin teachers, things break. If I just press the timeslot from the original page, things are fine, so something is wrong with the next/previous buttons. I don't know what exactly will help you debug and my brain is too tired to start debugging, but this is the error I get. The "8/14/2018 11:59" it mentions is the date for the "teachers missing as of ______". Let me know what other information would be helpful to reproduce/ help debug.

image

PR #2609, opened 2018-08-25
add background check and coteacher choices to classreg and minors policy

lukeshimanuki:
minors policy acknowledgement: asks whether mit affiliate, recently checked, or commits to getting checked (i'm having trouble getting it to actually save the selection though, if someone could help with this)
classreg: asks whether you commit to finding a coteacher or commit to observing as many hours as you teach


esp/esp/program/modules/handlers/minorspolicymodule.py
```diff
@@ -38,15 +46,27 @@ def minorspolicyacknowledgement(self, request, tl, one, two, module, extra, prog
         context = {'prog': prog}
         if request.method == 'POST':
             context['form'] = minorspolicyacknowledgementform_factory(prog)(request.POST)
+            selection = context['form']['background_check']
```
betaveros: I think the reason your code wasn't working was here: `selection` is a `BoundField` rather than a string; you want `selection.data`.

lukeshimanuki:
replaced by #2613

PR #2611, opened 2018-08-26
Handle Twilio exceptions gracefully

### LINKED ISSUE #2573: Handle invalid phone number and other exceptions from Twilio more gracefully

willgearty:
This handles Twilio exceptions and includes them in the logs when sending text messages. This also then catches whether there is an error when texting teachers for teacher checkin. To save space, the text on the teacher checkin page only says whether there was an error or not.

I tested this successfully with the phone number 000-000-0000.

Fixes #2573.

### DIFF
--- a/esp/esp/program/modules/handlers/teachercheckinmodule.py
+++ b/esp/esp/program/modules/handlers/teachercheckinmodule.py
@@ -51,7 +51,6 @@ from django.template.loader import render_to_string, select_template
 from django.db.models.aggregates import Min
 from django.db.models.query   import Q
 from datetime import datetime, timedelta
-from twilio.rest import TwilioRestClient
 
 import collections
 import json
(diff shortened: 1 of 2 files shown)


esp/esp/program/modules/handlers/teachercheckinmodule.py
```diff
@@ -169,8 +168,11 @@ def ajaxteachertext(self, request, tl, one, two, module, extra, prog):
                 sec = ClassSection.objects.get(id=request.POST['section'])
                 teacher = PersistentQueryFilter.create_from_Q(ESPUser, Q(username=request.POST['username']))
                 message = "Don't forget to check-in for your " + one + " class that is scheduled for " + sec.friendly_times(include_date = True)[0] + "!"
-                GroupTextModule.sendMessages(teacher, message, True)
-                return {'message': "Texted teacher"}
+                log = GroupTextModule.sendMessages(teacher, message, True)
+                if "error" in log:
```
benjaminjkraft: Will this cause problems if the teacher's name contains "error", or something?  It's a little hacky but maybe fine.

PR #2612, opened 2018-08-28
Manage class duration

### LINKED ISSUE #2025: Class durations are not shown in manageclass or classsearch

willgearty:
Adds ability to edit the duration of a class on the manage page.

Fixes #2025.

### DIFF
--- a/esp/esp/program/modules/forms/management.py
+++ b/esp/esp/program/modules/forms/management.py
@@ -46,6 +46,9 @@ class ClassManageForm(ManagementForm):
                 prefix = kwargs['prefix'] + '-'
             initial_dict = self.load_data(self.cls, prefix)
             super(ClassManageForm, self).__init__(data=initial_dict, *args, **kwargs)
+            if self.cls.hasScheduledSections():
+                self.fields['duration'].widget.attrs['disabled'] = True
+                self.fields['duration'].widget.attrs['title'] = "At least one section of this class has already been scheduled"
         else:
             super(ClassManageForm, self).__init__(*args, **kwargs)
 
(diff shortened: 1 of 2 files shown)


kkbrum:
Changing the class duration isn't updating the durations of the sections that already exist

willgearty:
Should now update the section durations @kkbrum

willgearty:
One thought I did have, though, was if we want to do anything special when an admin tries to change the duration of a scheduled class? Like tell them "no"?

kkbrum:
We probably at least want a warning message asking if they want to proceed knowing this could cause:
- teachers to be double booked (if lengthening class)
- students to be double booked (if lengthening) or have an empty block (if shortening)
- rooms to be double booked (if lengthening)
There's little harm in the case of shortening a block, so I think we should potentially let admins proceed if they understand the potential implications?

A warning that goes above and beyond would probably check teacher and room schedules and make sure double booking isnt happening, and checks whether students have signed up yet. That's probably much effort though and a simple warning asking the admin to be certain is probably fine?

willgearty:
Will changing the duration of a scheduled section actually impact many things? If it's already scheduled, the room resource is already set up for the original duration, right? And so it might not even impact schedules, etc? I'm leaning towards just a little note next to the duration field, but I guess I could be persuaded to have a more dramatic and situational warning.

kkbrum:
Oh hmm i just realized what i was originally thinking was that the changed duration would auto update the scheduling, but that wouldnt make sense because it's unclear whether to shorten/lengthen from the end or the beginning of the time. I think I agree now with your original thought of preventing people from changing duration (at least on this page) of a scheduled class... there's not really a good way to deal with it so might as well force the admin to unschedule and then do it so that nothing gets messed up.

willgearty:
That sounds good to me! Check it out now and let me know what you think.

PR #2613, opened 2018-08-28
add background check and observer option in minors policy form

lukeshimanuki:


willgearty:
This looks like a test error that we were getting on main a bit ago. Try merging main and see if it fixes the test?

lukeshimanuki:
i'll defer to @betaveros on whether we want to do that

benjaminjkraft:
This is a merge to mit-prod so feel free to do whatever you want, you get to keep the pieces if it breaks :^] .

benjaminjkraft:
Oh, except, I should note: this looks like it may at least in theory require migrations.  Don't put migrations in mit-prod if they're not also identical in main, it's a pain to clean up.  Feel free to make a separate PR to merge just the model changes, I don't see a problem with having those in main if it's convenient.

benjaminjkraft:
Django has made the (IMO questionable) choice that any changes to models, even if they don't actually affect the DB schema, require migrations.  So the changes to `EVENT_CHOICES` technically do.  It may be possible to just pretend they don't, though, since the DB won't care -- just be careful about creating any migrations on top of these as django will try to include these changes too.  (But I guess I just suggested you avoid migrations on top of these anyway!)

betaveros:
There are already much worse migrations on mit-prod, so...

PR #2618, opened 2018-09-07
Program specific teacher reg grade ranges

### LINKED ISSUE #2617: Grade ranges for teacher reg should be program specific
#2414 uses Tag.getTag to retrieve the potential grade range options, but really it should be Tag.getProgramTag in case the choices should be different for different programs.

willgearty:
Looks for grade ranges for the specific program first, then global.
Fixes #2617.


benjaminjkraft:
Thanks!

PR #2619, opened 2018-09-17
Make links for adding class more obvious

wishdasher:
It's so small right now.

image


willgearty:
Closing because I've included these changes in #3249.

PR #2621, opened 2018-09-20
Fix class import error when using grade ranges with new minimum grade for program

willgearty:
Stanford ran into an error that only arises when you try to import an old class (that included the old minimum grade) while using the grade ranges option and a new, higher minimum grade for the new program. I originally had the other grade fields hidden in the form, but when the minimum grade for the program is changed and the old imported class included the old minimum grade, that old minimum grade is then hidden in the `grade_min` field, which then causes an error each time the form is submitted. So here I went all the way and deleted the unnecessary grade fields instead of just hiding them (`grade_ranges` when using minimum and maximum grades, `grade_min` and `grade_max` when using grade ranges).

### DIFF
--- a/esp/esp/program/controllers/classreg.py
+++ b/esp/esp/program/controllers/classreg.py
@@ -1,7 +1,7 @@
 from esp.mailman import add_list_member
 from esp.program.models import Program, ClassSubject, ClassSection, ClassCategories, ClassSizeRange
 from esp.middleware import ESPError
-from esp.program.modules.forms.teacherreg import TeacherClassRegForm
+from esp.program.modules.forms.teacherreg import TeacherClassRegForm, TeacherOpenClassRegForm
 from esp.resources.forms import ResourceRequestFormSet
 from esp.resources.models import ResourceType, ResourceRequest
 from esp.tagdict.models import Tag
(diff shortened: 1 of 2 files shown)


willgearty:
Looks like this might have broken walk-in registration, so I'm currently investigating a fix for that.

willgearty:
OK, so that fixes one problem, but then reveals another: users can't import walkins from a program with a smaller minimum grade or greater maximum grade than the current program. Looking into how to fix that, but I think the easiest fix might be to go to no `grade_min` or `grade_max` fields on the open class form and then fill in the grades upon form submission/class creation.

willgearty:
OK, I think that should do it. I've tested importing regular classes and walkins and creating new regular classes and walkins (all for a new program with a greater minimum grade than the program from which the classes are being imported).

benjaminjkraft:
Yeah, this seems nicer -- I hadn't realized we did it that way.  If you can just make extra sure you've tested both import and create for both programs with a grade range and with min/max, go ahead and merge!

willgearty:
The following work on my dev server:

- Walk-in import for program with grade min/max (uses program min/max no matter original class's grades)
- Walk-in import for program with grade ranges (uses program min/max no matter original class's grades)
- New walk-in for program with grade min/max (with program min/max)
- New walk-in for program with grade ranges (with program min/max)
- Regular import for program with grade min/max (if either grade limit is outside program grade limits, field is blank on import)
- Regular import for program with grade ranges (if grades don't match any range options, field is blank on import)
- New regular class for program with grade min/max (with set grades)
- New regular class for program with grade ranges (with set grades)

Issue #2568, opened 2018-04-26
Module to integrate with Eventbrite


willgearty:
Sadly, Stanford can't use the credit card module and needs to use Eventbrite to collect credit card payments instead. It would be nice to integrate this with the website (right now we just have a widget on a template-overridden /extraform page). It looks like Eventbrite has an [API](https://www.eventbrite.com/developer/v3/), which is good, but it doesn't look like you can actually process payments through the API. Payments could be made through a widget of some sort, and then we could receive notification of the successful payments via [webhooks](https://www.eventbrite.com/developer/v3/api_overview/webhooks/) and then somehow associate these payments with particular accounts?

For the moment, I'm just putting this here to remind me to look into this possibility at some point in the not-too-distant future (unless this would also help other programs).

benjaminjkraft:
Yeah, probably we would link users to the eventbrite page, and use the webhook to know that they paid.

willgearty:
Resources for later:

[Eventbrite Python SDK](https://github.com/eventbrite/eventbrite-sdk-python)
[Eventbrite Webhooks Tutorial](https://www.eventbrite.com/developer/v3/resources/webhooks_tutorial/) (and [code](https://github.com/eventbrite/webhooks-tutorial-flask))
[Eventbrite Webhooks](https://www.eventbrite.com/developer/v3/api_overview/webhooks/)

willgearty:
#2623 is a sufficient (and more flexible) fix for this at the moment, but I'll leave this open in case someone (incl. me) wants to work on something with webhooks at a later time.

PR #2623, opened 2018-10-01
Script for settling accounting balance for list of students

willgearty:
In case you have students that have paid for the program in advance through some external site (e.g. Eventbrite {#2568}), this allows you to take a list of those students' usernames and settle their accounts for a given program. This will then result in them having a $0 balance, which can then be displayed on the website or on printables (e.g. schedules). The list should be provided as a csv with a column titled 'username'. All other columns will be ignored.


esp/useful_scripts/csv_student_payment.py
```diff
@@ -0,0 +1,61 @@
...
+from esp.accounting.controllers import IndividualAccountingController
+
+import csv
+import os
+import argparse
+
+from datetime import datetime
+
+
+parser = argparse.ArgumentParser(description = "Add accounting records for students that have already paid")
```
benjaminjkraft: Usually good practice is to wrap the entire script in
```
def main():
    ...
```
and then at the end you do
```
if __name__ == '__main__':
    main()
```
so that you can import the script without running it.

PR #2640, opened 2018-11-10
Fix onsite_create

willgearty:
This fixes a couple bugs in `onsite_create` (the new student registration in onsite). I've already pushed it to SR10 at the request of Miles and UMBC.


benjaminjkraft:
Eek, thanks!

PR #2641, opened 2018-11-11
Stats for registered and approved classes

willgearty:
This removes cancelled and rejected classes from the calculations of the class-hours and class-student-hours in the admin dashboard and teacher big board. I think this makes more sense when setting a goal for a program and comparing across multiple programs.

### DIFF
--- a/esp/esp/program/modules/handlers/jsondatamodule.py
+++ b/esp/esp/program/modules/handlers/jsondatamodule.py
@@ -755,7 +755,9 @@ teachers[key].filter(is_active = True).distinct().count()))
 
 
         shours = 0.0
+        shours_app = 0.0
         chours = 0.0
+        chours_app = 0.0
         crhours = 0.0
         ## Write this as a 'for' loop because PostgreSQL can't do it in
         ## one go without a subquery or duplicated logic, and Django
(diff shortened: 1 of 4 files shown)


benjaminjkraft:
Huh, not sure.  It could be the `sections__` is doing something funky.  You'd have to post a bit more context for the graph for me to tell.  In any case, this looks fine if you can figure out the bug!

willgearty:
Alright, I'll look into it. Also, now that I think about it, would it make sense to include both numbers? All hours and non-cancelled hours?

willgearty:
That weird bug was related to lunch and is now fixed in #2685. I have an idea for how to better display the registered vs. approved numbers, so let me give that a shot.

willgearty:
OK, all set to be reviewed!

esp/templates/program/modules/bigboardmodule/bigboard.html
```diff
@@ -24,86 +24,104 @@
...
+      },
+      yAxis: [{
+        title: {text: null},
+        min: 0,
+        maxPadding: 0.01,
+        endOnTick: false,
+      }
+      {% if right_axis_data %}
+      ,{
+        title: {text: 'class-hours'},
```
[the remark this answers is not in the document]
willgearty: Yeah, because we now have two series for each of the right axes. Any ideas for how to keep this general?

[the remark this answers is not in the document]
willgearty: Never mind, I figured out a better way to do everything.

willgearty:
OK, I removed a lot of the duplicated code, generalized the bigboard template a LOT, and added functionality to show a blank graph when there isn't any data (instead of when there was just nothing or a blank spot before).

esp/esp/program/modules/handlers/jsondatamodule.py
```diff
@@ -689,6 +689,16 @@ def message_requests():
...
     message_requests.method.cached_function.depend_on_model('dbmail.MessageRequest')
 
+    @staticmethod
+    def calc_hours(classes):
+        hours = {"class-hours": 0, "class-student-hours": 0, "class-registered-hours": 0}
+        for cls in classes:
+            if cls['subject_duration']:
+                hours["class-hours"] += float(cls['subject_duration'])
+                hours["class-student-hours"] += float(cls['subject_duration']) * (float(cls['class_size_max']) if cls['class_size_max'] else 0)
+                hours["class-registered-hours"] += float(cls['subject_duration']) * float(cls['subject_students']) / float(cls['num_sections'])
```
kkbrum: Why do we divide by the number of sections here?

willgearty:
OK, it looks like some themes have default line heights, which mess up the display of the numbers on the big board (I've documented this in #2695). I'll work on that in a separate PR.

esp/esp/program/modules/handlers/jsondatamodule.py
willgearty: That is a good question...

willgearty: Oh, I just remembered why: `subject_duration` is the sum of the durations of the sections of the class. Each student only attends 1 section, so we need to divide `subject_duration` by `num_sections` to get the actual duration of the class that the student is registered for.

PR #2643, opened 2018-11-17
Beep for mobile device student checkin

willgearty:
This adds a beep for when a barcode is scanned using the mobile device checkin (#2539). The beep only occurs if the scanner detects a new barcode, so you won't get a whole bunch of beeps if you accidentally scan the same barcode multiple times.
I also increased the font size of the checkin message based on feedback from testing of this feature at Stanford.


benjaminjkraft:
Heh, nice idea!

PR #2649, opened 2019-01-06
Sort fillslot like catalog

### LINKED ISSUE #2645: Category fixes for fillslot
I previously fixed a bunch of catalog sorting stuff in #2597, but those fixes need to be added to the `fillslot` logic as well.

willgearty:
This takes the fixes for category sorting and display from #2597 and implements them in the `fillslot` logic. I also removed the old class sorting logic from the fillslot code, since we want to be consistent and sort the classes as they are sorted in the normal catalog (which is already done via `ClassSubject.objects.catalog()`).

Fixes #2645.

### DIFF
--- a/esp/esp/program/modules/handlers/studentclassregmodule.py
+++ b/esp/esp/program/modules/handlers/studentclassregmodule.py
@@ -422,6 +422,7 @@ class StudentClassRegModule(ProgramModuleObj):
         is_onsite = user.isOnsite(self.program)
 
         #   Override both grade limits and size limits during onsite registration
+        #   Classes are sorted like the catalog
         if is_onsite and not 'filter' in request.GET:
             classes = list(ClassSubject.objects.catalog(self.program, ts))
         else:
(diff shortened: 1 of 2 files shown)


esp/esp/program/modules/handlers/studentclassregmodule.py
```diff
@@ -430,19 +431,28 @@ def fillslot(self, request, tl, one, two, module, extra, prog):
...
-
         categories = {}
 
         for cls in classes:
-            categories[cls.category_id] = {'id':cls.category_id, 'category':cls.category_txt if hasattr(cls, 'category_txt') else cls.category.category}
+            categories[cls.category_id] = {'id':cls.category_id, 'category':cls.category_txt if hasattr(cls, 'category_txt') else cls.category.category, 'symbol':cls.category.symbol}
+
+        # Are the classes sorted by category? If so, by which aspect of category?
+        # Default is to sort by category symbol
+        catalog_sort = 'category__symbol'
```
hwatheod: Can you factor out the logic in lines 441-450 into its own function to avoid duplicate code between fillslot and catalog_render?

willgearty: Done!

PR #2653, opened 2019-01-16
Selective resource import

### LINKED ISSUE #2457: Import a subset of classrooms from previous program
Right now you can import the entirety of a previous program's classrooms, but it would also be helpful to import just a subset of classrooms from a previous program.

Requested by Stanford.

willgearty:
This allows for selectively importing resources from a previous program (timeslots, resource types, classrooms, and floating resources). Instead of always importing an entire program's-worth of resources, admins can now select which of the old resources they'd like to import.

I've successfully tested importing all resources, some resources (where they aren't already imported), some resources (where they are already imported), and no resources.

Fixes #2457.

### DIFF
--- a/esp/templates/program/modules/resourcemodule/timeslot_import.html
+++ b/esp/templates/program/modules/resourcemodule/timeslot_import.html
@@ -47,7 +47,7 @@ You have chosen to import the timeslots from {{ past_program.niceName }} to {{ p
         <td class="underline">{{ t.start.date }}</td>
         <td class="underline">{{ t.start.time }}</td>
         <td class="underline">{{ t.end.time }}</td>
-        <td class="underline"><input type="checkbox" name="to_import" value="{{ t.short_description }}_{{ t.start|safe }}" checked></td>
+        <td class="underline"><input type="checkbox" name="to_import" value="{{ t.old_id }}" checked></td>
     </tr>
     {% endfor %}
     </table></td></tr>
--- a/esp/templates/program/modules/resourcemodule/timeslot_import.html
+++ b/esp/templates/program/modules/resourcemodule/timeslot_import.html
@@ -7,6 +7,14 @@
     <link rel="stylesheet" href="/media/styles/forms.css" type="text/css" />
 {% endblock %}
 
+{% block xtrajs %}
+<script>
+function toggle(source) {
+  $j('input[name=to_import]').prop('checked', source.checked);
+}
+</script>
+{% endblock %}
+
 {% block content %}
 
 <h1>Manage Resources for {{ prog.niceName }}</h1>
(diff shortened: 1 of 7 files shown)


willgearty:
Doh! How silly of me to forget a screenshot!

image

esp/templates/program/modules/resourcemodule/timeslot_import.html
```diff
@@ -26,22 +34,24 @@ <h2>Confirm Timeslot Import</h2>
...
         <td class="underline"><b>Start</b></td>
         <td class="underline"><b>End</b></td>
+        <td class="underline"><b>Import? <input type="checkbox" onclick=toggle(this) checked></b></td>
     </tr>
     {% for t in new_timeslots %}
     <tr>
         <td class="underline">{{ t.start.date }}</td>
         <td class="underline">{{ t.start.time }}</td>
         <td class="underline">{{ t.end.time }}</td>
+        <td class="underline"><input type="checkbox" name="to_import" value="{{ t.short_description }}_{{ t.start|safe }}" checked></td>
```
hwatheod: Why is this particular field t.start marked safe while nothing else in the rest of your changes is?

willgearty: It's a date-time object, which well get auto-formatted by Django in the template, but I just want the python string representation of the object so that it matches the logic in the handler. It's definitely possible there's a more appropriate template tag, but safe seemed to work for my purposes.

hwatheod: Got it.  Can you make a note of this in a comment?  It's a little bit confusing since this is not what `safe` is usually used for.

Otherwise, looks good, feel free to merge.

willgearty: Actually, now that I think about it, I'm not sure why I didn't do this in the first place, but it might just be easiest to switch to the IDs of the *old resources* that are getting copied to the new program. My original intent was to try to pick something that would be unique to the *new resources* (in this case, I assumed the combination of the short description and start time should be unique), but since we are looping over the old resources anyways, I believe I can just use the IDs, which are guaranteed to be unique. I think I originally thought of using IDs, but immediately dismissed it because they would have been the IDs of the *new resources*, which will be different when saving because the objects weren't saved the first time through.

willgearty: Ah, I remember why it wouldn't work: the IDs of the *old resources* aren't pushed to the preview page. I'll just need to add them as another attribute to the *new resources* when pushing them to the preview page (and only then).

willgearty:
OK, I've tested this new version, and it seems to work well. @hwatheod, can you take another quick look? Thanks!

hwatheod:
Merged. Using the old ids is much better than creating custom unique ids on the fly.

PR #2655, opened 2019-01-21
Allow .ico uploads on filebrowser

### LINKED ISSUE #2608: Allow uploading css files to filebrowser
Right now you need to upload as .txt and then change the name to .css.

willgearty:
This extends the default allowed filetypes for filebrowser uploads to include .ico, .css, and .js, since these are all common filetypes that chapters might want to upload to change or supplement default files (including theming, scripting, and favicons).

Fixes #2608.

### DIFF
--- a/esp/esp/django_settings.py
+++ b/esp/esp/django_settings.py
@@ -366,6 +366,21 @@ CYBERSOURCE_CONFIG = {
 #   (not just '/media/uploads/' which is the default)
 FILEBROWSER_DIRECTORY = ''
 
+FILEBROWSER_EXTENSIONS = {
+    'Image': ['.jpg','.jpeg','.gif','.png','.tif','.tiff','.ico'],
+    'Document': ['.pdf','.doc','.rtf','.txt','.xls','.csv','.css'],
+    'Video': ['.mov','.wmv','.mpeg','.mpg','.avi','.rm'],
+    'Audio': ['.mp3','.mp4','.wav','.aiff','.midi','.m4p'],
+    'Script': ['.js'],
+}
+
+FILEBROWSER_SELECT_FORMATS = {
+    'file': ['Image','Document','Video','Audio','Script'],
+    'image': ['Image'],
+    'document': ['Document'],
+    'media': ['Video','Audio'],
+}
+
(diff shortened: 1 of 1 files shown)


benjaminjkraft:
I'm not actually sure if this is a good idea.  (For js/css, ico is fine.)  I worry that admin custom js/css can hook into lots of things we don't expect and break things in surprising ways on updates.  In general, if someone has written enough js/css that they want to put it in a real file rather than inline in editable text, it's probably worth actually making a pull request so as to make things that much more maintainable.

willgearty:
At least for CSS, I'm mostly familiar with Stanford's site, which has a custom theme built around CSS files. However, I guess that only reinforces your argument about making the site weird for updates (since Stanford's site has been notoriously weird). I thought all of these were uploaded through the filebrowser, but it actually looks like they are part of the stanford-prod branch. So after thinking about it, I think I agree with you here.

For JS, I mostly did it because there are so many JS files and it adds a nice little "Script" type label once it is defined in the settings, but it also then allows you to upload JS files, and JS scripts can just be included in editable text or template overrides, so I guess I agree with you here as well.

Ultimately, perhaps you are right in that we should only add the .ico functionality. The only reason I can see for including .css and .js is that it adds the nice little type labels to the filebrowser so the user isn't so overwhelmed with all of the unlabelled stuff. But ultimately, it's probably just opening a can of worms that we don't want to deal with in the future.

benjaminjkraft:
Yeah, custom themes are an exception.  We could even make that explicit, if you have that setting set then you are allowed to upload at least css.

For JS, I do realize it would make putting JS in your editable text and template overrides simpler.  I just don't think we want to encourage that -- it made e.g. #2047 require a lot more care because I had to check all the sites for JS or TOs that might use it.

Anyway, I don't have to be the final arbiter on this, just wanted to point out a reason we might not want to.

willgearty:
I'd be fine with adding the ability to upload .css and/or .js to a chapter's local_settings.py on a case-by-case basis. And I can reduce this PR to just .ico.

benjaminjkraft:
Sorry for the delay in approving -- this seems fine!

PR #2656, opened 2019-01-21
Printable of classes sorted by teachers

### LINKED ISSUE #1220: Printable: list of classes by teacher
It would be useful to have a printable for a list of classes, sorted by teacher.

willgearty:
Uses the last name of the first teacher of each class to sort classes. If a class has no teacher(s) (e.g. Lunch), it is excluded.

Fixes #1220.

### DIFF
--- a/esp/esp/program/modules/handlers/programprintables.py
+++ b/esp/esp/program/modules/handlers/programprintables.py
@@ -400,6 +400,22 @@ class ProgramPrintables(ProgramModuleObj):
 
     @aux_call
     @needs_admin
+    def classesbyteacher(self, request, tl, one, two, module, extra, prog):
+        def cmp_teacher(one, other):
+            cmp0 = cmp(one.get_teachers()[0].last_name.lstrip(), other.get_teachers()[0].last_name.lstrip())
+
+            if cmp0 != 0:
+                return cmp0
+
+            return cmp(one, other)
+
+        def filt_teacher(cls):
+            return len(cls.get_teachers()) > 0
+
+        return self.classesbyFOO(request, tl, one, two, module, extra, prog, cmp_teacher, filt_teacher)
+
+    @aux_call
+    @needs_admin
     def classesbyroom(self, request, tl, one, two, module, extra, prog):
         def cmp_room(one, other):
             qs_one = one.initial_rooms()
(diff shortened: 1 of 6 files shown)


esp/esp/program/modules/handlers/programprintables.py
```diff
@@ -398,6 +398,22 @@ def cmp_title(one, other):
 
         return self.classesbyFOO(request, tl, one, two, module, extra, prog, cmp_title)
 
+    @aux_call
+    @needs_admin
+    def classesbyteacher(self, request, tl, one, two, module, extra, prog):
+        def cmp_teacher(one, other):
+            cmp0 = cmp(one.get_teachers()[0].last_name.lstrip(), other.get_teachers()[0].last_name.lstrip())
```
benjaminjkraft: Huh, so this is interesting -- do you want to sort the teacher names?  Or do you want to change things to include an entry for each class under each teacher name?  Right now this seems like it will sort by the name of an arbitrary teacher, which is not great, and even sorting by the name of the alphabetically-first teacher is not ideal.

willgearty: Hmm...This is certainly the *easiest* way to do it, although just taking the alphabetically-first teacher would be just as easy. However, I agree it probably makes the most sense to have an entry for each teacher? But that won't be doable with `classesbyFOO` as it currently exists. I guess all we would need to do is add a little bit of code in `classesbyFOO` to duplicate classes for each teacher (and assign that teacher as an attribute).

benjaminjkraft: Yeah, it's true.  Not sure how much it matters to admins, it just seems like it would be very confusing to me to have a list of classes by teacher that doesn't list my class under my name because I have an alphabetically-earlier coteacher.

willgearty:
@benjaminjkraft I couldn't make up my mind which way would be best, so I implemented both! One way uses the first teacher from the sorted list of teachers of each class to sort the class list. The other way splits out the teachers and sorts by teacher last name. Let know what you think.

esp/templates/program/modules/programprintables/classes_list.html
```diff
@@ -56,8 +56,13 @@
  <td sorttable_customkey="{{ cls.id }}">{{ cls.emailcode }}</td>
  <td>{{ cls.title }}</td>
  <td>{{ cls.class_info|truncatewords_html:10 }}
+ {% if split_teachers %}
+ <td>{{ cls.split_teacher.name }}</td>
```
benjaminjkraft: I wonder if it would actually be clearer to show all teachers of the class -- so you sort it under each teacher, but show all of them.

willgearty:
Yeah, that all makes sense @benjaminjkraft. I've reduced it to just a single printable, sorted by each teacher, but then showing all teachers. To make it clear which teacher it was sorted on, I bolding their name and phone number. Let me know what you think!

willgearty:
@benjaminjkraft Can you take a quick look at the updates to this?

benjaminjkraft:
Cool, thanks!

Issue #2630, opened 2018-10-12
Remove rawgit links


benjaminjkraft:
https://rawgit.com/
needs to happen sometime by Oct 2019.

PR #2660, opened 2019-01-27
Move from rawgit to jsdelivr

willgearty:
I've replaced all rawgit links that I could find in the codebase with links to [jsdelivr](https://www.jsdelivr.com/rawgit).


willgearty:
Fixed #2630.

PR #2663, opened 2019-02-03
return 0 instead of None for amount_requested in accounting/controllers.py

hwatheod:
Fixes the following exception occurring in amount_due() line 568:

  "unsupported operand type(s) for -: 'NoneType' and 'Decimal'"


willgearty:
Yep, looks good to me!

PR #2664, opened 2019-02-03
Remove unused HttpResponse in set_donation_amount

hwatheod:
We are seeing exceptions like:
   Decimal('20.00') is not JSON serializable
when forming the HttpResponse. But the response is not actually used.
set_donation_amount is only called as an Ajax call from these places:
  templates/program/modules/creditcardmodule_stripe/cardpay.html
  templates/program/modules/donationmodule/donation.html
and both of them discard the result.


willgearty:
Yep, looks good to me!

PR #2665, opened 2019-02-05
WYSIWYG GUI for QSD Editing

### LINKED ISSUE #2652: GUI for QSD/Editible text creation
The average admin is not overly familiar with HTML or Markdown, and the code-like syntax can be intimidating. It would be neat if we could have a more GUI-like interface (such as the one GitHub uses) for basic text formatting, link insertion, and graphics addition. Not a high priority, but probably a useful tool for many users

willgearty:
This adds a very nice GUI to the QSD editing components of the website (including editable text on template pages and the .edit.html pages for static pages). I made sure that there is still a way to edit the source code directly for those that are familiar with HTML/markdown. Information about the GUI is [here](https://xdsoft.net/jodit/). I included almost all of the buttons that were offered, but let me know if any should be removed.

Fixes #2652.

Examples:

image

image

image

### DIFF
--- a/esp/public/media/default_styles/qsd.css
+++ b/esp/public/media/default_styles/qsd.css
@@ -2,19 +2,9 @@
 .hidden {
   display: none;
 }
-.hidden {
-  display: none;
-}
-.hidden {
-  display: none;
-}
-.hidden {
-  display: none;
-}
 
 textarea.qsd_editor {
     width: 550px;
-    overflow: auto;
     border: 1px dashed #cccccc;
     border-top: none;
     margin: 8px 0 0;
@@ -36,33 +26,22 @@ div.qsd_header
     border: 1px dashed #cccccc;
     margin: 0px;
     height: 16px;
-    overflow: hidden;
     text-align: center;
 }
 
-div.hidden, div.qsd_edit_hidden
+div.hidden, div.qsd_edit_hidden, div.qsd_view_hidden
 {
     height: 0px;
     visibility: hidden;
+}
+
+div.qsd_buttons {
+    text-align: center;
 }
 
 div.qsd_edit_visible
 {
-    font-family: Arial, Helvetica, sans-serif;
-    font-size: 12px;
-    color: #cccccc;
-    padding: 0px;
-    border: 1px dashed #cccccc;
-    margin: 0px;
-    overflow: hidden;
-    text-align: center;
     visibility: visible;
-}
-
-div.qsd_view_hidden
-{
-    height: 0px;
-    visibility: hidden;
 }
 
 div.qsd_view_visible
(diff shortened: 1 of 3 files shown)


kkbrum:
I love this very much. I did not succeed in breaking anything, and the code looks good, so I think this is set!

Couple observations:
- the "add file" button doesn't make much sense to me, since you have to add the URL of a file and then it just links that URL. So it is essentially the same as the "add link" button, so can probably be removed?
- eventually (not on this PR probably), it would be nice to have the "add image" button accept uploads from the local computer and upload them to the filebrowser for you and display them. The adding image from URL is still a great feature though and i imagine coding up something to add the image from the computer to the filebrowser through this gui is a lot of work since it is not standard.

willgearty:
Actually, now that I'm looking around, it looks like there is the capability to upload images. It looks like there are two separate options:

1. Upload an image as a Base64 image, and the data for the image is stored in the HTML (see the option discussed in [this issue](https://github.com/jodit/jodit-angular/issues/9)). This is plug-and-play, although apparently there might be [problems](https://medium.com/snapp-mobile/dont-use-base64-encoded-images-on-mobile-13ddeac89d7c) using base64 on mobile?
2. Upload an image to the server using php (?), with documentation and examples [here](https://xdsoft.net/jodit/doc/types/IUploaderOptions/) and a tutorial [here](https://xdsoft.net/jodit/v.2/doc/tutorial-uploader-settings.html). It also looks like you can use similar methods to implement a [filebrowser](https://xdsoft.net/jodit/v.2/doc/tutorial-filebrowser-options.html) for the server-hosted images?

@benjaminjkraft As an actual developer, are you familiar with either of these things?

Either way, if we want to get this merged and pushed to chapters as part of the SR10-fixes, we can probably implement the image upload at a later time.

kkbrum:
I'm happy with this solution, but one thing I noticed is that if I now look at the source code after uploading an image, the image takes up a tonnnnnnnnn of lines in the source code so it might be pretty annoying if someone wanted to work with the html after uploading something. But if they want to use source code, they could just go through the filebrowser to avoid having annoying image info there...

willgearty:
Maybe base64 isn't the way to go, then... It looks like there are lots of [problems](https://stackoverflow.com/questions/17288290/why-not-embed-images-as-base64-strings-in-wygiwys-editors) with it, including the fact that the images are actually larger than the original files, they can slow down loading the page, and they can't be cached (and now what @kkbrum mentioned in the source code editor). But it's really easy to include in this PR, and I think we should get this change to chapters as soon as possible, so either we remove the base64 entirely now and then implement the actual upload to the server in a later PR or keep it for now and then implement the server upload. Either way, it'll take me a bunch of work to implement the server upload stuff.

willgearty:
This seems useful: https://www.froala.com/wysiwyg-editor/docs/server/django/file-upload

kkbrum:
I vote for removing the base64 and merging the rest of this gui because the rest is fabulous! The base64 was cool last time when it worked but this time I keep getting errors when it tries to save my qsd pages:
image

willgearty:
Yeah, I agree. After looking into it, we're storing so much data in the QSD database objects, that it's causing some serious memory issues. I've removed the base64 capability. I've implemented a temporary solution for the time being (link goes to the filebrowser):

image

kkbrum:
I like it!

Issue #2564, opened 2018-04-23
Comm panel preview shows plain text as html (parameters only)


willgearty:
If you don't put `<html>` tags around your email, it will not be sent in HTML format, but the email preview will still be in HTML format. Ideally, the commpanel preview should look exactly how the email will look in the user's inbox, which should include not formatting HTML code if there are no `<html>` tags.

Comm Panel Preview:
![image](https://user-images.githubusercontent.com/7232514/39138812-6e9e5380-46d5-11e8-95fc-0d501a1c863c.png)

In Gmail:
![image](https://user-images.githubusercontent.com/7232514/39138712-6547d4fa-46d5-11e8-97e4-5a98afa09b38.png)

willgearty:
Ah, thanks for trying to reproduce. After trying to reproduce the behavior myself I realized that it's only if you use the various parameters (e.g. `{{program.teacher_schedule}}` in this case). If you just type random html it does not get formatted, but any code that is inserted by parameters ends up formatted in the preview, even without `<html>` tags.

900miles:
Yeah, that seems to be the case. It looks like the code for sanitizing html for the email preview is:
```
if '<html>' not in body:
    htmlbody = body.replace('<', '&lt;').replace('>', '&gt;').replace('\n', '<br />')
else:
    htmlbody = body
```
So I'm guessing the problem is that the code is imported by the parameters only after the code is sanitized, so the browser interprets it as real html code, but gmail doesn't.

I'm wondering if it would be more user-friendly to automatically correctly format parameter code as html code though, instead of trying to unformat it. That way people don't have to put \<html\> tags just to use parameters.

willgearty:
Even worse, I think, is that that code would render html in the email preview for something like this:

`<html><h1>Hello</h1>`

But of course that's missing a `</html>` tag, which would be problematic once sent.

Aside from the existing edge-cases that that code doesn't address, are you suggesting that we add `<html>` tags around the email if parameters are included? Because that seems reasonable to me, but theoretically that could result in converting other plain text to html in the resulting email?

900miles:
Oh, right, it's pretty easy to forget adding a closing ``</html>`` tag, which could be a problem if the preview shows it as valid html. 

Adding ``<html>`` tags around the email should theoretically fix the problem, but as you said it might mess with the other text in the email. I'm not sure if there's a way to render only part of an email as html and leave the rest untouched.

[...]

PR #2666, opened 2019-02-05
WYSIWYG GUI for Comm Panel

willgearty:
I added a GUI similar to that of #2665. In order for the GUI to have enough space, I converted the table of template tags into a dropdown menu. By selecting an option from the menu, it inserts the corresponding template tag into the email at the current cursor location. However, I ran into a couple problems. I'd appreciate any feedback and/or suggestions on the following:

- I had to insert the template tags within a `<span>` because createElement() requires a tag and `<span>` seemed the least intrusive. I think this should be fine, but please let me know if it won't be.
- The template tag insertion only works when the user is using the WYSIWYG panel, not the source code panel, of the GUI. I tried to make it work for both, but it was VERY messy.
- The only way to make the HTML actually work for the email is to include `<html>` tags, of course. However, these tags need to be manually added in the source code panel. AND, if you go back to the WYSIWYG panel, those tags disappear.

Example:

image

### DIFF
--- a/esp/esp/program/modules/handlers/commmodule.py
+++ b/esp/esp/program/modules/handlers/commmodule.py
@@ -100,16 +100,14 @@ class CommModule(ProgramModuleObj):
 
         MessageRequest.assert_is_valid_sendto_fn_or_ESPError(sendto_fn_name)
 
-        #   If they were trying to use HTML, don't sanitize the content.
+        # If they used the rich-text editor, we'll need to add <html> tags
         if '<html>' not in body:
-            htmlbody = body.replace('<', '&lt;').replace('>', '&gt;').replace('\n', '<br />')
-        else:
-            htmlbody = body
+            body = '<html>' + body + '</html>'
 
         contextdict = {'user'   : ActionHandler(firstuser, firstuser),
                        'program': ActionHandler(self.program, firstuser) }
 
-        renderedtext = Template(htmlbody).render(DjangoContext(contextdict))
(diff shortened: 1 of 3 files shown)


willgearty:
I did it!

image

kkbrum:
Other than the html tags problem, this seems great! 

I do think once the html tags get solved, the comm panel should default to having the GUI open and not the source code when you arrive at the page since it is less intimidating (idk if this is a setting, but in the QSD GUI it had the GUI open by default).

kkbrum:
Nvm on the defaulting business- it is pulling the last state of GUI vs source code like you intended (I didn't realize). That seems like a great feature.

Once the html tags get fixed and maybe the uploading images from the other PR gets smoothed out, don't forget to add the uploading images feature to the comm panel :)

willgearty:
I didn't implement the base64 image upload for this because apparently it isn't [well supported](https://stackoverflow.com/questions/48446098/sending-email-using-png-vs-base64-encoding/48461894#48461894) in email clients. The easiest thing for the `<html>` tags  would be to just include them around whatever text is submitted (since it will almost always be html/rich text now). This was discussed a little bit in #2564, and we just want to make we aren't adding the tags to emails that already have tags (in the case of a veteran just using the source code panel and including the tags around the email).

elshobokshy:
Hello, could you show us the code that made you able to do this ? (have a button that gives a dropdown of possible choices, and when you click one you get it shown as text inside textarea)

PR #2667, opened 2019-02-06
Update default bounces address

willgearty:


hwatheod:
Is this a new @learningu.org account or an alias pointing to the same gmail as before?  And is the reason for doing this to minimize our emails getting put into spam by gmail?

willgearty:
This is an email address that was created (by @benjaminjkraft, I believe) about a year ago to solve Stanford's email SPF issues. Yale recently started having/noticing similar issues which resulted in their emails getting sent to spam folders. I realized the next bounce email address was only ever updated in the Stanford local_settings.py file, so this makes the change for everyone (although I imagine we'll still want to check the local_settings.py files for all of the chapters the next time we update their sites).

PR #2668, opened 2019-02-11
Round program durations to match class durations

willgearty:
Class durations are rounded to two decimal places, but the program durations were not rounded at all. This caused mismatches in the new duration field (#2612) on the manage class field whenever the duration caused a value with more than two decimal places (e.g. 55 minutes).

I tested classes with durations of 45 minutes, 1 hour 45 minutes, 1 hour, 55 minutes, and 1 hour 55 minutes.

### DIFF
--- a/esp/esp/program/modules/forms/management.py
+++ b/esp/esp/program/modules/forms/management.py
@@ -58,7 +58,7 @@ class ClassManageForm(ManagementForm):
             csm = cls.class_size_max
         else:  csm = 0
         if cls.duration:
-            dur = Decimal(float(str(cls.duration)))
+            dur = cls.duration
         else:  dur = Decimal(0)
         self.initial = {
             prefix+'status': cls.status,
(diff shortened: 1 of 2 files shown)


esp/esp/program/modules/forms/management.py
```diff
@@ -76,7 +76,7 @@ def save_data(self, cls):
         cls.grade_min = self.cleaned_data['min_grade']
         cls.grade_max = self.cleaned_data['max_grade']
         if not cls.hasScheduledSections():
-            cls.duration = Decimal(str(self.cleaned_data['duration']))
```
kkbrum: Do you know why whoever originally wrote this code might have included the str? It seems that removing it did not break anything, but I still wonder why it was there in the first place?

willgearty: I believe that was part of my quick fix for the problems I encountered while fixing #2025. However, that was just a bandaid for the real problem, which is fixed by this rounding of all durations.

kkbrum: Okay!

PR #2681, opened 2019-02-21
Fixes login error when clicking the login button

### LINKED ISSUE #2474: Wrong password error is unhelpful
> You need to log in before you can access /myesp/login/.

is not a user-friendly error message for when you enter an incorrect password.

image

kkbrum:
Fixes #2474. If you hit the log in button on your own (without having entered the correct user & password), it won't tell you something dumb like "you need to log in to view myesp/signup" or "myesp/login" etc. It will only give you that message if you end up at the log in page because you went to teacher reg or student reg or something like that and got redirected to log in.

### DIFF
--- a/esp/esp/themes/theme_data/bigpicture/templates/users/loginbox_content.html
+++ b/esp/esp/themes/theme_data/bigpicture/templates/users/loginbox_content.html
@@ -24,6 +24,7 @@
   <div class="not_logged_in hidden">
     <form name="loginform" id="loginform" method="post" action="/myesp/login/" class="navbar-form form-inline">
       <input type="hidden" name="next" value="{{ request.path }}" />
+      <input type="hidden" name="initiated_login"/>
       <input type="text" name="username" id="user" size="8" value="" placeholder="Username" maxlength="255" class="inputbox" />
       <input type="password" name="password" id="pass" size="8" value="" placeholder="Password" maxlength="255" class="inputbox" />
 
--- a/esp/esp/themes/theme_data/bigpicture/templates/users/loginbox_content.html
+++ b/esp/esp/themes/theme_data/bigpicture/templates/users/loginbox_content.html
@@ -24,7 +24,6 @@
   <div class="not_logged_in hidden">
     <form name="loginform" id="loginform" method="post" action="/myesp/login/" class="navbar-form form-inline">
       <input type="hidden" name="next" value="{{ request.path }}" />
-      <input type="hidden" name="initiated_login"/>
       <input type="text" name="username" id="user" size="8" value="" placeholder="Username" maxlength="255" class="inputbox" />
       <input type="password" name="password" id="pass" size="8" value="" placeholder="Password" maxlength="255" class="inputbox" />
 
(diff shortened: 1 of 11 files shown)


willgearty:
We could do something similar to what we did for the `loginbox` and just include another hidden field? If we find that in the POST data, we know that they tried to login. In fact, since we want to treat this the same as if they used the wrong credentials for the `loginbox`, we could use the SAME hidden field in the login form, then just add an else statement to the template to tell the user that they must have used an incorrect username or password.

benjaminjkraft:
I'm confused why we need a hidden field at all here.  Can't we just check whether the method was POST?  I would think if you are redirected because you tried to go to a login-required page you would always be doing a GET, so a POST can only mean you tried to log in and failed.

benjaminjkraft:
Also, IMO it's fine to give a more specific error in our case.  User enumeration is pretty easy and we make no effort to prevent it in other contexts, so there's no need to do so here.

willgearty:
Hmm...I guess I may have been overthinking it. Are there any forms or buttons that use POST but would require a login upon submission? If not, @benjaminjkraft is probably right, in that we can catch all of these cases by just checking for `request.POST`.

willgearty:
Oh, now I remember why we might need extra hidden fields. There's three cases we want to catch and show messages for here:

1. The user used incorrect credentials in the main loginbox. In this case we just want to tell them they had incorrect credentials, not that they need to be logged in to get to a certain page.
2. The user used incorrect credentials when prompted to sign in for a particular page. In this case, we want to tell them they used incorrect credentials, but also that they need to login to access that particular page.
3. The user hasn't submitted credentials yet and is being prompted to sign in for a particular page. Here we just want the original message that says they need to log in to access the particular page.

benjaminjkraft:
Can't we tell 1 from 2 via whether there is a `next` param?

willgearty:
While the default loginbox doesn't have a `next` field...
https://github.com/learning-unlimited/ESP-Website/blob/6a2f2a968a4190c8d4a678b60ea624ff8e2b6cef/esp/templates/users/loginbox_content.html#L38-L39
...most of the themes have `next` fields in their loginboxes:
https://github.com/learning-unlimited/ESP-Website/blob/6a2f2a968a4190c8d4a678b60ea624ff8e2b6cef/esp/esp/themes/theme_data/circles/templates/users/loginbox_content.html#L33-L34
https://github.com/learning-unlimited/ESP-Website/blob/6a2f2a968a4190c8d4a678b60ea624ff8e2b6cef/esp/esp/themes/theme_data/bigpicture/templates/users/loginbox_content.html#L25-L26
https://github.com/learning-unlimited/ESP-Website/blob/6a2f2a968a4190c8d4a678b60ea624ff8e2b6cef/esp/esp/themes/theme_data/fruitsalad/templates/bubblesfront.html#L79-L80
https://github.com/learning-unlimited/ESP-Website/blob/6a2f2a968a4190c8d4a678b60ea624ff8e2b6cef/esp/esp/themes/theme_data/fruitsalad/templates/main.html#L137-L138

kkbrum:
I changed it so we don't use a hidden field to check if login is initiated since there is no GET when login is initiated so I just use that. 

I'll look into showing better errors about username or password being wrong soon too.

kkbrum:
@benjaminjkraft  I tested this a bunch and it works well, but could you take a look that this is a fine approach in terms of coding?

willgearty:
Works perfectly for me! I especially think users will find the wrong username/pw feature very helpful!

PR #2684, opened 2019-02-23
Fix advanced checkin

### LINKED ISSUE #2365: Advanced checkin doesn't seem to be checking people in
Reported by USC Splash.  You search the person, and click check in, and it just silently does nothing.

willgearty:
Now properly checks students in, allows for updating a student's checkin, redirects to search for another user after each checkin, and shows a message to alert the admin that the checkin was successful. 

Fixes #2365.

### DIFF
--- a/esp/esp/program/modules/handlers/onsitecheckinmodule.py
+++ b/esp/esp/program/modules/handlers/onsitecheckinmodule.py
@@ -80,7 +80,7 @@ class OnSiteCheckinModule(ProgramModuleObj):
         if event=="paid":
             self.updatePaid(False)
 
-        recs = Record.objects.get_or_create(user=self.student,
+        recs, created = Record.objects.get_or_create(user=self.student,
                                             event=event,
                                             program=self.program)
         recs.delete()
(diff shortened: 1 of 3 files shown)


esp/esp/program/modules/handlers/onsitecheckinmodule.py
```diff
@@ -242,23 +242,31 @@ def ajaxbarcodecheckin(self, request, tl, one, two, module, extra, prog):
...
+            if user:
+                self.student = user
+                for key in ['attended','paid','liab','med']:
+                    if key in request.POST:
+                        self.create_record(key)
+                    else:
+                        self.delete_record(key)
+                message = "Check-in updated for " + user.username
+            else:
+                error = True
```
benjaminjkraft: Maybe set message here to user not found or something?

willgearty: That's the message that gets displayed when `error = True`:
https://github.com/learning-unlimited/ESP-Website/blob/9000414a8d60722cd9bf0483c1f9b0da9f675d5f/esp/templates/users/usersearch.html#L25-L27

PR #2685, opened 2019-02-24
Teacher big board fixes

### LINKED ISSUE #2662: Teacher big board fails if a class has 0 sections
I'm not sure how a class gets to such a state, but apparently if a class has 0 sections, it breaks [this line](https://github.com/learning-unlimited/ESP-Website/blob/d76640cf00dec98837176dee31c52395ac1dbacb/esp/esp/program/modules/handlers/teacherbigboardmodule.py#L150), because the annotation `Sum('sections__duration')` equates to `None`. I guess we can just filter the ClassSubjects such that we only get classes with at least 1 section?

Reported by Northwestern.

willgearty:
The teacher big board could break during the calculation of the hours if there was a class with no sections. I've fixed that and a bug that I think I reported on the old slack channel but never here. The class-hour lines would stop abruptly on the plot because they excluded lunch classes but the other calculations didn't. Now everything excludes lunch classes.

Fixes #2662.

### DIFF
--- a/esp/esp/program/modules/handlers/teacherbigboardmodule.py
+++ b/esp/esp/program/modules/handlers/teacherbigboardmodule.py
@@ -4,6 +4,7 @@ import subprocess
 
 from django.db.models.aggregates import Min, Sum
 from django.db.models.query import Q
+from django.db.models import Count
 
 from argcache import cache_function_for
 from esp.program.models import ClassSubject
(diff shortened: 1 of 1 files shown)


benjaminjkraft:
Good finds!  There's probably some value to deduplicating this code a bit, but it doesn't have to be now.

PR #2686, opened 2019-02-25
Fix layout and JS for missingteachers

### LINKED ISSUE #2605: Missing teachers page overlaps instructions with class search and administration links in fruitsalad theme
There are a useful set of instructions/ keyboard shortcuts on the onsite checkin teachers page, but they overlap the class search and the links to the backend, so I can't press the link to the admin pages from here.

image

willgearty:
I moved the hovering box over to the top right, so it doesn't conflict with the admin toolbar in some themes. I also noticed some problems with the JS (e.g. could only use arrow keys if you werent't focused on the input but the input would focus after each arrow keys press, Ctrl-Z would undo the input text before undo-ing the check-in). I've fixed it all up so it now works as described in the hovering box. I also added a small change that the input text will reset if the user clicks outside of the input and then starts typing again. If the input is focused, then the user will need to use backspace, Ctrl-A, etc. to edit the value like normal.

Fixes #2605.

### DIFF
--- a/esp/public/media/scripts/onsite/teacher_checkin.js
+++ b/esp/public/media/scripts/onsite/teacher_checkin.js
@@ -210,23 +210,13 @@ $j(function(){
         });
     }
 
-    $j(document).keypress(function(e){
-        if(e.which==13 && e.shiftKey){
-            $j(".selected .checkin").click();
-            e.preventDefault();
-            input.val("");
-        }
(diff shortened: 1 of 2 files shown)


willgearty:
BTW, it would be great if the reviewer of this (probably @kkbrum ) could test this in multiple browsers (i.e. Chrome and Firefox) to check that both browsers have the same key codes!

PR #2687, opened 2019-02-26
Fix handling of finaid form tag

### LINKED ISSUE #2620: Financial Aid fields tag doesn't hide questions
The finaid tag `finaid_form_fields` will hide the text boxes but not the questions, so if you use it, then the questions just float on the page without any space to answer them, which is unintuitive and unhelpful. I believe this could be as simple a fix as a single conditional in https://github.com/learning-unlimited/ESP-Website/blob/main/esp/templates/program/modules/financialaidappmodule/application.html that controls whether or not to display the relevant row of the table.

willgearty:
The form now hides the field labels if the fields don't exist in the form. I also added some documentation for the tag (mostly because I never knew this existed, so I imagine I wasn't alone). Finally, I added some filtering for the tag processing to only include fields from the tag that actually exist in the `FinancialAidRequest` model.

Fixes #2620.

Note to reviewer: you might need to comment out [this line](https://github.com/learning-unlimited/ESP-Website/blob/finaid-form-tag/esp/esp/program/modules/handlers/financialaidappmodule.py#L89) in order to access the financial aid application on a dev server

### DIFF
--- a/esp/esp/program/modules/handlers/financialaidappmodule.py
+++ b/esp/esp/program/modules/handlers/financialaidappmodule.py
@@ -104,7 +104,7 @@ class FinancialAidAppModule(ProgramModuleObj):
                 model = FinancialAidRequest
                 tag_data = Tag.getTag('finaid_form_fields')
                 if tag_data:
-                    fields = tuple(tag_data.split(','))
+                    fields = tuple(field.strip() for field in tag_data.split(',') if hasattr(FinancialAidRequest(), field.strip()))
                 else:
                     fields = '__all__'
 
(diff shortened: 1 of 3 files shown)


willgearty:
First of all, don't get me started about `explaination`, but it's in the models now, so it would be a pain to change (eww migrations). Second of all, what exactly is the tag value that isn't working for you? I'm pretty sure I tried combinations of all four possible values, and they all worked for me.

milescalabresi:
I couldn't get ``reduced_lunch`` to show up

willgearty:
Works for me...
image

milescalabresi:
What did you enter in the tag field?

milescalabresi:
Ah, okay I was entering a Python list by mistake (including square brackets). It looks good now

PR #2689, opened 2019-02-27
Prevent autofill of list_selector

### LINKED ISSUE #2680: Browser autofills UserSearchControl filters
I've never encountered this myself, but I've now heard two cases of people's browsers autofilling the filter fields in the UserSearchControl panel (commpanel, arbitrary user list, group text). Specifically, the browsers fill the username filter which results in a final list of just a single user (the admin using the panel). We should mark these filter fields such that we [prevent autofilling](https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion) by browsers.

Reported by an admin at SplashCon and by an admin at Stanford.

willgearty:
I believe this should fix any autofill issues with the `list_selector.html` template used by the commpanel, arbitrary user list generator, and group text modules. Apparently there's  some politics around the specifics of the `autocomplete` field, specifically with how Chrome handles them. [caniuse](https://caniuse.com/#feat=input-autocomplete-onoff) specifies that Firefox and IE should follow the `autocomplete="off"` setting for anything but a password field, so the `autocomplete="off"` in the `form` tags should cover those browsers. Chrome, on the other hand, apparently will just ignore this setting completely. [Their suggestion](https://bugs.chromium.org/p/chromium/issues/detail?id=468153#c164) is to use a slightly modified keyword as I have done (`autocomplete="new-username"`), such that they don't use their data to autofill the field. This isn't supported by Firefox, but I'm hoping that the `form` tag addition will take precedence in that case?

I personally can't test this since I didn't have the problem before and am not able to reproduce the problem on my dev server.

Fixes #2680.

### DIFF
--- a/esp/templates/users/usersearch/list_selector.html
+++ b/esp/templates/users/usersearch/list_selector.html
@@ -13,7 +13,7 @@
         {% endif %}
     </ul>
     <div id="tab_basic">
-        <form id="form_basic_list" method="POST" action="{{ action_path }}">
+        <form id="form_basic_list" method="POST" action="{{ action_path }}" autocomplete="off">
 
         <div id="basic_step_container">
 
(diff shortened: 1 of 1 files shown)


benjaminjkraft:
Shrug, ok!

PR #2691, opened 2019-03-03
Custom forms work even when users are deleted

### LINKED ISSUE #2673: Custom form responses without user ids break response views
Reported by UC Merced. I'm honestly not sure how a customform response was submitted with an attached `user_id`, but if this happens, it breaks any of the views that pull the response data (e.g. /customforms/responses/ and /customforms/exceldata/). This is ultimately due to the fact that [this line](https://github.com/learning-unlimited/ESP-Website/blob/stable-release-10/esp/esp/customforms/DynamicForm.py#L653) will only pull the users for responses with valid `user_id` values, but [this line](https://github.com/learning-unlimited/ESP-Website/blob/stable-release-10/esp/esp/customforms/DynamicForm.py#L661) then goes through all of the `user_id` values, including the `None` one(s) and breaks due to a `KeyError`. I was able to fix this particular instance by deleting the response with the invalid `user_id`, which was nontrivial, since you need to access the separate CustomForms database to get to the response objects.

willgearty:
Custom forms were breaking if a user associated with a response was deleted. We probably shouldn't be deleting users anyways, but just in case this happens again, this is a fix for that (we just skip including user data in the response in such cases, but we can still fill in all of the other information).

Fixes #2673.

### DIFF
--- a/esp/esp/customforms/DynamicForm.py
+++ b/esp/esp/customforms/DynamicForm.py
@@ -657,7 +657,7 @@ class FormHandler:
             link_instances_cache={}
 
             # Add in user if form is not anonymous
-            if not form.anonymous:
+            if not form.anonymous and response['user_id']:
                 user = users[response['user_id']]
                 response['user_id'] = unicode(response['user_id'])
                 response['user_display'] = user.name()


benjaminjkraft:
Seems reasonable!

Issue #2550, opened 2018-04-15
Combination list formatting messed up


kkbrum:
![image](https://user-images.githubusercontent.com/25392075/38775280-5d1fb25a-404c-11e8-8561-15051daa3510.png)

The "and" "or" "not" checkboxes are... not where they should be... If you click on the word, it works, but is not pretty.

hwatheod:
Interesting, I've noticed a problem like this on Stanford's site for a while but I always assumed it was because of Stanford's custom theme.

kkbrum:
@milescalabresi could you have done something during the Yale site update with our customization that caused this?  It was fine before the update. Is this a problem on other sites as well or just Stanford and now Yale?

milescalabresi:
It's not your customization... I checked a few other sites, and they're having the same problem when you use combination lists :(

benjaminjkraft:
Comparing to MIT ESP's site, which is not on SR10 yet I believe and where this looks fine, it looks like some of the jquery-ui styles that we used to have are gone.  I would guess this is due to jquery-ui upgrades (#2450), cc @betaveros.  Probably they changed the classes we need to apply or something.

kkbrum:
@willgearty @benjaminjkraft  i know this is a lot to ask, but this comm panel bug is super super annoying and I would be forever happy if someone took a look at this. I dont know anything about jquery-ui but i can also try a bit.

willgearty:
Not sure of a more permanent solution, but I figured out that the `ui-helper-hidden-accessible` class is what is ultimately causing the checkboxes to disappear (at least on the Stanford site). Removing the class doesn't seem to have any negative effects that I can see.

So the easiest thing at this point would be to just make a template override where the checkboxes don't have that class. Adding this template override for `program/modules/commmodule/commpanel_new.html` worked, although it's far from gorgeous:

```r
{% extends "users/usersearch/support.html" %}

{% block title %}Communications Portal{% endblock %}

{% block content %}

<h2>E-mail a list of users</h2>
<p>This page will help you e-mail a set of users in your database, with filtering options specific to {{ program.niceName }}.  Choose from three different approaches to select your users:
    <ul>
        <li><b>Basic list</b>: A set of users that all have something in common.</li>
        <li><b>Combination list</b>: A Boolean combination of basic lists.</li>
        <li><b>Previous e-mails</b>: Start from the settings that were used for a previous e-mail message.</li>
    </ul>
</p>

[...]

PR #2692, opened 2019-03-03
Tweak some commpanel formatting

willgearty:
While looking into #2550, I noticed some little things that could use tweaking:

1. I spaced out the AND/OR/NOT buttons a little (they were overlapping before)
2. I allowed the combination list dropdown to expand in width to show more of the text for the selected option.


benjaminjkraft:
Ok!

PR #2693, opened 2019-03-04
Sendgrid and anti-spam fixes

### LINKED ISSUE #2514: Move to sendgrid or another commercial email sender
This, along with perhaps some changes to how we set the various from-fields on the emails we send, could help us get marked as spam a lot less.  Copying what I wrote to web-team:

> Email sending is a dark art, and there are a whole lot of things one has to do right, not to mention just not get people to mark one's email as spam.  We could probably improve some things if folks want to spend the time to mess around with the headers the comm panel sets and set up DKIM/SPF/DMARC properly, but it's an unending mess and there are some things it's harder to do well.  If we really want to improve this, we should move to sending via some commercial provider like Sendgrid who will manage all this crap for us.  (If cost is an issue, we could have it as an option that Splashes can reimburse LU for.)  I think MIT ESP may be interested in this sort of thing as well.

Sendgrid is the provider I'm most familiar with, but there are a lot of others, depending on what features we want.  For example, some providers might provide a much nicer interface we could directly give chapters access to send email to simple lists like "all teachers".

willgearty:
This adds a package to interface with sendgrid's REST API as an Email Backend. In order to use this backend instead of our default backend, you will need to change the `local_settings.py` file as described [here](https://github.com/sklarsa/django-sendgrid-v5). I can share the API key with whoever ends up reviewing this.

Fixes #2514.


willgearty:
I'm going to use this branch for implementing other anti-spam things, so this doesn't need to be reviewed yet. For future reference, those things include:

- Include a plaintext version of HTML emails (looks like we would use the `EmailMultiAlternatives` subclass of `EmailMessage`)
- Include a list-unsubscribe header

benjaminjkraft:
Seems fine so far!  We'll probably need to find a way to handle secrets such that we don't need to manually add this every time we set up a new site.

willgearty:
Closing this in favor of #3721. We already have sendgrid implemented on the main branch, and that new PR is handling spam-related things.

Issue #2282, opened 2016-12-21
fruitsalad top-left contact info missing on login page


benjaminjkraft:
e.g. https://esp.mit.edu/accounts/login/?next=/manage/programs/, the program name and contact info normally in the top-left corner is missing.

willgearty:
This seems fixed?

willgearty:
It looks like `theme` isn't in the context, which results in some things not displaying.

willgearty:
Closed by #2699.

PR #2699, opened 2019-03-12
Add contact info in fruitsalad theme on login page

kkbrum:
Fixes issue #2282 by using the render_to_response function from esp/esp/utils/web.py which adds theme data to the context when not yet logged in.


willgearty:
Works perfectly for me!

PR #2704, opened 2019-03-17
Round Event durations to match the rounding of class/section durations

### LINKED ISSUE #2696: Class durations don't match time slot durations due to rounding
We round class durations to two decimal places, but we don't really round anything else to two decimal places, so this can cause problems with scheduling and adding co-teachers in cases where the lengths of timeblocks result in decimals with more than two decimal places (e.g. 50 minutes).

I think one of the biggest offenders is here, where we check whether a teacher can teach a class:
https://github.com/learning-unlimited/ESP-Website/blob/9896724526d6506f81531a8f2f0394aa05bd9eea/esp/esp/program/models/class_.py#L1678-L1707

However, there must be similar problems in the code for the scheduler. The scheduling diagnostics also need to be fixed:
https://github.com/learning-unlimited/ESP-Website/blob/9147b7912921642812fa3fc4bfdbca35eaac9665/esp/esp/program/modules/handlers/schedulingcheckmodule.py#L218-L227

Basically, any time we call `duration` and then make a comparison, we should make sure we are rounding consistently.

Alternatively, a lot of problems arise from calculating the lengths of time blocks, so we could just add rounding here:
https://github.com/learning-unlimited/ESP-Website/blob/9147b7912921642812fa3fc4bfdbca35eaac9665/esp/esp/cal/models.py#L75-L76

We'd then need to make sure we are using that method instead of reinventing the wheel each time we want the duration of an event (e.g. `conflicts()` above).

Reported by NEU.

willgearty:
I'm pretty sure this fixes all of the potential places that we might have different rounding between Event durations and class/section durations. At the very least, I've modified the `duration()` method, which will now always return a `timedelta` that is rounded appropriately. I think there may be other issues specifically in the AJAX scheduler, but I'll discuss that in a new issue.

Fixes #2696.

### DIFF
--- a/esp/esp/cal/models.py
+++ b/esp/esp/cal/models.py
@@ -73,7 +73,10 @@ class Event(models.Model):
         return self.name
 
     def duration(self):
-        return self.end - self.start
+        # Matches the rounding of class/section durations
+        dur = self.end - self.start
+        hrs = round(dur.total_seconds() / 3600.0, 2)
+        return timedelta(hours = hrs)
 
     def duration_str(self):
         dur = self.end - self.start
@@ -106,7 +109,10 @@ class Event(models.Model):
         event_list = list(event_list)
         event_list.sort(key=lambda x:x.start)
         if len(event_list) > 0:
-            return event_list[-1].end - event_list[0].start
+            # Matches the rounding of class/section durations
+            dur = event_list[-1].end - event_list[0].start
+            hrs = round(dur.total_seconds() / 3600.0, 2)
+            return timedelta(hours = hrs)
         else:
             return timedelta(seconds=0)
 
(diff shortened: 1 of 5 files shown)


willgearty:
This pull request **introduces 1 alert** when merging e5263fffcbdbfe08ad8707b7905abcc050392409 into 764f5c7b387d43b8d11f5d1061caa9c8b7dbdb23 - [view on LGTM.com](https://lgtm.com/projects/g/learning-unlimited/ESP-Website/rev/pr-de327e4255f8ea93790d8bd1b43dafef454b5b55)

**new alerts:**

* 1 for Unused import

---

*Comment posted by [LGTM.com](https://lgtm.com)*

willgearty:
Awesome, thanks for testing! It would be awesome if you could try to replicate the scenario reported by NEU. I think they had a class that covered every timeslot (with 50 minute classes), and they weren't able to add a co-teacher?

kkbrum:
I tested it on NEU's situation and it works great for adding coteachers, etc!

To confirm your AJAX suspicion, the scheduler claims the class is 3 blocks instead of 2, so that is definitely still a problem.

willgearty:
Awesome, thanks!! (And boo ajax)

PR #2708, opened 2019-03-21
fix numpy version check

byronxu99:
The numpy version check wasn't implemented correctly. It failed for version 1.16.0


willgearty:
Looks good to me! Thanks!

PR #2709, opened 2019-03-22
Teacher event signup layout redesign

willgearty:
This is the start of a layout redesign for the `event_signup` page for teachers. The starting point comes from a template override from Stanford, but I think I can make it better (also, I need to think about the consequences of having so many buttons if there are lots of interview slots, but I feel like anything would be better than a dropdown with 50 options).

Here's what it looks like now:
image

### DIFF
--- a/esp/templates/program/modules/teachereventsmodule/event_signup.html
+++ b/esp/templates/program/modules/teachereventsmodule/event_signup.html
@@ -2,27 +2,177 @@
 
 {% block title %}Teacher Preparation Events for {{ prog.niceName }}{% endblock %}
 
-{% block content %}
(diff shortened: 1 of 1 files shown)


willgearty:
Maybe something like one of these would be good:
https://github.com/llazzaro/django-scheduler
https://github.com/bitlabstudio/django-calendarium

Actually, this looks like a good place to start:
https://alexpnt.github.io/2017/07/15/django-calendar/

kkbrum:
@willgearty do you want to make an issue for someone to complete this PR?

willgearty:
Superseded by #5253

PR #2711, opened 2019-03-30
classesbyFOO and sectionsbyFOO fixes/changes

### LINKED ISSUE #2710: classesbyFOO and sectionsbyFOO break if unapproved classes are specified
If you supply an ID for a class that is not approved (or cancelled in the case of sectionsbyFOO with `cancelled` in the `GET`), it fails because of the indexing here:
https://github.com/learning-unlimited/ESP-Website/blob/f741555ca6bcb884a682451d587b6b0828e4e320/esp/esp/program/modules/handlers/programprintables.py#L342

https://github.com/learning-unlimited/ESP-Website/blob/f741555ca6bcb884a682451d587b6b0828e4e320/esp/esp/program/modules/handlers/programprintables.py#L374

Of course, this behavior should be fixed, since it's very easy for an unapproved class to be specified. The easiest scenario would be to just skip any of these classes. However, maybe classesbyFOO and sectionsbyFOO should include all classes by default, with specific filters specified in the `GET` parameters (such as `cancelled` or `approved`). The functions would filter by the supplied IDs _first_ (if any are supplied), then filter by the cancelled or approved filters afterwards.

Oh, also, there's a bug in the links from #2654. `classesbytime` actually uses section IDs instead of class IDs (even though it's called `classesbytime`). Perhaps it would make more sense to have both `GET` options be possible, in case someone wants to supply class IDs, like in the case of the class search page.

willgearty:
I fixed the classesbyFOO and sectionsbyFOO functions so they now don't break when unapproved class or sections IDs are supplied. I also reordered the filtering to what I believe is the most efficient order. Finally, I changed the default of these functions to return all classes that are not cancelled or rejected (i.e. `status >= 0`), but that can be overridden by various `GET` parameters (I'm happy to discuss if the anyone thinks that there is a better default).

To make it easier for admins to utilize these different options, I've added a large number of new links to the printables page.

I tested all of the different options for all of the different types of printables, and they all seem to work properly for my setup.

Fixes #2710.

### DIFF
--- a/esp/templates/program/modules/programprintables/options.html
+++ b/esp/templates/program/modules/programprintables/options.html
@@ -51,13 +51,47 @@ Please select from options below.
 
 <h3>Admin Binder Stuff</h3>
 <ul>
-<li><a href="./classesbytime" title="Classes by Time">Class Sections sorted by time</a>
+<li><a href="./classesbytime" title="Classes by Time">Class Sections Sorted by Time</a>
     <ul>
     <li><a href="./classesbytime?grade_min=7&grade_max=8" title="Classes by Time">Limit to middle school (grades 7-8)</a></li>
     <li><a href="./classesbytime?grade_min=9&grade_max=12" title="Classes by Time">Limit to high school (grades 9-12)</a></li>
+    <li><a href="./classesbytime?accepted" title="Classes by Time">Accepted Only</a></li>
+    <li><a href="./classesbytime?scheduled" title="Classes by Time">Scheduled Only</a></li>
+    <li><a href="./classesbytime?accepted&scheduled" title="Classes by Time">Accepted and Scheduled Only</a></li>
+    <li><a href="./classesbytime?cancelled" title="Classes by Time">Cancelled Only</a></li>
     </ul>
 </li>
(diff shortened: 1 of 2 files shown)


esp/templates/program/modules/programprintables/options.html
```diff
@@ -51,13 +51,47 @@ <h3>Receipts</h3>
...
+    <li><a href="./classesbytime?accepted" title="Classes by Time">Accepted Only</a></li>
+    <li><a href="./classesbytime?scheduled" title="Classes by Time">Scheduled Only</a></li>
+    <li><a href="./classesbytime?accepted&scheduled" title="Classes by Time">Accepted and Scheduled Only</a></li>
+    <li><a href="./classesbytime?cancelled" title="Classes by Time">Cancelled Only</a></li>
     </ul>
 </li>
-<li><a href="./classesbytime?cancelled" title="Classes by Time">Cancelled Class Sections</a></li>
+<li><a href="./classesbyid" title="Classes by ID">Class Subjects Sorted by ID</a></li>
+    <ul>
+    <li><a href="./classesbyid?grade_min=7&grade_max=8" title="Classes by ID">Limit to middle school (grades 7-8)</a></li>
```
kkbrum: Can you make all the grade limited ones follow the same capitalization pattern as the rest?

Issue #1402, opened 2014-10-17
Find availability of a set of teachers


benjaminjkraft:
Check availability will show you the availability of a single teacher; the schedule will show you the _remaining_ availability of the teachers of a class.  It would be nice to have something in one interface or the other to show you data like check availability, but intersected over the teachers of a class.

luac:
Is this still needed, since the scheduler shows unavailable blocks differently from currently-teaching blocks?

willgearty:
Fixed by #2714.

PR #2714, opened 2019-04-08
Class availability page

### LINKED ISSUE #2638: Availability page for class
Similar to #1402, but perhaps a little more expanded. At the moment, since the availability checks aren't amazing (#106, #176, #1780), often you need to check all of the availabilities of all of the teachers for a class. It would be much easier if you could see all of this in one place for a single class. Maybe something like doodle.com, where you can see the availabilities of each teacher and can check which teachers are limiting the availability of the class. This information must already be determined, since it's used for the time availability spreadsheet, so I think it just needs a UI.

willgearty:
This implements an availability page for a class. It looks just like the teacher/volunteer availability pages, but shows whether the teachers for a class are available or not for the time blocks:
image

If there are teachers that aren't available for a particular time block, hovering over that time block will show which teachers are already teaching:
image

and/or which teachers are not available:
image

If a section of the class is scheduled, hovering over that time block will show that:
image

I added links to this page from the `manageclass_class_info.html` template, which is included in the `/manageclass` and `/classsearch` pages. I also included a link to this page from the manage coteachers page. Finally, I included links in the scheduler, so you can go directly to this page while scheduling classes. While I was at it, I added return links to the class manage page from the manage coteachers and class availability pages.

Fixes #2638 and #1402.

### DIFF
--- a/esp/esp/program/modules/handlers/adminclass.py
+++ b/esp/esp/program/modules/handlers/adminclass.py
@@ -35,7 +35,9 @@ Learning Unlimited, Inc.
 from esp.program.modules.base import ProgramModuleObj, needs_admin, aux_call
 from esp.program.modules.handlers.teacherclassregmodule import TeacherClassRegModule
 
+from esp.cal.models import Event
 from esp.program.models import ClassSubject, ClassSection, ClassFlagType
+from esp.tagdict.models import Tag
 from esp.users.models import ESPUser, User
 
 from esp.utils.web import render_to_response
@@ -420,7 +422,8 @@ class AdminClass(ProgramModuleObj):
                                    'ajax': ajax,
                                    'txtTeachers': txtTeachers,
                                    'coteachers': coteachers,
-                                   'conflict': conflictinguser})
+                                   'conflict': conflictinguser,
+                                   'program': prog})
 
     @aux_call
     @needs_admin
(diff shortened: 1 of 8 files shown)


willgearty:
This pull request **introduces 1 alert** when merging 4458379c9ab3efee3cdd519ec4b68c98c609271b into 33a71d682523ddb24c9b5d15578b4d527eddda32 - [view on LGTM.com](https://lgtm.com/projects/g/learning-unlimited/ESP-Website/rev/pr-2b522aacae67de615f2327fb11c9e43e52d4a030)

**new alerts:**

* 1 for Unused local variable

---

*Comment posted by [LGTM.com](https://lgtm.com)*

kkbrum:
I like this a lot. One thought I have is that in my mind, "unavailable" and "teaching something else" are kind of different because the approaches to fixing them are different. Do you think it might make sense to specify that someone is teaching some other class as opposed to calling them unavailable if that is the case?

kkbrum:
On the class availability page, would we want to note somewhere the length of the class? Just so we know if it's a 2 block class, that we are looking for 2 consecutive blocks of availability? just for easy reference

willgearty:
Thanks for the great ideas @kkbrum! The hover details now distinguish between whether a teacher is already teaching or just unavailable. I've also added the length of the class to the page. I've updated the images in the OP to reflect these changes.

kkbrum:
Could we put a line break before teaching here?

image

willgearty:
@kkbrum Let me know if that works.

PR #2719, opened 2019-04-13
Fix black cell backgrounds in scheduler

### LINKED ISSUE #2361: Change colors for visibility on AJAX Scheduler
UCLA requested on the feature request spreadsheet:

>Scheduler currently lists section ID (eg. A9s2) in black font and on a dark grey background. Very hard to see.
Recommend changing background color. If possible, also allocate one color to each subject (eg. light blue for math classes) to give variety to a timeslot.
If possible, allow user to customise colors as well.

willgearty:
This fixes the coloring of cells (and text color) in the AJAX Scheduler for programs with class IDs with fewer than 4 digits. The new logic pads the class IDs with zeros to get at least 2 digits to work with, then uses the last 2 digits of the resulting IDs. ~~Therefore, programs with class IDs with fewer than 3 digits can expect very blue/green colors, but red is incorporated once class IDs have at least 3 digits.~~

Before:
image

After:
image

Fixes #2361.

### DIFF
--- a/esp/public/media/scripts/ajaxschedulingmodule/ESP/CellColors.js
+++ b/esp/public/media/scripts/ajaxschedulingmodule/ESP/CellColors.js
@@ -5,6 +5,17 @@
  * @method color(section)
  * @method textColor(section)
  */
+var colors = palette('tol-rainbow', 100);
+
+function hexToRGB(hex) {
+    var result = /^#?([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})$/i.exec(hex);
+    return result ? {
+        r: parseInt(result[1], 16),
+        g: parseInt(result[2], 16),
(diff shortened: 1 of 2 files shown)


kkbrum:
This looks fine, but maybe instead of padding with 0s and getting lots of blues we could just do if statements based on the number of digits in the class code and for the ones with less digits, calculate colors in some smart way such that the colors are more fun? I don't really like that the colors are going to be so sad until programs get 100 classes :(

willgearty:
OK, I've changed the approach to use a pre-established palette of 100 colors and to rotate through them based on the last two digits of the class code (using a prime number and modulus to make consecutive classes have pretty different colors). Let me know what you think! (I've updated the image in the OP)

PR #2724, opened 2019-04-21
Show scheduled classes on availability form, even if unavailable

### LINKED ISSUE #2716: Availability module does not show a scheduled class if teacher is not available
If a class has been scheduled at a time that a teacher is not available, the timeslot in their availability will be grey ("Not Available") instead of blue ("Scheduled to Teach"). The hover popup still works though.

willgearty:
Instead of only looking for the checked checkboxes to set the classes for the different cells, we just check all of the checkboxes, which allows for cells to have the teaching CSS class (the checkbox is disabled) even if the teacher isn't available (the checkbox isn't checked).

Fixes #2716.

### DIFF
--- a/esp/public/media/default_styles/availability.css
+++ b/esp/public/media/default_styles/availability.css
@@ -84,6 +84,10 @@ td.group, table.group {
   background: #42b3f4;
 }
 
+.teaching:not(.canDo) {
+  border: solid 2px red !important;
+}
+
 .headerText {
   font-size: 18px;
 }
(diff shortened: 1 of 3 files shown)


willgearty:
I've moved your first two requests to issue #2738 (since it has to do with the class availability page, not the individual teacher availability page).

willgearty:
Thanks @kkbrum for the great ideas. I've implemented the changes for this page, and I'll get to the other ideas for the class availability page next. As suggested, I added a red border to the cells where the teacher is scheduled to teach but isn't available:

image

I also made it so that you can click on those cells to make the teacher available (and then submit it to update their availability). Previously, you could not change the availability for any cells that had a scheduled class. Ultimately, I added this feature more so admins could more easily update teachers' availabilities (without having to go to the admin pages), but it's probably fine if a teacher does it too.

PR #2732, opened 2019-05-06
upgrade pillow from 3.3.1 to 3.3.3 (CVE-2016-9189, CVE-2016-9190)

hwatheod:
This showed up in a security alert within github.


willgearty:
It sounds like 3.3.2 [introduced a bug](https://github.com/python-pillow/Pillow/blob/master/CHANGES.rst), which is fixed in 3.3.3, maybe we should use that instead?

hwatheod:
good call.  changed to 3.3.3.

willgearty:
Looks good to me!

PR #2743, opened 2019-05-13
Add csv templates for some printables

### LINKED ISSUE #2730: sectionsbyFOO and teachersbyFOO call .csv templates which don't exist
`sectionsbyFOO` calls `sections_list.csv`, which doesn't exist:
https://github.com/learning-unlimited/ESP-Website/blob/e89f47c92808217209c8b64d06e323a6cf951806/esp/esp/program/modules/handlers/programprintables.py#L372-L376

`teachersbyFOO` calls `teacherlist.csv`, which doesn't exist:
https://github.com/learning-unlimited/ESP-Website/blob/e89f47c92808217209c8b64d06e323a6cf951806/esp/esp/program/modules/handlers/programprintables.py#L483-L487

Stanford has template overrides for both of these `.csv` templates, so I'll just upload those (and clean them up if required).

willgearty:
There were references in the code to `csv` templates that didn't exist, so I added them (mostly based on template overrides Stanford had). I also tweaked some of the existing `html` templates and made `classesbyFOO` more flexible by allowing for a user-specified template like `sectionsbyFOO` and `teachersbyFOO`.

Fixes #2730.


kkbrum:
Similar to how next to "Student Schedules" it says "TEX" "PS" etc, I think the printables with CSV vs HTML options should have those listed next to them so people are aware and can access them without altering the URL. At the very least though, there could be a sentence at the top of the page explaining which printables have CSV options and to add "/csv" to the URL. Otherwise no one will ever know?

I think CSVs are mainly useful in terms of files, not for viewing in the browser, so I think if the CSV option is selected, it should auto download the file like it does for "All Classes Spreadsheet" and "Viable Times". I don't really think there's a point to showing the CSV in the browser instead of downloading it??

In the CSV templates, every entry gets quotation marks around it, which I don't think is ideal?

willgearty:
All reasonable requests! I've added links to `.csv` versions of all of the applicable printables to `/printoptions`. I also made it so when you go to the url, it actually downloads as a `.csv` file instead of showing it in the browser.

kkbrum:
The column header and the column itself in the teacherlist csv aren't aligned (there's an empty column in the header before "time of first class").

image

kkbrum:
Should the names of the csv files across the variations of the sections/class subjects ones be different to signify what you clicked on? I'm imagining a scenario where an admin unfamiliar with the classes in the program clicks on 9-12 by accident instead of 7-8 and since there is no grade column in the CSV would never notice the difference? I think having the names overlap across the various sortings is fine because sorting doesn't make a difference/ is obvious, but i think naming the files based on grade/accepted status might be helpful?

kkbrum:
Adding columns to the csvs that make it clear which option you selected is fine too :)

willgearty:
Thanks for the feedback @kkbrum! I couldn't reproduce the problem you had with the column alignment, but I changed some things in the `teacherlist.csv` to maybe help? Let me know if it's fixed for you now (and if not, can you send me the actual `.csv` file?). I think the filename has the potential to get really long if we want to include all of the information, so I went ahead and added the columns you suggested (Min Grade, Max Grade, and Status) to both the `html` and `csv` versions.. Let me know what you think!

PR #2766, opened 2019-06-26
Fix alignment of gender field

### LINKED ISSUE #2760: Gender field alignment is off
If you set the ‘student_profile_gender_field’ Tag to True, the students are asked for their gender on the profile, but the alignment is off compared to other fields: 
image

kkbrum:
Fixes #2760


willgearty:
Looks good to me!

PR #2773, opened 2019-07-06
Filter program options for various import forms

### LINKED ISSUE #2757: Whoops when importing from program with no volunteer timeslots
Got a Whoops! page when importing volunteer timeslots from a program that didn't have any. An error pop up instead would be great.

willgearty:
This filters the possible programs that you can import things from for the resource and volunteer management pages. For each form, it will only include the programs that have things to import for that feature. It also excludes the current program, since it doesn't really make sense to import from a program to itself (and the heading of the form also says "Import from a previous program", which combined could confuse people).

I do think this makes #2768 obsolete, but we can keep it just in case someone gets to the page somehow without using the form.

Fixes #2757.


kkbrum:
Welp 

image

willgearty:
Can you try it now @kkbrum ?

kkbrum:
Womp 

image

willgearty:
How about now, @kkbrum?

willgearty:
:D

PR #2774, opened 2019-07-08
Skip calculating teacher big board graph data if missing any timestamps

### LINKED ISSUE #2753: Teacher big board errors if classes lack timestamps
Understandably, the teacher big board doesn't work for programs before we implemented class registration timestamps (#2013). However, admins can still add the module and attempt to go to the view, which results in a fairly unhelpful error (but ultimately related to the absence of timestamps for classes). Even if classes are not timestamped for a program, we should still be able to at least calculate the static statistics? Then either skip figuring out the graph data, only use the timestamped classes for it, or give the untimestamped classes a timestamp based on the first or last timestamp? (the second option would be weirdly misleading since the numbers won't match the static statistics)

willgearty:
Many old programs have classes without timestamps (pre-2016). If this is the case, we'll just skip calculating the data that would be used for the graph on the teacher big board (even if there are some classes with timestamps, because any solution for those would either be misleading because we'd need to assign an arbitrary timestamp or it wouldn't match up with the static numbers because we'd skip the ones without timestamps). I also added a little fix to `get_hours()` just in case there are some (but not all) classes without timestamps for when we are calculating the static numbers (which we want to do regardless of whether there are timestamps or not).

Fixes #2753.

### DIFF
--- a/esp/esp/program/modules/handlers/teacherbigboardmodule.py
+++ b/esp/esp/program/modules/handlers/teacherbigboardmodule.py
@@ -1,5 +1,4 @@
 import datetime
-import operator
 import subprocess
 
 from django.db.models.aggregates import Min, Sum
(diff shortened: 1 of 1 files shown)


willgearty:
I removed some of the timestamps for a couple classes for a test program and it worked fine for me.

PR #2779, opened 2019-07-14
Fix classroom import with incomplete availability

### LINKED ISSUE #2670: Incomplete availability makes classroom available for only one time slot
When you don't select the complete availability checkbox when importing classrooms, you see the correct availabilities in the preview, but only get a single timeslot of availability upon saving. It's possible this is due to #2653, but I'm not quite sure. Reported by Yale (on main after #2653 was merged).

willgearty:
When I implemented the selective import (#2653), I actual broke the functionality of importing classrooms with matching availability (incomplete availability). This _should_ fix it again.

Fixes #2670.

### DIFF
--- a/esp/esp/program/modules/handlers/resourcemodule.py
+++ b/esp/esp/program/modules/handlers/resourcemodule.py
@@ -357,28 +357,29 @@ class ResourceModule(ProgramModuleObj):
                 for i in range(min(len(ts_old), len(ts_new))):
                     ts_map[ts_old[i].id] = ts_new[i]
 
-                #   Iterate over the resources in the previous program
-                for res in past_program.getClassrooms():
-                    furnishing_dict[res.name] = set()
-                    #   If we know what timeslot to put it in, make a copy
-                    if res.event.id in ts_map:
-                        new_res = Resource(
-                            name = res.name,
-                            res_type = res.res_type,
-                            num_students = res.num_students,
-                            is_unique = res.is_unique,
-                            user = res.user,
-                            event = ts_map[res.event.id]
-                        )
-                        #   Check to avoid duplicating rooms (so the process is idempotent)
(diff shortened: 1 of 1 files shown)


hwatheod:
What is "incomplete availability" ? I couldn't figure it out from the code changes.

willgearty:
If you don't select the "complete availability" checkbox when importing classrooms.

esp/esp/program/modules/handlers/resourcemodule.py
```diff
@@ -357,28 +357,29 @@ def resources_classroom_import(self, request, tl, one, two, module, extra, prog)
...
-                        else:
-                            new_res.old_id = res.id
-                        if import_furnishings:
-                            new_furns = self.furnishings_import(res, new_res, self.program, import_mode, to_import)
-                            furnishing_dict[res.name].update(new_furn.res_type.name + (" (Hidden)" if new_furn.res_type.hidden else "") + ((": " + new_furn.attribute_value) if new_furn.attribute_value else "") for new_furn in new_furns)
-                        resource_list.append(new_res)
+                #   Iterate over the classrooms in the previous program
+                for resource in past_program.groupedClassrooms():
+                    for event in resource.timegroup:
+                        furnishing_dict[resource.name] = set()
```
hwatheod: Is it right for this to be inside the inner ("event") loop?  It seem likes this will reset the furnishing_dict for this resource.name for every item in the inner loop.  And the furnishing_dict doesn't get used inside the loop anywhere (only updated on line 381), so something seems off.  In the "complete availability" section, this line is inside the outer loop.

willgearty: `furnishing_dict` is only used to show the furnishings in the import preview, and it's at the classroom level, so you're probably right here. And since it's a set, we don't need to worry about duplicates from line 381.

willgearty:
That should do it @hwatheod .

Issue #2541, opened 2018-04-07
Managing global resources is unintuitive


willgearty:
Copied from #2432:

> A bug I noticed:If I create a resource, save it (by clicking "update resource"), then it shows up on the list of resource types underneath ("resource types for _____"). Then I click "edit" for that resource, check the "global" box (not sure what that checkbox means), and click "update resource" again, the resource I just added disappears from the list of resource types shown on the page. It's not really deleted though; it still shows up in the admin interface. (from @hwatheod)

> Global refers to whether a resource type is associated with a particular program (not global) or not (global). The behavior that you are describing appears to be the intended behavior of the original code, given that the default behavior for getResourceTypes() is to exclude global resources. So, when you set the resource type to global, it was no longer included in the resource types available in the form. I've modified the code such that it follows the logic of the teacher class reg module, in that it will show global resource types if the allow_global_restypes tag is set to True. Otherwise, the form will only show resources associated with the current program. Oddly, the code was set up to handle global resource types and the template already had logic to identify global resource types, so perhaps the getResourceTypes() default was changed at some point or something else changed? (from @willgearty)

It seems like this functionality could be better, but it's not clear to me what the _default_ functionality should be here. Should the option to make resources global be removed from the interface? Should we just show all resources (from the current program and global)?

willgearty:
Fixed by #2781.

PR #2781, opened 2019-07-22
Hide global resources option by default

### LINKED ISSUE #2756: Explain "Global" resource tag with help text
For adding a resource, would be nice if “Global?” was explained like “Only one?” and “Hidden?” are

willgearty:
Ultimately, this adds help text to the global option in the resource types manage page. But while doing that, I realized it didn't really make much sense to have this option if the `allow_global_restypes` tag was set to False, since you would make the resource type then it would seemingly disappear (as discussed by @hwatheod in https://github.com/learning-unlimited/ESP-Website/pull/2432#issuecomment-333393148). I thought about just including a little tip in the help text about the tag, but then the help text seemed more confusing than before for chapters that aren't super familiar with tags. So, in the end I decided to actually hide the `is_global` field when that tag doesn't exist or is set to `False`. While I was at it, I realized the use of allow_global_restypes was really inconsistent, so I updated all the instances I could find to use getBooleanTag with a default of `False` and updated the tag documentation.

Fixes #2756.

### DIFF
--- a/esp/esp/program/models/__init__.py
+++ b/esp/esp/program/models/__init__.py
@@ -906,7 +906,7 @@ class Program(models.Model, CustomFormsLinkModel):
             exclude_types += [ResourceType.get_or_create('Classroom')]
 
         if include_global is None:
-            include_global = Tag.getTag('allow_global_restypes')
+            include_global = Tag.getBooleanTag('allow_global_restypes', default = False)
 
         if include_global:
             Q_filters = Q(program=self) | Q(program__isnull=True)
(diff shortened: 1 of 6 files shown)


willgearty:
Oh, I guess this also fixes #2541.

willgearty:
Yeah, those resource types are created when a new website is made in that bit that you linked to. The tag is then used to determine whether to show those (and other global resource types) which I added in https://github.com/learning-unlimited/ESP-Website/pull/2432:
https://github.com/learning-unlimited/ESP-Website/blob/90ab715d88a1da5e48b13d048d4118c503ac391c/esp/esp/program/modules/handlers/resourcemodule.py#L637-L638
I think if chapters don't have the tag set, then we shouldn't confuse them by showing those on the resources page, just like we aren't going to show them the option in the form.

PR #2783, opened 2019-07-26
Separate handlers for availability and check availability modules

### LINKED ISSUE #2770: Userview availability link shows without availability module
Userview "check availability" link shows even if the availability module isn't enabled for that program.

willgearty:
The teach and manage availability modules were still in the same handlers, which meant you couldn't check specifically for one of them with `prog.hasModule()`. @kkbrum split them into separate handlers and I cleaned up the code afterwards.

Fixes #2770.


hwatheod:
Can you please fix the Python alerts from LGTM ?  I think they're all about unused imports and you just need to delete them.

willgearty:
I think that should be all of the unused imports @hwatheod.

willgearty:
I think we can ignore the "Except block handles BaseException" alert.

PR #2789, opened 2019-08-06
Check for check availability module when linking to edit_availability

### LINKED ISSUE #2786: Check for availability checker module before linking to conflicting teacher's edit availability
On the setting teachers admin page, if a user has a conflicting availability, I made it link to their edit availability page a while ago, but if the availability checker module is not enabled we should remove that link (otherwise you get a nice broken link oops)

willgearty:
Just in case the module isn't enabled, we should check before including links to the `edit_availability` page.

Fixes #2786.


willgearty:
Should be good now @kkbrum

PR #2791, opened 2019-08-07
Added blank teacher info for coteachers added manually

### LINKED ISSUE #2759: Update teacher links for a coteacher that was added without full teacher info
For the teacher links on userview, it uses the last program where there is teacher info. But if you add a coteacher and they are now teaching for a later program than they have filled out info for, it’d be cool to have the links for that program :) (so basically take the latest program out of the ones they have info for and the ones they are teaching for)

kkbrum:
Automatically add a teacher info if a coteacher is added but hasn't filled it out. This would really only happen if they started a basic account/ had an account from previous programs and then an admin manually added proper availability for the program and added them to a class. 

Fixes #2759

### DIFF
--- a/esp/esp/program/modules/handlers/adminclass.py
+++ b/esp/esp/program/modules/handlers/adminclass.py
@@ -46,6 +46,9 @@ from esp.program.modules.forms.management import ClassManageForm, SectionManageF
 from django.http import HttpResponseRedirect, HttpResponse
 from esp.middleware import ESPError
 from esp.program.controllers.classreg import ClassCreationController
+from esp.users.models import TeacherInfo
+from esp.program.models import RegistrationProfile
+
 
 """ Module in the middle of a rewrite. -Michael """
 
@@ -396,6 +399,10 @@ class AdminClass(ProgramModuleObj):
             if cls.conflicts(teacher):
                 conflictinguser = teacher
             else:
+                if not RegistrationProfile.getLastForProgram(teacher, prog).teacher_info:
+                    lastProf = RegistrationProfile.getLastForProgram(teacher, prog)
+                    lastProf.teacher_info = TeacherInfo.addOrUpdate(teacher, lastProf, {'graduation_year': ''})
+                    lastProf.save()
                 coteachers.append(teacher)
                 txtTeachers = ",".join([str(coteacher.id) for coteacher in coteachers ])
                 ccc.associate_teacher_with_class(cls, teacher)


esp/esp/program/modules/handlers/adminclass.py
```diff
@@ -396,6 +399,10 @@ def coteachers(self, request, tl, one, two, module, extra, prog):
             if cls.conflicts(teacher):
                 conflictinguser = teacher
             else:
+                if not RegistrationProfile.getLastForProgram(teacher, prog).teacher_info:
+                    lastProf = RegistrationProfile.getLastForProgram(teacher, prog)
+                    lastProf.teacher_info = TeacherInfo.addOrUpdate(teacher, lastProf, {'graduation_year': ''})
```
willgearty: Is there a reason you set graduation_year to `''`? Shouldn't we be passing all of the information that the teacher info should have? That includes `graduation_year`, `affiliation`, `major`, `shirt_size`, and `shirt_type`.

esp/esp/program/modules/handlers/adminclass.py
```diff
@@ -396,6 +399,10 @@ def coteachers(self, request, tl, one, two, module, extra, prog):
             if cls.conflicts(teacher):
                 conflictinguser = teacher
             else:
+                if not RegistrationProfile.getLastForProgram(teacher, prog).teacher_info:
+                    lastProf = RegistrationProfile.getLastForProgram(teacher, prog)
```
willgearty: This and the previous line are the same, meaning you're making the same call twice. Why not just move this line outside of the if loop, then do `if not lastProf.teacher_info`?

esp/esp/program/modules/handlers/adminclass.py
```diff
@@ -396,6 +399,19 @@ def coteachers(self, request, tl, one, two, module, extra, prog):
             if cls.conflicts(teacher):
                 conflictinguser = teacher
             else:
+                lastProf = RegistrationProfile.getLastForProgram(teacher, prog)
```
willgearty: If a teacher has a profile for a previous program, but not this program, this gets that old profile and spoofs it as being for this program, but then it doesn't save it (if the old profile is more than 5 days old for some reason). Instead of messing with what is presumably a really old decision, you could just make sure to always save this `lastProf` here by bumping the `lastProf.save()` below out one tab.

willgearty: Actually, it might be better to fix the problem at the source, since it does seem to be a little backwards (it probably makes more sense to not save if the other profile is less than five days old, not more). But just to be sure we can still do my suggestion above as well.

kkbrum: @willgearty I did the easy version, bumping the lastProf.save() over, but maybe we want to "fix the problem at the source" and fix the backwards logic in a different PR

willgearty: Sounds good. I'll test this as soon as I get a chance. You should feel free to make that other PR whenever you get a chance. I think the only change that is needed is to change the `<` to a `>` (or vice versa).

willgearty:
So, we should probably add the same functionality to the teach module, just in case an admin updates a teacher's availability, but not their profile, then a teacher tries to add them to a class.

PR #2794, opened 2019-08-23
Spoof regprof only when _less_ than 5 days old

willgearty:
When calling `getLastForProgram` for a user that doesn't have a registration profile for the specified program, it creates a new registration profile object for that program. However, before we were only saving this new registration profile when the one it was based on was last updated _less_ than 5 days ago. However, the intention here is to reduce the number of profiles that are made (and to try to be better about accessing the most recently updated information), so we probably want to save the new registration profile only when the one it was based on was last updated _more_ than 5 days ago.

### DIFF
--- a/esp/esp/program/models/__init__.py
+++ b/esp/esp/program/models/__init__.py
@@ -1429,7 +1429,7 @@ class RegistrationProfile(models.Model):
             regProf.program = program
             if regProf.id is not None:
                 regProf.id = None
-                if (datetime.now() - regProf.last_ts).days <= 5:
+                if (datetime.now() - regProf.last_ts).days >= 5:
                     regProf.save()
         else:
             regProf = regProfList[0]


hwatheod:
I don't understand the 5-day cutoff. If a profile isn't associated with a program, regardless of the age, is there a reason why we have to keep it around instead of overwriting it? What would it be used for? Perhaps I don't have a good overall picture of how profiles are used, and in particular, under what circumstances we need multiple profiles for a user.

willgearty:
Great points @hwatheod. I was originally going to try to just have a single registration profile for each user as you suggested, and then just change the program associated with the profile as they went from program to program. However, then I realized (thanks to @kkbrum) that that would break the "Teachers/students with profiles for this program" filter for the comm panel and arbitrary user list modules, because as soon as a user joins another program, they would no longer have a profile for the old program (which could be problematic when there are multiple concurrent programs). Aside from this functionality, though, it doesn't seem like anything else _needs_ a separate registration profile for each program, unless I'm missing something. We could theoretically add a new record for "Has a profile for program" or something which we could set when a profile is set for a program. Then we wouldn't need to worry about overriding the registration profile, because the record would still be there.

At the very least, it doesn't seem necessary to keep profiles that aren't associated with a program, so I've changed it here to always coopt those when making a profile for a new program, instead of making a new registration profile. Furthermore, I discovered that the /myesp/profile form always creates/saves a profile that isn't associated with a program (and if none exists, it makes one). Because we never need a profile that isn't associated with a program (if we have other profiles that are associated with programs), I added the most recent commit which now uses the most recent profile even if it is associated with a program.

As far as the 5 days cutoff, that was just the original cutoff that I kept (but fixed). We could probably do 7 days to make it a nice full week or even 30 days (a month) if we are fine with assuming information is usually up-to-date for that long.

In the end, the new functionality here provides some nice user-side benefits:
1. When a user has just made an account and filled out the normal profile form, they often then go to register for a program. Before, they would then have to submit the program-specific profile form again, but now this will just assume the info is already up-to-date and the profile module will be marked as completed, sending them to the next step automatically.
2. When a user registers for one program, they normally have to fill out/submit the profile form. If they then register for a concurrent program, before they would have to submit another profile form for the new program, but now the profile module will again be marked as completed and skip to the next step of registration automatically.

If the most recent profile (provided they don't have a profile for the program of interest) is older than the cutoff, they will be prompted to fill out the profile form like usual.

PR #2801, opened 2019-09-09
Generic program links for most recent/current program

### LINKED ISSUE #2793: Generic registration links
It would be cool to have a link like `[site].learningu.org/Splash/studentreg` that would take you to the student registration page for the most "current" program (a la #2234 and #2504). It could even use the "Splash" bit to determine which type of program you want (in case there are multiple programs at the same time like at MIT or Yale), such that you could have e.g. `[site].learningu.org/Splash/studentreg` and `[site].learningu.org/Sprout/studentreg` and they would always link to the most current "Splash" and "Sprout" programs, respectively. This would hopefully also work for teacher and volunteer registration.

I'm not really sure how this would work, but it'd be really cool and a great way to make sure links stay current.

willgearty:
This creates a system to generate/follow links for specific views within the most recent/current program (merely the one that is latest in time). The links are of the form `[site].learningu.org/[tl]/[one]/current/[view]`, where `[site]` is the specific chapter site; `[tl]` is "teach", "learn", "manage", or "volunteer"; `[one]` is the program type (e.g. "Splash", "Sprout", "HSSP"); and `[view]` is the specific page/view (e.g. "teacherreg", "studentreg", "dashboard", etc).

I should mention that any of the links on the page will then direct the user to a link with "current
 replaced with the actual instance (e.g. "2019_Spring"). I'm not sure if this is a good thing or a bad thing. We could make it maintain the "current" part, but that would require modifying a lot of templates.

The hope is that this will make it easier to set generic links on pages or even in emails.

Fixes #2793.

### DIFF
--- a/esp/esp/context_processors.py
+++ b/esp/esp/context_processors.py
@@ -21,21 +21,34 @@ def email_settings(request):
     return context
 
 def program(request):
-    path_parts = request.path.lstrip('/').split('/')
-    if len(path_parts) > 3:
-        program_url = '/'.join(path_parts[1:3])
-        if Program.objects.filter(url=program_url).count() == 1:
-            return {'program': Program.objects.get(url=program_url)}
+    if getattr(request, "program", None):
+        return {'program': request.program}
+    elif getattr(request, "prog", None):
+        return {'program': request.prog}
+    else:
+        path_parts = request.path.lstrip('/').split('/')
+        if len(path_parts) > 3:
+            program_url = '/'.join(path_parts[1:3])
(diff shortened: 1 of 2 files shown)


willgearty:
Those originally used the URL to determine the program and school year to include in the template context. Now, if there is a program specified in the request, it uses that instead (which gets around the fact that the URL doesn't have a specified program). I only noticed it once I started testing various student reg pages which broke because they didn't have the program or school year in the context. Those should be the only two context variables that use the URL.

PR #2803, opened 2019-09-11
Remove phase zero from checkboxes, simplify template handling

### LINKED ISSUE #2736: Phase zero checkbox link directs to error page
After running the phase zero lottery, you want to keep the module enabled such that only winners of the lottery can access the other student registration modules. If the user was not a winner, if/when they try to access `/studentreg` they will see a page that says "We're sorry, but you were not selected in this program's student lottery. We hope you will try again next time." This is the desired behavior. However, a student will also see this same page if they were selected and then click the phase zero link in the checkbox module list.

image

This is not really the desired behavior, and I imagine this could be very confusing to students.

willgearty:
Ultimately, all of this is to simplify the handling of the phase zero module. The first part of this is to remove the module from the checkboxes so students don't go back after they've progressed through to class registration. Then I fixed the edge case where students might navigate to the /studentregphasezero url, in which case they now get a `HttpResponseRedirect` to the main student registration view.

Finally, I realized that there was no need for multiple templates based on whether or not the lottery has been run or whether the student still had the permissions to enter the lottery. Now everything funnels to 1 of 3 templates (rather than 5): 1) The submission page where students first enter the lottery, 2) The confirmation page where students can see their lottery status (including whether or not they can still change lottery groups, whether or not the lottery has been run, and if so, if they weren't selected), or 3) The error page if the lottery is over and the student never entered. I think this will make things a lot easier for users because there is only one layout that they need to get accustomed to (and it will always show their lottery confirmation/information/members in case they need that information when corresponding with the chapter).

Fixes #2736.


willgearty:
Thanks for the comments, @hwatheod! I've added comments for each case in the python and html (it indeed adds up to the seven cases that you mentioned). I also realized I wasn't catching the case where the lottery isn't open yet, so I added that. Usually this is caught by the `@meets_deadline` decorator, but we want students to be able to access this module after their permission has expired (the permission is for entering the lottery), so we can't use that.

PR #2804, opened 2019-09-15
Include debug toolbar when morphed

### LINKED ISSUE #1193: Debug Toolbar Should Be Usable While Morphed
While morphed, most features of the site, including the debug toolbar, think you are the morphed user, rather than the morphing admin. This means you can't use the debug toolbar to figure out problems in a student's account. It would be nice to fix this logic.

willgearty:
This allows for admins to use the debug toolbar even when they are morphed into a user.

Note: this is only necessary when `settings.DEBUG = False` (i.e. not a development server) because otherwise the debug toolbar is always on, no matter whether the user is an admin, morphed, etc..

Fixes #1193.

### DIFF
--- a/esp/esp/middleware/debugtoolbar/middleware.py
+++ b/esp/esp/middleware/debugtoolbar/middleware.py
@@ -45,7 +45,8 @@ class ESPDebugToolbarMiddleware(DebugToolbarMiddleware):
         # proxy caching. See Github issue #739.
         enabled = ((not request.is_ajax()) and settings.DEBUG_TOOLBAR and (
                 (settings.DEBUG and not request.session.get('debug_toolbar') == 'f') or
-                (request.session.get('debug_toolbar') == 't' and request.user.isAdmin())))
+                (request.session.get('debug_toolbar') == 't' and
+                (request.user.isAdmin() or request.user.isMorphed()))))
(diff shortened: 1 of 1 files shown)


esp/esp/middleware/debugtoolbar/middleware.py
```diff
@@ -45,7 +45,8 @@ def custom_show_toolbar(request):
         # proxy caching. See Github issue #739.
         enabled = ((not request.is_ajax()) and settings.DEBUG_TOOLBAR and (
                 (settings.DEBUG and not request.session.get('debug_toolbar') == 'f') or
-                (request.session.get('debug_toolbar') == 't' and request.user.isAdmin())))
+                (request.session.get('debug_toolbar') == 't' and
+                (request.user.isAdmin() or request.user.isMorphed()))))
```
hwatheod: Is this right?  I don't see a function `isMorphed` defined anywhere in the codebase.  However, there is a function `is_morphed` defined in `esp/esp/users/models/__init__.py` , which I think is the right function to call.

willgearty: Whoops, good catch! I even wrote down that function name in the issue...

PR #2805, opened 2019-09-15
Custom teacher check-in text message

### LINKED ISSUE #2734: Make teacher text in /missingteachers customizable
Right now, the teacher text that is sent from the `/missingteachers` view is hard-coded (e.g. "Don't forget to check-in for your Splash class that is scheduled for Sun, May 05, 10:00am--11:45am!"):

https://github.com/learning-unlimited/ESP-Website/blob/176a2758abe1efc180ab8d16ca92e0572c8e8b98/esp/esp/program/modules/handlers/teachercheckinmodule.py#L155-L179

We probably don't want to allow admins to specify messages that include actual python like in the hard-coded message, but we could allow for a form field (or tag) that allows them to specify some sort of message that is just a string.

willgearty:
This was requested by a lot of chapters at SplashCon this past year. I've moved the teacher check-in text message to a template so now chapters can customize it how they wish! I'm happy to add other context variables if the reviewer thinks that would be useful, but I think I've covered the important ones (`prog`, `one`, `two`, `sec`, and `teacher`).

Fixes #2734.

### DIFF
--- a/esp/esp/program/modules/handlers/teachercheckinmodule.py
+++ b/esp/esp/program/modules/handlers/teachercheckinmodule.py
@@ -47,7 +47,7 @@ from esp.users.models    import ESPUser, PersistentQueryFilter, Record, ContactI
 from esp.cal.models import Event
 from django              import forms
 from django.http import HttpResponse, HttpResponseRedirect
-from django.template.loader import render_to_string, select_template
+from django.template.loader import render_to_string, get_template
 from django.db.models.aggregates import Min
 from django.db.models.query   import Q
 from datetime import datetime, timedelta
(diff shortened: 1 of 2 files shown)


esp/esp/program/modules/handlers/teachercheckinmodule.py
```diff
@@ -167,7 +167,9 @@ def ajaxteachertext(self, request, tl, one, two, module, extra, prog):
             if 'username' in request.POST and 'section' in request.POST:
                 sec = ClassSection.objects.get(id=request.POST['section'])
                 teacher = PersistentQueryFilter.create_from_Q(ESPUser, Q(username=request.POST['username']))
-                message = "Don't forget to check-in for your " + one + " class that is scheduled for " + sec.friendly_times(include_date = True)[0] + "!"
+                template = get_template(self.baseDir() + 'teachertext.txt')
+                context = {'prog': prog, 'one': one, 'two': two, 'sec': sec, 'teacher': teacher}
```
hwatheod: If you read the template from a file in the codebase, how will chapters customize it?  If you read it from a tag instead, it would be easier for chapters to customize. Since this is a text message, it will be relatively short so putting the entire thing in a tag is reasonable.

hwatheod: Oh, I just read the original issue #2734 and realized that they could use a template override.

PR #2420, opened 2017-08-18
Adds asterisk for required profile questions

willgearty:
Fixes #1109

image

Issue #2435, opened 2017-09-05
Fields marked required in __init__ don't have required class


willgearty:
If a field is marked as required in the `__init__` of a form (e.g. #2427), the `input` does not have `class="required"`.
This is problematic if we are marking the required fields like in #2420.

If we ever upgrade to django 1.10, this would be resolved as required fields automatically have the `required` HTML attribute.

willgearty:
Further, marking a field as not required in the `__init__` does not remove `class="required"` from the field, so you need to manually remove it (e.g. `self.fields['address_street'].widget.attrs['class'] = ""`)

willgearty:
This should be resolved now that we're on Django 1.11, but will need to check.

PR #2806, opened 2019-09-18
Required fields bug

### LINKED ISSUE #2629: Not all required profile form fields marked as required
Noticed by BC Splash. Looks like this relates to some fields that are processed on form submission. For example, we only require one phone number, so neither of the fields is marked as required in the form, but then the form submission errors when neither is submitted.

There appears to be a similar problem with the address fields, where the address may not be required if a certain tag is set (teacher_address_required), so maybe we are checking that on submission? Although these fields were marked when I tested this in #2420, so I'm a little confused.

image

kkbrum:
This PR addresses edge cases in required fields that were not caught by the broad * adding behavior in #2420. This fixes #2629. 

This is not super great code, but builds upon the spaghettiness of the current code? Let me know if anyone has better ideas. The fixes involve:

- starting the address fields as required and then changing them to not required based on tag behavior instead of vice versa (this is trivial but allows the fields to be caught by the * behavior)
- adding a blank option to otherdropdown fields that are not required
- adding asterisk manually to otherdropdown fields that are basically required
- adding asterisk manually to the "Phone numbers" label above the home/cell fields, one of which is required

PS. I made decisions in this PR about which other drop down fields I felt should be required vs not-- please let me know if any of my decisions are unwise.
- Required: teacher affiliation with school
- Not required: student transportation, student post-high school plans, student heard of esp

### DIFF
--- a/esp/esp/users/forms/user_profile.py
+++ b/esp/esp/users/forms/user_profile.py
@@ -49,10 +49,10 @@ class UserContactForm(FormUnrestrictedOtherUser, FormWithTagInitialValues):
     phone_day = USPhoneNumberField(required=False)
     phone_cell = USPhoneNumberField(required=False)
     receive_txt_message = forms.TypedChoiceField(coerce=lambda x: x =='True', choices=((True, 'Yes'),(False, 'No')), widget=forms.RadioSelect)
-    address_street = StrippedCharField(required=False, length=40, max_length=100)
-    address_city = StrippedCharField(required=False, length=20, max_length=50)
-    address_state = forms.ChoiceField(required=False, choices=zip(_states,_states), widget=forms.Select(attrs={'class': 'input-mini'}))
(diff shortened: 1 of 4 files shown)


willgearty:
@hwatheod I think the asterisk behavior is due to a Django bug: https://github.com/learning-unlimited/ESP-Website/pull/2427.

kkbrum:
For the dropdowns, the problem was that if you mark the "Otherdropdownwidget" as required, it requires you to write something in the "other" field as well, so that's no good, but currently they were implemented without a blank option, so the dropdowns were basically all required since they defaulted to one of the options. So I added blank options to the not required ones and manually marked required ones even though i didn't physically make them required due to the weird behavior with the "other" box if they are physically required.

As Will mentioned, the tag stuff seems to be a django thing so there's not much I can do about it unless we upgrade the django version (issue #2435). I agree that the code depending on the order of operations is not great, but I'm not sure when/if we plan on upgrading django?

Issue #956, opened 2014-01-17
User search controller cannot correctly combine filters


pricem:
Due to the use of combined Q objects, the UserSearchController (used by the communications panel and user list generator) does not return correct sets of users when multiple filtering criteria are used ("combination list").  This is an architectural issue and it may require some discussion to come up with a good solution.

Supriyo760:
Hi @willgearty and team,

I was reviewing this older issue and the linked problems (#1655, #1718, #1810) and I'd love to take a crack at finally resolving it.

**The Root Cause:** The bug stems from a well-known "gotcha" in the Django ORM regarding spanning multi-valued relationships (as @benjaminjkraft correctly identified years ago). When `UserSearchController` combines `Q()` objects using `&` for the same related model (e.g., `Record` events or `ResourceRequest`s), Django applies both conditions to a single `JOIN`ed row.

For example, asking for "Attended" `AND` "Confirmed Reg" generates SQL requiring a single `Record` row to simultaneously have `event='attended'` AND `event='reg_confirmed'`, which is impossible and returns 0 users.

**Proposed Solution:** Instead of trying to hack the `Q` objects to force subqueries, we should shift the architecture of `UserSearchController` to manipulate `QuerySet`s when performing `AND` operations across multiple distinct criteria blocks:

1. **QuerySet Intersection:** Instead of doing `Q1 & Q2` and feeding it into one `.filter()`, we can generate separate `QuerySet`s for each major criteria block and intersect them: `qs1.filter(pk__in=qs2)` or `qs1.intersection(qs2)`.

2. **Refactoring `query_from_postdata`:** Update the core combination logic so that "AND" lists generate base subsets of `ESPUser`s which are then iteratively filtered down, rather than building one gigantic conflicting `Q` object.

3. **Validation:** We can write a few specific unit tests in `test_usersearch.py` that intentionally select two mutually exclusive events (like the Spark 2013 example) and verify the intersection logic correctly returns users who possess both distinct records.

Since this touches the core of the communications panel and user list generator, I'll be sure to run the full test suite and test it thoroughly on a local database.

Could you assign this to me? I'd love to implement this architectural cleanup!

willgearty:
Sure, that all sounds good, thanks @Supriyo760 !

PR #2808, opened 2019-10-02
Add statistic to dashboard for students who enrolled in past & fix teachers who have taught before

### LINKED ISSUE #2441: Statistics on first-time vs. returning students
We need statistics on how many enrolled/attended students were first-time students and how many have attended a previous splash (similar to the teacher statistic "Teachers who have taught for a previous program")

kkbrum:
Fixes #2441 . Adds queries for students who enrolled in a past program, attended a past program, and students who have enrolled for the first time (this could be done via a combo list, but having it stand alone means it will show up on the dashboard).

Note that there seem to be unnecessary steps in the code to generate the Q objects for the past, but the past Q objects must be user based and not based on class subjects / registrations, otherwise when someone tries a combo list with a current class subject / registration, the Q objects will combine and search for class subjects / registrations that are simultaneously for a past and current program, which is impossible. By making the past Q objects be user based, this conflict will not occur.


esp/esp/program/modules/handlers/studentclassregmodule.py
```diff
@@ -149,6 +149,10 @@ def students(self, QObject = False):
         Enrolled = Q(studentregistration__relationship__name='Enrolled')
         Par = Q(studentregistration__section__parent_class__parent_program=self.program)
         Unexpired = nest_Q(StudentRegistration.is_valid_qobject(), 'studentregistration')
+        past_programs = [x for x in Program.objects.all() if len(x.dates()) and x.dates()[0] < self.program.dates()[0]]
+        Past = Q(studentregistration__section__parent_class__parent_program__in=past_programs)
+        past_users = ESPUser.objects.filter(Past).values('id')
```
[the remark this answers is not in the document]
hwatheod: line 178 does do a distinct, but only in the `else` branch. Not sure if that's okay.

milescalabresi:
Could this also include students who have attended? If so, then this PR fixes #2842.

willgearty:
@kkbrum what still needs to be done for this PR?

kkbrum:
This should be done now, but I need to do a bit more testing to make sure everything works as intended!

kkbrum:
So if I just put &s or |s or ~s in my shell_plus to combine queries, I think everything works correctly. 

However, if I load my devsite, and try something like "teachers who have taught before" & "teachers who have submitted a class", I get a list index out of range error coming from [here](https://github.com/learning-unlimited/ESP-Website/blob/323326e8b9aa3c6480be811815022e335bf18534/esp/esp/users/controllers/usersearch.py#L315). I know that the children of my qobject is only a list of length 1 and not 2, but I'm unclear about what the point of this chunk is, so I'm not very comfortable changing it. 

Note that if I do something like "teachers teaching a nearly full class" & "teachers who have submitted a class", I get the same error but I didn't change either of those queries in this PR. In fact, I checked this on Yale's site and it gives an error as well. So this isn't my fault, but I'm a bit scared to touch that code without understanding why it was put there.

willgearty:
After digging into this a bit, it's first important to note that `children` does not refer to the number of objects returned by the query (these are Q objects, so they aren't querying the database yet, that happens when you put them in  `.filter()`). Rather, it refers to the subqueries that make up the Q object (either within a single `Q` call or with `&`s/`|`s). For example, for the `class_nearly_full` query, it consists of two subqueries:
https://github.com/learning-unlimited/ESP-Website/blob/bdf76c7e047dd08cf8f5e6975358e7574158fb37/esp/esp/program/modules/handlers/teacherclassregmodule.py#L138-L141

`Q_nearly_full_teacher .children[1]` would return `Q_isteacher` in this example.

For the vast majority of qobjects, they have at least 2 subqueries (because we are usually filtering by at least program and something else), but the "has submitted a class" query only has a single subquery:
https://github.com/learning-unlimited/ESP-Website/blob/bdf76c7e047dd08cf8f5e6975358e7574158fb37/esp/esp/program/modules/handlers/teacherclassregmodule.py#L129

This is why that line of code errors when that query is included in the combination list. Note that if you use that for the original list and then use the "teachers teaching a nearly full class" as your AND clause, it should proceed (although not necessarily with the expected result).

Now, back to this bit of code. It looks like this bit of code was written very specifically to handle qobjects that deal with records (see the discussion here: https://github.com/learning-unlimited/ESP-Website/issues/956).

For example, when you look at the 'student_custom_form' query:
https://github.com/learning-unlimited/ESP-Website/blob/c6a6ace7304f636af23e92c4c11e106f9c03d654/esp/esp/program/modules/handlers/studentcustomformmodule.py#L80

`qo.children[1]` returns `('record__event', 'student_extra_form_done')` (even though it is first in the Q object, it seems to go second in the children for some reason).

Therefore, we can infer that this code is not intended to be very flexible. It was intended to specifically handle multiple combination lists that all deal specifically with records. A quick fix for the error you are encountering would therefore be to add another if statement between lines 314 and 315 to check if there are at least 2 children in the qobject.

However, that won't solve the larger problem here, which is that we now have more queries that have overlapping filters. Unfortunately, it's not immediately clear to me if we can just broaden the code here to handle other cases, if we need handle specific cases differently, or if we need to scrap it all and try again.

willgearty:
Maybe the solution is to change all queries to filter on user ID (like you've done with some of them here)? I'm guessing that would have a bit worse performance, but maybe worse performance is better than confusing/incorrect behavior?

kkbrum:
I'm having a bunch of "or" teacher combo lists come up empty even though they shouldnt be (and aren't in my shell). I can't figure out why though!

willgearty:
Hmm...is it possible the problem is caused by the default grade filters that we now have?

willgearty:
Should be fixed now!

esp/esp/program/modules/handlers/teacherclassregmodule.py
```diff
@@ -190,6 +188,7 @@ def teacherDesc(self):
             'class_full': """Teachers teaching a completely full class""",
             'class_nearly_full': """Teachers teaching a nearly-full class (>%d%% of capacity)""" % (100 * capacity_factor),
             'taught_before': """Teachers who have taught for a previous program""",
+            'taught_before_and_now': """Teachers who have submitted a class and who have taught for a previous program""",
```
willgearty, line 191: Shouldn't this just be a combo list search?

kkbrum: yes. but i dont know how to add something to the dashboard without doing it that way-- let me know if there's a way.

willgearty (approved):
Looks good to me! I checked the code, ran some AUL queries, and double-checked the AUL and dashboard numbers are the same.

PR #2809, opened 2019-10-02
Add option to include unreviewed classes in scheduling diagnostics

### LINKED ISSUE #2718: Include unreviewed classes in scheduling diagnostics
I don't know if this applies to any other programs, but Stanford sometimes doesn't approve classes until AFTER scheduling takes place. Unfortunately, almost all of the scheduling diagnostics only use approved classes, so they aren't much use to us right after we schedule classes. For us at least, it would be great if unreviewed classes were included in the scheduling diagnostics (at least where it makes sense).

willgearty:
This adds a button to the scheduling diagnostics page which allows you to switch between 1) only including approved classes and 2) including unreviewed and approved classes.

(I also included a backlink to the main manage page)

Fixes #2718.

### DIFF
--- a/esp/esp/program/modules/handlers/schedulingcheckmodule.py
+++ b/esp/esp/program/modules/handlers/schedulingcheckmodule.py
@@ -10,6 +10,8 @@ from esp.users.models import ESPUser
 from esp.tagdict.models import Tag
 from esp.cal.models import Event
 
+from esp.middleware.threadlocalrequest import get_current_request
+
 import json
 import re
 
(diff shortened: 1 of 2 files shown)


willgearty:
That function actually grabs it's own day of classes, so it's not dependent on this change:
https://github.com/learning-unlimited/ESP-Website/blob/0d597d0827ddfcbdf6505a1c19f63fa337571605/esp/esp/program/modules/handlers/schedulingcheckmodule.py#L275

hwatheod:
I see.  It's different from all the others which use the `self._all_class_sections`.

PR #2810, opened 2019-10-09
Usersearch sort and include last program

### LINKED ISSUE #2395: Display or sort by last program or similar information in user search
Sometimes when we want to look up a person by their first or last name (e.g. if we only know their parent's name), we get a long list of users. Usually we're looking them up because they are attending or have attended our current or most recent program, so displaying or sorting by this information somehow might help us disambiguate.

willgearty:
This makes the table on the usersearch results page sortable (using `sorttable.js`). I included the last program (with a profile) as a column, separated the name column into first and last name columns, and sorted the user list by the last program for the initial population of the table. I also added some styling to make it look nicer and moved the link to the userview page to the username field.

image

Fix #2395.

### DIFF
--- a/esp/templates/users/userview_search.html
+++ b/esp/templates/users/userview_search.html
@@ -80,7 +80,7 @@ tbody > tr:hover {
         {% endif %}
         {% endwith %}
       </td>
-      <td sorttable_customkey="{{ user.get_last_program_with_profile.dates.0|date:'Ymd' }}">{{ user.get_last_program_with_profile }}
+      <td sorttable_customkey="{{ user.get_last_program_with_profile.dates.0|date:'Ymd' }}">{{ user.get_last_program_with_profile }}</td>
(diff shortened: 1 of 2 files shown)


esp/templates/users/userview_search.html
```diff
@@ -19,42 +19,72 @@
...
+        {% if types %}
+        {% for type in user.getUserTypes %}
+        <span class="{{ type }}">{{ type }}</span>{% if not forloop.last %}, {% endif %}
+        {% endfor %}
+        {% else %}
+        <span class="no_usertype">None</span>
+        {% endif %}
+        {% endwith %}
+      </td>
+      <td sorttable_customkey="{{ user.get_last_program_with_profile.dates.0|date:'Ymd' }}">{{ user.get_last_program_with_profile }}
```
hwatheod: Does this need a closing `</td>` ?

willgearty: Whoops, good catch, thanks!

PR #2816, opened 2019-10-20
Add UI for program, crmi, and scrmi settings

### LINKED ISSUE #2659: UI for program settings
Consensus from SplashCon 2019 is that very important program settings should not be hidden in the admin panel. We need a UI for altering these settings so admins don't need to worry about sifting through the admin pages. Such settings include:

- Capacity options
- Open reg
- Cancellation
- Class capacity multiplier/offset
- Cancel button/message

willgearty:
Adds a friendlier (than the admin panel) user interface for admins to change settings associated with the program (such as those that are set in the new program creation wizard), teacher registration, and student registration. The settings are in groupings that made sense to me, but I'm happy to move things around. Also, if either the program type or program term are changed (resulting in a change in the base url for the program), the user is redirected to the new settings page for the updated program.

Fixes #2659.


willgearty:
Hmm... Should we also have a form with fields for some of the more common program-specific tags?

esp/esp/program/modules/handlers/admincore.py
```diff
@@ -83,6 +84,69 @@ def main(self, request, tl, one, two, module, extra, prog):
...
+                    form = StudentRegSettingsForm(request.POST, instance = scrmi)
+                    scrmi_form = form
+                if form.is_valid():
+                    form.save()
+                    #If the url for the program is now different, redirect to the new settings page
+                    if prog.url is not old_url:
+                        return HttpResponseRedirect( '/manage/%s/settings' % (prog.url))
+
+        #Set up any other forms on the page
+        if submitted_form is not "Program Settings":
```
hwatheod: Why do you use `is not` here (and in several lines below) rather than `==`?  You want string equality, not pointer equality.

esp/esp/program/modules/forms/admincore.py
```diff
@@ -44,3 +49,59 @@ def save_data(self):
...
+    student_reg_end   = None
+    def __init__(self, *args, **kwargs):
+        super(ProgramSettingsForm, self).__init__(*args, **kwargs)
+
+    def save(self):
+        prog = self.instance
+        LineItemType.objects.filter(text='Program admission',program=prog
+        ).update(amount_dec=Decimal('%.2f' % self.cleaned_data['base_cost']))
+        prog.sibling_discount = self.cleaned_data['sibling_discount']
+        return super(ProgramSettingsForm, self).save()
```
hwatheod: Why do we need this code in this form?  Shouldn't the `save` logic already be somewhere else?

willgearty: The `ProgramCreationForm`'s `save` only processes the model fields associated with the `Program` model, including the `name` and `url` fields. We then set the permission and accounting details after the fact in `commit_program`:
https://github.com/learning-unlimited/ESP-Website/blob/36158a4eb5a0214fa8c9612cbcc47c0436f014aa/esp/esp/program/views.py#L504-L506

https://github.com/learning-unlimited/ESP-Website/blob/32ed5e9816c99cdbe5f8555765e25b3d40cd2697/esp/esp/program/setup.py#L66

esp/esp/program/modules/handlers/admincore.py
willgearty: Whoops, must have gotten my python and django template logic mixed up here. Thanks for catching it!

willgearty:
Anything else @hwatheod ?

PR #2818, opened 2019-10-23
Classes Dashboard sorting

### LINKED ISSUE #2367: Dashboard won't sort classes by size or requests
The "size" and "special teacher requests" (message to directors or planned purchase) sort options on the dashboard don't work as expected. See [here](https://github.com/learning-unlimited/ESP-Website/blob/stable-release-8/esp/public/media/scripts/program/modules/adminclass.js#L312) for the Javascript that controls all of the sorting options (the other three appear to work as far as I can tell).

milescalabresi:
Allows sorting classes by multiple fields. Fixes #2367

### DIFF
--- a/esp/public/media/scripts/program/modules/adminclass.js
+++ b/esp/public/media/scripts/program/modules/adminclass.js
@@ -290,6 +290,7 @@ function createStatusButtonTd(clsObj) {
 function createClassRow(clsObj)
 {
     var tr = $j(document.createElement('tr'));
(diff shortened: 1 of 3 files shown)


willgearty:
I see you made changes to sorttable.js. Does that affect other implementations of sorttable?

milescalabresi:
It shouldn't. The changes I made move the up/down arrow icon inline with the column name, rather than being a child node. That should be cosmetic unless some other use assumes the arrow icon will be in its own row

PR #2825, opened 2019-10-30
Add floating resource choices

### LINKED ISSUE #2615: Floating resource choices
Admins should be able to specify particular resource type choices for their floating resources, just like #2547.

willgearty:
This adds a choice field to the floating resource form now that #2311 is implemented (and to follow up on #2547 where this was implemented for the classroom form and #2432 where this was implemented for the resource type form). Just as in #2547, the choices field is changed to a dropdown menu if the selected resource type has set options. The selected choice is also now displayed in the table of floating resources.

I also fixed a couple bugs with the floating resource form when editing floating resources (I guess no one really does that, because they haven't been reported before to my knowledge). And I fixed a couple bugs with my previous implementation in #2547 related to the initialization of the classroom form.

I tested this pretty thoroughly. The creation and editing of floating resources seems to work perfectly, and the creation and editing of classrooms still works as it did before.

Fixes #2615.

### DIFF
--- a/esp/esp/program/modules/forms/resources.py
+++ b/esp/esp/program/modules/forms/resources.py
@@ -109,10 +109,11 @@ class EquipmentForm(forms.Form):
     name = forms.CharField()
     times_available = forms.MultipleChoiceField()
     resource_type = forms.ChoiceField()
+    choice = forms.CharField(label = "Choice (optional)", required=False, max_length=50)
 
     def __init__(self, *args, **kwargs):
         if isinstance(args[0], Program):
-            self.base_fields['resource_type'].choices = setup_furnishings(args[0].getResourceTypes())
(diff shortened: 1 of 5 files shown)


kkbrum:
So I tried to import floating resources from a previous program (either full avail or not-- same in both cases), but the times were messed up for the import-- it showed up correctly for the preview, but then it only imported the first block instead of all the blocks in the preview

willgearty:
Hmm...that sounds like https://github.com/learning-unlimited/ESP-Website/issues/2670, which means it probably just needs a similar fix to that in https://github.com/learning-unlimited/ESP-Website/pull/2779

willgearty:
That should do it @kkbrum. Thanks for being thorough and let me know if you notice anything else.

Issue #2385, opened 2017-05-08
Error in renaming classrooms


milescalabresi:
Reported by Princeton: when they change the name of a classroom in which a class is already scheduled, the room number is not recognized as part of the classroom name, and the AJAX scheduler will not load because of an uncaught TypeError. It is not clear if this error is restricted to their site or this program instance.

willgearty:
Maybe related to #1130 and/or #1391?

benjaminjkraft:
Unclear if it's related to #1391.  #1130 was a similar issue with the old scheduler which presumably is gone now and replaced by this one.

@milescalabresi what do you mean by "the room number"?  Are they making a specific type of change here?

benjaminjkraft:
Actually, I realize, it may be this is because of the same underlying caching issue as the old scheduler.  But in any case let's keep this one to be clearer that it's for the new scheduler.

milescalabresi:
Yeah, they were changing the room number from "buildingname 3" to "buildingname 03" to test without confusing anyone as to the location. That seemed to break the scheduler. Here's a screenshot
![image](https://cloud.githubusercontent.com/assets/5506191/26026611/751a2040-37cc-11e7-962c-749173107455.png)

willgearty:
I know this was a long time ago, but was the problem fixed by flushing the cache?

willgearty:
I remember someone at SplashCon mentioning that changing a class name caused a bunch of problems with the scheduler. Maybe @milescalabresi or @bhorowitz was there and remembers who it was/more details?

Issue #1408, opened 2014-10-17
Filtering rooms in the scheduler


benjaminjkraft:
This used to be a feature, and there have been a number of requests for it to return, bigger and better than ever.  Possible filters include:
- size range
- resources
- name
- building (or start of name)

Issue #1391, opened 2014-10-17
Scheduler stores classrooms by name rather than ID


benjaminjkraft:
This means that if a classroom is renamed, you can't schedule anything into it until refreshing the scheduler.

willgearty:
Is this related to #1130?

benjaminjkraft:
It looks like this is still true in the new scheduler, although I'm not totally sure.

willgearty:
I attempted to fix this, but then realized that "classrooms" don't really have IDs. Classrooms as they are designed in the database are a set of resources, each of which has a unique ID (and a unique resource group ID). Therefore, I think this will be a bit bigger change than I was anticipating given how ingrained classrooms names are in the code.

willgearty:
Actually, looking at the code for assigning rooms to classes, it looks like we only really need to know one of the IDs of any of the resources for a classroom:
https://github.com/learning-unlimited/ESP-Website/blob/cbba79cec28f0a8bdb5054236295750fff4adc37/esp/esp/program/models/class_.py#L677-L679

PR #2834, opened 2019-11-10
Filter classrooms in the scheduler

willgearty:
This adds a third tab to the ajax scheduler to allows you to filter classrooms by capacity 
(min/max), resources/values (regex), and name (regex). The classes that don't match the filters are hidden.

While I was at it, I fixed the problem where classrooms were stored as names instead of IDs. Since classrooms don't have true IDs, I used the minimum ID of the resources that make up the classroom. This should fix a lot of problems related to renaming classrooms during scheduling. I was going to modify the ajax change log models, but it doesn't really seem necessary for the amount of work required, so I kept the room field as `room_name` but now it will store the room's ID. This shouldn't be a problem.

Also, I'm pretty sure this will break the automatic scheduling assistant, but I honestly don't have enough experience with it to know what to adjust to fix it. I'd be happy to work with @jerrywu64 to integrate these changes.

Also also, I may have to fix some tests or the ajax scheduler spec, but I'm pushing for now to start the review process.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/1408.
Also fixes https://github.com/learning-unlimited/ESP-Website/issues/1391 and probably fixes https://github.com/learning-unlimited/ESP-Website/issues/2385.

### DIFF
--- a/esp/esp/program/modules/handlers/ajaxschedulingmodule.py
+++ b/esp/esp/program/modules/handlers/ajaxschedulingmodule.py
@@ -117,15 +117,15 @@ class AJAXSchedulingModule(ProgramModuleObj):
 
         return self.makeret(prog, ret=True, msg="Schedule removed for Class Section '%s'" % cls.emailcode())
 
-    def ajax_schedule_assignreg(self, prog, cls, timeslot_ids, classroom_names, user=None):
+    def ajax_schedule_assignreg(self, prog, cls, timeslot_ids, classroom_ids, user=None):
         if len(timeslot_ids) < 1:
             return self.makeret(prog, ret=False, msg="No times specified!, can't assign to a timeblock")
 
-        if len(classroom_names) < 1:
+        if len(classroom_ids) < 1:
             return self.makeret(prog, ret=False, msg="No classrooms specified!, can't assign to a timeblock")
 
-        basic_cls = classroom_names[0]
-        for c in classroom_names:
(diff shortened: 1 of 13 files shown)


kkbrum:
This doesn't feel super reproduceable, but I'm having lots of problems with my entire table of classrooms and classes disappearing when I add certain resources to the rooms (cache flushed, hard refreshed, etc to no avail). Basically it seems it's fine if I add the resources that I created, which in this particular case were hidden global resources with text typed into the choice field, but if i add one of the default global options like computers with one of the drop down choices selected, then my entire scheduler decides I have no rooms/classes. I don't really know what's going on and can't give super specific instructions to reproducing it, but I do think there's a bit of weird behavior with resources and the scheduler...

willgearty:
Can you see if that behavior occurs when you are on main @kkbrum?

willgearty:
Alternatively, @kkbrum, could you just take a screenshot of what the browser console looks like after you load the page?

kkbrum:
image

kkbrum:
on Main: 
image

kkbrum:
Main works after flushed cache, branch doesnt after flushed cache and fab refresh. config:

image

image

willgearty:
Very strange...I'll look into this when I get a chance!

willgearty:
Should be fixed now! Give it another go when you get a chance!

PR #2840, opened 2019-11-17
Specify number of unique items for a floating resource

### LINKED ISSUE #2832: Specify number for each floating resource
I'm really hoping that https://github.com/learning-unlimited/ESP-Website/issues/2744 will encourage more people to use floating resources (since right now I'm pretty sure only MIT uses them). However, one thing I'm worried about is in order to accommodate common floating resources (e.g. cables, markers, etc), you would currently need to add many individual floating resources with unique names. I could maybe see this being ideal in a situation where each item needs to be kept track of (serial numbers/unique IDs on projectors/etc), but for things like cables or markers, you probably just need to know how many of that floating resource you have.

The good thing is, I don't think we need to change much in terms of the codebase. You can already have multiple floating resources with the same name and same event object. For the resources page, the backend just collapses these. We can use that to our advantage and just count how many objects are collapsed:
https://github.com/learning-unlimited/ESP-Website/blob/d998b14f80cc8a153b897081250825b5c421b4be/esp/esp/program/models/__init__.py#L717-L733

We could include this as an attribute for each floating resource like we include `timegroup`. Then we just need to add a form field for the floating resources, and make that many of resources in the db when the form is submitted. For #2744, we can just assign unique resources and we should be able to count how many of a given floating resource are assigned/unassigned.

willgearty:
This adds a field to the floating resources form that allows the user to specify the number of unique floating resources that will be available (e.g. 5 expo markers or 15 VGA cables). It doesn't change the models because it just makes that many duplicate resource objects for each of the specified timeslots. I also fixed a little HTML copy-paste bug I introduced in https://github.com/learning-unlimited/ESP-Website/pull/2817, so a resource section should always be open when using that specific resource section.

Fixes #2832.

### DIFF
--- a/esp/esp/program/models/__init__.py
+++ b/esp/esp/program/models/__init__.py
@@ -724,6 +724,7 @@ class Program(models.Model, CustomFormsLinkModel):
                 result[c.name].furnishings = c.associated_resources()
                 result[c.name].sequence = c.schedule_sequence(self)
                 result[c.name].prog_available_times = c.available_times_html(self)
+                result[c.name].num_items = c.number_duplicates()
             else:
                 result[c.name].timeslots.append(c.event)
 
--- a/esp/esp/program/modules/forms/management.py
+++ b/esp/esp/program/modules/forms/management.py
@@ -2,6 +2,7 @@
 from django import forms
 
 from esp.cal.models import Event
+from esp.middleware import ESPError
 from esp.resources.models import ResourceType, Resource
 from esp.program.modules.handlers.grouptextmodule import GroupTextModule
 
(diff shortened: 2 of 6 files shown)


kkbrum:
When I tried to add the floating resource that I made with num=3 to a class via the manage class page, I got the following: 

image

willgearty:
Hmm...is the floating resource that you tried to assign available at the time for which the section is scheduled?

willgearty:
I think that should fix your problem, @kkbrum. Now we'll at least throw a useful error telling the user that the resource isn't available at the section's time.

Issue #1355, opened 2014-10-09
Class flag filtering in scheduler


benjaminjkraft:
Designated class flags should show up, possibly without detail, in the AJAX scheduler.

willgearty:
Class flags are now included in the scheduler, but flag filtering would be great!

PR #2844, opened 2019-11-20
Add flag, resource, and admin class filters to scheduler

willgearty:
This adds a few more class filters to the ajax scheduler, including a field to search flags, a field to search resources, and a checkbox to filter for classes that have at least one teacher that is an admin (suggested by @kkbrum).

Fixes https://github.com/learning-unlimited/ESP-Website/issues/1355


kkbrum:
In a series of adding a flag, changing a resource type from hidden to not, and requesting a resource for a class, I now have the issue with the blank scheduler.  I think I undid the changes I made one by one but its still broken, so i'm not quite sure what change caused it. The console looks like this: 
image

Perhaps this is similar to the issue I was having with #2834? Maybe merging main into this branch will help?

willgearty:
Can you give it another go now, @kkbrum?

willgearty:
Any luck @kkbrum ?

Issue #2626, opened 2018-10-06
Admin page to monitor email sending


benjaminjkraft:
Our email framework is, ah, not the most stable.  But even when it is, admins have very little visibility into it.  We should make a page that has a list of emails sent in the last month, and for each:
- whether the message request is processed
- how many emails it will send
- how many of those are sent
We can also include useful text like "if you see something on this page, but it isn't sending, please do not re-send it!  instead tell websupport."

PR #2852, opened 2019-11-27
Admin page to monitor emails

willgearty:
This adds a page that admins can use to more easily (and quickly) monitor email requests that have been sent through the comm panel. It includes information about the email request (subject, text, sender), whether the request has been processed, how many recipients there are/will be, and how many emails have been sent so far.

By default the view only pulls email requests from the last 30 days, but that can be overrided with the `start_date` GET parameter (as described in the template).

Fixes https://github.com/learning-unlimited/ESP-Website/issues/2626.

### DIFF
--- a/esp/templates/admin/emails.html
+++ b/esp/templates/admin/emails.html
@@ -1,0 +1,104 @@
+{% extends "main.html" %}
+
+{% block title %}Monitor Emails{% endblock %}
+{% block content_title %}Monitor Emails{% endblock %}
+
+{% block stylesheets %}
+{{ block.super }}
+<style>
+table.emails-table {
+  border: 1px solid black;
+}
+  
+table {
+  width: 100%;
+  table-layout: fixed;
+  word-wrap: break-word;
+  border-collapse: collapse;
+}
+
+th {
+  background: black;
+  color: white;
+}
+
+tr.email-header > th, tr.email-header > td {
+  border-bottom: 1px solid #ddd;
+}
+
+th, td {
+  padding-top: 5px;
+  padding-left: 5px;
+  padding-bottom: 5px;
+  padding-right: 5px;
+}
+
+tr.email-header:hover {
+  background-color: #f5f5f5;
+}
+
+tr.email-details {
+  display: none;
+  border-left: 1px solid #ddd;
+  border-right: 1px solid #ddd;
+  border-bottom: 1px solid #ddd;
+}
+</style>
+{% endblock %}
+
+{% block content %}
+
+<p>This page lets you monitor emails that have been sent through the website via a program's comm panel within the last month.</p>
+<p>You can extend the search by appending "?start_date=MM/DD/YYYY" to the URL.</p>
+<p>If you notice any problems (emails aren't getting processed, emails won't send, emails are not sent to everyone, etc.), please contact <a href="mailto:websupport@learningu.org">websupport</a>.</p>
+
+{% if not requests %}
+<p>There are no emails to monitor.</p>
+{% else %}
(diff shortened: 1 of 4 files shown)


esp/templates/admin/emails.html
```diff
@@ -0,0 +1,104 @@
...
+            <td>+</td>
+            <td><b>{{ req.subject }}</b></td>
+            <td style="text-align:center;">{% if req.processed %}&#10003;{% else %}&#10005;{% endif %}</td>
+            <td style="text-align:center;">{{ req.num_rec }}</td>
+            <td style="text-align:center;">{{ req.num_sent }}</td>
+        </tr>
+        <tr class="email-details">
+            {% autoescape off %}
+            <td colspan="3" style="vertical-align:top;">{{ req.msgtext }}</td>
+            {% endautoescape %}
```
hwatheod: Is `autoescape off` really the right thing to do? It would only be right if `req.msgtext` was already HTML-escaped, but (I think) if you send plaintext rather than HTML messages, it would not be HTML-escaped.

willgearty: Hmm...I think I was doing this so that we display the tags properly (and don't try to process them in the template). Is there a more appropriate way to do that?

hwatheod: Not sure what you mean. The purpose of escaping is so that, if e.g. someone writes `<tr>` in an email, then it gets converted to `&lt;tr&gt;` in the HTML generated by the template, and then the browser will render it as the text `<tr>` in the web page. If you turn autoescape off, then it will generate `<tr>` directly in the HTML, and then the browser will interpret `<tr>` as an HTML in the web page, which will mess up the entire page and is not what you want.  

The only reason for turning autoescape off is if the code had already HTML-escaped the variable `reg.msgtext` (i.e. replaced `<tr>` with `&lt;tr&gt;` before being passing the variable to the template).

There's a simple way to test this, which is to send a message with some HTML tags in it, and see how it renders.

willgearty: Oh, I meant the django template tags. But also, we want to render the email in full HTML (to go along with how admins can now write emails in rich text). When testing this, I used some dummy emails that had lots of HTML tags in them (from previous Stanford emails) and they rendered perfectly in this view. I guess the only concern is that, if the email has incorrectly used HTML tags that it could cause problems with rendering the page (but the rich text/source code editor in the comm panel should help to limit this because it automatically makes close tags).

hwatheod: Got it. The only thing is, there can be something of a [security hole](https://en.wikipedia.org/wiki/Cross-site_scripting) if you inject HTML directly from a database into the web page, if some of the contents of the database is attacker-controlled.  In this case, if some injected a `<script ...>` tag into an email that pointed to a malicious script, then any admin viewing this page would run that script. 

Most of the text of the email is controlled by admins so the risk is minimal (I don't think an admin would try to attack other admins), but we do (or do we?) substitute the actual student or teacher's first name for tags like `{{ first_name }}`.  Since student/teacher accounts can be created by random people, the risk a little higher.  (To be clear, they could create an account where they put HTML tags in their name.)  

Does the `reg.msgtext` have the names substituted in, or does it still have the string `{{ first_name }}` in it ?  If it still says `{{ first_name }}` then we are okay. Otherwise we are taking this risk.

willgearty: reg.msgtext still has the string {{ first_name }} in it. The substitutions are made in the email texts. So I think the only concern would be admin-on-admin attacks.

hwatheod: Ok, then I am fine with merging this.

Issue #2557, opened 2018-04-17
Standardize barcodes on student schedules


willgearty:
Based on discussion in #2506, it's unclear whether the implementation of barcodes on student schedules is standardized across chapters. Even if it is, it would be good to implement this on main so that other chapters have the option of using them.

willgearty:
We probably don't want to have the barcodes there by default, so maybe use a tag to turn them on?

PR #2854, opened 2019-11-30
Student schedules with barcodes

willgearty:
I used the template overrides from Stanford and Yale to make a new student schedule LaTeX template. It has a barcode on it for use with the barcode scanning student checkin module. (also, wow, student schedules haven't been updated since 2011)

image

Fixes https://github.com/learning-unlimited/ESP-Website/issues/2557.

### DIFF
--- a/esp/templates/program/modules/programprintables/studentschedule.tex
+++ b/esp/templates/program/modules/programprintables/studentschedule.tex
@@ -71,7 +71,7 @@ Student Schedule for \\
 {% for cls in student.classes %}\hline
 {% if cls.friendly_times|length_is:"0" %}N/A{% endif %}{{ cls.friendly_times|join:", " }} \newline
 \footnotesize{ {% if cls.initial_rooms|length_is:0 %}N/A{% else %}{{ cls.prettyrooms.0|texescape }}{% endif %} } &
-{% if cls.title|length_is:"0" %}N/A{% endif %} {{ cls.title|truncatewords_char:45|texescape }} \newline
+{% if cls.event_type.description == "Compulsory" %}{{ cls.description }}{% elif cls.title|length_is:"0" %}N/A{% else %}{{ cls.title|truncatewords_char:45|texescape }}{% endif %} \newline
 \footnotesize{ {{ cls.parent_class.getTeacherNames|join:", "|truncatewords_char:50|texescape }} }  &
 \texttt{% templatetag openbrace %}{{ cls.emailcode }}{% templatetag closebrace %} \\
 {% endfor %}\hline
--- a/esp/templates/program/modules/programprintables/studentschedule.tex
+++ b/esp/templates/program/modules/programprintables/studentschedule.tex
@@ -79,16 +79,16 @@ Student Schedule for \\
 
 \end{center}
 
-\begin{center}	
-\begin{tabularx}{17cm}{X c}	
-\multicolumn{2}{c}{\small	
-\textit{Please see your map for building directions, or ask anyone for help.}	
-\normalsize } \\	
-\multicolumn{2}{c}{\small	
-\textit{If you are not signed up for a full day of classes, we encourage you to add more!}	
-\normalsize } \\	
-~ & ~ 	
-\end{tabularx}	
+\begin{center}
+\begin{tabularx}{17cm}{X c}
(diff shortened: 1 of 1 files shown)


esp/templates/program/modules/programprintables/studentschedule.tex
```diff
@@ -1,88 +1,99 @@
...
+\begin{center}	
+\begin{tabularx}{17cm}{X c}	
+\multicolumn{2}{c}{\small	
+\textit{Please see your map for building directions, or ask anyone for help.}	
+\normalsize } \\	
+\multicolumn{2}{c}{\small	
+\textit{If you are not signed up for a full day of classes, we encourage you to add more!}	
+\normalsize } \\	
+~ & ~ 	
+\end{tabularx}
```
hwatheod: Can you get rid of all the extra spaces (or maybe they're tabs) at the end of these lines?

hwatheod: This refers to all of lines 82-91... apparently github only shows 4 lines even though I highlighted all the lines.

esp/templates/program/modules/programprintables/studentschedule.tex
```diff
@@ -1,88 +1,99 @@
...
+\begin{tabularx}{\textwidth}{|m{1.7in}|X|m{0.55in}|}
 \hline
-\multicolumn{3}{c}{
-\rowstyle{\color{white}\cellcolor{black} } 
-\textbf{Class Schedule for {{ student.name|texescape }} } }  \\
-\multicolumn{1}{|c|}{\textbf{Class Time} } & \textbf{Room} & \multicolumn{1}{c|}{\textbf{Class} } \\
+\textbf{Time and Room} & \textbf{Class and Teacher} & \textbf{Code} \\
 {% for cls in student.classes %}\hline
-{% ifequal cls.event_type.description "Compulsory" %}
-  \multicolumn{1}{|c|}{ {{ cls.start|date:"f" }} -- {{ cls.end|date:"f A" }} } & \multicolumn{2}{|c|}{ \textbf{\textit{ {{ cls.description }} } } }
```
hwatheod: This deleted code, are we sure no one is using this "Compulsory" logic anymore? It seems to print the full description rather than teachers/room/class title. I don't know what it's for but I'm guessing it would be used for e.g. an "opening ceremony" or similar event that all students attended before going to classes.

esp/templates/program/modules/programprintables/studentschedule.tex
willgearty: Fixed!

esp/templates/program/modules/programprintables/studentschedule.tex
willgearty: Hmm...we'll still show them in the new format, so I don't think we're losing anything unless the description is really long? I guess I could add it back in and show the full description without teachers?

hwatheod: The difference is that the old code is printing `cls.description` but it seems your new code is not printing `cls.description` at all (not even a truncated version of it).

willgearty: Ah, you are totally correct. What if I put the description where the teacher info goes if the class is of the "Compulsory" event type?

hwatheod: That sounds like the right thing to do.

willgearty: Ok, after some digging, I now see what's going on. In this case, we inject timeslots into the class list if they are compulsory (because we don't have classes for them). That explains why we don't need to grab `cls.parent_class.get_meeting_times.0` here to check the `event_type`. Looking at other instances of where we check for whether a class is "Compulsory", it actually looks like we we use the `short_description` and/or `description` from the timeslot:
https://github.com/learning-unlimited/ESP-Website/blob/ed0733c3c797c1fc56bb4d556b9bf72b4acf848b/esp/templates/users/student_schedule.html#L104-L106
`short_description` is the timeblock information, which we already have in the first column, so, I've just replaced the class title with the `description` and the teacher line will be blank.

hwatheod: Ok, looks good.

Issue #1593, opened 2015-02-14
Teacher availability search includes student accounts in results


jerrywu64:
This is the thing you can access in all modules -> check teacher availability where you search for a teacher. If someone has both a teacher account and a student account, it'll show both and you'll be confused. Not sure if other types of accounts will also show up.

mgersh:
Similarly, student check-in (the non-barcode one) also shows teachers, which can be confusing for programs that do student and teacher check-in in the same room.

willgearty:
It looks like both of these modules use `AjaxForeignKeyNewformField`:
https://github.com/learning-unlimited/ESP-Website/blob/82d1e03c6322166f724b732893d94b54106584ae/esp/esp/program/modules/forms/onsite.py#L26-L27
and
https://github.com/learning-unlimited/ESP-Website/blob/82d1e03c6322166f724b732893d94b54106584ae/esp/esp/program/modules/handlers/checkavailabilitymodule.py#L71
https://github.com/learning-unlimited/ESP-Website/blob/c42712632733061d4ee23092829c60049eac2ccf/esp/esp/users/forms/generic_search_form.py#L5-L7

The student checkin form queries on registration profile users, and the teacher availability form queries on all ESPUsers. Both of these use the default ajax function to autocomplete the form, in this case `ESPUser.ajax_autocomplete()`:
https://github.com/learning-unlimited/ESP-Website/blob/c42712632733061d4ee23092829c60049eac2ccf/esp/esp/users/models/__init__.py#L183

There's probably two options to resolve this:

1. Use a different original model to query on. For example, for the teacher availability form, we could use `TeacherInfo.user.field`, and for the student checkin form, we could use `StudentInfo.user.field`. I'm pretty sure that will then only query users that have teacherinfos or studentinfos, respectively.
2. Make a new ajax function that filters the users to only users within a particular group (in this case, teachers or students). One function could be `ajax_autocomplete_teachers`, the other `ajax_autocomplete_students`, for example. Then make sure to include this as an argument to `AjaxForeignKeyNewformField`.

willgearty:
Well, #1 doesn't actually work. It uses then entire model field, even if it's specified via another model.

PR #2855, opened 2019-12-02
Generic user search forms for teachers and students

willgearty:
I added two new search forms that act like `GenericSearchForm`, but filter the possible users (for the autocomplete) to only teachers or students. I then replaced any old instances of `GenericSearchForm` that should only search for teachers or students.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/1593.


esp/esp/users/models/__init__.py
```diff
@@ -202,12 +202,23 @@ def ajax_autocomplete(cls, data):
 
         query_set = cls.objects.filter(q_names | q_username | q_id)
 
+        if group:
+            query_set = query_set.filter(groups=group)
```
hwatheod: Couldn't a user belong to more than one group? (e.g. both "Student" and "Splash Fall 2019 winners") Then will this query do the right thing?

willgearty: Yep, this query still works. My admin/teacher account showed up in the autocomplete for the teacher form.

Issue #1125, opened 2014-03-05
Two phase lottery: approved but unscheduled classes show up


luanthe:
The lottery interface should show scheduled classes only.

willgearty:
I think we'd want to change the returned list of sections here:
https://github.com/learning-unlimited/ESP-Website/blob/a0a1e50fbee39598313869d917f43ffd889cc30b/esp/esp/program/modules/handlers/jsondatamodule.py#L182
and maybe here:
https://github.com/learning-unlimited/ESP-Website/blob/a0a1e50fbee39598313869d917f43ffd889cc30b/esp/esp/program/modules/handlers/jsondatamodule.py#L367

willgearty:
Probably want to make sure those json calls aren't used anywhere else, first, though.

PR #2861, opened 2019-12-14
Remove unscheduled sections and classes with no sections to show from two-phase reg

willgearty:
This removes unscheduled sections from the two-phase registration list. Then, if a class no longer has any sections to show, the entire class is removed from the two-phase registration list (I had to move the class code to after the section code to make this possible).

Fixes https://github.com/learning-unlimited/ESP-Website/issues/1125.


esp/public/media/scripts/catalog-new.js
```diff
@@ -290,10 +290,38 @@ var CatalogViewModel = function () {
...
         for (var key in data.classes) {
             var cls = data.classes[key];
-            if (cls.status <= 5) {
+            var shown_secs = 0;
+            for (var sec of cls.sections) {
+                // count how many of the sections are still shown
+                if (sec in data.sections) {
+                    shown_secs++;
+                }
+            }
```
hwatheod: Since the subsequent code only cares whether shown_secs == 0, an optimization is possible here by breaking out as soon as you find the first section.  In fact, you only really need a class_has_sections_shown boolean.

willgearty: I think the newest commit should address this, thanks!

Issue #1616, opened 2015-03-07
Onsite main page shows modules that aren't enabled


benjaminjkraft:
It seems to be static, so it can link to things that don't exist (such as the grid).

luac:
Alternatively, maybe module views that aren't enabled should display a message to admins that says to enable the relevant module.

benjaminjkraft:
Ooh, ++.  That might also make it possible to make module view routing more static, since whether the module is enabled would change what is displayed, but not what module gets the call.

willgearty:
It looks like only the grid and the scrolling class list are static, everything else is populated based on the onsite modules that are enabled;
https://github.com/learning-unlimited/ESP-Website/blob/cdf3663db8ebdbb198686a406c5ec7b2e8d1452c/esp/templates/program/modules/onsitecore/mainpage.html#L24-L36

willgearty:
One potential fix for this is to override the default `makeLink` method with a custom one for [onsiteclasslist.py](https://github.com/learning-unlimited/ESP-Website/blob/main/esp/esp/program/modules/handlers/onsiteclasslist.py)

PR #2865, opened 2019-12-28
Fix onsite mainpage to reflect enabled modules

willgearty:
This fixes the onsite main page so that only links to modules that are enabled are included. The tricky part was the fact that the page originally included three links to views that are included in the `OnSiteClassList` module, so I needed to make a custom `makeLink` method for that module to return multiple list items.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/1616.


hwatheod:
Can you explain the `makeLink` code? It looks to me that each time you call it, you will be repeating the `classchange_grid` and `classList` items.

willgearty:
`makeLink` will only ever be used to populate module lists (like on the program mainpage for "manage" modules, on the onsite main page for "onsite" modules, and on the student/teacher registration mainpages). The default `makeLink` method only creates a link for the `main_call`, but we have some `aux_call`s that we want to link to for the `OnSiteClassList` module. Since the only time this module's `makeLink` is called is in the onsite module list, we can just follow that format to include multiple list items in the `makeLink` return.

In an ideal world, we would have a method/property that defines which calls should be listed, then loop over those and call the `makeLink` method for each of those, but I'm not entirely sure how the latter would work in a template.

PR #2870, opened 2020-01-06
Sorttable fix

milescalabresi:
Revert sorttable.js to its previous state but keep arrows inline with headers on the Dashboard


milescalabresi:
Note to self/reviewer: check that the arrows still switch and go in the correct direction when reversing sort. (Cf https://github.com/learning-unlimited/ESP-Website/pull/2818/commits/c02bda02bdd4cb6a67a64fa1ec537b5f212ed786)

willgearty:
Looks good to me on the dashboard and on other pages that use `sorttable` (I checked the teacherlist printable).

PR #2897, opened 2020-01-15
Be more clear about student phone numbers

### LINKED ISSUE #2863: Clarify student phone number field
For information like the phone number in the student profile, often the parent phone will get put in. I think it might make sense to make it more clear that the cell phone we are asking for is the student phone. Maybe we could add a yellow box like there is above the email address field and have it say "Students: Please put in your cell phone information, if you have one. If not, please leave this field blank. There will be a field for parent phone numbers in the parent section." Also I think we could rename the whole section from "Contact information" to "Student contact information" since often parents will put in their names instead of the student names as well.

This is useful for Yale splash who needs student cell phone numbers.

kkbrum:
Fixes #2863. I added an alert before asking for the phone numbers specifying they should be student numbers, and also made the student phone numbers not required (since not all students have cell phones).


esp/esp/users/forms/user_profile.py
```diff
@@ -69,7 +69,7 @@ def __init__(self, *args, **kwargs):
 
     def clean(self):
         super(UserContactForm, self).clean()
-        if self.user.isTeacher() or Tag.getBooleanTag('request_student_phonenum', default=True):
+        if self.user.isTeacher():
```
willgearty: Maybe it would be more appropriate to change the default for the tag?

esp/templates/users/profiles/usercontact.html
```diff
@@ -1,6 +1,6 @@
 <!-- User ContactInfo -->
 
-<h3>Contact Information</h3>
+<h3>{% ifequal profiletype "student" %}Student {% endifequal %}Contact Information</h3>
```
willgearty: What about this instead?
```suggestion
<h3>{{ profiletype|capfirst }} Contact Information</h3>
```

esp/esp/users/forms/user_profile.py
willgearty: Do we really think students won't have either phone type?

esp/templates/users/profiles/usercontact.html
```diff
@@ -74,7 +74,14 @@ <h4>Address:</h4>
 
 {{ form.address_postal }}
 
-<h4>Phone number{% if form.phone_day and form.phone_cell %}s{% endif %}:<font color="red">*</font></h4>
+{% ifequal profiletype "student" %}
+<div class="alert">
+    <button type="button" class="close" data-dismiss="alert">x</button>
+    <strong>Students:</strong> Please enter <b>your</b> phone numbers below. If you do not have {% if form.phone_day %}a home phone {% endif %}{% if form.phone_day and form.phone_cell %}or {% endif %} {% if form.phone_cell %}a personal cell phone{% endif %}, please leave this blank. There will be other boxes for you to enter your guardian and emergency contact phone numbers later in the form.
```
willgearty:
```suggestion
    <strong>Students:</strong> Please enter <b>your</b> phone numbers below. If you do not have {% if form.phone_day %}a home phone {% endif %}{% if form.phone_day and form.phone_cell %}nor {% endif %} {% if form.phone_cell %}a personal cell phone{% endif %}, please leave this blank. There will be other boxes for you to enter your guardian and emergency contact phone numbers below.
```

esp/esp/users/forms/user_profile.py
kkbrum: I think default true makes sense because we do want to ask for their phone number in most cases? But yes, in New Haven at least, we get many students without cell phones. Not sure about home phones but those are getting rarer and rarer anyways.

willgearty: If you change the default to "False", that means the value will be "False" if no tag is set. If someone wants to request student phone numbers, then they would need to set the tag as "True". As you have it now, the tag does nothing.

esp/templates/users/profiles/usercontact.html
```diff
@@ -1,6 +1,6 @@
 <!-- User ContactInfo -->
 
-<h3>Contact Information</h3>
+<h3>{{ profiletype|capfirst }} Contact Information</h3>
```
willgearty: Now that we've included this (which I think is a good change), will users be confused when the next section says "Teacher/Student/Guardian/Educator Information"? Should we maybe change those templates' headers?

esp/templates/users/profiles/usercontact.html
```diff
@@ -74,7 +74,14 @@ <h4>Address:</h4>
...
 
-<h4>Phone number{% if form.phone_day and form.phone_cell %}s{% endif %}:<font color="red">*</font></h4>
+<h4>Phone number{% if form.phone_day and form.phone_cell %}s{% endif %}:{% if profiletype != "student" or require_student_phonenum %}<font color="red">*</font>{% endif %}</h4>
+
+{% if profiletype == "student" and not require_student_phonenum %}
+<div class="alert">
+    <button type="button" class="close" data-dismiss="alert">x</button>
+    <strong>Students:</strong> Please enter <b>your</b> phone numbers below. If you do not have {% if form.phone_day %}a home phone {% endif %}{% if form.phone_day and form.phone_cell %}nor {% endif %} {% if form.phone_cell %}a personal cell phone{% endif %}, please leave this blank. You will enter your guardian and emergency contact phone numbers in the sections below.
+</div>
+{% endif %}
```
willgearty: I think we want the alert no matter what, but if the fields are required only show the first sentence. Otherwise, show the whole thing. Also, I think you want to get rid of the space here: `{% endif %} {% if form.phone_cell %}`

esp/esp/tagdict/__init__.py
```diff
@@ -32,7 +32,8 @@
     'current_theme_params': (False, ""),
     'theme_template_control': (False, ""),
     'current_theme_palette': (False, ""),
-    'request_student_phonenum': (True, ""),
+    'request_student_phonenum': (True, "Whether to request a student home phone number"),
+    'require_student_phonenum': (True, "Whether to require a student phone number"),
```
willgearty: Maybe mention the default behavior?

esp/templates/users/profiles/usercontact.html
```diff
@@ -74,7 +74,14 @@ <h4>Address:</h4>
 
 {{ form.address_postal }}
 
-<h4>Phone number{% if form.phone_day and form.phone_cell %}s{% endif %}:<font color="red">*</font></h4>
+<h4>Phone number{% if form.phone_day and form.phone_cell %}s{% endif %}:{% if profiletype != "student" or require_student_phonenum %}<font color="red">*</font>{% endif %}</h4>
```
willgearty: I think we only want the red asterisk if a) the user is a teacher or b) the user is a student and `require_student_phonenum` is true, to match the behavior here:
https://github.com/learning-unlimited/ESP-Website/blob/3fbdf8d0e1eb63f04f6bcd28505e8254569d90fa/esp/esp/users/forms/user_profile.py#L72-L74

esp/esp/users/forms/user_profile.py
```diff
@@ -69,7 +69,7 @@ def __init__(self, *args, **kwargs):
 
     def clean(self):
         super(UserContactForm, self).clean()
-        if self.user.isTeacher() or Tag.getBooleanTag('request_student_phonenum', default=True):
+        if self.user.isTeacher() or Tag.getBooleanTag('require_student_phonenum', default=True):
```
willgearty:
```suggestion
        if self.user.isTeacher() or (self.iser.isStudent() and Tag.getBooleanTag('require_student_phonenum', default=True)):
```

Issue #2891, opened 2020-01-12
Printables page has too many links


willgearty:
SplashCon 2020 attendees commented that the printables page now has way too many links. Some suggestions:

- Have collapsible divs/expandable options
- Have form fields at the top of the page that would filter/populate links on page (e.g. min/max grade, CSV vs. PDF)

willgearty:
For the record, it currently looks like this:
![image](https://user-images.githubusercontent.com/7232514/72685612-7da7e900-3ab1-11ea-8dc5-f7761eb89b2b.png)

PR #2901, opened 2020-01-19
Clean up printables page

willgearty:
This is my first attempt to clean up the printables page. The most notable problem before was the class/section lists section, so I've condensed that down to just the main links. Admins can now use the dropdown menus to modify the links as they please (this gives more flexibility than before). I also moved a couple items around to places that were more appropriate. Finally, I made the very bottom bit about "Happy Printables" a QSD field, so it can be modified by chapters as they see fit (since it's currently very MIT-centric).

I was thinking about including the new collapsible divs from https://github.com/learning-unlimited/ESP-Website/pull/2817, but I was worried people would be upset because then everything would be hidden and it might take more work to find what they are looking for.

Definitely open to feedback and I'd be happy to condense more if we think it's appropriate/possible.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/2891 (see this issue for the old appearance).

New appearance:
image

### DIFF
--- a/esp/templates/program/modules/programprintables/options.html
+++ b/esp/templates/program/modules/programprintables/options.html
@@ -21,15 +21,14 @@
     $j(function () {
         $j("#class_list_options :input").change(function (){
             // Get user-selected options
-            var fields = $j("#class_list_options :input")
+            var fields = $j("#class_list_options :input");
             var get = "";
-            var length = fields.length
             // Create GET string
             fields.each(function (index, element) {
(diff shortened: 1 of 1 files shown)


willgearty:
@milescalabresi @nrjean @kkbrum Could I get some feedback on this? Not looking for a full review, just looking to see if this is a good direction and seeing what other changes you think might help.

kkbrum:
It looks great to me!

Issue #2874, opened 2020-01-11
Onsite checkout module


willgearty:
The idea here is to have a module that marks that a student has physically left the program (for either an indeterminate or determinate amount of time). The module should show a student's schedule so the admin can then unenroll them ("expire") from any number of their future classes (but not necessarily all of them, because they might have a planned return time). This could have a couple consequences in other modules:

1. Students that have been checked out will not be marked as checked-in in the various onsite check-in modules/pages
2. Students that have been checked out should not show up as missing in the attendance module (https://github.com/learning-unlimited/ESP-Website/pull/2868)
3. Others?

Some things to think about:

1. If a program has a multi-day program, should a student be considered "checked out" for future days until they are "checked in"?
2. How should this affect the student onsite webapp (https://github.com/learning-unlimited/ESP-Website/pull/2507)? How about the attendance page for teachers (https://github.com/learning-unlimited/ESP-Website/pull/2868 and https://github.com/learning-unlimited/ESP-Website/pull/2866)?

Requested by Yale.

willgearty:
![2020-01-12 13 29 56](https://user-images.githubusercontent.com/7232514/72224287-c5c18b80-3546-11ea-802f-57191af04633.jpg)

PR #2903, opened 2020-01-21
Onsite checkout module

willgearty:
This module adds a new record to keep track of students that have checked _out_ of a program (as opposed to check _in_). The module allows an onsite user to select a student with a search bar, then choose which classes, if any, the student will be unenrolled from when they are checked out. The three options given do not affect the `checked_out` record that is created, rather they are short-hands to select the correct classes in the schedule below for unenrollment. There is also an option to check out all students that have checked in (with multiple confirmation checks).

The module itself is done, but I marked this as "Work in progress" because the record doesn't actually affect anything else on the site at this point.

image

Fixes https://github.com/learning-unlimited/ESP-Website/issues/2874, and thanks to the SplashCon admins that helped design this!

### DIFF
--- a/esp/esp/program/modules/handlers/onsitecheckoutmodule.py
+++ b/esp/esp/program/modules/handlers/onsitecheckoutmodule.py
@@ -1,0 +1,120 @@
+
+__author__    = "Individual contributors (see AUTHORS file)"
+__date__      = "$DATE$"
+__rev__       = "$REV$"
+__license__   = "AGPL v.3"
+__copyright__ = """
+This file is part of the ESP Web Site
+Copyright (c) 2007 by the individual contributors
+  (see AUTHORS file)
+
+The ESP Web Site is free software; you can redistribute it and/or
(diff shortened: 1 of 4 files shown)


kkbrum:
When you select a specific student who is not yet checked in, I think a little popup should come up saying that the student has not been checked in to the program and asking you whether you want to also check them in before you check them out

kkbrum:
If I type in a name but don't select it from the dropdown (either its not there or im just lazy), and hit "checkout", I get this: 

image

willgearty:
I believe the newest commit should address both of these concerns. Let me know if it works!

kkbrum:
Yep, the bug is fixed and I see the note you added. I think for the note (for students who arent checked in), it might look more complete if you add a "Caution: " to the front and an "!" at the end.

willgearty:
You got it, @kkbrum!

willgearty:
@kkbrum ?

PR #2905, opened 2020-01-27
HTML Barcodes on back of nametags

### LINKED ISSUE #2873: Barcodes on the Back of Student Nametags
It would be helpful to us (Yale program) to have barcodes on the back of student nametags. This would be paired with the new attendance system (https://github.com/learning-unlimited/ESP-Website/pull/2868) as if the new attendance app allows for scanning into the website (which, again, ideally it would for us if it's possible to put that in), it may be more convenient to scan nametags rather than student schedules and would also be helpful in case students lose their schedules. It's also possible that this is something we could do ourselves with template overrides--I am not entirely sure whether that's within its capabilities or not (if it is, we would appreciate guidance in creating that override).

kkbrum:
Adds an option to put barcodes on the back of nametags associated with user IDs. The HTML barcode generating code is taken from [here](https://www.codeproject.com/Articles/146336/Creating-a-Code-39-Barcode-using-HTML-CSS-and-Java).

One small thing is that the first page front and back, when printed with my computer's default settings, are ever so slightly off from each other. The top line of the first page is a bit under the top line of the back page. This won't matter for the barcodes because they are in the middle of the nametag, but isn't ideal. I played around but I'm not sure how to fix it-- I think it has something to do with margins, because if you print with no margins, then the top line of both pages aligns.

Fixes #2873.

### DIFF
--- a/esp/esp/program/modules/forms/admincore.py
+++ b/esp/esp/program/modules/forms/admincore.py
@@ -1,10 +1,15 @@
+from decimal import Decimal
 from django import forms
 from django.contrib import admin
 from django.utils.safestring import mark_safe
+from form_utils.forms import BetterModelForm
 
+from esp.accounting.models import LineItemType
 from esp.cal.models import Event
-from esp.program.models import RegistrationType
 from esp.program.controllers.lunch_constraints import LunchConstraintGenerator
+from esp.program.forms import ProgramCreationForm
+from esp.program.models import RegistrationType, Program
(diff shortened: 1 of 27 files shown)


willgearty:
Please resolve conflicts.

willgearty:
Oh, and I know this isn't part of your PR, but could we put the "instructions" at the top of the page in a QSD field? Like I did in https://github.com/learning-unlimited/ESP-Website/pull/2901?

PR #2911, opened 2020-02-02
Fix barcodes on nametags

kkbrum:
Oops I broke the barcodes I just put on nametags when trying to change some text before... turns out it changed more than just the text... Fixed them now :'( 

Related to #2905 and #2873.


willgearty:
Looks good to me!

Issue #2831, opened 2019-11-06
Add UI for program tags


willgearty:
Tags are a black box. Let's make them less of a black box, especially for the more commonly used/requested ones. I'm thinking it should be pretty easy to automate a form that pulls from the `all_program_tags` dictionary in [tagdict/\_\_init\_\_.py](https://github.com/learning-unlimited/ESP-Website/blob/main/esp/esp/tagdict/__init__.py). This gives us information about the format of the tag (boolean or not) and the help text associated with each tag.

Of course, that then means that we would need to add a little more documentation to the dictionary there so we can just pull the help text directly from it (and of course make sure the list is complete/exhaustive). It might also be good to make sure the default value is listed in the help text.

Also, program tags can still be global, so it might make sense to have a "Is Global?" checkbox for each one? Otherwise, we'd just associate it with the specific program.

I'm thinking this could go right below the other settings forms added in https://github.com/learning-unlimited/ESP-Website/pull/2816.

PR #2916, opened 2020-02-08
User Interface for Tags

willgearty:
This adds two separate user interfaces for admins to alter tag values (one at /manage/tags for global tags and one at /manage/one/two/tags for program tags). It shows a form that is populated with fields for each tag listed in `all_program_tags` or `all_global_tags` in [tagdict/\_\_init\_\_.py](https://github.com/learning-unlimited/ESP-Website/blob/main/esp/esp/tagdict/__init__.py). I also expanded the information in the dictionaries in the init file to now list the default value, whether the tag should be shown in this new form, and which category the tag falls into. I went through the main branch and added any missing tags to the init file, then added (or edited) help text for every tag. Finally, I went through the existing tags and converted some to boolean and moved others between the global and program lists.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/2831

Here's a preview of the global tags form (the program tags form looks similar):
image


willgearty:
This is done! The one weird thing that I'm trying to figure out is that if you go to /manage/tags, it redirects you to /manage/tags/index.html. This doesn't actually cause any issues, but it's a little frustrating.

willgearty:
This is done! The one weird thing that I'm trying to figure out is that if you go to /manage/tags, it redirects you to /manage/tags/index.html. This doesn't actually cause any issues, but it's a little frustrating. I think it has to do with some urlpattern, but the two that I see that add index.html don't seem to be to blame (I commented them both out to no effect). Maybe @hwatheod has an idea?

hwatheod:
This is awesome work to collect and document all the tags!

Unfortunately, I have no idea about the index.html issue. I agree it's somewhat annoying.

esp/esp/tagdict/tests.py
```diff
@@ -5,10 +5,10 @@
...
 
 # Make test-only tags not raise warnings
-all_global_tags['test'] = (False, "")
-all_program_tags['test'] = (False, "")
-all_global_tags['test_bool'] = (True, "")
-all_program_tags['test_bool'] = (True, "")
+all_global_tags['test'] = (False, "", None, "manage", True)
+all_program_tags['test'] = (False, "", None, "manage", True)
+all_global_tags['test_bool'] = (True, "", False, "manage", True)
+all_program_tags['test_bool'] = (True, "", False, "manage", True)
```
kkbrum: Why are the test-only tags "True" for showing on tag settings page?

willgearty: My thinking was that this would make them go through more of the new code for testing purposes (but of course they'll never actually show up on the real page).

willgearty:
@nrjean noted that the examples have single quotes, but JSON requires double quotes around strings, so I'll probably need to change the outer quotes to single quotes and make the example quotes double quotes.

esp/templates/program/modules/admincore/settings.html
```diff
@@ -15,20 +15,20 @@
 <h1>Program Settings for {{ program.niceName }} (ID: {{ program.id }})</h1>
 
 {% inline_program_qsd_block program "manage:main_top" %}
-Welcome to the program settings page for {{ program.niceName }}. This page lists most of the settings that you may want to change to make the program run how you would like.
+Welcome to the program settings page for {{ program.niceName }}. This page lists most of the settings that you may want to change to make the program run how you would like. If you can't find the setting here, try checking out the <a href="/manage/{{ program.getUrlBase }}/tags/">tag settings page</a> or reach out to <a href="mailto:websupport@learningu.org">websupport</a>.
```
kkbrum: Link to global tags as well?

esp/templates/program/modules/admincore/tags.html
```diff
@@ -0,0 +1,87 @@
...
+
+{% block content %}
+
+{% load render_qsd %}
+
+<h1>Tag Settings{% if program %} for {{ program.niceName }} (ID: {{ program.id }}){% endif %}</h1>
+
+{% if program %}
+{% inline_program_qsd_block program "manage:main_top" %}
+Welcome to the tag settings page for {{ program.niceName }}. This page lists other miscellaneous settings that you may want to change to make the program run how you would like. The main program settings can be found on <a href="/manage/{{ program.getUrlBase }}/settings/">the page here</a>.
```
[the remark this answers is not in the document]
kkbrum: Link to global tags page as well?

esp/templates/program/manage_programs.html
```diff
@@ -34,6 +34,8 @@ <h2>Hello Admins!</h2>
           <li><a href="/manage/statistics/">View statistics about students</a></li>
           
           <li><a href="/manage/emails/">Monitor comm panel email status</a></li>
+          
+          <li><a href="/manage/tags/">Modify tag settings (generic miscellaneous settings, for experts only)</a></li>
```
kkbrum: I feel like this (and some of the other links as well) should exist outside of a QSD that could get erroneously editted out/not updated/etc. Could we make a more official and permanent link somewhere?

kkbrum:
This relates to previous PR, but why arent the first two left justified like the rest?
image

kkbrum:
In these two forms + settings form, my eyes see the help text as relating to the field below the proper one-- could you add spacing so that the help text more clearly relates to the current item?
image

kkbrum:
Could this have better explanation of what this means?
image

esp/templates/program/modules/admincore/settings.html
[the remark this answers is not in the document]
willgearty: Whoops, good catch!

esp/templates/program/modules/admincore/tags.html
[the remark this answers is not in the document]
willgearty: Oh sure

esp/templates/program/modules/admincore/settings.html
willgearty: Sure!

esp/templates/program/manage_programs.html
willgearty: I've been feeling the same way for quite a while, now (in fact, since we had to go through and manually change everyone's QSDs on this page during the last stable release). Really, we probably want everything in this QSD to be part of the template, then have a QSD below it for chapters to customize.

willgearty:
I believe that should address all of your comments, @kkbrum.

willgearty:
Also, I figured out the weird index.html bug, and it should be resolved as of the most recent commit!

PR #2917, opened 2020-02-09
Mit walkin imports fix

alexjli:
Bug fix for walkins imports for Spark 2020:
Allows importing walkins from previous programs without setting teacher reg hidden fields
Instead sets walkin hidden field values to default when copying the class


willgearty:
By the way, I'm pretty sure this was fixed in https://github.com/learning-unlimited/ESP-Website/pull/2621

PR #2918, opened 2020-02-11
Lower bigboard Threshhold

### LINKED ISSUE #2899: Lower bigboard graph threshold
Right now, the threshold to begin graphing is 25, which is too high for many smaller programs. We don't want to include test accounts, so we're thinking the 3-5 range. Relevant code is here https://github.com/learning-unlimited/ESP-Website/blob/77e43165a442fe05efc865a1ef80eeb00c25c6de/esp/esp/program/modules/handlers/bigboardmodule.py#L68

nrjean:
Fixes #2899 

Changed threshold of graphing on big board to 5. Also changed the amount of registrations that are dropped to 4 from the beginning and 0 from the end.

### DIFF
--- a/esp/esp/program/modules/handlers/bigboardmodule.py
+++ b/esp/esp/program/modules/handlers/bigboardmodule.py
@@ -65,7 +65,7 @@ class BigBoardModule(ProgramModuleObj):
             ("signed up for classes", [(1, time) for time in self.times_classes(prog)]),
         ]
 
-        timess_data, start = self.make_graph_data(timess, 10, 10, 5)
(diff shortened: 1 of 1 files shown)


hwatheod:
What exactly does this threshold filter out?  Would 5 be too small for a larger program?  Would it be better to make it a configurable parameter (tag) ?

nrjean:
It was supposed to filter out test accounts I believe. Thought is this shouldn't have an effect on larger programs because larger programs shouldn't have to make more test accounts. In the case they do - the program should be large enough that the test accounts would get lost in the noise.

PR #2868, opened 2020-01-02
Student attendance (teacher and admin interfaces)

willgearty:
This implements a set of tools for teachers and admins to manage student class attendance.

This adds a new page (`section_attendance`) that allows teachers to take attendance for a particular section (including unenrolled students) (see details in picture below). The page is linked to from the main teacher registration page (for sections that are approved and scheduled). The page has a dropdown menu for the teacher to select a section for attendance.

(I also used the new template tag from https://github.com/learning-unlimited/ESP-Website/pull/2860)

I also added an onsite module that summarizes attendance statistics/details (see details in picture below). The page has a similar dropdown to select a particular timeslot for attendance. The page is linked to from the main onsite page (provided the module is enabled).

Fixes https://github.com/learning-unlimited/ESP-Website/issues/2864.

Teacher interface:
image

Admin interface:
image

PR #2925, opened 2020-02-25
Support recurring classes for attendance

willgearty:
This extends the work I did in #2868 to support classes that occur on multiple days but rely on the same timeslot (e.g. Sprout classes). The first step was to make it so that attendance now expires at midnight on the day it is set. This then makes it so we can set or check attendance for specific days. I added a field that allows admins to pick this date for the onsite attendance module (the default is the day of the timeslot). For teachers, they won't notice any difference, except that the students they marked as attending for previous instances of the class will no longer be marked as attending. To make sure all of this is clear, I've made the code a lot clearer with more verbose comments, and I've made the headings for the various tables on the page much clearer and more precise. If a day other than the default is selected, this will be noted in lots of places on the page as a reminder.

image

### DIFF
--- a/esp/esp/program/modules/handlers/onsiteattendance.py
+++ b/esp/esp/program/modules/handlers/onsiteattendance.py
@@ -65,9 +65,10 @@ class OnSiteAttendance(ProgramModuleObj):
             if len(timeslots) == 1:
                 timeslot = timeslots[0]
                 context['timeslot'] = timeslot
-                if request.GET and "when" in request.GET:
+                when_get = request.GET.get("when", "")
+                if when_get != "":
                     # if a date is specified, use that date and the end time of the timeblock
-                    date = datetime.datetime.strptime(request.GET["when"], "%Y-%m-%d").date()
+                    date = datetime.datetime.strptime(when_get, "%Y-%m-%d").date()
                     when = datetime.datetime.combine(date, timeslot.end.time())
                 else:
                     # otherwise, just use the end time (and date) of the timeblock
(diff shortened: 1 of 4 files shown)


kkbrum:
If I click the little blue x in the preferred date box, and wait a couple seconds, this error pops up:

image

kkbrum:
Could you check whether you added code to expire students at end of day who are marked Onsite/AttendedClass when a teacher marks them attended and they weren't previously enrolled? I don't think you did but I could also be confused. Also, if those get expired, would the student still remain actively enrolled in the class (for future week schedule's/ posterity sake)? If not, maybe we need to add double registrations for them so that an attended registration gets expired but the onsite registration remains active?

willgearty:
Hmm...I'm not seeing this little "x" in my browser, but that should be a pretty easy fix (just check that there's a date specified for the when variable).

For your second comment, that list sounds correct to me. My thinking was that a student was enrolled via attendance, not registration, so we'd want to always know that, regardless of what day it is (even for a future instance of the class, the student was still originally enrolled via attendance, not registration), but if we only want to know that for the original day (again, all of this was designed for non-Sprout classes), I can change that (by either expiring the onsite/attendedclass relationship at the end of the day or filtering onsite_srs to only the specified day). I can see it both ways, so we just need to make a decision one way or the other.

kkbrum:
I think it makes sense to expire the onsite/attendedclass registration, since beyond the day of that registration, I would consider a student "enrolled" in a class.

willgearty:
I think that should address both concerns, @kkbrum!

Issue #2890, opened 2020-01-12
Implement survey creation in survey module


willgearty:
Right now it just says "This is not implemented yet...". We should probably actually implement it. Maybe something like the resources page(s) where you can create question types, then make questions with those question types. Then make surveys that include those questions. (Or something like that)

willgearty:
It would also be nice to a have a "Survey Preview". Either as a separate page or as an alternate view on the survey creation page.

PR #2929, opened 2020-03-08
Add UI for building surveys

### LINKED ISSUE #1146: Some survey question types render in bold.
See /learn///survey for a program with a survey; multiple choice questions have bold question text, while numeric ratings don't.  This appears to just be a difference between the different templates for no particular reason.

willgearty:
This adds a new UI that makes it much much easier for admins to create teacher/student surveys for programs. The main page provides forms for users to create new surveys and questions for those surveys. There is also a form to import a survey from another program. Both of these work very similarly to the resources page. When a user goes to create a question, the form will dynamically change which fields the user needs to populate based on the desired question type.

Speaking of which, I've also created a migration to make sure we add the currently viable question types to the chapters' databases. Theoretically, others could be created locally, but this at least adds the ones that we have templates for (speaking of which, I cleaned up some of the really old survey question templates).

Below the forms, the existing surveys and questions are listed in sequence order. For the questions, there's the option to view the raw database fields or the rendered questions (same on the import confirmation page).

Here's a preview:
image

Fixes https://github.com/learning-unlimited/ESP-Website/issues/2890.
Also fixes #1146.


kkbrum:
The Survey drop down for Step 2 isnt intuitively reactive, as in when I add a survey in Step 1, it takes a lot to get it to appear in the drop down (ie, I would expect it to appear once I hit "create/save survey" in step 1, but i need to refresh the page or hit some other sequence of buttons it seems)

kkbrum:
I'm not sure "name" is the correct name for the question field... Wouldn't just "Question" make more sense? I would expect "name" to be like a short version or something, not the full thing

esp/templates/program/modules/surveymanagement/manage.html
```diff
@@ -0,0 +1,176 @@
...
+    <table align="center" cellpadding="0" cellspacing="0" width="100%">
+        <tr><th colspan="6">Questions for {{ program.niceName }}</th></tr>
+        <tr>
+            <th class="selector active" colspan="3" width="50%" data-panel="details">Question Details</th>
+            <th class="selector" colspan="3" width="50%" data-panel="rendered">Rendered Questions</th>
+        </tr>
+        <tr id="details-panel"><td colspan="6"><table width="100%" cellpadding="0" cellspacing="0">
+        <tr>
+            <td width="5%"><b>Seq</b></td>
+            <td width="30%"><b>Name</b></td>
```
kkbrum: I think just "Question" would be a better name for this field. Name makes it sound like just a shorted keyword, not the full blown question.

esp/templates/program/modules/surveymanagement/manage.html
```diff
@@ -0,0 +1,176 @@
...
+    </form>
+    <br />
+    <form method="post" action="/manage/{{ program.url }}/surveys/manage?obj=survey">
+        {% if survey_form.id %}
+        <input type="hidden" name="survey" value="{{ survey_form.id }}" />
+        {{ survey_form.id }}{% endif %}
+        <table align="center" cellpadding="0" cellspacing="0" width="100%">
+            <tr><th colspan="2">Add/modify a Survey</th></tr>
+            {{ survey_form.as_table }}
+            <tr><td colspan="2" align="center"><input class="fancybutton" type="submit" value="Create/Save Survey" /></td></tr>
```
kkbrum: When this button gets pressed, can you auto-update the survey dropdown for step 2? I seem to need to refresh/hit some other things to get it to update rather than just creating a new survey.

esp/templates/program/modules/surveymanagement/manage.html
```diff
@@ -0,0 +1,176 @@
...
+        <tr><th colspan="6">Questions for {{ program.niceName }}</th></tr>
+        <tr>
+            <th class="selector active" colspan="3" width="50%" data-panel="details">Question Details</th>
+            <th class="selector" colspan="3" width="50%" data-panel="rendered">Rendered Questions</th>
+        </tr>
+        <tr id="details-panel"><td colspan="6"><table width="100%" cellpadding="0" cellspacing="0">
+        <tr>
+            <td width="5%"><b>Seq</b></td>
+            <td width="30%"><b>Name</b></td>
+            <td width="25%"><b>Question Type</b></td>
```
kkbrum: For the question types that don't have anything after "includes," can we remove the word "includes"?

image

esp/esp/survey/migrations/0002_auto_20200301_1048.py
```diff
@@ -0,0 +1,27 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+def create_question_types(apps, schema_editor):
+    db_alias = schema_editor.connection.alias
+    QuestionType = apps.get_model("survey", "QuestionType")
+    QuestionType.objects.using(db_alias).get_or_create(name="Instruction Text", _param_names = "Text")
+    QuestionType.objects.using(db_alias).get_or_create(name="Long Answer", _param_names = "Rows|Columns")
```
kkbrum: I sorta understand that rows will make the input box longer, but can there just be a default for the columns business? It feels really unnatural to specify the columns for a long answer question.

kkbrum: Actually I'm not even sure changing the columns does anything at least in the rendered questions display in this builder.

esp/templates/program/modules/surveymanagement/manage.html
```diff
@@ -0,0 +1,180 @@
...
+    <script type="text/javascript">
+        var question_types = {{ question_types|safe }};
+    </script>
+{% endblock %}
+
+{% block content %}
+<h1>Survey Management for {{ program.niceName }}</h1>
+
+<p>
+You can use this page to set up surveys for teachers or students of the program. You will first need to create a survey, then you can add questions to it.
```
kkbrum: Every time I add a question by hitting "create/save," it's overriding the current question on my survey even though I'm trying to add a new question?

esp/esp/survey/migrations/0002_auto_20200301_1048.py
```diff
@@ -0,0 +1,27 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+def create_question_types(apps, schema_editor):
+    db_alias = schema_editor.connection.alias
+    QuestionType = apps.get_model("survey", "QuestionType")
+    QuestionType.objects.using(db_alias).get_or_create(name="Instruction Text", _param_names = "Text")
+    QuestionType.objects.using(db_alias).get_or_create(name="Long Answer", _param_names = "Rows")
```
kkbrum: Even though I applied the migrations and refreshed the cache, I still can't get the columns parameter to go away :(

esp/templates/program/modules/surveymanagement/manage.html
willgearty: OK, I think I know what's going on. I'll take a look when I get a chance!

esp/esp/survey/migrations/0002_auto_20200301_1048.py
willgearty: Yeah, the migration isn't really designed to _fix_ the QuestionTypes, just to set them up in the first place.

esp/templates/program/modules/surveymanagement/manage.html
willgearty: Should be all set with the latest commit!

Issue #2909, opened 2020-02-01
Use "checked_out" record in various places


willgearty:
Now that we have a "checked_out" record, we want to reflect that status in places on the site that check whether students are checked in for the program. Instead of just checking for the "attending" record, we'll want to see whether those records are more recent than any "checked_out" records for the same users. Places that come to mind include:
- attendance module (teacher and admin interfaces?)
- student webapp
- student big board (see https://github.com/learning-unlimited/ESP-Website/issues/2894)
- teacher webapp?
- checkin module(s) maybe

willgearty:
Really, we should probably just have a generic function somewhere (maybe in the program models) that determines whether a student is "checked in" at the current moment. This would mean that:

1. the student has been checked in at least once for the program at some point, and
2. the student has not been checked out since the most recent check-in

willgearty:
The default of the function should probably be to just return `True` or `False` (`True` meaning that the student is checked in and hasn't been checked out), but it may also be useful to have a slightly more verbose option where the function returns `not_checked_in`, `checked_in`, or `checked_out`.

PR #2935, opened 2020-03-22
Use checked out records

willgearty:
In https://github.com/learning-unlimited/ESP-Website/pull/2903 I implemented a new module that can be used to check out students from the program (e.g. they are leaving for lunch). This can be useful for keeping track of students whereabouts or for class changes when trying to determine if a class has enough room to enroll another student. In this PR, I've used the new "checked_out" record that was implemented in that PR to affect a number of other modules that depend on attendance records:

1. The onsite attendance module now filters out checked out students for some of its boxes.
2. For the onsite checkin page, it now lists the first time the student was marked as attending a program and also the most recent time that student was marked as such. Also, unchecking the "attending" checkbox and submitting the form checks out the student.
3. For the barcode checkin page, only students that are currently checked in will result in an error message (previously it was any student that was marked as attending).
4. In the onsite checkout module, if a student is currently checked out, a warning is show to the user (previously it was just if the student had never been marked as attending).
5. Attendance numbers in the grid-based class changes and grid-based status pages are now based on students that are not checked_out.
6. There is a new "currently checked out students" option for the comm panel/arbitrary user list/etc.
7. The teacher attendance page (including within the teacher webapp) now only marks a student as checked in if they are not checked out.

For future development, I've also created two useful functions within the program models (`isCheckedIn()` and `checkedOutStudents()`).

Fixes https://github.com/learning-unlimited/ESP-Website/issues/2909.


nrjean:
Is there anyway we could use the checked_out record to get a per day attendance statistic for programs that run more than one day?

willgearty:
I she @nrjean. This PR is for all the little things that needed to be changed, but I think the daily stats will require a bit more work, including redesigning the student big board. That will all be implemented when I address #2984 separately.

esp/templates/program/modules/onsitecheckinmodule/checkin.html
```diff
@@ -49,13 +49,16 @@ <h1>Advanced Checkin &mdash; For &quot;{{module.student.name}}&quot;</h1>
...
+    {% if module.timeCheckedIn != module.lastCheckedIn %}
+    <br />{{module.student.name}} last checked in at <b>{{module.lastCheckedIn}}</b>.
+    {% endif %}
   </p>
   <form name="checkin" method="POST" action="{{request.path}}">
     <table>
       <tr>
 	<td><label for="attended"><strong>Attending? (i.e. here right now?)</strong></label></td>
-	<td><input type="checkbox" name="attended" id="attended" {% if module.hasAttended %}checked="checked"{% endif %} /></td>
+	<td><input type="checkbox" name="attended" id="attended" {% if module.isAttending %}checked="checked"{% endif %} /></td>
```
kkbrum, line 61: If you uncheck this box when you don't have the check out module enabled, it seems like it still adds a check out record, instead of deleting the attended record which is what I would imagine should happen when the check out module not enabled?

willgearty: Hmm...I hadn't thought about that. We only want to make checkout records if the checkout module is enabled?

kkbrum: Okay nvm- we'll call the checkout records part of the base website and the actual checkout module is only the fancy interface page. So then this is fine that we're making check outs. However-- would we maybe want to allow someone to undo an attended record on this page that they made accidentally? By making a "undo latest attended record for this student" box and a "checkout student" box?

esp/templates/program/modules/onsitecheckinmodule/checkin.html
```diff
@@ -49,13 +49,16 @@ <h1>Advanced Checkin &mdash; For &quot;{{module.student.name}}&quot;</h1>
 {% else %}
 <div style="margin:10px">
   <p>
-    {{module.student.name}} already checked in at <b>{{module.timeCheckedIn}}</b>.
+    {{module.student.name}} first checked in at <b>{{module.timeCheckedIn}}</b>.
```
kkbrum, line 52: Instead of having "first checked in at:" and "last checked in at:" could we include info about the checkouts too? I think a nice way to have the full information would be 
"Student checked in during the following times:
(start1) to (end1)
(start 2) to present"
or something? So that we know what happened between that first and last check in? Like whether they were present at x time.

esp/esp/program/modules/handlers/onsiteclasslist.py
```diff
@@ -188,7 +188,8 @@ def register_student(self, request, tl, one, two, module, extra, prog):
     @needs_onsite
     def checkin_status(self, request, tl, one, two, module, extra, prog):
         resp = HttpResponse(content_type='application/json')
-        data = ESPUser.objects.filter(record__event="attended", record__program=prog).distinct().values_list('id')
+        students = prog.students(True)
+        data = ESPUser.objects.filter(students['attended'] & ~students['checked_out']).distinct().values_list('id')
```
kkbrum, line 192: It doesn't seem to be filtering out checked out students from the counts of students checked in on grid based class changes for me...

esp/esp/program/modules/handlers/studentregcore.py
```diff
@@ -91,6 +93,7 @@ def students(self, QObject = False):
 
         retVal = {'confirmed': ESPUser.objects.filter(q_confirmed).distinct(),
                   'attended' : ESPUser.objects.filter(q_attended).distinct(),
+                  'checked_out': ESPUser.objects.filter(q_checked_out).distinct(),
```
kkbrum, line 96: It's giving me 0 students in the AUL even though I have students checked out...

esp/templates/program/modules/onsitecheckinmodule/checkin.html
willgearty: So we want this for every single set of checkins and checkouts?

esp/esp/program/modules/handlers/onsiteclasslist.py
willgearty: This seems related to your next comment.

esp/esp/program/modules/handlers/studentregcore.py
willgearty: Have the students been checked back in since they were checked out? Are they checked out for the program you are using? This worked for me, so I'm not really sure how to reproduce the problem you're having.

esp/templates/program/modules/onsitecheckinmodule/checkin.html
kkbrum: I think so. I wouldnt expect there to be that many per student (maybe a couple per day of program) and its useful information

esp/esp/program/modules/handlers/studentregcore.py
kkbrum: They are currently checked out for the program on hand (haven't been checked back in)

willgearty: Are the checked out students excluded properly in the attendance module?

kkbrum: Yeah they were. But also, I just rechecked the AUL/ grid based class changes in an effort to help you debug and now the checked out students appear properly (removed from grid and added to AUL). When I tested this the other day, I had refreshed the cache & hard refreshed. Are there any other time-based things in place that might have made it not work right away (~5 min after checking out the students)?

willgearty: Umm...it's possible there are supercached things that aren't cleared when you flush the cache and are only updated once you run fab manage:update. Anywho, glad to hear this is resolved, thanks!

willgearty: Actually, can you try checking out a student and seeing if the AUL is updated immediately? (since you have your dev server running now)

kkbrum: I just tried checking out a student, and grid based & AUL don't change (even after cache clear & hard refresh) whereas attendance does. Personally I think this is a problem and not resolved? Because grid-based is supposed to be accurate to the minute in terms of class changing students and if you checkout students and they stay there for some unknown period of time, it seems a little useless to me

willgearty: OK, I can update grid-based class changes to use the same code as the attendance module. I'm not sure I can make the AUL update faster, unfortunately.

kkbrum:
We forgot about the web app version of class changes (need to account for checked out students [here ](https://github.com/learning-unlimited/ESP-Website/blob/f1f1f8e8c66e222bb3e126cafca79aa3cbee798c/esp/esp/program/models/class_.py#L953) but I'll do it in my PR for issue #2968.

Issue #2953, opened 2020-04-15
teacher_profile_hide_fields tag problems


willgearty:
1. When the `teacher_profile_hide_fields` tag is set, it properly hides the fields in the form, but not their labels (presumably because we are rendering them separately in the template and aren't checking if the fields exist first).
2. When the tag is processed, it currently doesn't `strip()` the individual comma-separated values, so if you have spaces after your commas, the following field names won't be processed (would probably best to just add a `strip()` call).
3. When fields are hidden due to this tag being set, the profile form can not be submitted.

PR #2958, opened 2020-04-18
Fix teacher_profile_hide_fields tag problems

willgearty:
First, we now strip any whitespace in the tag when it is processed (just a little thing that I think will reduce stress related to using the tag). Then I made sure the field labels are hidden if their respective fields are hidden. Then I fixed the processing of the form for when the `affiliation` field was being hidden due to this tag.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/2953


esp/esp/users/forms/user_profile.py
```diff
@@ -398,7 +399,7 @@ class TeacherProfileForm(UserContactForm, TeacherInfoForm):
     """ Form for teacher profiles """
     def __init__(self, *args, **kwargs):
         super(TeacherProfileForm, self).__init__(*args, **kwargs)
-        for field_name in Tag.getTag('teacher_profile_hide_fields', default='').split(','):
+        for field_name in [x.strip() for x in Tag.getTag('teacher_profile_hide_fields', default='').split(',')]:
```
hwatheod, line 402: I would also recommend converting the field name to lowercase, to further normalize (double check that all the existing fields are lowercase).

willgearty: Yep, all of the existing fields for the TeacherInfoForm are lowercase, so I've gone ahead and added `.lower()`. Thanks, @hwatheod!

Issue #2947, opened 2020-04-13
Ways to export phase zero data and manually pick winners


willgearty:
Right now the student lottery ("phase zero") is completely random, picking lottery tickets randomly until grade capacities are met. However, as mentioned in https://github.com/learning-unlimited/ESP-Website/issues/2642, it's possible programs will want a more complicated method for picking winners. Since there's any number of ways programs may want to do this, the easiest and most useful functionality would be to have a way to do the following:

1) Download lottery ticket data (probably a csv/tsv file with each row corresponding to the username of the students on a ticket). An alternative to this would be to have an option in the AUL to have the lottery ticket ID in the data dump.
2) Specify a list of usernames that should be selected to win the lottery (instead of running the built-in algorithm). This can probably just be an expandable `textarea` on the Phase Zero Management page.

With these two additions, programs will be able to download the lottery data, cross reference it with other data from the site, develop their own algorithm or method for selecting winners, then specify those winners when the lottery is "run".

PR #2963, opened 2020-04-21
Manual phase zero lottery

willgearty:
This implements two changes, both of which are useful for programs using the student lottery (phase zero) module:
1. Adds an AUL field that pulls the student lottery ticket ID for a student.
2. Adds an alternate option to the phase zero management page to run a _manual_ lottery in which the winners are specified by the admin.

The idea here is that you can use the AUL, perhaps in combination with a download of custom form data, to then do some more complicated weighting/algorithm for the lottery offline. Once you've selected the winners, you can then specify them in the form, and then the module will add them to the proper user group. There is also an option to also accept any other students in the winners' groups.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/2947.

image


esp/esp/program/modules/handlers/studentregphasezeromanage.py
```diff
@@ -58,32 +58,56 @@ def module_properties(cls):
...
+                if student.isStudent():
+                    if post.get('groups'):
+                        records = PhaseZeroRecord.objects.filter(program=prog, user=student).order_by('time')
+                        if records.count() > 0:
+                            group_members = records[0].user.all()
+                            winners.user_set.add(*group_members)
+                    else:
+                        winners.user_set.add(student)
+                else:
+                    messages.append(username + "is not a student")
```
hwatheod, line 110: Do you think it's a better idea to validate all users first and reject without making any changes at all if any of them are invalid (user doesn't exist, not a student, etc.) ?  An invalid user points to an error in the manual lottery algorithm, and since it's difficult to undo the lottery process, it might be better to just reject completely.

willgearty: That's a good idea. I'll change the code around to check the students first, then add to the winners group only if all checks pass.

willgearty:
OK, I've tested all of the possible error cases, and they all properly prevent the lottery from being finalized.

PR #2966, opened 2020-04-25
More tags to hide profile fields

willgearty:
Until now, we only had the `teacher_profile_hide_fields` tag for hiding fields in the teacher profile form. This adds a tag for each of the remaining user types, each of which allows you to hide any profile form fields for those user types. I then resolved all (I'm pretty sure) of the problems that arose when fields were removed from the profile form.


willgearty:
@kkbrum What was the tag value that you were using?

kkbrum:
image

esp/templates/users/profiles/usercontact.html
```diff
@@ -69,11 +78,11 @@ <h4>Address:</h4>
         {% endif %}
     </div>
 </div>
-    
 {% endif %}
 
 {{ form.address_postal }}
 
+{% if form.phone_day or form.phone_cell %}
```
kkbrum, line 85: If I remove phone_cell for student_reg, i can't even load the profile page anymore:
image

esp/templates/users/profiles/studentinfo.html
```diff
@@ -36,7 +37,9 @@ <h3>Additional Student Information</h3>
     {% endif %}
     </div>
 </div>
+{% endif %}
 
+{% if form.graduation_year %}
```
kkbrum, line 42: a) I don't really think we should allow removing graduation_year from student_reg since then new accounts wont have grades and wont be able to access programs... 
b) If i do remove it from student_reg, i get an error when I submit the profile form:
image

willgearty: Oh, that's a good idea. Any other fields that fields that you think should be "undeletable"?

willgearty: I've added email to the list. Those are the only two that I think we NEED to have?

willgearty:
Ok, I think I've fixed all of the problems you brought up, @kkbrum. Can you give it another look?

willgearty:
So I just realized that this makes it so you can't hide the graduation_year field for teachers. Not sure if that's desired.

kkbrum:
Potentially someone might want to delete grad year if they are letting teachers be anyone in the community? Could we make separate sets of _undeletable_fields for each profile type? And include email in all of them and grad year in just the student one?

willgearty:
OK, how about that @kkbrum ?

Issue #2835, opened 2019-11-10
Onsite programs list not loading


milescalabresi:
Northeastern reported that https://neptun.learningu.org/myesp/onsite/ failed to load the upcoming program. It said there were multiple upcoming programs but then didn't link to any, as it should have. Controlling code is here https://github.com/learning-unlimited/ESP-Website/blob/main/esp/esp/web/views/myesp.py#L261. Unclear if the "TODO" comment if related.

milescalabresi:
The source of the problem is twofold. The "multiple programs" text on [this page](https://github.com/learning-unlimited/ESP-Website/blob/main/esp/templates/program/pickonsite.html) is just bogus and should be changed. The underlying problem, though, is that the "onsite" permission is assigned only to the `onsite` user. Proposal is to add this permission to the admin role when the site is set up and backport this fix to the existing sites.

Issue #2374, opened 2017-04-23
Permission.program_by_perm doesn't look at global permissions


benjaminjkraft:
This means that e.g. /myesp/onsite (linked in bigpicture's sidebar) will only show you programs where you have explicit onsite permissions, even if you have global onsite/admin permissions.

PR #2836, opened 2019-11-13
Make admins onsite users by default for all programs

milescalabresi:
Fixes #2835

PR #2967, opened 2020-04-26
Fix myesp/onsite for admins

willgearty:
This fixes the `myesp/onsite` page for admins by checking for global permissions within `Permission.program_by_perm()` (before, `program_by_perm` only checked for program-specific permissions, but the ultimate permissions that admins have are not tied to a program). I included some other changes from https://github.com/learning-unlimited/ESP-Website/pull/2836 that looked useful, although I left the program list as-is without any filtering.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/2835 and fixes https://github.com/learning-unlimited/ESP-Website/issues/2374.


esp/esp/users/models/__init__.py
```diff
@@ -2549,6 +2549,10 @@ def program_by_perm(cls,user,perm):
         implies = [perm]
         implies+=[x for x,y in cls.implications.items() if perm in y]
 
+        #Check for global permissions (should only work for non-deadlines and admins)
+        if any([cls.user_has_perm(user, x) for x in implies]):
+            return Program.objects.all()
+
```
hwatheod, line 2555: Can you explain why this is the right thing to do?

willgearty: Below this we check for which specific programs the user has the specified permission, but there are a couple permissions that can also be global (e.g. Administrator has every permission globally). Those aren't caught in the code below this, so we need to check if the user has any global permissions that imply the specified permission (`user_has_perm` checks for global permissions when there is no program argument specified).

hwatheod: `user_has_perm` says the following.  Do you need to set `program_is_none_implies_all` to True in this call?

`:param program_is_none_implies_all:
            If True, treat Permission objects with program=None as if they are
            global across all programs. Return True if the user has a
            Permission object with program=program or with program=None.
            If False, do not treat Permission objects with program=None as if
            they are global across all programs. Only return True if the user
            has a Permission object with program=program.`

willgearty: I remember getting hung up on that, but I think because `program=None`, then it _only_ checks for global permissions anyways:
```python
:param program:
            Check for permission for `name` on this program.
            If program is None, check only for Permission objects with
            program=None.
            If program_is_none_implies_all is False, check only for Permission
            objects with program=program.
            If program_is_none_implies_all is True, check for Permission
            objects with program=program or program=None.
```

You can see that in `q_permissions_on_program` (which gets called by `user_has_perm`) that the Q filter is already Q(program=program), so the added Q filter when `program_is_none_implies_all=True` wouldn't do anything:
https://github.com/learning-unlimited/ESP-Website/blob/4bc9787dc344d7ea15974ad9386ea98e89fca5d9/esp/esp/users/models/__init__.py#L2405-L2407

hwatheod: Ok, got it.

Issue #2715, opened 2019-04-08
Big board for phase zero


willgearty:
Either include phase zero on the big board or add a big board to the phase zero management module. Phase zero records do have timestamps, so we could use those, although I guess we would have to assume that all of the students in a lottery ticket joined at that time (probably not a good assumption, but probably fine for these purposes). Alternatively, we could create a new record that tracks when each student enters the student lottery.

willgearty:
The problem here is that the phase zero records don't have timestamps. Maybe we just need to make a new record "Entered the Student Lottery" or something and use those to calculate a line for the student big board.

PR #2973, opened 2020-05-02
Phase zero graph

willgearty:
This replaces the number and list of usernames on the phase zero management page with a bigboard-esque graph of lottery registration over time.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/2715.

Here's a picture of the graph on the cloud site:
image


esp/esp/program/modules/handlers/studentregphasezeromanage.py
```diff
@@ -106,8 +107,12 @@ def phasezero(self, request, tl, one, two, module, extra, prog):
         context['role'] = role
         q_phasezero = Q(phasezerorecord__program=self.program)
         entrants = ESPUser.objects.filter(q_phasezero).distinct()
-        context['entrants'] = entrants
-        context['nentrants'] = len(entrants)
+
+        recs = PhaseZeroRecord.objects.filter(program=prog)
+        timess = [("number of lottery students", [(rec.user.count(), rec.time) for rec in recs])]
```
hwatheod, line 112: Probably I'm just misunderstanding something, but what does this `.count()` do ?  `rec.user` is already referring to a single user so what does it mean to count?

willgearty: `user` is actually a ManyToMany field in this case (I guess in hindsight, `users` would have been better):
https://github.com/learning-unlimited/ESP-Website/blob/4bc9787dc344d7ea15974ad9386ea98e89fca5d9/esp/esp/program/models/__init__.py#L2052-L2056
So `rec.user` is a set of users and `count()` is just counting how many users are in each `PhaseZeroRecord`.

willgearty: Googling around it appears `rec.user.all().count()` may also work, but I think `.count()` will either be as efficient or more efficient..

hwatheod: Okay, I get it now.  The `user` as opposed to `users` threw me off.

PR #2986, opened 2020-05-05
Webapp class changes based on enrollment vs program vs class attendance numbers

### LINKED ISSUE #2949: Check outs not updating grid based class changes/ AUL
Checking out students isn't immediately updating the relevant AUL/grid based class changes as it should from PR #2935 (tested on dev3). It does update after some time, but not within say ~10 minutes and not from flushing the cache.

### LINKED ISSUE #2968: Allow class changes based on class attendance
Now that we have a module for class based attendance, it might be good to integrate that data into the class change interfaces. My thoughts are:

- If the class attendance module is enabled, we add a fourth number to grid based class changes that represents students checked into class. Ie it would be "students checked in to class/ students checked in to program/ students enrolled/ capacity." This would essentially give a hierarchy of numbers for admins to look at. If the first number is 0, use the second number. If that's 0, use the third number. The first two numbers should be very close if we assume students go to their designated classes, but if we have the extra data, we might as well add it?

- For the webapp, there is currently a tag that lets you switch between using attended vs enrolled numbers for deciding whether a class is full to students. Ideally, we want to have the most accurate number for open class spots and have minimal tag-switching throughout the day. I think this means the tag should have three options:

1. "Enrollment"
Only use enrollment numbers. (For programs who don't want to use the below options because they mark some students as attended but don't think their data is good enough to allow extra students into classes or something.)

2. "Enrollment, program_attendance, class_attendance" progression system.
I envision this doing the following, and being useful for programs where students are expected to attend most of the day and think they take good attendance:

- Until 5 minutes after the start of the program AND at least one person has been marked attended for the program, use enrollment numbers. This allows for the bulk of student check-in to be completed before switching to attendance numbers, and safeguards against cases where attendance doesn't get put in to the website until later (or never).
- Switch to program attendance numbers for all classes.
- If it's 5 minutes or more into a class AND at least one person has been marked as attending a class, use class attendance numbers for that class. This allows for using the most up to date information available while safeguarding against teachers who forget to take attendance as well as people walking in right at the start of class/slightly after. (Note that it's okay if a program doesn't take class attendance, it will just never switch into this regime.)

3. "Enrollment, class_attendance" progression system.
I envision this being useful for programs where there is no sign in procedure at the beginning of the day so students are not all marked attending early on and where attendance for the whole program is not mandatory. For example, Cloud's rainstorm will only take class attendance, and given the lottery based system, the multiple timezones, and the online nature, I do not think the assumption that everyone in the program will get checked in during the first block is valid. Thus, we don't want to ever really rely on the program attendance numbers. It would work as above but skipping the middle step, ie:

- Use enrollment numbers by default.
- If it's 5 minutes or more into a class AND at least one person has been marked as attending a class, use class attendance numbers for that class.

kkbrum:
Fixed #2968 to incorporate class attendance into class changes. Grid based will now show 4 numbers, with the new number being class attendance. Class changes in the webapp will be allowed based on class attendance numbers depending on a combination of tags. 

The tags are: 

- switch_time: If set, this is the time at which class changes will be based on program attendance numbers if available.
- switch_lag: If set, this is the amount of minutes after a class starts at which we will use class attendance numbers if available. 

The hierarchy is that we try to use class attendance, then program attendance, then enrollment numbers depending on the tags, what time it is, and whether attendance has been taken.

In the process of setting up caching for this, I also fixed #2949 (the time variable was messing with the caching, so I just added a function to only use the current time). I also added checkedInStudents to parallel checkedOutStudents to make things easier.


kkbrum:
@willgearty  I got this working, but I want to be careful about not calling the database too many times but am not good with all the caching/whatnot. Could you look at the three TODOs I added and give some suggestions/advice? Thanks <3

esp/esp/program/models/class_.py
```diff
@@ -1111,7 +1115,27 @@ def firstBlockEvent(self):
     def isFull(self, ignore_changes=False, webapp=False):
         if len(self.get_meeting_times()) == 0:
             return True
-        if webapp and Tag.getBooleanTag('count_checked_in_only', program = self.parent_program, default = False):
+
+        now = datetime.datetime.now()
+        switch_time = None
+        if Tag.getProgramTag('switch_time_program_attendance', program=self.parent_program):
+            try:
+                switch_hour, switch_min = [int(x) for x in Tag.getProgramTag('switch_time_program_attendance', program=self.parent_program).split(":")]
```
willgearty, line 1123: Instead of doing this precarious splitting, probably cleaner to use `strptime()`: https://docs.python.org/3/library/datetime.html#strftime-strptime-behavior

esp/esp/program/models/class_.py
```diff
@@ -1111,7 +1115,27 @@ def firstBlockEvent(self):
...
+                switch_time = datetime.datetime(now.year, now.month, now.day, switch_hour, switch_min)
+            # TODO: Is this the best way to deal with someone specifying the tag incorrectly?
+            except ValueError:
```
willgearty, line 1126: If you make the change I suggested above, you'll want to catch whatever error would be thrown for an incorrectly formatted time. But otherwise this looks like a nice way to catch incorrectly specified tag values.

esp/esp/program/models/class_.py
```diff
@@ -1111,7 +1115,27 @@ def firstBlockEvent(self):
...
         if len(self.get_meeting_times()) == 0:
             return True
-        if webapp and Tag.getBooleanTag('count_checked_in_only', program = self.parent_program, default = False):
+
+        now = datetime.datetime.now()
+        switch_time = None
+        if Tag.getProgramTag('switch_time_program_attendance', program=self.parent_program):
+            try:
+                switch_hour, switch_min = [int(x) for x in Tag.getProgramTag('switch_time_program_attendance', program=self.parent_program).split(":")]
+                switch_time = datetime.datetime(now.year, now.month, now.day, switch_hour, switch_min)
```
willgearty, line 1124: Instead of doing this precarious splitting, probably cleaner to use strptime(): https://docs.python.org/3/library/datetime.html#strftime-strptime-behavior

esp/esp/program/models/class_.py
willgearty, line 1126: If you make the change I suggested above, you'll want to catch whatever error would be thrown for an incorrectly formatted time. But otherwise this looks like a nice way to catch incorrectly specified tag values.

esp/esp/program/models/class_.py
```diff
@@ -1111,7 +1115,27 @@ def firstBlockEvent(self):
...
+                pass
+        switch_lag = None
+        if Tag.getProgramTag('switch_lag_class_attendance', program = self.parent_program):
+            try:
+                switch_lag = int(Tag.getProgramTag('switch_lag_class_attendance', program = self.parent_program))
+            except ValueError:
+                pass
+        # TODO: Don't call num_students_attending() twice, but also don't call it unless needed?
+        if webapp and switch_lag and now >= (self.start_time_prefetchable() + timedelta(minutes=switch_lag)) and self.num_students_attending() >= 1:
+            num_students = self.num_students_attending()
```
willgearty, line 1136: Hmm...yeah, that's tricky. I can't think of a way to avoid the two calls here, but we could probably cache `num_students_attending()` above so that it's only one database call maximum. `self.students()` might already have some caching going on, too, so if that's the case, you wouldn't even need to worry about it.

esp/esp/program/models/class_.py
```diff
@@ -1111,7 +1115,27 @@ def firstBlockEvent(self):
...
+        # TODO: Don't count the number of students attending the program for every class...
+        elif webapp and switch_time and now >= switch_time and ESPUser.objects.filter(record__event="attended", record__program=self.parent_program).distinct().count() >= 1:
```
willgearty, line 1138: Is the idea that you only want to see if any of the students that are enrolled in this class are attending the program? If that's the case, you could just put `id__in=self.students['enrolled']` in the filter call, or something like that.

kkbrum: Nah so basically I only want to switch to program attendance numbers if some program attendance has been inputted (guarding against the event that a program wants to switch to program attendance at 10am but something goes wrong on day-of and they haven't inputted any attendance by then and then every class appears empty). Right now, when it checks whether **each class** is full, i check if there are any students marked as attending the **entire program** (the same call over and over again for each class).

An idea is maybe to add an attribute to the parent_program object like parent_program.any_attending() which would be False until there is at least one program attendance record and would only update if it's currently False & the records are changed? Is that possible? Do you have other ideas for how to check that at least 1 program attendance has been inputted without naively counting over and over?

esp/esp/program/models/class_.py
```diff
@@ -1111,7 +1115,27 @@ def firstBlockEvent(self):
...
+            except ValueError:
+                pass
+        switch_lag = None
+        if Tag.getProgramTag('switch_lag_class_attendance', program = self.parent_program):
+            try:
+                switch_lag = int(Tag.getProgramTag('switch_lag_class_attendance', program = self.parent_program))
+            except ValueError:
+                pass
+        # TODO: Don't call num_students_attending() twice, but also don't call it unless needed?
+        if webapp and switch_lag and now >= (self.start_time_prefetchable() + timedelta(minutes=switch_lag)) and self.num_students_attending() >= 1:
```
willgearty, line 1135: Can you add a comment for each of these conditions to explain in english what the condition is?

esp/esp/program/models/class_.py
willgearty: I don't _think_ it's possible to only change the attribute when the number of records changes from 0 to 1. The only way  I can think of to cache such an attribute/function would check the number of records _every_ time the records are changed (but I'm also not an expert on our caching practices). Ultimately, I imagine `Record.objects.filter(event="attended", program=self.parent_program).exists()` would be a pretty quick DB query, but one possible false positive would be if there are records from when registration was being tested.

esp/esp/program/models/class_.py
```diff
@@ -1111,8 +1125,36 @@ def firstBlockEvent(self):
...
+        # Mode 1: Base "fullness" on class attendance numbers if:
+        # 1) using webapp, 2) 'switch_lag_class_attendance' tag is set properly
+        # 3) it is currently past the class start time + however many minutes specified in tag
+        # 4) at least one student has been marked as attending the class
+        if webapp and switch_lag and now >= (self.start_time_prefetchable() + timedelta(minutes=switch_lag)) and self.num_students_attending() >= 1:
+            num_students = self.num_students_attending()
+        # Mode 2: Base "fullness" on program attendance numbers if:
+        # 1) using webapp, 2) 'switch_time_program_attendance' tag is set properly
+        # 3) it is currently past the time specified in tag
+        # 4) at least one student has been marked as attending the program
```
kkbrum, line 1154: Happy to change condition 4 to be at least 5 students attending the program if that seems good in order to avoid problems with test accounts?

esp/public/media/scripts/onsite/ajax_status.js
```diff
@@ -914,7 +914,7 @@ function render_table(display_mode, student_id)
                     new_td.append($j("<span/>").addClass("room").html(section.rooms));
                 }
                 //  TODO: make this snap to the right reliably
-                new_td.append($j("<span/>").addClass("studentcounts").attr("id", "studentcounts_" + section.id).html(section.num_students_checked_in.toString() + "/" + section.num_students_enrolled + "/" + section.capacity));
+                new_td.append($j("<span/>").addClass("studentcounts").attr("id", "studentcounts_" + section.id).html(section.num_students_attending.toString() + "/" + section.num_students_checked_in + "/" + section.num_students_enrolled + "/" + section.capacity));
```
kkbrum, line 917: TODO: Only show the class attendance number if the class attendance module is enabled.

willgearty:
I think you'll need to make a migration for that new field.

willgearty:
Even though there may be some things that still require migrations, you should only be adding a migration for the things relevant to this PR here. Looks like you can just remove that other migration file.

esp/esp/tagdict/__init__.py
```diff
@@ -129,7 +129,8 @@
     'volunteer_help_text_comments': (False, 'If set, overrides the help text for the comments field for the volunteer form', None, 'volunteer', True),
     'volunteer_require_auth': (True, 'Do volunteers need to have accounts before signing up? (If not, one will be created when they sign up)', False, 'volunteer', True),
     'program_center': (False, 'The geographic center for a program, following the form {lat: 37.427490, lng: -122.170267}. Used for the teacher and student onsite webapps.', '{lat: 37.427490, lng: -122.170267}', 'manage', True),
-    'count_checked_in_only': (True, 'Should the student onsite webapp use checked in students (instead of registered students) to determine if a class is full?', False, 'learn', True),
+    'switch_time_program_attendance': (False, 'At what time should the student onsite webapp use program attendance if available (instead of enrollment) to determine if a class is full? If blank, program attendance numbers will not be used. Format: HH:MM', None, 'learn', True),
```
willgearty, line 132: I assume this is 24 hour time?

esp/public/media/scripts/onsite/ajax_status.js
```diff
@@ -1211,6 +1211,7 @@ function populate_students()
         new_student.first_name = data.students_list[i][2];
         new_student.grade = 0;
         new_student.sections = [];
+        new_student.sections_attending = [];
```
willgearty, line 1214: Is this ever used?

esp/public/media/scripts/onsite/ajax_status.js
```diff
@@ -1293,15 +1294,19 @@ function populate_counts()
...
-            if (num_students > data.sections[sec_id].num_students_enrolled)
-                data.sections[sec_id].num_students_enrolled = num_students;
+        else {
+            data.sections[sec_id].num_students_attending = num_students_attending;
+            if (data.sections[sec_id].num_students_enrolled != num_students)
+            {
+                //  console.log("Warning: Section " + sec_id + " claims to have " + num_students + " students but " + data.sections[sec_id].num_students_enrolled + " are enrolled.");
+                if (num_students > data.sections[sec_id].num_students_enrolled)
+                    data.sections[sec_id].num_students_enrolled = num_students;
+            }
```
willgearty, line 1309: I know you didn't write this part, but I'm a little confused about what's going on here. Is this saying that we only update the enrollment numbers if they are _higher_ than they were before? Shouldn't we just always update them like you've done with `num_students_attending`?

willgearty:
Testing results:

- [x] Grid-based class changes visual changes
- [x] Grid-based class changes class attendance numbers update
- [x] Grid-based class changes class program attendance numbers update
- [ ] Webapp shows correct isFull status based on student enrollment
- [ ] Webapp shows correct isFull status based on student program attendance (with tag set)
- [ ] Webapp shows correct isFull status based on student class attendance (with tag set)

willgearty:
OK, everything I tested (see above) worked. I checked to make sure all of the different conditions had to be met before the two different switches occurred (e.g. I tried set tag but no checked in students). If there's anything you think I missed, feel free to let me know and I can do more testing.

I made some comments about the code and caching above, but overall everything seems to update appropriately for what I tested.

Finally, while testing I noticed that we turn classes red when they "Full" on the grid-based class changes page. Should this "Full" status reflect the same "Full" status of the webapp/`isFull`? Right now I think it's strictly triggered when `num_enrolled == capacity`, but do we want to change the condition as you've done for `isFull`?

esp/esp/program/models/class_.py
```diff
@@ -976,6 +978,13 @@ def count_enrolled_students(self):
 
     enrolled_students = DerivedField(models.IntegerField, count_enrolled_students)(null=False, default=0)
 
+    @cache_function
+    def count_attending_students(self):
+        return self.num_students(verbs=['Attended'], use_cache=False)
+    count_attending_students.depend_on_row('program.StudentRegistration', lambda reg: {'self': reg.section})
```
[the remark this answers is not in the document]
kkbrum: I don't think so-- a student attending a class doesnt depend on whether the student is checked in to the program as far as I can tell?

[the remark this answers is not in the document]
willgearty: Yep, my bad, got mixed up with program attendance

esp/public/media/scripts/onsite/ajax_status.js
kkbrum: So num_students is the derived field of enrollment for the section whereas data.sections[].num_students_enrolled originally comes incrementing the enrollment of the section one by one as they loop through student enrollments in line 1257 (populate enrollments). So if they don't match they take the higher of the two. For students checked in to the program, they only do the looping, not the derived field, whereas for class attendance, I only do the derived field because it seems fastest & easiest. Perhaps this code needs to be cleaned up in the future...

Issue #1470, opened 2014-11-15
Onsite computes section capacities differently from section.capacity


benjaminjkraft:
This appears to happen if the class is assigned to multiple rooms with different capacities; see for example E65s1 on brandeis.learningu.org

PR #2998, opened 2020-05-17
Fix and unify section capacity calculation

willgearty:
This PR does two things:

1. Change from using a section's initial rooms to using all rooms to calculate its capacity (to fix the edge case where a section takes place in multiple classrooms and any of the latter classrooms are smaller than the initial classroom. @benjaminjkraft says this should even increase performance (https://github.com/learning-unlimited/ESP-Website/issues/1470#issuecomment-68031030).
2. Use `ClassSection.capacity` to populate capacities in the grid-based class changes (instead of doing the same calculation in the JS using the classrooms that the section is in). Since `ClassSection._get_capacity` does all of the checks that we used to do in the JS, I've removed all of that extra, unnecessary code.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/1470.


hwatheod:
So confused by the existing code here, so I am not really sure what these changes are fixing.

1. There are cases where a single section is split between two rooms?
2. The `_get_capacity` code in ClassSection: It seems like the complicated logic only happens when `max_class_capacity` is not available, which already seems like a very corner case. How can this happen and why wouldn't it be fixed by just asking the teacher for the maximum capacity? 
3. The `_get_room_capacity` code in ClassSection: it seems to be adding up the number of students for all of the classrooms for a *section*. That calculation doesn't make sense to me. I must be misunderstanding something.

willgearty:
Well, I didn't write the original code, but I'll try my best to clear things up.

1. Apparently this has been used before and is supported by the sure (you could imagine a case where a teacher wants to physically move the class from one classroom to another during the class)
2. I think the confusion here stems from the difference between the maximum capacity associated with a class and the maximum capacity associated with a section. The former is set when a teacher fills out the class registration form. The latter is usually null, but can be set to a specific value in the manage class page for a given section. When a value is set, it overrides all other settings and is used as the capacity for the section, otherwise, `_get_capacity` does a bunch of other calculations to figure out a section's capacity based on class capacity and room capacity.
3. Ah, good catch, we probably should be taking the minimum here. I think the original idea was to use `initial_rooms` and sum the capacities if those (in case for some reason a section was in multiple rooms at the same time, which also appears to be supported), but since we're now using all classrooms, I guess what we really want is the minimum sum of any classrooms that occur at the same time.

willgearty:
Sorry this took so long, but I think that change should address the 3rd issue you brought up @hwatheod. Let me know what you think!

Issue #2744, opened 2019-05-13
Make floating resources assignable at teacher checkin


willgearty:
Right now floating resources can only be assigned on the `/manageclass/` page, and I highly doubt anyone even uses that, which means floating resources as a model seems useless at the moment. However, what is a common occurrence is that teachers check-out items at teacher check-in, such as supplies, cables, projectors, etc. In the cases that I've seen, most splashes maintain a google spreadsheet to keep track of these check-outs. But this is literally what floating resources is designed to be used for. So we should have a way to assign (checked-out) these resources on the teacher check-in page (`/missingteachers`), and then a way to mark them as returned (checked-in).

willgearty:
I imagine this will work just like flags.

PR #3012, opened 2020-06-02
Add resource checkout to /missingteachers

willgearty:
This allows admins to checkout floating resources (e.g. projector, speaker, microphone) to teachers through the teacher checkin interface. These resources can then be marked as returned when the teachers return them.

However, at the moment I need to fix how this works because resources are event-specific, and I haven't accounted for that (but we probably don't want to confuse admins with that, so we'll want to collapse the associated resources for the actual interface). We also probably want a page where we can see which items haven't been returned yet.

image

Fixes https://github.com/learning-unlimited/ESP-Website/issues/2744.


willgearty:
OK, this now assigns resources appropriately for each timeslot of a section (and also makes sure you can only choose resources that are available for the entire section). However, due to the way resources work, if you have multiple assignments of the same resource to a section, when you remove them, you have to remove them all. We probably want a new `ResourceAssignment` field like `res_group` for resources that groups the assignments in case we only want to delete a single assignment 'group'.

willgearty:
OK, I've added a grouping field for the resource assignments, so we can now confidently only delete a single set of resources when it is deleted on the teacher checkin page.

The last thing I'm thinking of including is a little status box somewhere on the page that would show any unreturned items for classes that have ended (they would look the same as when they under the class and checking the check box would remove them from that status box). Does something like that sound good @kkbrum or @milescalabresi? And if so, where on the teacher checkout page should I put it?

milescalabresi:
I think that sounds useful! Yeah I'd want it towards the top since keeping track of missing items is usually a priority

esp/esp/program/models/__init__.py
```diff
@@ -978,9 +978,9 @@ def getFloatingResources(self, timeslot=None, queryset=False):
             result = self.collapsed_dict(res_list)
             return [result[c] for c in result]
 
-    def getAvailableResources(self, timeslot):
+    def getAvailableResources(self, timeslot, queryset=False):
         #   Filters down the floating resources to those that are not taken.
-        return filter(lambda x: x.is_available(), self.getFloatingResources(timeslot))
+        return filter(lambda x: x.is_available(), self.getFloatingResources(timeslot=timeslot, queryset=queryset))
```
kkbrum, line 983: The number of resources available seems to only pertain to a block. So if you have three laptops available, you can only check out 3 laptops in a block, but then if you go to block 2, even if those laptops weren't returned, you can still check out 3 more laptops, at least from what I can tell. The resource numbers available should account for the resources checked out during previous blocks that havent been returned yet. (Important because teachers dont return from block 1 until after block 2 teachers have already checked in)

esp/esp/program/modules/handlers/resourcemodule.py
```diff
@@ -674,6 +678,79 @@ def resources(self, request, tl, one, two, module, extra, prog):
         #   Display default form
         return render_to_response(self.baseDir()+'resource_main.html', request, context)
 
+    @aux_call
+    @needs_admin
+    def newassignment(self, request, tl, one, two, module, extra, prog):
```
kkbrum, line 683: This is an oddball but I have a class with two teachers. If I assign 1 resource to the class and refresh the page, suddenly the resource shows up twice for that class? But if I go to the next block it only shows up as one resource again. :/ Not entirely sure it has to do with the two teachers thing, but it doesnt happen for my single teacher class so that's my only guess why the resource fake-doubles.

esp/templates/program/modules/teachercheckinmodule/missingteachers.html
```diff
@@ -211,6 +272,12 @@ <h1>Missing Teachers for {{ program.niceName }}</h1>
                 </div>
                 <button class="add-flag btn">add flag</button>
             {% endif %}
+            <div class="assignment-name-list fqr-class-assignments">
+                {% for assignment in section.resourceassignments.all %}
+                    {% include "program/modules/resourcemodule/assignment_name.html" %}
+                {% endfor %}
+            </div>
+            <button class="add-assignment btn" data-section="{{ section.id }}" data-geturl="/manage/{{program.getUrlBase}}/getavailableequipment/">checkout resource</button>
```
kkbrum, line 280: If a teacher was checked in a previous block, then on the checkin page the teacher's box is minimized saying they are already checked in and then the only way to assign them resources for that block (suppose they didn't need an adapter for their first class but need one for their second class) is to uncheck them and then assign the resource and check them in again. So if teachers are checked in previously, we should collapse the checking in part but NOT the resource checkout part. I think?

esp/esp/program/modules/handlers/resourcemodule.py
willgearty: That is...very strange...I've got classes with multiple teachers, and I haven't experienced this problem in my testing. Also, the number of teachers should not be a factor when the resources are assigned. Is it definitely only one resource that is assigned (in the admin panel)?

esp/templates/program/modules/teachercheckinmodule/missingteachers.html
willgearty: The classes aren't collapsed, the list of sections to be displayed only includes those with at least one unchecked-in teacher. So should we just list all of the sections (ordered as 1) sections with no checked-in teachers, 2) sections with some checked-in teachers, 3) sections with all teachers checked-in) and not have the checked-in teachers box at the bottom?

esp/esp/program/models/__init__.py
willgearty: Hmm...this will not be easy due to how resources work. Resources (and resource assignments) are timeslot-specific, so they don't really know about resources of other timeslots. I guess what I'm saying is it will be pretty hacky to make this work (and in my mind, if all of a resource is already checked out and not returned in real life, an admin would be smart enough to not check it out to another teacher).

kkbrum: I guess that's a fine assumption for the time being then!

esp/templates/program/modules/teachercheckinmodule/missingteachers.html
kkbrum: That seems like a good solution!

esp/esp/program/models/__init__.py
willgearty: OK, I'm happy to return to this at some point, because I do think it would be a nice thing to include, but I don't think the investment will be worth the return at this point.

willgearty:
OK, we're now showing all sections on the page, categorized by whether none, some, or all of their teachers are checked in. I've moved the "Undo Check-in" boxes up to the sections themselves, disabled the "Text Teacher" buttons if the teachers are checked-in, and removed the entire "Teachers checked in" table.

I also found the bug that @kkbrum mentioned (just an oversight on my part from the assignment group change) and cleaned up the CSS a little bit.

esp/templates/program/modules/teachercheckinmodule/missingteachers.html
```diff
@@ -220,6 +220,30 @@ <h1>Missing Teachers for {{ program.niceName }}</h1>
 </tr>
```
kkbrum, line 220: It wont let me comment on lines 205/207, but now that we include checked in teachers in this same box, can you remove the word "Missing" from the super label for the box? since it now includes teachers that aren't missing

Issue #3008, opened 2020-05-31
Make onsite attendance tables collapsable


willgearty:
When you have a lot of students that have been checked in during a time block, the tables can get really long. We should probably have the tables collapsed by default if they have over some number of rows and have a way to expand them in that case.

Requested by Cloud.

Issue #3007, opened 2020-05-31
"Record attendance" button on onsite attendance page is confusing


willgearty:
People think you are supposed to choose an option from the dropdown then immediately click the button (but actually, choosing something from the dropdown reloads the page for the selected timeslot). We should probably add some text to clarify that the button takes you to the non-timeslot specific section attendance page.

PR #3022, opened 2020-06-20
Clean up onsite attendance UI

willgearty:
This cleans up the onsite attendance UI in a number of ways:

1. This moves the "section attendance" button on the landing page to above the dropdown menu. I've also changed the text to make it clear that the two are an "either or" situation (before people were trying to use the dropdown and then immediately click the button).
2. Wrap the tables in collapsible divs. From my experience looking at this page for larger programs, these tables can be obnoxiously large, so now they are hidden by default and can be revealed when necessary.
3. Make the tables sortable, because why not?

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3007 and fixes https://github.com/learning-unlimited/ESP-Website/issues/3008.

### DIFF
--- a/esp/templates/program/modules/onsiteattendance/attendance.html
+++ b/esp/templates/program/modules/onsiteattendance/attendance.html
@@ -5,6 +5,12 @@
 {% block stylesheets %}
     {{ block.super }}
     <link rel="stylesheet" href="/media/styles/forms.css" type="text/css" />
+    <link rel="stylesheet" href="/media/styles/expand_display.css" type="text/css" />
+{% endblock %}
+
+{% block xtrajs %}
+    {{ block.super }}
+    <script src="/media/scripts/sorttable.js"></script>
 {% endblock %}
 
 {% block content %}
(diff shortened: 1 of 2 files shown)


esp/templates/program/modules/onsiteattendance/attendance.html
```diff
@@ -5,6 +5,12 @@
 {% block stylesheets %}
     {{ block.super }}
     <link rel="stylesheet" href="/media/styles/forms.css" type="text/css" />
+    <link rel="stylesheet" href="/media/styles/expand_display.css" type="text/css" />
+{% endblock %}
+
+{% block xtrajs %}
+    {{ block.super }}
+    <script src="/media/scripts/sorttable.js"></script>
```
kkbrum, line 13: maybe add a note to the top of student attendance page after the explanation saying "All tables are sortable by the columns if you click on the headers." since its not super obvious that this is a feature but it is very cool and helpful

willgearty:
OK, those two things should be addressed. I've also made the section attendance tables sortable, made the "Back to Attendance Summary" button go to a specific timeslot if one was selected on the section attendance page, removed "Student" from a lot of the column headers when it was implied already, and split out the student names to first and last names.

Issue #3010, opened 2020-06-01
Survey questions are too squished on webapp


willgearty:
Both the questions themselves and the answer choices are way too squished, sometimes with only single words on each line:

![image](https://user-images.githubusercontent.com/7232514/83449909-ffbdfc00-a419-11ea-8dfd-e8f56ef2177e.png)

Reported by Cloud.

willgearty:
I'm thinking the easiest way to do this is to change everything to `<div>`s with `float:left`, so if one `<div>` is too large, it should be automatically moved to the next line.

willgearty:
I think almost every question type would benefit from at least having the question on a separate row from the answer choices.

Issue #2985, opened 2020-05-05
Favorite class survey question has no dummy data for rendered version in survey management


willgearty:
The "Favorite Class" survey question type depends on the classes that a student/teacher has taken/taught to create a list of options, but when we are displaying it on the survey management page, it doesn't have a list of classes to use:

![image](https://user-images.githubusercontent.com/7232514/81026153-c4da9f80-8e3e-11ea-9629-1dbb4b63a0ec.png)

We should probably just give it a dummy list to show what it will look like on this page.

willgearty:
This also occurs on the survey import confirmation page.

PR #3044, opened 2020-06-29
Updated survey format/presentation

dylanfernandezdelara:
Survey questions are now on a separate line from the answers. I also changed the use of table cells to div containers so the presentation is vertical. Added a style class to forms.css and webapp.css that is implemented in the questions as a div class.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3010.


dylanfernandezdelara:
Issues that still need fixing: 

-favorite.class.html is completely broken. When you make a survey with the "favorite class" question it doesn't give you an option to add any text to the question and instead displays "{% for l in lst %} {{ l.emailcode }}: {{ l.title }}
{% endfor %}". I wasn't sure how to fix this...  

-numeric_rating.html does not display the "Middle Text". I added a "{% if forloop.middle %} {{ middle_text }}{% endif %}" mirroring the existing statement that uses .first or .last, however this didn't display the text either.

willgearty:
That first issue sounds like https://github.com/learning-unlimited/ESP-Website/issues/2985.

dylanfernandezdelara:
I ran into an issue with writing a commit so for some reason that uploaded twice. For this change,

willgearty:
@dylanfernandezdelara it looks like you might have added the dummy database to this PR. Can you remove it, please?

kkbrum:
You currently have a check failing due to lint, which in most cases is just extra spaces, but could be other stylistic/programming things. To see what's making it fail, click on "Details" next to one of the Travis CI's that's failing,  then click on the job number that failed (has a little red x), and scroll down in the job log. Right before the error, it'll specify what's wrong and on what line of what file. In this case it's a trailing whitespace. Once you fix that (or create any new commit to a pull request) the check will run again and hopefully pass! If not, just keep checking the logs to find out what's wrong!

esp/esp/survey/templatetags/survey.py
```diff
@@ -1,3 +1,4 @@
+from __future__ import division
```
willgearty, line 1: Is this necessary? If you're worried about getting a float (or not) from your division, you can just divide by a float (or not) (e.g. 3/2 = 1, 3/2. = 1.5).

esp/esp/survey/templatetags/survey.py
```diff
@@ -47,6 +49,14 @@
 
 register = template.Library()
 
+@register.filter
+def midValue(sizeLs0):
+    sizeLst = int(sizeLs0)
+    if sizeLst%2 == 1:
+        return math.ceil((sizeLst / 2))
+    else:
+        return sizeLst/2+0.1
```
willgearty, line 58: I think you're just worried about getting a float here, so `sizeLst/2.` would be better.

esp/templates/survey/questions/checkboxes.html
```diff
@@ -2,10 +2,10 @@
 {% load survey %}
 <table class="fullwidth">
 <tr>
-    <td width="30%"><b>{{ name }}</b></td>
     <td>
+    <b>{{ name }}</b>
     {% for choice in list %}
-        <div style="padding-left:2em; text-indent:-2em;">
+        <div class = "floatLeft">
```
willgearty, line 8: I think you might want to preserve the padding and indent here (it was added to implement a hanging indent).

dylanfernandezdelara: Yep, I will add it back.

dylanfernandezdelara:
I'm not sure in what context a short answer question would be used in. On the web app the space for a response is very small, but I left it that way in case this question is only used for one to two word responses. Also, for the favorite class question I updated the formatting to match the others, however it still doesn't have a "dummy list" so it doesn't display any classes right now.

esp/templates/survey/questions/favorite_class.html
```diff
@@ -2,11 +2,12 @@
...
 <tr>
-    <td width="30%"><b>{{ name }}</b></td>
-    <td width="70%">
-        {% templatetag openblock %} for l in lst {% templatetag closeblock %}
+    <td>
+    <div style = "text-align: left; font-weight: bold;"> {{ name }} </div>
+    <div style="text-align: left; padding-left:2em; text-indent:-2em;">{% templatetag openblock %} for l in lst {% templatetag closeblock %}
         <input type="checkbox" name="question_{{ id }}" value="{% templatetag openvariable %} l.id {% templatetag closevariable %}" />{% templatetag openvariable %} l.emailcode {% templatetag closevariable %}: {% templatetag openvariable %} l.title {% templatetag closevariable %}<br />
         {% templatetag openblock %} endfor {% templatetag closeblock %}
+    </div>
```
willgearty, line 10: I think you want each checkbox/answer pair in its own div like with the other multiple choice question types below, otherwise you end up with something like this:
image

willgearty: If you're having trouble debugging this, you can register dummy classes through teacher registration

esp/templates/survey/questions/long_answer.html
```diff
@@ -2,7 +2,9 @@
 {% load survey %}
 <table class="fullwidth">
 <tr>
-    <td width="35%"><b>{{ name }}</b></td>
-    <td><textarea rows="{{ rows }}" name="question_{{ id }}{% if for_class %}_{% templatetag openvariable %} id {% templatetag closevariable %}{% endif %}"></textarea></td>
+    <td>
+    <div style = "text-align: left; font-weight: bold;"> {{ name }} </div>
+    <div style="text-align: left; padding-left:2em; text-indent:-2em;"><textarea rows="{{ rows }}" name="question_{{ id }}{% if for_class %}_{% templatetag openvariable %} id {% templatetag closevariable %}{% endif %}"></textarea></div>
```
willgearty, line 7: It might make sense to give this `width: 100%`, since there's no way to specify the default width of the box.

esp/templates/survey/questions/short_answer.html
```diff
@@ -2,7 +2,9 @@
 {% load survey %}
 <table class="fullwidth">
 <tr>
-    <td width="35%"><b>{{ name }}</b></td>
-    <td><input type="text" size="{{ input_length }}" name="question_{{ id }}{% if for_class %}_{% templatetag openvariable %} id {% templatetag closevariable %}{% endif %}"/></td>
+    <td>
+    <div style = "text-align: left; font-weight: bold;"> {{ name }} </div>
+    <div style="text-align: left; padding-left:2em; text-indent:-2em;"><input type="text" size="{{ input_length }}" name="question_{{ id }}{% if for_class %}_{% templatetag openvariable %} id {% templatetag closevariable %}{% endif %}"/></div>
```
willgearty, line 7: I think you can specify the `font-size` CSS attribute for the input and it will make the font (and box) bigger. It looks like the font is pretty small on the webapp and the normal survey, so might be fine to hardcode it here (instead of in the separate css files), but also keep in mind that we want the webapp font size to be reasonable for the user's device size (in most cases I used `vh`, you can see examples in webapp.css).

PR #3053, opened 2020-07-07
Show labels on/near results histogram for labeled numeric rating survey questions

### LINKED ISSUE #3009: Show labels on/near results histogram for labeled numeric rating survey questions
Right now you only see the numbers on the histogram, so it's not clear what the options were labelled as. We should either include the labels on the histogram or include them elsewhere in the results section for that question.

Requested by Cloud.

dylanfernandezdelara:
Fixes #3009 Added labels for the histogram under statistics. This is only displayed if the question is a labeled numeric rating and is validated using .count to check is the list has elements.


dylanfernandezdelara:
For some reason it uploaded my changes from the previous pull request. I'll pull from the main branch and push my changes again.

dylanfernandezdelara:
I pulled from the main branch and it says my branch was up to date. Also, on the GitHub desktop app is says I have only made one change... so I'm not sure what happened

willgearty:
Should be all set now! I cherry picked the new commit from this PR to a clean branch and then force pushed it to this PR.

esp/templates/survey/answers.html
```diff
@@ -48,9 +48,34 @@
...
+        it is assumed to be of labeled numeric rating type and not just numeric rating. -->
+        <td width = "100%">
+            <b>Statistics:</b>
+            <ul>
+                <li>Average: {{ q.answers|unpack_answers|average }}</li>
+                <li>Std. deviation: {{ q.answers|unpack_answers|stdev }}</li>
+                {% if q.question.per_class %}
+                    <li>Average for all classes: {{ q.question.global_average }}</li>
+                {% endif %}
+            </ul>
```
willgearty, line 65: It would be ideal if we could avoid this code duplication (should be able to just move the response labels portion to down below).

esp/templates/survey/answers.html
```diff
@@ -10,7 +10,7 @@
 {% load survey %}
 {% load main %}
 
-<table width="100%">
+<table width="600">
```
willgearty, line 13: We've been trying to move to percentages for widths lately. It's more accessible.

esp/templates/survey/answers.html
dylanfernandezdelara: Gotcha, I was confused and thought that numeric rating didn't have labels but I see that this is the question type with lower, middle, and upper labels.

dylanfernandezdelara:
I did "git pull" from this branch before I started editing and it still uploaded all of my commits from other pull requests :/. Not sure why this happened again. I realized that the if statement I had previously was true for the question type Yes/No which wouldn't make sense since there isn't a histogram. I changed this to just check if it is of type numeric rating. Hope this fixes everything.

willgearty:
Did you also run `git reset --hard origin/HistogramLabels` like I suggested above?

dylanfernandezdelara:
Ok thank you Will. I only did "git push" because it said that my branch was up to date so I assumed I didn't need to do a hard reset. I'll do that next time.

willgearty:
Your local branch is still messed up, you need to sync it with the github version (the remote). You could also delete the local branch then do `git fetch origin`.

dylanfernandezdelara:
"Cannot delete branch 'HistogramLabels' checked out at '/Users/dylanfernandezdelara/devsite'"
I tried deleting the local branch and this is the error I am receiving.

willgearty:
You can't delete the branch that you are on. You'll need to checkout another branch (`git checkout main`) first.

dylanfernandezdelara:
Ok, I deleted the branch and did 'git fetch origin'

esp/templates/survey/answers.html
```diff
@@ -59,6 +62,13 @@
                     <li>Average for all classes: {{ q.question.global_average }}</li>
                 {% endif %}
             </ul>
+            <!-- Creates a key that provides details about the Histogram Labeling-->
+            <b>Response Labels:</b>
```
willgearty, line 66: Looks like now this shows up for numeric rating questions, but we don't have any labels to loop through. Maybe it would make sense to move the lower_text/middle_text/upper_text to here (and remove it from lines 39-46 above). You'll need an if/else here to check if the question has each of the texts (or a list instead if it's a labeled numeric rating question).

dylanfernandezdelara:
This new implementation should have numbered labels (1 = , 2 = , ...) for labeled numeric questions under Statistics data. For numeric rating questions, labels (max of 3 labels— lower, middle, upper) should appear instead of the numbered labels (1 = , 2 = , ...). If the length of the ratings is odd then the middle label will appear and if even, then the only the lower and upper labels will appear. This is consistent with the survey questions which only display all three labels if the number of possible responses/ratings is odd.

esp/templates/survey/answers.html
```diff
@@ -59,6 +54,28 @@
...
+            {% for label in q.question.get_params.list %}
+            <ul>
+                <li> <b> {{ forloop.counter }}: </b>{{ label }} </li>
+            </ul>
+            {% endfor %}
```
willgearty, line 63: The `<ul>` tags should go outside the forloop, otherwise it looks like this:
image
```suggestion
            <ul>
            {% for label in q.question.get_params.list %}
                <li> <b> {{ forloop.counter }}: </b>{{ label }} </li>
            {% endfor %}
            </ul>
```

willgearty: Once that is done, I think we'll want to wrap the whole thing in an if loop to make sure the list exists (otherwise you get an empty `<ul>` that takes up a little space).

esp/templates/survey/answers.html
```diff
@@ -59,6 +54,28 @@
...
+                {% with q.question.get_params as params %}
+                <ol style = "font-weight: bold;">
```
willgearty, line 65: Might be good to wrap the `<ol>` in an if loop that checks at least one of the lower middle or upper texts exists? Otherwise we get an empty `<ol>` that takes up a little space.

esp/templates/survey/answers.html
```diff
@@ -59,6 +54,28 @@
...
                 {% endif %}
             </ul>
+            <!-- Creates a key that provides details about the Histogram Labeling-->
+            <b>Response Labels:</b>
+            {% for label in q.question.get_params.list %}
+            <ul>
+                <li> <b> {{ forloop.counter }}: </b>{{ label }} </li>
+            </ul>
+            {% endfor %}
+                {% with q.question.get_params as params %}
```
willgearty, line 64: This `with` could probably be expanded around the loop above so we are only calling `get_params` once instead of twice.

esp/templates/survey/answers.html
dylanfernandezdelara: When I was testing this even when the list doesn't have labels, the list still has number_of_ratings amount of objects in the list (even if they are empty). Do you want me to NOT display anything if number_of_ratings is zero? The problem is that the list won't be empty because number_of_ratings denotes how many items there are in the list even if the user doesn't give them labels.

willgearty: Oh, I was mostly talking about when the question is of type numeric rating, we want to skip this part (and not include an empty `<ul>`), and vice versa when the question is of type labelled numeric rating (see my comment below).

Issue #3006, opened 2020-05-31
Add links to filtered comm panel/AUL search from /classsearch


willgearty:
We should have links to the comm panel/AUL with the user ID filter pre-populated based on the results of a `/classsearch` query. This will make it really easy to make complicated class queries and then email the teachers or get information about them.

Requested by Cloud.

PR #3058, opened 2020-07-12
Add links to comm panel, AUL, and group text module on /classsearch

willgearty:
I added links from /classsearch to the comm panel, arbitrary user list, and group text module such that you can use the /classsearch query to filter those three modules for the teachers of the classes (this populates the fields via the js in https://github.com/learning-unlimited/ESP-Website/pull/2868/commits/ff525daf72dc7c3bf345c2cfaa3a7800b1d1eb0e). Theoretically, this now makes it possible to put together an arbitrarily complex /classsearch query and then use the results of that to email the teachers (or get their info or text them).

I also removed some of the classes on the other buttons because the varied colors were getting to be a little overwhelming (h/t @kkbrum).

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3006.

### DIFF
--- a/esp/templates/program/modules/classsearchmodule/class_search.html
+++ b/esp/templates/program/modules/classsearchmodule/class_search.html
@@ -7,6 +7,15 @@
     <link rel="stylesheet" type="text/css" href="/media/styles/forms.css"/>
     <link rel="stylesheet" type="text/css" href="/media/styles/flags.css"/>
     <link rel="stylesheet" type="text/css" href="/media/styles/query-builder.css"/>
+    <style>
+    .btn-group a {
+        color: #333333 !important;
+        text-decoration: none !important;
+    }
+    .dropdown-menu a:hover {
+        color: #ffffff !important;
+    }
+    </style>
 {% endblock %}
 
 {% block xtrajs %}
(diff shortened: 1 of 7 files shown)


kkbrum:
I feel like you could have predicted my comment: This is great, but... why not add the same functionality with students? Then if there's a flag of like "needs permission slip" admins can use this to email the students! etc.

esp/templates/program/modules/classsearchmodule/class_search.html
```diff
@@ -41,17 +42,25 @@ <h1>Search Classes for {{ program.niceName }}</h1>
...
 <div class="btn-group">
-    <a class="btn btn-approve" href="./classflagdetails?clsids={{ IDs|join:',' }}" target="_blank">class flags</a>
-    <a class="btn btn-unreview" href="./classflagdetails?comments&clsids={{ IDs|join:',' }}" target="_blank">class flags (with comments)</a>
+    <a class="btn" href="./classflagdetails?clsids={{ IDs|join:',' }}" target="_blank">class flags</a>
+    <a class="btn" href="./classflagdetails?comments&clsids={{ IDs|join:',' }}" target="_blank">class flags (with comments)</a>
+</div>
+</br>
+<div class="btn-group">
+    <a class="btn" href="./selectList?recipient_type=Teacher&userid={% for cls in queryset %}{% for teacher in cls.teachers.all %}{{ teacher.id }}{% if not forloop.last %},{% endif %}{% endfor %}{% if cls.teachers.exists and not forloop.last %},{% endif %}{% endfor %}" target="_blank">Get Teacher Information</a>
+    <a class="btn" href="./commpanel?recipient_type=Teacher&userid={% for cls in queryset %}{% for teacher in cls.teachers.all %}{{ teacher.id }}{% if not forloop.last %},{% endif %}{% endfor %}{% if cls.teachers.exists and not forloop.last %},{% endif %}{% endfor %}" target="_blank">Email Teachers</a>
```
kkbrum, line 60: It preselects "teachers", but doesnt hide the options that come with having "students" selected.

image

willgearty:
Thanks for the comments @kkbrum. I've fixed that bug and I added student links, but I got worried about having more buttons, so I condensed the buttons into dropdown menus:
image

kkbrum:
This looks and works great! I like it a lot. However, I am slightly concerned that people have no way of knowing the difference between attended and onsite/attendedclass as well as any of the other terms and so will be misinformed when choosing the group and will get incorrect results. Would it be possible to add hover text explaining each of the registration types? In both the comm panel filter and in class search? Or some other way to make sure that people are selecting the correct registration type/ combination for their desired output?

willgearty:
OK, let me know what you think of the new changes @kkbrum

kkbrum:
Looks great! Thanks for this. I'm currently getting hover text on the classsearch page but not on the "Class registrations" filter for the AUL. But I'm so pleased with the display names that it's okay if putting hover text on that page is not feasible, but maybe worth a quick look to see if it's easy?

willgearty:
Done!

PR #3061, opened 2020-07-14
Update gitignore and add untracked scripts

### LINKED ISSUE #3042: Add backups and media files to .gitignore
It makes it harder to identify real changes on the server when there are a bunch of backup files (which should be there) that git complains about, so tell git to ignore them.

milescalabresi:
This PR is meant to clean up the untracked files on the server when we run `git status`. It adds some extensions to `.gitignore` (backups and caches) so they don't clutter up the list, and it adds some useful scripts that were previously just floating around untracked on the server to the codebase. The file I deleted had a single line that referenced another file that I kept.

Fixes #3042


willgearty:
Can you please add an actual description for the PR and explain why you deleted a random file?

milescalabresi:
This PR is meant to clean up the untracked files on the server when we run `git status`. It adds some extensions to `.gitignore` (backups and caches) so they don't clutter up the list, and it adds some useful scripts that were previously just floating around untracked on the server to the codebase. The file I deleted had a single line that referenced another file that I kept.

.gitignore
```diff
@@ -28,6 +28,15 @@ theme_compiled.css
 *_admin_thumbnail.jpg
 *_admin_thumbnail.png
 local_fabfile.py
+*.sql
+*.tar
+*.gz
+*.bak
+*.orig
+esp/public/media/uploaded/*
```
willgearty, line 36: This is already in the list

willgearty:
Looks like @benjaminjkraft [made the symlink](https://github.com/learning-unlimited/ESP-Website/commit/d470f96ceb732dde11f37fa695860c72dce464dc) between the crazy_excel_thing and another file at one point in order to give it a better name but to also allow people to keep using the old name. It's probably been long enough that we can remove the symlink at this point. The additions to the list look reasonable (except for the one comment I made).

Finally, do you know if all of the "useful scripts" you've added actually work? We only want to commit scripts if they do what they say they do (I have no idea how old some of these are).

.gitignore
milescalabresi: oops thanks for the catch

milescalabresi:
I removed some that don't do anything as well as the crazy_excel rename because that already existed as `export_all_surveys.py`. I wrote the email one and used it last fall. I tested the donation one and added some flexibility just now

Issue #2980, opened 2020-05-03
Volunteer registration form is missing fields


willgearty:
The volunteer registration form is missing the shirt and comment fields. I'm guessing I accidentally removed these when I remodeled the template to use the new availability layout.

willgearty:
Update: I did NOT remove these, they just have never been there.

PR #3067, opened 2020-07-17
Add shirt and comment fields to volunteer form template

willgearty:
This adds the shirt (size and style) and comment fields to the volunteer form template. They were already in the form itself, just never rendered in the template. Now they are rendered if the fields are included in the form (which is tag-controlled).

I also added a little bit of help text to the timeslot portion of the form.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/2980.

### DIFF
--- a/esp/templates/program/modules/volunteersignup/signup.html
+++ b/esp/templates/program/modules/volunteersignup/signup.html
@@ -100,17 +100,18 @@ The directors have been notified that you are no longer able to volunteer at {{ 
         {{ form.phone }}
         {% if form.shirt_size %}
         {{ form.shirt_size.label_tag }}
+        {{ form.shirt_size.errors }}
         {{ form.shirt_size }}
         {% endif %}
         {% if form.shirt_type %}
         {{ form.shirt_type.label_tag }}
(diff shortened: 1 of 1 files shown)


esp/templates/program/modules/volunteersignup/signup.html
```diff
@@ -79,43 +79,61 @@ <h3>Volunteer Shifts Cancelled</h3>
...
+        {{ form.email.label_tag }}
+        {{ form.email.errors }}
+        {{ form.email }}
+        {{ form.phone.label_tag }}
+        {{ form.phone.errors }}
+        {{ form.phone }}
+        {% if form.shirt_size %}
+        {{ form.shirt_size.label_tag }}
+        {{ form.shirt_size }}
+        {% endif %}
```
hwatheod, line 104: Are there no errors for shirt_size ?

willgearty: Whoops, good catch @hwatheod, thanks! I also realized I had them in the wrong order (relative to the existing code).

Issue #3016, opened 2020-06-07
Multiple admins taking class attendance simultaneously


kkbrum:
For large lecture classes, it would be useful if multiple admins/teachers could take class attendance simultaneously.

Requested by Cloud.

willgearty:
See https://github.com/learning-unlimited/ESP-Website/issues/3004

Issue #3004, opened 2020-05-30
Update section attendance without submit button


willgearty:
We should use post/ajax calls to marks students as attending whenever a checkbox is checked instead of requiring the teacher/admin to click the submit button.

Requested by Cloud.

PR #3068, opened 2020-07-20
Attendance without form submission

willgearty:
When a teacher or onsite user now checks off (or unchecks) a student as attending a class, it sends a POST request to the server and updates the attendance accordingly (similar to how the teacher checkin page works). This means that users no longer need to submit the page to save the attendance. It also means that multiple users can be marking attendance for the same class at the same time and there will be no conflicts.

One thing that bugs me a little is that the status messages are kinda long and obnoxious, but I can't really think of a different way to display them (I though about a whole separate column, but that might result in the other columns getting squished) nor can I think of much shorter status messages.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3004 and fixes https://github.com/learning-unlimited/ESP-Website/issues/3016.

### DIFF
--- a/esp/esp/program/modules/handlers/teacherclassregmodule.py
+++ b/esp/esp/program/modules/handlers/teacherclassregmodule.py
@@ -329,13 +329,13 @@ class TeacherClassRegModule(ProgramModuleObj):
         if 'student' in request.POST and 'secid' in request.POST:
             students = ESPUser.objects.filter(username=request.POST['student'])
             if not students.exists():
-                json_data['message'] = 'User not found!'
+                json_data['error'] = 'User with username %s not found!' % request.POST['student']
             else:
                 student = students[0]
                 json_data['name'] = student.name()
                 sections = ClassSection.objects.filter(id=request.POST['secid'])
                 if not sections.exists():
-                    json_data['message'] = 'Section not found!'
+                    json_data['error'] = 'Section with ID %s not found!' % request.POST['secid']
                 else:
                     section = sections[0]
                     json_data['secid'] = section.id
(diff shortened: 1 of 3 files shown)


esp/esp/program/modules/handlers/teacherclassregmodule.py
```diff
@@ -317,6 +304,79 @@ def process_attendance(section, request, prog):
...
+                                    srs = StudentRegistration.objects.filter(user = student, section = section, relationship = rt)
+                                    if srs.count() > 0:
+                                        sr = srs[0]
+                                        sr.unexpire()
+                                    else:
+                                        sr = StudentRegistration.objects.create(user = student, section = section, relationship = rt)
+                                    if rt.name=='OnSite/AttendedClass':
+                                        sr.end_date = today_max
+                                        sr.save()
+                        json_data['message'] = '%s is marked as attending.' % student.name()
```
kkbrum, line 377: Honestly, I don't think these status messages are necessary. I think the red disable hover and the "updating attendance" are useful to make the user understand that something is happening, but I feel like unless something went wrong, the check/uncheck symbolizes everything and the status is redundant. With a long list of students, the status messages appearing for every attending student will get super annoying

kkbrum: My other thought is to have the status messages appear for 10 seconds and then disappear but I think it's too much work and they aren't necessary anyways

willgearty:
OK, now we only show errors and show them as alerts instead of inline text. In the event of an error, the checkbox status is reverted. The messages that used to be shown inline are now shown in the console log for those that are really interested.

PR #3070, opened 2020-07-24
Check for guardian info before providing it in AUL

### LINKED ISSUE #2972: Guardian email field breaks AUL
Looks like the lookup field should be "e_mail" not "email" here:
https://github.com/learning-unlimited/ESP-Website/blob/6a1bb903527ddbad02ba409c7161f6ef0bb9cfc0/esp/esp/program/modules/handlers/listgenmodule.py#L132-L134

kkbrum:
Fixes #2972 to check that a student has guardian info provided before returning the guardian name, cell phone, or email in the AUL.

Note that I was only able to replicate the original issue on dev3 (tried Yale, cloud, my devserver as well), but the dev3 logs said "AttributeError: 'NoneType' object has no attribute 'phone_cell'" and putting dev3 on this branch fixed the AUL generation (for all students in database).

### DIFF
--- a/esp/esp/program/modules/handlers/listgenmodule.py
+++ b/esp/esp/program/modules/handlers/listgenmodule.py
@@ -132,15 +132,18 @@ class UserAttributeGetter(object):
 
     def get_guardian_email(self):
         if self.profile.student_info:
-            return self.profile.contact_guardian.email
+            if self.profile.contact_guardian:
+                return self.profile.contact_guardian.email
 
     def get_guardian_name(self):
         if self.profile.student_info:
-            return self.profile.contact_guardian.name
+            if self.profile.contact_guardian:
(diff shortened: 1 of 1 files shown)


hwatheod:
The original ticket #2972 says that the field should be `e_mail` rather than `email` - is that correct?

kkbrum:
Not sure why, but I tested it and it works either way-- is there a preference for which I should use?

hwatheod:
I guess it works either way because there is this code in `esp/esp/users/models/__init__.py`

`e_mail = models.EmailField('Email address', blank=True, null=True, max_length=75)`

which is followed later by this code which points `email` to `e_mail`:

`email = property(lambda self: self.e_mail)`

So it's fine, you can leave it as is.

Issue #3035, opened 2020-06-23
Add /teach and /learn survey links somewhere


willgearty:
From @kkbrum:

> Yeah maybe add a "student link: ..." and "teacher link: ..." there so they know where to link students. but also i would think we could do the "is_step" thing like for webapp, where we could have a tag to add survey as a step to student reg that is greyed out/disabled until the program starts

PR #3071, opened 2020-07-25
Add student and teacher links to surveys

willgearty:
I added links to the student and teacher surveys (and the teacher survey review page) to the survey management page (if there are multiple surveys for a single category, we specify the survey ID for each link). I also changed the `isStep` behavior for the survey module such that it will be included in the registration steps if 1) the `survey_isstep` tag is set to True and 2) the first class block of the program has started.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3035.

### DIFF
--- a/esp/esp/program/modules/handlers/teachersurveymodule.py
+++ b/esp/esp/program/modules/handlers/teachersurveymodule.py
@@ -32,17 +32,12 @@ Learning Unlimited, Inc.
   Phone: 617-379-0178
   Email: web-team@learningu.org
 """
-from esp.program.modules.base import ProgramModuleObj, needs_teacher, needs_student, needs_admin, usercheck_usetl, meets_deadline, meets_grade, main_call
-from esp.program.modules import module_ext
+from esp.program.modules.base import ProgramModuleObj, needs_teacher, meets_deadline, main_call
 from esp.tagdict.models import Tag
-from esp.utils.web import render_to_response
 from esp.users.models    import ESPUser
 from django.db.models.query   import Q
-from esp.middleware     import ESPError
-from esp.survey.models  import QuestionType, Question, Answer, SurveyResponse, Survey
 from esp.survey.views   import survey_view, survey_review, survey_graphical, survey_review_single
 
-import operator
 import datetime
 
 class TeacherSurveyModule(ProgramModuleObj):
@@ -75,7 +70,7 @@ class TeacherSurveyModule(ProgramModuleObj):
                 self.program.getSurveys().filter(category__in = ["learn","teach"]).exists())
 
     @main_call
-    @usercheck_usetl
+    @needs_teacher
     @meets_deadline('/Survey')
     def survey(self, request, tl, one, two, module, extra, prog):
         if extra is None or extra == '':
--- a/esp/esp/program/modules/handlers/teachersurveymodule.py
+++ b/esp/esp/program/modules/handlers/teachersurveymodule.py
@@ -45,32 +45,18 @@ from esp.survey.views   import survey_view, survey_review, survey_graphical, sur
 import operator
 import datetime
 
-class SurveyModule(ProgramModuleObj):
+class TeacherSurveyModule(ProgramModuleObj):
     """ A module for people to take surveys. """
 
     @classmethod
     def module_properties(cls):
         return [ {
-            "admin_title": "Student Surveys",
-            "link_title": "Surveys",
-            "module_type": "learn",
-            "seq": 20,
-            "choosable": 1,
-        }, {
             "admin_title": "Teacher Surveys",
-            "link_title": "Survey",
+            "link_title": "Teacher Surveys",
             "module_type": "teach",
-            "seq": 15,
+            "seq": 9999,
             "choosable": 1,
         } ]
-
-    def students(self, QObject = False):
-        event="student_survey"
-        program=self.program
-
-        if QObject:
-            return {'student_survey': Q(record__program=program) & Q(record__event=event)}
-        return {'student_survey': ESPUser.objects.filter(record__program=program, record__event=event).distinct()}
 
     def teachers(self, QObject = False):
         event="teacher_survey"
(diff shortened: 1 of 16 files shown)


esp/esp/tagdict/__init__.py
```diff
@@ -136,6 +136,7 @@
     'switch_time_program_attendance': (False, 'At what time should the student onsite webapp use program attendance if available (instead of enrollment) to determine if a class is full? If blank, program attendance numbers will not be used. Format: HH:MM where HH is in 24 hour time.', None, 'learn', True),
     'switch_lag_class_attendance': (False, 'How many minutes into a class should the student onsite webapp use class attendance numbers if available (instead of enrollment or program attendance) to determine if a class is full? If blank, class attendance numbers will not be used.', None, 'learn', True),
     'student_lottery_group_max': (False, 'What is the maximum number of students that can be in the same lottery group? (set to 1 to not allow groups)', '4', 'learn', True),
+    'survey_isstep': (True, 'Should the student and teacher surveys be shown as steps in student and teacher registration respectively once the event has started?', False, 'manage', True),
```
kkbrum, line 139: Why is the survey_isstep a program tag while webapp_isstep is a global tag? I think these isstep tags should be consistent, otherwise it makes them hard to find as an admin. I think as long as we import tags when we import a program, I would probably prefer both to be program tags.

esp/templates/program/modules/surveymanagement/main.html
```diff
@@ -16,16 +16,24 @@ <h1>Surveys for {{ program.niceName }}</h1>
...
 <ul>
-{% for s in program.getSurveys %}
+{% for s in surveys %}
     <li><a href="/manage/{{ program.getUrlBase }}/surveys/manage?obj=survey&op=edit&id={{ s.id }}">{{ s.name }}</a> (click to edit)
+    <ul>
+    {% if s.category == "teach" %}
+        <li><a href="/teach/{{ program.getUrlBase }}/survey{% if counts.teach > 1 %}?survey_id={{ s.id }}{% endif %}">Link for teachers</a></li>
+    {% elif s.category == "learn" %}
+        <li><a href="/learn/{{ program.getUrlBase }}/survey{% if counts.learn > 1 %}?survey_id={{ s.id }}{% endif %}">Link for students</a></li>
+        <li><a href="/teach/{{ program.getUrlBase }}/survey/review{% if counts.learn > 1 %}?survey_id={{ s.id }}{% endif %}">Link for teachers to review results</a></li>
```
kkbrum, line 34: Perhaps this link should also get added to teacher registration (if student surveys exist)

esp/esp/tagdict/__init__.py
kkbrum: Looking at the code changes I don't have any clue as to how to do this, but ideally even with the "survey_isstep", the step should only show up if a student/teacher survey exists for the program. For example if "isstep" is true and there's actually only a teacher survey, it would be not great for the student survey link to show up in student registration and have this error pop up:
image

kkbrum: The step in student registration for "surveys" shows up smack in the middle of the list. I think it should probably be the last item. Could we update the seq for the module to make it lower on the list?

willgearty:
OK, I think that should address all of your comments @kkbrum!

First, I've split the survey program module into two separate modules, one for student registration and one for teacher registration (a la #2125). I made a migration that hopefully changes existing modules such that programs with the surveys modules will keep those modules and will just use the new handlers (although I had a hard time testing this, so please let me know if it works in review). Ultimately, nothing should behave differently on the frontend with this change, but separating the two modules to separate handlers does allow us to have separate `isStep()` functions, which also means we can have separate isStep tags (#3074). I'll make separate tags for the webapps in a separate PR.

Second, I've added a link for teachers to go to their survey review page. This required a little bit more work than just adding a link because we want to make sure we show the survey page (and not an error page) in cases where there is no teacher survey but there are student survey results for the teacher to review.

Finally, I fixed a bug that I discovered where we weren't linking to the correct page for section or general surveys if there were multiple surveys of the same category for a single program.

esp/templates/survey/survey_form.html
```diff
@@ -22,20 +22,24 @@
...
+                    <td class="underline" align="center">{% if not general_available %}<span style="color: grey">Not available</span>{% elif general_done %}<span style="color: green">Completed</span>{% else %}<a class="btn fancybutton no-dec" href="{{ request.path }}?{% if survey_id %}survey_id={{ survey_id }}&{% endif %}general">Take Survey</a>{% endif %}</td>
                 </tr>
             {% endif %}
         </table>
+        <br><br>
     {% else %}
         <p class="message">This survey does not have any questions.</p>
     {% endif %}
+    {% if student_results %}
+        <p class="message">You can view the results of the student survey(s) for your class(es) by <a href="{{ request.path }}/review">going here</a>.
```
kkbrum, line 41: If there are multiple teacher surveys, then this shows up only after you click on one of the teacher surveys, which is weird to me since the student survey is not connected to the teacher surveys. So at the very least I think it should show up on the page with the list of multiple surveys instead of when you click one of them. 

However, I've realized since my original review that I would prefer if the survey results were linked in the "Manage your classes" boxes at the bottom of the techerreg page instead. Since there's a box per class and that box should contain everything about the class, I think that's the most logical place for a teacher to find their survey results. I think I would just add a button "View survey results" under the "Email ___ Students" button (if survey results exist).

esp/esp/program/modules/migrations/0023_auto_20200730_2006.py
```diff
@@ -0,0 +1,80 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+def set_my_defaults(apps, schema_editor):
```
kkbrum, line 6: this worked for me!

esp/esp/program/modules/handlers/teachersurveymodule.py
```diff
@@ -78,17 +61,16 @@ def teachers(self, QObject = False):
...
-        return {'student_survey': """Students who filled out the survey"""}
-
     def teacherDesc(self):
         return {'teacher_survey': """Teachers who filled out the survey"""}
 
     def isStep(self):
-        return False
+        return (Tag.getBooleanTag('teacher_survey_isstep', program=self.program, default=False) and
+                self.program.getTimeSlots()[0].start < datetime.datetime.now() and
+                self.program.getSurveys().filter(category__in = ["learn","teach"]).exists())
```
kkbrum, line 70: Just a reminder that if you implement my comment about moving the survey results link, then this should just check for a teacher survey

esp/esp/program/modules/handlers/teachersurveymodule.py
```diff
@@ -78,17 +61,16 @@ def teachers(self, QObject = False):
...
-    def studentDesc(self):
-        return {'student_survey': """Students who filled out the survey"""}
-
     def teacherDesc(self):
         return {'teacher_survey': """Teachers who filled out the survey"""}
 
     def isStep(self):
-        return False
+        return (Tag.getBooleanTag('teacher_survey_isstep', program=self.program, default=False) and
+                self.program.getTimeSlots()[0].start < datetime.datetime.now() and
```
kkbrum, line 69: I checked this for you and it works as desired!

esp/templates/survey/survey_form.html
kkbrum: (you can change that last condition to be just if a student survey exists. If there are no results I'm okay with having the link be there and say "no results yet! check back later" when clicked)

willgearty:
OK, I added the links as requested to the main teacher registration page (each link is specific to a class). I also added the survey review link to the "Choose a Survey" page if there are multiple teacher surveys. I figured it couldn't hurt to keep it on the original page, so I didn't change the isStep condition. Thanks as always for the great feedback @kkbrum!

Issue #3003, opened 2020-05-29
Rejected classes aren't shown on /teacherreg or /userview


willgearty:
If a class is marked as rejected, it no longer shows up on a teacher's /teacherreg page or on their /userview page. This can be really confusing for teachers since they no longer see the classes they registered on the main teacher registration page. It can also be really confusing for admins, because they no longer see any classes on the teacher's userview page.

Reported by Cloud.

Issue #2992, opened 2020-05-11
"Accepted but hidden" sections show as "Rejected" in teacher registration


willgearty:
In the "Manage Your Classes" section of the teacher registration page, if a section has been marked as "Accepted but hidden" by an admin, the section will be listed as "Rejected" for the teacher (not the class, just the section).

![image](https://user-images.githubusercontent.com/7232514/81574941-110c6f00-936c-11ea-842f-810d5711ab01.png)

hwatheod:
What is the desired behavior?  I've never understood what "Accepted but hidden" is meant to be used for.

willgearty:
My understanding is you use "Accepted but hidden" if you want the class to be "Accepted" but don't want it to show up in the catalog/student registration? This seems most useful for testing, I suppose, although I guess you might want to have a class that only some students can attend or something?

PR #3073, opened 2020-07-28
Fix class and section status inconsistencies

willgearty:
This fixes a few separate issues, some of which were reported:

1. Previously we had changed `isAccepted` for classes to also return `True` if the class was "accepted but hidden. I've copied this behavior over to sections.
2. We weren't including rejected classes on the teacher registration and userview pages.
3. We weren't distinguishing between "Cancelled" and "Rejected" status on the teacher registration and userview pages.
4. We weren't identifying "accepted but hidden" classes or sections as "hidden" on the teacher registration and userview pages.

It's certainly possible there are other places we repeat numbers 2 through 3, so please let me know if anything comes to mind during review!

Fixes https://github.com/learning-unlimited/ESP-Website/issues/2992 and fixes https://github.com/learning-unlimited/ESP-Website/issues/3003.

### DIFF
--- a/esp/esp/program/models/class_.py
+++ b/esp/esp/program/models/class_.py
@@ -1210,7 +1210,8 @@ class ClassSection(models.Model):
     def friendly_times_with_date(self, raw=False):
         return self.friendly_times(raw=raw, include_date=True)
 
-    def isAccepted(self): return self.status == ACCEPTED
+    def isAccepted(self): return self.status > 0
+    def isHidden(self): return self.status == HIDDEN
     def isReviewed(self): return self.status != UNREVIEWED
     def isRejected(self): return self.status == REJECTED
     def isCancelled(self): return self.status == CANCELLED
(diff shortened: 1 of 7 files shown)


kkbrum:
1. Lower on the manage page in the dropdowns for the class options, it says "accepted but hidden" but at the top of the manage page, it still just says accepted. could we change that to specify the hiddenness?

image to accepted but hidden, it did not change the status of the section as appearing in the "information for section 1" box on the manage page. This seems to contradict what I expected and the note above the class options box that says "Please use the form below to alter blah blah. Note that changes made to the class will propagate to each of the sections, but you can later change each section independently."

willgearty:
OK, I can't make the page override the section statuses without breaking the ability to have custom section statuses at all (without a much bigger change to how we handle the forms on the manage class page which is definitely outside the scope of this PR). So instead I modified the help text to reflect the actual behavior. If you definitely want the other behavior, feel free to make a separate issue. I also fixed the status at the top of the page as requested.

Issue #2482, opened 2018-01-14
Automatic email confirmation when financial aid is approved


bhorowitz:
Request from Cornell Splash:

When a financial aid request is confirmed, automatically email the student saying it has been approved.

willgearty:
I think this would be pretty easy given that we now have a financial aid approval module.

PR #3078, opened 2020-08-04
Send email when financial aid request is approved

willgearty:
When a financial aid request is approved (whether it be automatically in the financial aid app, through the financial aid approval module, or through the admin panel [this is new]), we now send a brief email to the student telling them that their request was approved.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/2482.

### DIFF
--- a/esp/esp/accounting/admin.py
+++ b/esp/esp/accounting/admin.py
@@ -75,6 +75,7 @@ def finalize_finaid_grants(modeladmin, request, queryset):
         grant.finalize()
 class FinancialAidGrantAdmin(admin.ModelAdmin):
     list_display = ['id', 'request', 'user', 'program', 'finalized', 'amount_max_dec', 'percent']
+    readonly_fields=('finalized',)
     list_filter = ['request__program']
     search_fields = default_user_search('request__user')
     actions = [ finalize_finaid_grants, ]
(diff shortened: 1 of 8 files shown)


kkbrum:
For me, I get the email preview in the console only when the financial aid request is automatically approved due to free lunch and not when I approve the request through the back end (although a grant is generated when i approve through the back end). I'm currently having a bit of trouble testing through the financial aid approval module due to missing/incorrect CSRF token errors so not sure what happens in that case.

willgearty:
I just double checked and I got an approval email (via sendgrid and recorded in the console) when I tested each of the following:

- Autoapproval in the finaid app module due to free lunch
- Approval through the admin panel (note, this is with the dropdown on the request list, not on the page for a specific request)
- Approval through the finaid approval module

Note that the approval process will fail silently if the request is already approved. Also, the note in the console should only say something like "INFO: Sent mail to foo@bar.com", you may not get the entire text of the email in the console unless your server is set up that way.

kkbrum:
Ah I see now. I do get the emails for the approval through the dropdown in the admin panel, but not on the page for a specific request. Would it be possible to have an approval email sent if it is approved through the page for a specific request?

willgearty:
OK, after a whole bunch of documentation and source code reading, I came up with a solution that uses the new `approve()` method when a new financial aid grant is added to an individual financial aid request (including using the specified `amount_max_dec` and `percent` values). Let me know what you think, @kkbrum!

PR #3083, opened 2020-08-09
Add dummy data for survey questions that need it

willgearty:
I added some dummy classes and a dummy section for the context of the survey questions when they are rendered on the survey management pages. This was specifically aimed at the favorite classes survey question type, but this copies how survey questions are rendered in the actual survey, so it should catch any other edge cases (apparently you can include django template tags for the fields of the section in a survey question if it is a per-class question?).

Fixes https://github.com/learning-unlimited/ESP-Website/issues/2985.

### DIFF
--- a/esp/templates/program/modules/surveymanagement/import.html
+++ b/esp/templates/program/modules/surveymanagement/import.html
@@ -9,7 +9,7 @@
 {% endblock %}
 
 {% block content %}
-
+{% load survey %}
 <h1>Survey Management for {{ program.niceName }}</h1>
 
 <h2>Confirm Survey Import</h2>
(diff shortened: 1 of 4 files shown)


esp/templates/program/modules/surveymanagement/import.html
```diff
@@ -62,7 +62,7 @@ <h2>Confirm Survey Import</h2>
     {% for question in questions %}
         <tr>
             <td class="underline">{{ question.seq }}</td>
-            <td class="underline" colspan="4">{{ question.render }}</td>
+            <td class="underline" colspan="4">{% if question.per_class %}{{ question.render|substitute:section }}{% else %}{{ question.render|uselist:classes }}{% endif %}</td>
```
kkbrum, line 65: I get this error when trying to import a survey:

image

Issue #2952, opened 2020-04-15
Asterisks still show even when address not required


willgearty:
When the `teacher_address_required` tag is set to False (the default), the fields properly aren't required when the form is submitted, but the form fields still have the "required" class which results in them having asterisks next to their labels.

PR #3086, opened 2020-08-15
Fix asterisks when address not required

willgearty:
This manually removes the "required" CSS class from the address fields when they aren't required (due to the `teacher_address_required` tag being `False`). This removes the asterisks from the labels in the form. This is all necessary due to a django bug reported in https://github.com/learning-unlimited/ESP-Website/issues/2435.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/2952.


kkbrum:
This is magical. One note though is that in doing this, we lost the formatting of the length of the state and zipcode boxes. Would it be possible to retain that formatting? If not, my vote is that it's okay as is, but if there's an easy fix then that would be great.

With asterisks (nice zip/state formatting):
image

Without asterisks (lost formatting of boxes):
image

willgearty:
Whoops, my bad! Should be fixed now @kkbrum!

PR #3104, opened 2020-09-04
Check for payment, not amount due, on extracosts/donation

### LINKED ISSUE #1315: Extra costs erroneously blocks page with "You've already paid" message
This is because we aren't sophisticated enough to let a student switch options after they've already submitted a credit card payment. This is an intended feature.

However, the module can also prevent you from using it, even if you haven't paid. This might be because:
- The cost of the program is $0.
- You've received 100% financial aid.

The module is only checking that you owe $0, and isn't actually checking to see if you've made any credit card payments (which can be found by looking for Transfer objects to receivable for the given program).

If we actually check for credit card payments, then the page should work properly again.

kkbrum:
Fixes #1315. 

On the extra costs/ donation modules, it used to say "whoop you've already paid" even if a program was free or if fin aid was granted because it was checking that the amount due was <=0. Now I switch to checking whether there is any credit card payment instead.

The [has_paid function](https://github.com/learning-unlimited/ESP-Website/blob/43dc04958b5bfa9518ec69b7b6b3b98b8831b803/esp/esp/accounting/controllers.py#L564) in the iac already is based on amount_paid, which sums the credit card transfers, so I switched to using that and got rid of the have_paid functions to reduce code duplication, except in the case where it adds caching.

### DIFF
--- a/esp/esp/program/modules/handlers/donationmodule.py
+++ b/esp/esp/program/modules/handlers/donationmodule.py
@@ -176,7 +176,7 @@ class DonationModule(ProgramModuleObj):
         # credit card payment has occured. For now, just do the same thing we
         # do in other accounting modules, and don't allow changes after payment
         # has occured.
-        if iac.amount_due() <= 0:
+        if iac.has_paid():
             raise ESPError("You've already paid for this program.  Please make any further changes onsite so that we can charge or refund you properly.", log=False)
 
         form = None
--- a/esp/esp/program/modules/handlers/studentextracosts.py
+++ b/esp/esp/program/modules/handlers/studentextracosts.py
@@ -81,10 +81,6 @@ class StudentExtraCosts(ProgramModuleObj):
         super(StudentExtraCosts, self).__init__(*args, **kwargs)
         self.event = "extra_costs_done"
 
-    def have_paid(self):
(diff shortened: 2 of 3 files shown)


hwatheod (approved):

milescalabresi:
I know I'm late to this party, but does this account for students who receive finaid? The `has_paid()` function doesn't look like it does. Is that the behavior we want?

kkbrum:
I brainstormed about this with Will, and we figured that the intent is to keep people from paying by credit card and then messing with their extra costs options. For financial aid, we figured that its changing the options after receiving fin aid is fine as long as the program is assigning financial aid robustly. For example, if programs wanted to not pay for shirts or whatnot, they could put in the program admission amount in the financial aid grant amount and then its okay for a student to add a shirt or whatnot after they receive financial aid and the grant amount would still be correct. If you feel differently, feel free to make an issue and i can change it or add a tag setting so that financial aid is also checked for.

milescalabresi:
Oops, I guess I missed that conversation. That makes a lot of sense. I think the tag idea isn't bad but probably isn't high priority. Any program that is thinking about it that carefully can probably just assign financial aid carefully as well. Thanks

Issue #2348, opened 2017-03-10
Available fields dependent on selected users for selectList


willgearty:
At present, it is just a list of the implemented fields, with no indication as to which will actually return information or are relevant to the selected users.
In particular, some fields are specific to students or teachers, and don't work for other users, even if other users have fields of the same name (e.g. "Graduation Year").
The fields that the admin can select from should therefore be dependent on the users that were selected (even if using a combination of users).

willgearty:
We could handle this like we handle the filters on the page beforehand (with JS):
https://github.com/learning-unlimited/ESP-Website/blob/0a849d3424f5aab7c9c9aed5bb89a6af161c039a/esp/public/media/scripts/commpanel.js#L65-L86

PR #3110, opened 2020-09-12
Filter listgen fields based on recipient type

willgearty:
This filters the vast list of available fields in the list gen module to only those that are relevant based on the recipient type. In the case of a combo list, we still show all of the fields in the form.

While I was at it, I also cleaned up the way we were hiding some of the user filters in the user search modules, fixed a couple of bugs (there were supposed to be initial values for the list gen fields form, the filters were all supposed to be collapsed at initialization, and there was a string handling error when there were no GET parameters in the URL), and cleaned up the styling of the list gen fields form.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/2348.


kkbrum (approved):
Wonderful!

Issue #2978, opened 2020-05-03
Select users to plot on map


willgearty:
It would be a cool if there was a view where you could select users (using the UserSearchController) that were then plotted on a map, like here: https://www.highcharts.com/maps/demo/map-drilldown.

PR #3112, opened 2020-09-15
Generate map of users

willgearty:
This adds a new module that lets an admin use the UserSearchController to select a set of users and then plots the geographic distribution of those users at the state and zip code level. If, for some reason, you only have zip code data (e.g. Cloud/Rainstorm), we use the zip code data to populate the state counts based on a csv table (if a zip code crosses state lines, it should be counted for both states).

image

Fixes https://github.com/learning-unlimited/ESP-Website/issues/2978.


kkbrum (approved):
Fancy!

kkbrum (changes requested):
Wait jokes unapprove. For https://dev3.learningu.org/manage/Splash/2013_Fall/usermap, it should be showing 772 users for enrolled students, but then it says there are 1201 from new york? 1201 > 772...

willgearty:
Looks like the query I was making was interacting strangely with some of the queries that were made by the UserSearchController, causing users to be multiplied based on the USC query rows (e.g. when you picked the "Enrolled" list, each user would be multiplied by the number of enrollments they had for the program). I've fixed it by creating a brand new user filter based on the ids of the users from the USC filter.

kkbrum (approved):
Fabulous!

Issue #3030, opened 2020-06-23
/missingteachers search is bad


willgearty:
We really should be searching all teacher name text, not just from the beginning of names.

PR #3113, opened 2020-09-18
Improve teacher checkin search

willgearty:
Previously, the search on the teacher checkin page (`/missingteachers`) would only search the very beginning of a teacher's name. Now it permits any regular expressions (including plain text searches) of teacher names and class titles/codes. I updated the search help text accordingly.

image

I also fixed some bugs/layout issues that I ran across:

1. The search no longer captures input when you are trying to type into a flag comment field.
2. The search was previously supposed to search the class title, but we were using `section.name` for the checkin button's name attribute. I replaced this with `section.title`, the valid property.
3. We had the same issue as #2 where we were listing the emailcode for each class. I fixed it, but that caused spacing issues, so I combined it with the title row (as just `{{ section }}`) and moved it to the top of each class section.

image

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3030.


hwatheod (commented):
Can you please add comments or constant definitions indicating which of the key codes correspond to which keys?

willgearty:
I didn't change any of them, but that's probably a good idea regardless. I'll get to it ASAP!

hwatheod:
> I didn't change any of them, but that's probably a good idea regardless. I'll get to it ASAP!

Sorry, I didn't realize that code was already there before. Switching to "split" diff and "hide whitespace changes" made that clearer.

willgearty:
I knew there was the unified/split option, but I never realized there was a "hide whitespace changes" option in the diff view! Anywho, I've added comments to explain what the seemingly random if statements are checking for.

esp/templates/program/modules/teachercheckinmodule/missingteachers.html
```diff
@@ -155,16 +155,22 @@ <h1>Missing Teachers for {{ program.niceName }}</h1>
...
-        {{ section.parent_class.emailcode }}: {{ section.name }}
+    <td class="section-detail-header" title="Click to toggle class details" colspan="5">
+        {{ section }}
+        (<a href="{{section.get_absolute_url|urlencode}}">manage</a> | <a href="{{section.get_edit_absolute_url|urlencode}}">edit</a>)
     </td>
+</tr>
+<tr>
+    <!-- <td rowspan="{% if section.teachers %}{{ section.teachers|length }}{% else %}1{% endif %}" class="section-code"> -->
+        <!-- {{ section.parent_class.emailcode }}: {{ section.title }} -->
+    <!-- </td> -->
```
hwatheod, line 173: Can you delete this commented out code ?

willgearty: Whoops, good catch!

hwatheod (approved):

Issue #3114, opened 2020-09-20
Make isStep dependent on presence of module options


willgearty:
Most teacher/student registration modules are included as steps in registration as soon as they are enabled. However, some modules, such as the teacher training sign up or student lunch selection modules shouldn't be shown if there are no training slots or lunch times to choose from. We should update isStep for these modules (and others like them) accordingly (see the recent example for the survey modules in https://github.com/learning-unlimited/ESP-Website/pull/3071).

PR #3115, opened 2020-09-24
Add isStep dependencies to some teacher and student modules

willgearty:
This adds custom `isStep` functions to a number of teacher and student modules that will effectively hide the modules in registration if they are not set up correctly. For example, if there are no lunch times set up, we shouldn't show the lunch selection module, even if it is enabled. I believe this should cover all of the modules where such a change would be appropriate (although I'd be happy to be proven wrong), except for two: the combined student and teacher custom form handler. As is, a single `isStep` can't be written for this handler since the two modules would be dependent on the existence of different tags. The handler will need to be split in a future PR (see https://github.com/learning-unlimited/ESP-Website/pull/2125).

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3114.

### DIFF
--- a/esp/esp/program/modules/handlers/availabilitymodule.py
+++ b/esp/esp/program/modules/handlers/availabilitymodule.py
@@ -228,6 +228,9 @@ class AvailabilityModule(ProgramModuleObj):
 
         return render_to_response(self.baseDir()+'availability_form.html', request, context)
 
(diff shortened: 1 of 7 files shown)


hwatheod (approved):
Looks good, although I didn't think about if there were any other places where a similar change would be appropriate.

PR #3117, opened 2020-09-25
Attended student class-hours on dashboard

### LINKED ISSUE #2066: Add attended student class hours to dashboard
Right now, only capacity and registered show up under the student class hours, but we actually use the number of class hours for _attended_ students for things like calculating votes for the chapter board (and it's nice to know for publicity, etc.). It wouldn't be hard to add this number to the dashboard, given that it's just summing a column from an arbitrary user list: take attended students and display class hours -- I think it's safe to assume that if they attended, then they came for approximately however many hours they're signed up for.

kkbrum:
Fixes #2066. Adds a statistic for the number of class hours enrolled by students that attended the program. Note that as mentioned in the issue, this is a crude estimate, that assumes any attended student attended all the classes for which they were enrolled. This could be improved in two different ways based on newer website features:
1. Use the check in / check out times to determine which of their enrolled classes they could have potentially attended
2. Use the class attendance instead of program attendance.

1 seems overly complicated for the time being and most programs probably aren't using the check out feature yet. 
2 is very very easy, but until most programs are using class attendance, won't be very useful. 
I figure the crude estimate is satisfactory for now? Especially since this is the number that would be used for chapter board votes, i think the crude estimate is best since otherwise programs that use the check out feature would be penalized over those that dont. 

As usual with the dashboard, the caching is very funky :/


willgearty:
Looks like you need to renumber your migration (and perhaps change the dependency within the migration) because another migration was added on main.

hwatheod (approved):

Issue #2990, opened 2020-05-05
Automatically pull default values for tags


hwatheod:
Right now, the `Tag.getTag` function requires you to specify the default value each time you call it.  But we have a catalog of tags with default values in the `tagdict/__init__.py` so there should be a way to pull it automatically in `Tag.getTag` without having the caller specify it each time (and possibly make a mistake).

willgearty:
I think there might be one or two tags that might have different defaults based on the context (e.g. `catalog_sort_fields`), so we would still want a way to specify a default, but otherwise fall back to the default in the `tagdict` catalog.

Issue #2964, opened 2020-04-21
Validators/widgets for tag settings


willgearty:
It would be really nice if we could specify validators or even entire widgets for each tag for the tag settings page. This would help prevent admins from specifying improper tag values which often can break the website (especially since tags are so exposed now on the new tag settings page). At this point, it might also make more sense to make the tagdict init lists actually be lists of dictionaries instead of lists of tuples.

PR #3128, opened 2020-10-02
Tag implementation improvements

willgearty:
This implements three major things for tags:

1. Converts the tuples to dictionaries in `tagdict/__init__.py`, which makes things like the next item much easier to add in the future.
2. Adds the ability to specify custom form fields/widgets/validators for specific tags for the tag UI. I've added a good number of custom fields for the tags where I though it was obvious what field type it should be or how we should validate the submitted values. There are still many that could use better validation, especially the JSON tags ([maybe something from here](https://djangopackages.org/grids/g/json-fields/)) and the tags with a specific set of options (e.g. teacher_profile_hide_fields, see the TODO).
3. Removes almost all default arguments from the uses of getTag, getProgram, and getBooleanTag in favor of pulling the defaults from the tag dictionaries in `tagdict/__init__.py`. This should help make the defaults more consistent when tags are implemented in the future.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/2990 and fixes https://github.com/learning-unlimited/ESP-Website/issues/2964.


esp/esp/tagdict/__init__.py
```diff
@@ -1,144 +1,943 @@
+from django import forms
+from django.forms import widgets
+from django.core.validators import RegexValidator
+from decimal import Decimal
+
+# Can't import this because of an import loop
```
hwatheod, line 6: Instead of defining it twice, can the other place(s) import this one instead? This file feels like a "config" file, it makes sense if all other code that needs _states imports it from here.

willgearty: It feels kinda weird defining the `_states` list in the tagdict config file to be used elsewhere, but I guess it would certainly resolve the import issue.

hwatheod (approved):

Issue #2057, opened 2016-02-29
Link to public view of past comm panel emails


benjaminjkraft:
It would be nice if we could (optionally enable) viewing a comm panel or built-in mailing list email in the browser.  This could have a number of uses, a few that come to mind:
- Putting a "view this email in a browser" link in HTML emails
- Linking to student logistics emails from the website in case students or parents didn't see them
- Linking to teachers' emails from the catalog, which someone suggested might be a great lightweight way to have some class materials be accessible to other students

willgearty:
This is a great idea!
We store the data for every email, so it shouldn't be too crazy to make a view to see any email based on its ID, right?

For teacher emails, are those actually saved the same way comm panel emails are saved?

PR #3129, opened 2020-10-08
Add public view for emails marked as public

willgearty:
This adds a public view for anonymous users to view the generic (non-parsed) version of an email sent through the comm panel. The view will only work for emails that have been marked as "Public" when sending them via the comm panel (there is a new checkbox). When the admin finalizes the email, the link to the public view will be shown on the "finished" page. The link to the public view is also shown on the email monitoring page (when the email is public). There's also a new tag in the comm panel to add a link to the public version of the email.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/2057.


hwatheod:
In the public email view, what happens to the fields like {{ firstname }} (or whatever) that are filled in with each student's information in the private emails?

esp/esp/dbmail/models.py
```diff
@@ -218,6 +224,7 @@ def createRequest(var_dict = None, *args, **kwargs):
 
         if var_dict is not None:
             new_request.save()
+            var_dict['request'] = new_request
```
hwatheod, line 227: What is this new line for?

willgearty: This adds the message request as a variable for the email rendering. I added a new tag to the comm panel (`{{ request.public_url }}`) that utilizes this variable.

hwatheod (approved):

Issue #1903, opened 2015-11-22
Search by parent email


btidor:
User search should allow searching for students by a parent's email address.

willgearty:
The commit by @btidor doesn't seem to work as desired (at least on the Stanford site), possibly because parent contact infos aren't being associated with the proper user ids.

willgearty:
In fact, after looking at the code a bit, we never assign a user object to [emergency](https://github.com/learning-unlimited/ESP-Website/blob/9147b7912921642812fa3fc4bfdbca35eaac9665/esp/esp/web/views/myesp.py#L167) or [guardian](https://github.com/learning-unlimited/ESP-Website/blob/9147b7912921642812fa3fc4bfdbca35eaac9665/esp/esp/web/views/myesp.py#L174) contact infos.

Issue #1075, opened 2014-02-19
ContactInfo Objects with Null User


jmoldow:
There are ContactInfo objects with a user field of None. The problem seems to be some calls to `ContactInfo.addOrUpdate` that forget to pass something for the `curUser` argument.

I'd like to propose the following:
- Fix all calls to `ContactInfo.addOrUpdate` to add the missing argument, and make the argument required instead of optional in the function definition.
- Add a migration to fill in the missing user field by taking it from the associated RegistrationProfile. If any ContactInfos are missing a registration profile, we can delete them, since there is no way to associate them with a user.
- Add a migration to make the user field null=False.

Any objections?

PR #3131, opened 2020-10-15
Give all contact infos a user

willgearty:
I've made the following changes:

1. All calls to `ContactInfo.addOrUpdate` now require a user object.
2. Added a migration to clean up old ContactInfos. This includes deleting any that aren't associated with registration profiles and adding user objects if they are missing them based on their associated registration profiles. This means ContactInfos for emergency contacts and guardians are now associated with the student accounts. Also, we were making blank emergency ContactInfos for every non-student, so those are deleted.
3. Made the `user` field required for ContactInfos (including a migration).
4. Resolved any fallout due to old code that assumed ContactInfos had a 1:1 relationship.

In addition to the benefits of all ContactInfos now having an associated user and deleting a bunch of unneeded/blank ContactInfos, this also makes it possible to perform a user search based on any information within emergency contact or guardian ContactInfos (e.g. find a student from their parent's email address), which has been requested many many times by chapters.

I've tested the migrations multiple times, but it would be good to get another set of eyes on them and their application to make sure no important data is accidentally deleted.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/1075 and fixes https://github.com/learning-unlimited/ESP-Website/issues/1903.


esp/esp/users/migrations/0020_contactinfo_cleanup.py
```diff
@@ -0,0 +1,44 @@
...
+            # Look for associated registration profiles with user info
+            if ci.as_user.exists():
+                ci.user = ci.as_user.latest('last_ts').user
+                ci.save()
+            elif ci.as_guardian.exists():
+                ci.user = ci.as_guardian.latest('last_ts').user
+                ci.save()
+            elif ci.as_emergency.exists():
+                if not (ci.first_name and ci.last_name):
+                    # We want to delete emergency contact info for non-students
```
hwatheod, line 20: Don't see the relationship between the code and the comment. How is having (or not having) a first name and last name the same as being (or not being) a student?

esp/esp/program/modules/handlers/mapgenmodule.py
```diff
@@ -78,17 +78,14 @@ def usermap(self, request, tl, one, two, module, extra, prog):
...
-                ).annotate(count = Count('contactinfo__address_state'))
-            states = {state['contactinfo__address_state']: state['count'] for state in states_raw if state['contactinfo__address_state']}
-
-            zipcodes_raw = ESPUser.objects.filter(id__in=[user.id for user in users]).select_related('contactinfo_set', 'contactinfo_set__address_zip'
-                ).values('contactinfo__address_zip'
-                ).order_by('contactinfo__address_zip'
-                ).annotate(count = Count('contactinfo__address_zip'))
-            zipcodes = {zipcode['contactinfo__address_zip']: zipcode['count'] for zipcode in zipcodes_raw if zipcode['contactinfo__address_zip']}
+            #   Only get contact infos that are for the actual users (not guardians or emergency contacts)
+            states_raw = ContactInfo.objects.filter(user__in=users, as_user__isnull=False
```
hwatheod, line 82: I don't quite understand this as_user condition.  How does it tell us that the ContactInfo is for the actual user?  I thought as_user was related to registration profiles.

hwatheod (commented):
For item 4, how did you look for all cases where the code assumes that ContactInfos and Users are 1:1 ?

esp/esp/program/modules/handlers/mapgenmodule.py
willgearty: as_user is the set of registration profiles that use the contact info as the `contact_user` relationship. Otherwise, they use them for `contact_emergency` or `contact_guardian` relationship (if the user is a student). If we find that the contact info has been used as the `contact_user` relationship, then we know this is the contact info for the actual user.

esp/esp/users/migrations/0020_contactinfo_cleanup.py
willgearty: Ah, so I'm trying to retroactively fix the problem caused by this old line (now moved into the student forloop):
https://github.com/learning-unlimited/ESP-Website/blob/0fa4528f5c484de79569d651f4dea0672214444a/esp/esp/web/views/myesp.py#L174.
We used to add a `contact_emergency` for all users, not just students. But we only ask for emergency contact info in the student version of the profile form, so it always resulted in a blank contactinfo. So here we are looking to see if a contact info has been used for `contact_emergency` but does not have a first and last name (indicating it was created by the above line for a non-student).

willgearty:
I searched GitHub for all the uses of either `ContactInfo` (for generating an object) or `contactinfo` (for queries). I also just checked any calls/queries that used `contact_user`, `contact_emergency`, or `contact_guardian` but they are all related to getting those objects for a single registration profile, so they should be fine. I also just checked and `as_user`, `as_guardian`, and `as_emergency` are never used in the code base before this.

esp/esp/program/modules/handlers/programprintables.py
```diff
@@ -609,7 +609,7 @@ def teachersbyFOO(self, request, tl, one, two, module, extra, prog, sort_exp = l
                 classes = new_classes
 
             # aseering 9-29-2007, 1:30am: There must be a better way to do this...
```
kkbrum, line 611: This comment can probably be deleted since you've changed the code (hopefully for the better)?

esp/esp/program/views.py
```diff
@@ -312,13 +312,15 @@ def find_user(userstr):
         if cleaned.isnumeric() and len(cleaned) == 10:
             formatted = "%s%s%s-%s%s%s-%s%s%s%s" % tuple(cleaned)
             user_q = user_q | Q(contactinfo__phone_day=formatted) | Q(contactinfo__phone_cell=formatted)
-
+        #try name (including parent/emergency contact)
```
kkbrum, line 315: Can you add a space after the # before the comment? And capitalize the first letter? :)

kkbrum (approved):
The migration worked well for me, cleaning up the contact infos without removing anything extra and assigning users as I would have expected it to. I tested creating a student account and the contact infos for the student, parent, and emergency contact all looked good with the affiliated user (and the user search allowed searching by parent & emergency contact info). I also tested that creating a teacher account only creates one contact info for the teacher. If there's anything else I should test, let me know, but otherwise this looks great!

PR #3133, opened 2020-10-17
upgrade psycopg2 to latest (2.8.6)

### LINKED ISSUE #3132: Update Python Postgres package psycopg2
According to https://github.com/mwielgoszewski/doorman/issues/138, our current version shouldn't actually be working, and it's not possible to install this version on the new server because it's been deprecated. It sounds like we want >=2.7 of psycopg2.

hwatheod:
Fix #3132.

### DIFF
--- a/esp/requirements.txt
+++ b/esp/requirements.txt
@@ -23,7 +23,7 @@ ipython==3.2.1 # Used for shell_plus
 Markdown==2.3.1 # Used in QSD
 numpy==1.7.1 # Used mainly in the lottery and class change controller
 pillow==3.3.3 # Required for ImageField, which we use in teacher bios
-psycopg2==2.6.1 # Talks to postgres
+psycopg2==2.8.6 # Talks to postgres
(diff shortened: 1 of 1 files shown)


milescalabresi:
@hwatheod were you able to test and/or verify in the changelog <https://pypi.org/project/psycopg2/#history> that it doesn't break anything or introduce additional dependencies that might?

hwatheod:
> @hwatheod were you able to test and/or verify in the changelog https://pypi.org/project/psycopg2/#history that it doesn't break anything or introduce additional dependencies that might?

I reviewed the "what's new" from 2.6.2 onwards at https://www.psycopg.org/docs/news.html and didn't see anything that would obviously affect us. To be honest, it's kind of hard to tell, because we never call psycopg2 directly ourselves; it's called on our behalf by Django when it accesses the DB.  The tests all pass, which is at least some check that things are working.

milescalabresi:
That's good enough for me. Since we upgraded to the most recent version of Django, I would hope that the most recent version of this is compatible

milescalabresi (approved):

Issue #3127, opened 2020-10-02
Manage class error when duration is exactly 1 hour long


willgearty:
When a class is exactly one hour long, an error shows up next to the duration form field on the manage class page. Not sure what's going on, since I thought this was fixed in https://github.com/learning-unlimited/ESP-Website/pull/2668.

Confirmed on Columbia's and Yale's sites.

![Screenshot_20201001-225616](https://user-images.githubusercontent.com/7232514/94886636-67706c00-0439-11eb-89ab-8a421f929087.png)

hwatheod:
Could not reproduce this in dev3.  It looks like these changes did not get merged into some sites.  For example, in the branch yale-sr12, the change isn't there.

Compare: https://github.com/learning-unlimited/ESP-Website/blob/yale-sr12/esp/esp/program/models/__init__.py  (lines 1025-1029)
with the changes in #2668  - yale-sr12 is still the old code.

willgearty:
Looks like I reverted the change in https://github.com/learning-unlimited/ESP-Website/pull/2704 in favor of broader duration changes. I see the same error for the dev3 class that you made @hwatheod:
![image](https://user-images.githubusercontent.com/7232514/95019581-c564a580-062b-11eb-9eb8-6a5dfd7be7c7.png)

PR #3139, opened 2020-10-23
Fix /manageclass duration error message

willgearty:
This fixes the error message shown when a class has a duration that is exactly some number of hours. `getDurations()` is only ever used in the manageclass form and the teacher registration forms, and I checked both and there are no errors and both are able to modify the field without any problems.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3127.


kkbrum (approved):
Stellar!

Issue #3136, opened 2020-10-22
Classroom timeslots on resources page are combined if adjacent


willgearty:
It looks like if the timeslots that are assigned to a classroom are exactly adjacent (no gap between), they are combined when they are listed in the table. This can be misleading as it makes it seem like the classrooms are only available for a single timeslot.

Reported by Brandeis.

PR #3140, opened 2020-10-23
Don't group timeslots on resources pages

willgearty:
This ungroups the timeslots on the resources pages (specifically for classrooms and equipment/floating resources), so that it's clearer which timeslots (as the column heading says) are available for the resource.

Old: 
image
New:
image

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3136.


kkbrum (approved):
Excellent!

Issue #3135, opened 2020-10-20
Phase Zero Management breaks if a user doesn't have a grade


willgearty:
Because of the way student registration modules work, students can access "future" modules before they have completed all required "past" modules if they have the specific link to the module (e.g. /learn/one/two/studentregphasezero). That means that a student can theoretically enter the phase zero lottery without filling out a profile. If this happens, the management page breaks because we can't figure out the student's grade (last line of the below code):

https://github.com/learning-unlimited/ESP-Website/blob/453dfef8bd34542412043e3f0edff8ca1608fe51/esp/esp/program/modules/handlers/studentregphasezeromanage.py#L151-L156

This second loop should be modified to check if a grade is in the dictionary first, then incrementing if so, and adding it if not (basically, combining the two for loops). If a student doesn't have a grade, getGrade returns 0, so we can just add that as another grade in the stats.

PR #3141, opened 2020-10-23
Catch students with invalid grades in phase zero

willgearty:
The phase zero management page would break when a student somehow hadn't filled out their profile. This fixes this and also alerts the admins to which students have invalid grades.

image

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3135.


kkbrum (approved):
Wonderful!

Issue #2392, opened 2017-06-01
Grade range increment is screwed up


AhaanRungta:
It seems like grades increment on July 31 (i.e. that's the graduation date). For this reason, we need to keep adding a increment_default_grade_levels tag for every Summer HSSP. We should just change this to like June 1 or something so that we don't have to deal with the rising 7th graders with this tag every single time.

willgearty:
Perhaps it would make more sense to have a tag that overrides when the grades increment, with the default being July 31?

PR #3143, opened 2020-10-30
Add tag to adjust date when grades increment

willgearty:
I added a new tag `grade_increment_date` that allows admins to adjust when student grades increment (default is July 31). This can be useful for chapters that have summer programs instead of always setting the `increment_default_grade_levels` tag. I've tested that dates before today maintain grades on the /userview page and dates after today result in grades - 1 on the /userview page, so it appears to be working.

image

Fixes https://github.com/learning-unlimited/ESP-Website/issues/2392.

### DIFF
--- a/esp/esp/tagdict/__init__.py
+++ b/esp/esp/tagdict/__init__.py
@@ -450,7 +450,7 @@ all_global_tags = {
         'is_boolean': False,
         'help_text': 'When should students\' grades rollover/increment?',
         'default': datetime.date(datetime.date.today().year, 7, 31),
-        'category': 'manage',
+        'category': 'learn',
         'is_setting': True,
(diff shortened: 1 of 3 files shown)


esp/esp/tagdict/__init__.py
```diff
@@ -445,6 +446,14 @@
         'category': 'manage',
         'is_setting': True,
     },
+    'grade_increment_date': {
+        'is_boolean': False,
+        'help_text': 'When should students\' grades rollover/increment?',
+        'default': datetime.date(datetime.date.today().year, 7, 31),
+        'category': 'manage',
```
kkbrum, line 453: PR works great, but I think this tag should be in the "student registration settings" portion of the global tags since that's where the other grade-related tags are, and this only affects students.

kkbrum (approved):
Spiffing!

Issue #3144, opened 2020-10-30
address_state field broken for international users for newer chapters


willgearty:
Many years ago we had an issue (https://github.com/learning-unlimited/ESP-Website/issues/696) where Canadians couldn't submit their profile because the field only permitted 2 characters (because it was originally designed to be for US states only). We fixed this in the databases once this was discovered with https://github.com/learning-unlimited/ESP-Website/commit/8d25d7b14b7fd1e72cd3ecd283d0e814249ea178. However, it looks like a) that migration was deleted in one of the more recent releases (c. Stable Release 7 in https://github.com/learning-unlimited/ESP-Website/pull/1779) and b) we never actually changed the models to reflect the migration.

Both of these problems have resulted in the users for **new chapters** (since the migration was deleted) not being able to submit the profile form with "Canada" as an option. I only discovered this when I went to add an "International" option for UR Splash, but it resulted in errors when the form was submitted. I've added a fix for this (including model changes and a migration) on the SR12 branch (https://github.com/learning-unlimited/ESP-Website/commit/10dffc09c5d663b157714f95503c5241c96a1193), and it will need to be adapted for SR13 due to migration indices changing.

Reported by UR Splash. I confirmed it is also a problem for Lexington Splash. It is NOT a problem for Stanford or Berkeley.

PR #3149, opened 2020-11-05
Add international option to states

willgearty:
This adds the change implemented in https://github.com/learning-unlimited/ESP-Website/commit/10dffc09c5d663b157714f95503c5241c96a1193 for main (couldn't just cherry-pick because the state list is somewhere else now and the migration needed a different number). The changes add an international option to the state list and make the address-state field support values that are up to 32 characters long. This also fixes newer chapters not allowing Canada to be selected from the state dropdown since they didn't have the old migration from https://github.com/learning-unlimited/ESP-Website/commit/8d25d7b14b7fd1e72cd3ecd283d0e814249ea178.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3144.


kkbrum (commented):
I know this was a quick urgent fix, but wouldn't we want to collect the country of international students? Obviously this doesn't fit the US-centric fields right now, but maybe just leaving the "international" option in the state but then having a "country" field that is only required to be filled out if international is selected or something

willgearty:
Yeah, that's probably a good idea. I've gone ahead and added a country field which is hidden unless the "International" state is selected. When the field is hidden, it's set to "United States", so we will get the country of all users the next time they fill out their profiles (and until then, the default is "United States"). For some reason, the new country field takes nearly 3 minutes to load on my dev server, so I've put the PR on dev3 for testing.

kkbrum (approved):
Magnificent!

Issue #3146, opened 2020-10-31
AJAX scheduler timeslot headers are weird


willgearty:
The headers are duplicated and offset in the AJAX scheduler:

![image](https://user-images.githubusercontent.com/7232514/97783993-62522a00-1b69-11eb-9c44-a9e298957cdf.png)

I think it might have something to do with the really long classroom names?

Reported by Berkeley. Works fine for Yale and Stanford.

PR #3150, opened 2020-11-05
Fix ajax scheduler header alignment

willgearty:
This fixes the width of the cells of the room column so the timeslot headers are properly aligned with the scheduling matrix. I also updated our version of fixed_table_rc.js to match that of https://github.com/meetselva/fixed-table-rows-cols/.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3146.


kkbrum (approved):
Cracking!

PR #3153, opened 2020-11-05
New user list filters

### LINKED ISSUE #46: Comm Panel -- new filters
- Number of class hours
- Whether student requested text messages
- Individual teaching and interviewing slots

(If you know how to add filters in general and you know how to query for something by hand, it's very easy to add a specific filter.  But you have to learn those two things first, which takes some effort.  So the "Easy" tag is conditional.)

### LINKED ISSUE #2376: Comm Panel Option for Second Day Teachers
This would be similar to the printables but would enable you to email only the teachers that are teaching on the second day of the program.

Requested by Stanford Splash.

kkbrum:
Fixes #46 and fixes #2376. 

This creates user list (list gen, comm panel, and group text) filters for:
- number of class hours of a student
- class times of a student
- teaching times of a teacher
- training/interview times of a teacher
- teachers of a particular student

Since you can now select teaching times of a teacher, this also solves the "filter for second day teachers" because you can select all the relevant time blocks for the 2nd day. 

Note that I did not test the group text panel as I don't have it set up, but it is the same as the comm panel and list gen, both of which I did test.

### DIFF
--- a/esp/esp/users/controllers/usersearch.py
+++ b/esp/esp/users/controllers/usersearch.py
@@ -46,8 +46,6 @@ from django.contrib.auth.models import Group
 
 import collections
 import re
-import operator
-
 
 class UserSearchController(object):
 
--- a/esp/esp/users/controllers/usersearch.py
+++ b/esp/esp/users/controllers/usersearch.py
@@ -35,7 +35,7 @@ from collections import defaultdict
 from esp.users.models import ESPUser, ZipCode, PersistentQueryFilter, Record
 from esp.middleware import ESPError
 from esp.utils.web import render_to_response
-from esp.program.models import Program, RegistrationType, StudentRegistration, ClassSubject
+from esp.program.models import Program, RegistrationType, StudentRegistration
 from esp.dbmail.models import MessageRequest
 from esp.utils.query_utils import nest_Q
 from esp.cal.models import EventType
--- a/esp/esp/users/controllers/usersearch.py
+++ b/esp/esp/users/controllers/usersearch.py
@@ -109,7 +109,6 @@ class UserSearchController(object):
 
             if 'class_times' in criteria:
                 class_times = criteria['class_times']
-                print(class_times)
                 if 'regtypes' in criteria:
                     student_verbs = criteria['regtypes']
                 else:
--- a/esp/esp/users/controllers/usersearch.py
+++ b/esp/esp/users/controllers/usersearch.py
@@ -231,7 +231,9 @@ class UserSearchController(object):
 
             if 'target_user' in criteria:
                 student_id = criteria['target_user']
-                if student_id:
+                if student_id == "invalid":
+                    raise ESPError('Please select a valid student whose teachers to email.', log=False)
+                else:
                     sections = [sr.section for sr in StudentRegistration.valid_objects().filter(user_id=student_id, relationship__name="Enrolled", section__parent_class__parent_program=program)]
                     if len(sections):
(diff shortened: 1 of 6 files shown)


esp/esp/users/controllers/usersearch.py
```diff
@@ -104,6 +107,29 @@ def query_from_criteria(self, user_type, criteria):
...
+                else:
+                    student_verbs = ['Enrolled']
+                times_filter = reduce(operator.or_, (Q(studentregistration__section__meeting_times__in=time, studentregistration__relationship__name__in=student_verbs)
+                                                     & nest_Q(StudentRegistration.is_valid_qobject(), 'studentregistration') for time in class_times))
+                Q_include &= times_filter
+                self.updated = True
+
+            if 'teaching_times' in criteria:
+                teaching_times = criteria['teaching_times']
+                times_filter = reduce(operator.or_, (Q(classsubject__sections__meeting_times__in=time) for time in teaching_times))
```
willgearty, line 123: Maybe I'm overlooking something here, but couldn't you just do `classsubject__sections__meeting_times__in=teaching_times`?

esp/esp/users/controllers/usersearch.py
```diff
@@ -104,6 +107,29 @@ def query_from_criteria(self, user_type, criteria):
...
+                Q_include &= times_filter
+                self.updated = True
+
+            if 'teacher_events' in criteria:
+                teacher_events = criteria['teacher_events']
+                events_filter = reduce(operator.or_, (Q(useravailability__event=event) for event in teacher_events))
```
willgearty, line 129: Same as above, could you just do `useravailability__event__in=teacher_events`

esp/templates/users/usersearch/list_selector.html
```diff
@@ -112,29 +118,62 @@ <h4 class="student"><a href="#">Grade</a></h4>
...
+                    {% endfor %}
+                </select></div>
+
+                <h4 class="student"><a href="#">Number of enrolled class blocks</a></h4>
+                <div>
+                Min number: <input type="text" size="3" name="hours_min" value="" /><br />
+                Max number: <input type="text" size="3" name="hours_max" value="" />
+                </div>
+
+                <h4 class="teacher"><a href="#">Student enrolled in class</a></h4>
```
willgearty, line 170: I think this could be a little more explicit that this filters to the teachers of classes in which students are enrolled

willgearty (changes requested):
Here are a few code comments. I'll actually test everything sometime today or tomorrow.

esp/esp/users/controllers/usersearch.py
willgearty: Also, I think it should be `meeting_times__id` since the `teaching_times` are event ids?

willgearty: OK, so at first I was like, "wait, how is this even working?". I tried running this line by myself using a test list of event integer ids, and it kept erroring (because `time` wasn't iterable). However, what I discovered is that `time` here is actually a string, so what happens is it treats the string as an iterable, looking for events with an id that matches each character of the string. This works fine if your events all have single digit ids, but once you get into larger ids, you're then filtering inadvertently to the single digit ids that make up the desired multidigit value. My two comments above should fix that. You'll want to do that here, above, and below.

willgearty (approved):
Works well for me (except for a related bug but I don't think it was caused by this PR)! Great job!

Issue #3145, opened 2020-10-31
Classes which are the wrong length scheduling check is broken


willgearty:
For some programs all of the classes are being listed as being the wrong length in the scheduling checks. I'm guessing this has to do with duration rounding?

Reported by Berkeley and Yale, but works for Stanford.

PR #3154, opened 2020-11-06
Fix rounding for wrong class length scheduling check

willgearty:
Since duration is now rounded to two decimal places, we need to round the calculated duration to two decimal places as well.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3145.


kkbrum (approved):
Glorious!

PR #1528, opened 2015-01-04
Improvements to StudentExtraCosts and DonationModule

pricem:
Joel did some work on this during the fall to address ticket #1327.  We would like to get this merged first before adding the other code changes he has for #1320 and #1319.  I followed up with some interface improvements to StudentExtraCosts, which were necessary for #1327 to be understandable to users, but also overdue in general.

The main change here is to add an "is_custom" boolean field to LineItemOptions.  If you check this in the admin interface, the students will be able to type in any dollar amount they please when they select that line item option.  

The main application of this is donations.  The donation module has also been cleaned up to use an actual Django form, although it is still possible to collect donations through the extracosts module.

Fixes #1327.

PR #3156, opened 2020-11-09
Add missing method to creditcardmodule

kkbrum:
Unfortunately it looks like #1528 had a missing method, breaking the credit card payment. I added it here and tested that it fixes the problem.

The error, reported by Yale, was:

File "/lu/sites/yale/esp/esp/program/modules/handlers/creditcardmodule_stripe.py", line 224, in charge_payment

AttributeError: 'CreditCardModule_Stripe' object has no attribute 'line_item_type'


willgearty (approved):
Whoops! Thanks for fixing this!

PR #3157, opened 2020-11-09
Allow partial financial aid when paying by credit card

### LINKED ISSUE #3159: Partial financial aid may break credit card payment
When a credit card payment is made, it attempts to link the payment to any transfers it paid for. In the process, there is a consistency check that the payment paid for a whole number of transfers (https://github.com/learning-unlimited/ESP-Website/blob/main/esp/esp/accounting/controllers.py#L620). However, with partial financial aid, this may not be the case, and then the credit card payment fails.

kkbrum:
Fixes #3159. This is a bit off-putting since I am removing a consistency check. 

The reason I think it's okay to remove this check is because this code simply links transfers to their payments, mostly by guessing. If someone is paying for a portion of a transfer, I think it's okay to link it to the payment that covered part of it (with the other part presumably being through partial financial aid). The other option would be to not link the transfer to the payment unless the payment covered the whole amount. But given that these links are mostly for admin convenience (I think?), I think that the conservative approach of linking partial payments is better than not linking anything.

Open to discussion!


kkbrum:
After further discussion with Will, we should probably do this the "right way," which would probably entail the following:
- allow the paid_in field to take multiple entries, thus allowing a transfer to be linked to both a financial aid and a payment transfer
- instead of assuming transfers with a paid_in entry have been fully paid for, actually sum the amounts of the transfers vs the payment amounts/ financial aid amounts to determine if there is a remaining balance on the transfer to be paid
- this would allow us to still do a consistency check at the end

kkbrum:
I believe this works, but I have NOT rigorously tested it yet

willgearty (approved):
This is fantastic, I think this is a really clean way to make this work. I was even able to manually add an item after paying and then pay a second time!

Issue #3148, opened 2020-11-03
"lottery preferences" value on big board include class enrollments


willgearty:
Not sure how I haven't noticed this before, but the big board reports the "lottery preferences entered", which I think implies the total number of priorities + the total number of stars. In reality, it's the total number of valid student registrations + the total number of stars:
https://github.com/learning-unlimited/ESP-Website/blob/417c66cb7b0c6148a60b4f26851debb2d1f5593c/esp/esp/program/modules/handlers/bigboardmodule.py#L149-L153

We probably want to filter the SRs to only priorities.

Reported by Northwestern.

PR #3160, opened 2020-11-12
Fix big board lottery preferences count

willgearty:
Before, we were counting all valid (unexpired) student registrations (oops). Now, we only count student registrations created by the class lottery or two-phase lottery modules.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3148.


kkbrum (approved):
Ace!

Issue #352, opened 2012-04-21
Account Management Page


jmoldow:
A few chapters have a QSD page which is identical to http://esp.mit.edu/myesp/accountmanage.html. In fact, all of the chapters using the MIT ESP template have a link to this page (although the QSD page itself doesn't exist on all of these sites; I recently created it on Clark's site).

This page has a few very useful links, and I think we should make it an actual page on the website. Since it wouldn't be a QSD page, it'd have to be /myesp/accountmanage or something similar.

PR #3163, opened 2020-11-13
Add account management page

willgearty:
This adds a uniform account management page to all themes (previously only the fruitsalad theme alluded to a QSD page that didn't even exist by default). I've added a new view at /myesp/accountmanage with a QSD section and the privacy policy:

image

I've also modified the login links for the various themes to link to this new page instead of the profile/password/etc pages.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/352 and fixes https://github.com/learning-unlimited/ESP-Website/issues/353


milescalabresi (approved):
One note: you need to recompile the theme for the changes to take effect since login_box is theme-controlled.

Issue #3164, opened 2020-11-16
Some list gen fields cause an error


willgearty:
When you select some list gen fields (I haven't checked exhaustively, but it looks like it might be the ones that are user-specific), it results in an error message when you submit the form. I think it may have been caused by the changes in https://github.com/learning-unlimited/ESP-Website/pull/3110.
![image](https://user-images.githubusercontent.com/7232514/99316288-b75c9400-2829-11eb-8f02-3cc2ad1f5617.png)

Doesn't occur on Yale's site, but occurred on my dev server while testing https://github.com/learning-unlimited/ESP-Website/pull/3153.

PR #3166, opened 2020-11-19
Fix list gen form

willgearty:
I forgot to update other `ListGenForm` calls when I added the `usertype` argument in https://github.com/learning-unlimited/ESP-Website/pull/3110. This fixes it so all fields can now be selected (again) in the list gen form.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3164.


kkbrum (approved):
Grand!

Issue #3142, opened 2020-10-27
Better UserSearchController groups filter


willgearty:
https://github.com/learning-unlimited/ESP-Website/pull/2223 added support for filtering users by user group. This should be expanded to:

- Support selecting multiple groups (see the multi-select in https://github.com/learning-unlimited/ESP-Website/pull/3001)
- Support excluding a single group or multiple groups

PR #3168, opened 2020-11-20
Include and exclude multiple user groups in USC

willgearty:
This modifies the user group filter in the user search controller to allow for admins to include or exclude users from multiple user groups. Additionally, I pulled out the filter accordion into it's own template so we never need to write the code twice ever again. I also fixed the logic of how exclusion queries were combined in the user search controller (before we were using `AND`s and now we use `OR`s). I guess this bug was never reported because no one ever used multiple exclusion queries before (e.g. exclude user groups and exclude students with fewer than 2 class blocks).

image

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3142.


kkbrum (approved):
Smashing!

Issue #3103, opened 2020-09-03
Volunteer registration form is missing help text


willgearty:
Looks like we have help text for most of the fields (and the help text for the comments field is even based on a program tag), but none of it is shown.

PR #3171, opened 2020-12-03
Fix volunteer reg form

willgearty:
This fixes the following for the volunteer registration form:

- Checkbox data was broken after Django 1.11 form widget changes (breaking the form's UI) (I checked and this field choice data extraction is not used anywhere else on the site)
- Previous offers loading was broken after Django 1.11 `values_list()` change
- Help text was missing for comments field (and help text for confirm checkbox was duplicated instead of being pulled from the form)
- Required fields were not marked in any way

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3103.


kkbrum (approved):
Ingenious!

Issue #3175, opened 2020-12-04
Auto-refresh button/form for scheduling checks


willgearty:
suggestion from @benjaminjkraft based on a suggestion from @emannes: I think with the new scheduling checks page it would be fairly easy to just add an auto-refresh button (with suitable warnings, of course), and then we could have the normal checks page, open to the appropriate check, and with auto-refresh turned on, projected somewhere.  Do people think that would be useful?

_Originally posted by @benjaminjkraft in https://github.com/learning-unlimited/ESP-Website/issues/1406#issuecomment-68613970_

PR #3185, opened 2020-12-11
Add autorefresh button to scheduling checks page

willgearty:
This adds a button to the scheduling checks page that, when clicked, causes all of the checks to refresh at some interval that is specified by the user (default is 30 seconds). When you click the button again, the refreshing stops.

image

image
(My dev server was turned off when I took these screenshots, hence the failed loading)

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3175.

### DIFF
--- a/esp/public/media/default_styles/scheduling_checks.css
+++ b/esp/public/media/default_styles/scheduling_checks.css
@@ -113,3 +113,14 @@ table {
 .alittlepspace {
 	padding-top: 8px;
 }
+
+input[name="refresh_interval"] {
+    width: 40px;
+    float: right;
+    margin-right: 10px;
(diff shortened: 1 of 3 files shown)


kkbrum (commented):
Looks good, but I'm not sure it's abundantly clear to a first time user what the number to the left of auto refresh is. Could you either add hover text to the input field to say that its the number of seconds between refreshes, or add the word "seconds" to the right of the input field? It's probably easy enough to guess the meaning, but I'd rather people didn't have to guess.

kkbrum (approved):
My bad, there is hover text, you just have to hover longer than I expected! Good to go!

Issue #3176, opened 2020-12-05
Add teacher filter to grid-based class changes


willgearty:
We should add teachers as an additional piece of info for each class (via checkbox) and should also make teachers a filterable field in the class search box.

Requested by @nrjean.

PR #3186, opened 2020-12-17
Add teacher filter and checkbox to grid based class changes

willgearty:
This adds a checkbox to the grid based class changes interface that allows users to add a list of teacher names to each class in the grid. It also adds teacher names as a possible filter for the search box.

image

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3176.


kkbrum (approved):
Masterly!

Issue #3179, opened 2020-12-06
"Hide past timeblocks" breaks grid based class changes columns


willgearty:
Looks like we only hide the old sections, instead of filtering them like we do with the category filters.

![image](https://user-images.githubusercontent.com/7232514/101286399-ec686080-37af-11eb-8575-0a41f8e5e4b7.png)

Reported by Cloud.

willgearty:
Looks like this is the one filter I didn't handle in https://github.com/learning-unlimited/ESP-Website/pull/3081:
https://github.com/learning-unlimited/ESP-Website/blob/2ceff5c26c966688a4d963d84a624a36a01f8cb3/esp/public/media/scripts/onsite/ajax_status.js#L924-L936

PR #3188, opened 2020-12-17
Fix hide past timeblocks option in grid based class changes

willgearty:
We now filter out both the timeblocks and sections based on whether they have started already (this includes classes that are still going on and classes in the student's schedule if a student is selected), then only render the remaining timeblocks and sections.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3179.

### DIFF
--- a/esp/public/media/scripts/onsite/ajax_status.js
+++ b/esp/public/media/scripts/onsite/ajax_status.js
@@ -834,21 +834,13 @@ function render_table(display_mode, student_id)
     var footers = $j(".timeslot_footers");
     var timeslots_ordered = Object.keys(data.timeslots)
     timeslots_ordered.sort((a, b) => (data.timeslots[a].startTimeMillis > data.timeslots[b].startTimeMillis) ? 1 : -1)
+    //exclude timeslots that have a start time 20 minutes prior to the current time (if we're not showing past timeslots)
+    if (settings.hide_past_time_blocks) {
+        timeslots_ordered = timeslots_ordered.filter(ts_id => (Math.floor((Date.now() - data.timeslots[ts_id].startTimeMillis)/60000)) < minMinutesToHideTimeSlot);
+    }
     for (var ts_id of timeslots_ordered)
     {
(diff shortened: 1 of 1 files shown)


kkbrum (approved):
Glorious!

BalaSubramaniam12007:
Hey @willgearty   @kkbrum 

I was exploring the "Hide past timeblocks" behaviour in the classchange_grid and noticed something that looks like an incomplete fix from this PR.

**What I observed**

`check_timeslots()`  returns `false` as soon as it finds *any* timeslot  that started more than 20 minutes ago:

``` js
if (differenceInMinutes > minMinutesToHideTimeSlot)
    return false;  // hides on ANY past timeslot
```
Real scenario:

1. It's 10:25 AM
2. A class runs 10:00–11:00 AND 11:00–12:00 (two blocks)
3. Block 1 (10:00 AM) is now 25 minutes old → past threshold
4. Block 2 (11:00 AM) hasn't started yet → still future
5. workbench.html sees block 1 is past → returns false > entire class disappears from the grid
6. Admin cannot see or modify a class that is actively running
7. Single-block classes are fine- they only have one timeslot, so "any past" = "all past". 

The bug only surfaces for multi-block classes.

I have a fix ready if this is confirmed as a bug. Happy to open a PR.

willgearty:
I don't know what workbench.html is (I don't believe we have a file with that name), but this is desired behavior. If a class starts in a past timeblock (but continues into present/future timeblock), we don't want to show that class since it has already started. Enrolling a student in a class that has already started means they wouldn't get the full experience and could end up interrupting the class. Therefore, if a user is filtering out past blocks, we don't want to show these multiblock classes.

BalaSubramaniam12007:
Thanks for the quick clarification that makes complete sense. A student enrolling mid-class would miss content and potentially disrupt the session, so hiding it entirely is the right call.

I also want to apologize for the sloppy write-up. 

Thanks again for taking the time to explain the intent behind the design.

Issue #3177, opened 2020-12-05
Big board and program-wide stats for check-in/attendance


willgearty:
The landing page for the attendance module is pretty bare right now:
![image](https://user-images.githubusercontent.com/7232514/101266953-52ee6f80-3719-11eb-9a2a-2625def50d2e.png)

We could probably spruce it up with some sort of plot of attendance/check-in over time. We could even include some sort of stats like those on the time slot-specific pages but for the entire program (beyond what is already included on the dashboard).

Requested by Cloud.

PR #3189, opened 2020-12-30
Attendance-over-time chart

willgearty:
This adds a plot of program and class attendance over time (by the hour) to the onsite attendance landing page. I've tested it with multi-hour classes and multi-week programs and it seems to work properly.

image

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3177.


kkbrum (approved):
Spectacular!

Issue #3183, opened 2020-12-10
Upgrade sorttable.js


willgearty:
Our current [sorttable.js](https://github.com/learning-unlimited/ESP-Website/blob/main/esp/public/media/scripts/sorttable.js) is from 2007. The internet has progressed a bit since then (most notably with HTML5), and it looks like [there's a fork](https://github.com/White-Tiger/sorttable.js) that has some cool benefits over the old version, including:

- Support for dynamic tables
- Support for multiple headers (?)
- Conforms to HTML5
- Reduced page-load footprint

We use this library in [quite a few places](https://github.com/learning-unlimited/ESP-Website/search?q=sorttable.js), so I'm making this it's own issue so that we replace them all at once.

PR #3190, opened 2021-01-06
Sorttable upgrade

willgearty:
This upgrades sorttable.js to a newer version (see https://github.com/White-Tiger/sorttable.js for benefits). The biggest benefit is the support of dynamic tables. I've also made some modifications to the code such that it always re-guesses the type (this is important for the dashboard table). I fixed all other uses of sorttable.js (most notably the HTML attribute for custom keys), and cleaned up some templates while I was at it.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3183.


willgearty:
The LGTM alerts are all related to the new sorttable.js code, so we should just ignore them.

kkbrum (approved):
Exceptional!

Issue #3147, opened 2020-11-02
Copy more stuff when using a template program


willgearty:
When we use a program as a template for a new program, we currently only copy the modules (and module settings with https://github.com/learning-unlimited/ESP-Website/pull/2703) and accounting information. Really, we should be copying the crmi, scrmi, and program tags, too. Basically, when a program is copied, it should work almost identically to the template program.

willgearty:
https://docs.djangoproject.com/en/1.11/topics/db/queries/#copying-model-instances

PR #3192, opened 2021-01-07
Copy crmi, scrmi, and tags from template programs

willgearty:
This builds on https://github.com/learning-unlimited/ESP-Website/issues/3147 and makes it so that the ClassRegModuleInfo settings, StudentClassRegModuleInfo settings, and all program tags are copied over from a template program to a new program. This means that a new program that uses another program as a template should function almost exactly (if not exactly) like the template program.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3147.


kkbrum (changes requested):
I'm getting this error. Is it possible that the sibling discount is a problem because it gets copied from the program tags AND from the program settings forms? It seems there is a field in each of them for sibling discount.

image

willgearty:
Whoops, should be fixed now!

esp/esp/program/views.py
```diff
@@ -536,13 +536,11 @@ def newprogram(request):
...
-                tags = Tag.objects.filter(content_type=ct, object_id=old_prog.id)
-                for tag in tags:
-                    tag.pk = None
-                    tag.id = None
-                    tag.target = new_prog
-                    tag.object_id = new_prog.id
-                    tag.save()
+                old_tags = Tag.objects.filter(content_type=ct, object_id=old_prog.id)
+                for old_tag in old_tags:
+                    new_tag, _ = Tag.objects.get_or_create(key=old_tag.key, content_type=ct, object_id=new_prog.id)
```
kkbrum, line 541: I now don't get an error, but I do think there's a bit of undesired behavior. The past program tag for sibling discount is overriding the tag created from the form for creating the new program, which means I enter a new value for the sibling discount in the program creation form but then it's really being set to the past value without telling me. I think instead of the "get_or_create", the code should check if that tag has already been created and only copy the old program's tag if it doesnt exist?  Not sure if this would negatively impact tags other than sibling discount though.

willgearty: OK, how about now @kkbrum?

kkbrum (approved):
Exceptional!

Issue #3178, opened 2020-12-05
Inconsistent survey question width


willgearty:
I'm not sure if this is theme specific, but on the normal survey interface (not the webapp interface), the questions have inconsistent widths:

![image](https://user-images.githubusercontent.com/7232514/101267138-479c4380-371b-11eb-890b-46818673f425.png)

Reported by Cloud.

willgearty:
I think these lines need to be included in the CSS that is used for that page:
https://github.com/learning-unlimited/ESP-Website/blob/4cedb38e1ae374fade08cec95e7148bfd72394c3/esp/public/media/default_styles/survey_manage.css#L1-L3

willgearty:
Ah yes, I see that is in the fruit salad less file:
https://github.com/learning-unlimited/ESP-Website/blob/dadcfbed66cb4ec522fafd137df48c280c6a63e5/esp/esp/themes/theme_data/fruitsalad/less/main.less#L652-L655

But not in any other theme less files.

willgearty:
In fact, I see that class used pretty often for tables, so that should probably just go into all themes.

PR #3193, opened 2021-01-07
Fullwidth fix

willgearty:
This adds CSS for all themes for the "fullwidth" class (this was previously only included in the fruitsalad theme but the class is used in lots of places). I then removed any fixes that had been implemented for one-off instances.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3178.


kkbrum (approved):
Wondrous!

Issue #2473, opened 2017-11-12
Cancel multiple sections at same time


willgearty:
It would be helpful, especially for classes with lots of sections, to be able to cancel multiple sections at once, perhaps with the same reason.
This is different from cancelling the class, because you might not want to cancel the entire class.

e.g. A teacher can no longer teach their classes on one of the days of a program (or, for one-day programs, can no longer teach in the afternoon).

PR #3194, opened 2021-01-09
Combine all cancel section forms into one

willgearty:
I've combined all of the cancel section forms into a single form so you can cancel multiple sections at the same time (using the same message and cancellation settings).

Fixes https://github.com/learning-unlimited/ESP-Website/issues/2473.

image


kkbrum (approved):
Flawless! (but this approval should probably be noted as pending until the tests run and pass?)

PR #3197, opened 2021-01-21
Fix plural names and capitalization

### LINKED ISSUE #2354: Fix pluralizations of models on admin page
Kind of related to #1321, but at least we should tell Django that (for example) the plural of "category" is "categories". I don't think Meta class changes would require database migrations or anything of that sort? Basically just do this: https://docs.djangoproject.com/en/1.8/ref/models/options/#verbose-name-plural

kkbrum:
Fixes #2354. While there, I also fixed inconsistent capitalization.


willgearty (approved):
All of the changes look good to me (and work fine on my dev server), thanks!

Issue #3181, opened 2020-12-06
Survey Results should be really be cached


nrjean:
See title - it takes over 1.5 minutes to load. Seems silly especially since most of the time survey result aren't coming in. Putting a button a the top that could refetch results and otherwise caching the page seems like a good solution to me.

nrjean:
Also if you have the debug tool bar enabled it take so long to load that I ran out of patience

PR #3201, opened 2021-01-29
Cache survey results

willgearty:
This caches the survey review pages for admins (html and pdf versions!), making them load a lot faster once they've been run once. The rendering for each survey is cached independently, and we only ever update the caches when new responses are saved (just like how each class on the catalog is cached independently).

Examples of speed up (on my dev server, using "CPU Time"):

- manage/one/two/surveys/review: ~20500 ms -> 500 ms (woohoo!)
- manage/one/two/surveys/review (only one of two surveys updated): ~20500 ms -> 9160 ms
- manage/one/two/surveys/review_pdf: hard to time since the debug toolbar is gone, but likely about the same speed up, if not faster because we aren't rendering the rest of the site

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3181.


willgearty:
And...looks like my current implementation breaks the individual teacher review functionality (it loads the same teacher regardless of who you pick).

willgearty:
OK, I've fixed the problem by mimicking the caching we do for the catalog (caching at the template level instead of the python level) to avoid any user-specific or form-specific parts. I'm still thinking about how to cache the page for teachers or for when we look up a teacher, but really those have always loaded pretty fast, so I'm not really sure we need to.

willgearty:
LGTM alerts are fine, that's how that file works (the unused imports make the caching work).

willgearty:
OK, I decided that we could do better. Now we cache each part of the survey results individually so we only refresh those parts when they have new relevant answers (the entire "individual questions" part is one part, and then each section is a separate part). So, for example, if a student fills out a survey for a particular section, only the cache for the part of the page that shows the results for that particular section gets refreshed.

Since this means we have a little more rendering to do each time the page is loaded, the **minimum time** to open the page is now a little longer, but the **average time** to open the page should be a lot shorter, especially during times when students are occasionally filling out surveys. This also means that we can cache the parts for teachers (and when we look up specific teachers), and the really cool part is that if a teacher has recently opened their page, the parts of the survey review for that teacher's sections will already be cached if an admin opens the full review (and vice versa).

Example times now:

- manage/one/two/surveys/review: ~25000 ms -> ~1500 ms
- manage/one/two/surveys/review (after submitting a section survey): ~1500 ms -> ~2300 ms
- manage/one/two/surveys/review_pdf: hard to time since the debug toolbar is gone, but likely about the same speed up, if not faster because we aren't rendering the rest of the site
- teach/one/two/surveys/review: ~1800 ms -> ~800 ms
- teach/one/two/surveys/review_pdf: again, hard to estimate, but seems about as fast as /review
- teach/one/two/onsitesurvey/review: ~3500 ms -> ~500 ms

Note: I believe the biggest component of loading the survey review page is creating and loading the histograms for numeric questions. The more histograms in a survey component, the longer that survey component will take to load if the cache is refreshed.

nrjean (approved):

PR #3204, opened 2021-02-11
Add more multiprogram statistics queries

### LINKED ISSUE #2922: Feature Request: Multiprogram Stats
Page that would allow admins to pull basic statistics from multiple program and output them.
This would allow admins to generate graphs of trends over multiple program without having to go to each program and run the statistics individually. Example output of this could look as follows: 

image

### LINKED ISSUE #3202: student big board check-in stat is wrong
The stat merely counts how many check-in records there have been for the program but doesn't take into account the fact that there can duplicates for individual users (now that the checkout module exists). I imagine all we need to do is add a distinct() call to the queryset.

willgearty:
This adds new queries to the /manage/statistics page to get multiprogram stats on teacher registration (# of teachers), class registration (# of classes), and student registration (# of students). Since these stats are basically already calculated for a single program for the big board modules, The results are shown in a neat table and in a graph (see screenshots below). I just made those static functions and call them for each of the selected programs. While I was at it, I made the form prettier, table-fied most of the outputs of the other queries, and fixed a minor bug on the student big board (#3202).

image
image
image

Fixes #2922 and fixes #3202

### DIFF
--- a/esp/esp/program/forms.py
+++ b/esp/esp/program/forms.py
@@ -180,6 +180,7 @@ class StatisticsQueryForm(forms.Form):
         ('hours', 'How many hours of class did the students take and when?'),
         ('student_reg', 'How many students registered?'),
         ('teacher_reg', 'How many teachers registered?'),
+        ('class_reg', 'How many classes were registered by teachers?'),
         #   (other queries here)
     )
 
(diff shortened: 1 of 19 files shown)


kkbrum:
I think this is all really cool! I am getting inconsistent results between the generated tables and my individual dashboards though. Could you check whether the table numbers are matching up with the dashboard for one of your programs?

Class registration:
- number of approved classes is too low (perhaps this is counting _scheduled_ classes?)
- number of class-student hours (both registered and approved) is too high 

Teachers: 
- number of teachers with approved classes is too low (perhaps counting _scheduled_ classes?)

Additionally, perhaps this is what's being counted, but I think having the second column type be scheduled numbers instead of approved numbers makes sense. Nearly all classes get approved, many fewer get scheduled, and that's what actually drives capacity and such so is arguably the most important. 

FInally, I dont understand what the meaning of the "__ matched by query" line is, so could we either remove that or make it more meaningful?

willgearty:
OK, I think I've fixed the discrepancies between the stats here and the teacher big board. Can you check and see if those problems are resolved for you, @kkbrum? Then I can go in and make the other changes you requested.

willgearty:
OK, I've added scheduled classes, etc. to the output for the teacher and classes queries. I also cleaned up the templates to make the query number a little clearer to the user. Let me know what you think @kkbrum!

kkbrum (approved):
Looks great! Doesn't quite match the dashboard, but that's a problem with the dashboard that I'll make an issue for. This looks great!

Issue #3161, opened 2020-11-12
Error pages are missing theme, settings, and current programs context


willgearty:
The errors that are handled here use the default django `render` function instead of our custom one that injects extra context:
https://github.com/learning-unlimited/ESP-Website/blob/89d7312e1b5ab4ce885314df48e80a73c024fda8/esp/esp/middleware/esperrormiddleware.py#L156-L212

willgearty:
Note: this is not for the "Whoops!" page or the "Error 404" page.

willgearty:
It looks like the the error page that you get when you search for a nonexistent user has this problem ("No user found by that name!" error page), so that seems like a reproducible example for testing.

PR #3212, opened 2021-02-19
Use custom render function to render error pages

willgearty:
This fixes error pages such that they properly display theme content, the admin toolbar, etc. It's often hard to intentionally produce one of these errors, but if you search for a user that doesn't exist, it should trigger it (for whoever reviews this, I would suggest viewing the page before and after checking out this PR).

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3161.


kkbrum (approved):
Flawless!

PR #3214, opened 2021-02-19
Fix top margin of catalog sorted by category

### LINKED ISSUE #3029: PDF catalog needs bigger margins
It appears the default catalog (sorted by category) has basically no margins, but other sorting methods appear to have margins (although they are a little small).

Also unofficially requested by Yale and Stanford for years.

kkbrum:
Fixes #3029. 

Looks like the geometry top margin didn't like the cm! The catalog sorted by category now matches the margins when sorted by timeblock.


willgearty (approved):
Looks good! Silly metric system!

Issue #3210, opened 2021-02-12
grid-based class changes uses old jquery-ui


willgearty:
We're currently using version 1.12.1 of jquery-ui on the rest of the site, but on the grid-based class changes page we use an older version. We should upgrade (although it looks like it will take some work).

PR #3225, opened 2021-02-22
Upgrade jquery-ui for grid based class changes

willgearty:
I thought this would be a much larger project than it ended up being. I only needed to fix one thing that was changed between jquery-ui 1.9 and 1.12 (`data("autocomplete")` -> `data("ui-autocomplete")`). The only visual change that I can see is that the student that is hovered over in the student dropdown is now highlighted in blue.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3210.


kkbrum (approved):
Lovely!

PR #3226, opened 2021-02-25
Changed subject to required

### LINKED ISSUE #3216: Subject should be required in comm panel
If you send an email through the comm panel without a subject, the sending fails. We.shouod make the subject field required to prevent this.

laurenmaclachlan:
Fixes #3216

Made Subject field required when sending emails through comm panel


hwatheod (approved):

hwatheod:
In the future, please give a better commit message than "Update step2.html". The commit message should tell the reader why you are updating the file.  I changed it to "Make subject field required in comm panel" in the final squash.

Issue #3023, opened 2020-06-21
Add UI for line item types


milescalabresi:
It would make extra costs (a.k.a. line items) such as lunch, shirts, photos, etc. more accessible if there were a UI where admins could add/remove select them. I think it'd be possible to recycle the code used for the resources module

willgearty:
I think @kkbrum volunteered to make a mockup for this. I'd be happy to implement after that.

kkbrum:
I agree with Miles that this could have the same layout as resources. Here's a mockup of the fields it would have, with some notes about line items in italics!

**Import line items from a previous program**
Program dropdown
_(I would exclude the auto-generated line items from those that get imported. Those are sibling discount, program admission, financial aid grant, and student payment. These all should not be modified via this page (sibling discount can be modified through program settings or tag settings pages currently, and the rest should not be changed))_

**Add/modify a line item** 
Name:
Cost:
Required checkbox (help text: Should this line item be automatically added for every student?)
Max quantity:
_(I would not include the “for payments” or “for finaid” checkboxes from the backend, as they are not meant to be checked for user-added line items.)_
Options: _(Same remove/ add option framework as with resources)_
	For each option you want to add:
		Description: 
		Cost: (help text: If blank, takes overall cost)
		Custom cost checkbox (help text: Should the student be allowed to specify a custom amount for this option?)

PR #3227, opened 2021-02-26
Frontend interface for line items

willgearty:
This adds a user interface in the frontend for admins to add, remove, edit, and import line items. It functions pretty similarly to the resources page.

image

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3023.


kkbrum (changes requested):
I'm having a really hard time reviewing this due to my devserver being bizarrely slow today so this review is a work in progress, but my first comment:

There's probably an extra/ missing div somewhere causing the bottom of the page to be whacky:

image

esp/esp/program/modules/forms/lineitems.py
```diff
@@ -0,0 +1,52 @@
...
+    }
+)
+
+class LineItemImportForm(forms.Form):
+    program = forms.ModelChoiceField(queryset=None)
+
+    def __init__(self, *args, **kwargs):
+        cur_prog = kwargs.pop('cur_prog', None)
+        super(LineItemImportForm, self).__init__(*args, **kwargs)
+        progs = LineItemType.objects.exclude(text__in=["Sibling discount", "Program admission", "Financial aid grant", "Student payment"]).values_list('program', flat = True).distinct()
```
kkbrum, line 48: There's also a "Donation to Learning Unlimited" line item that's autogenerated by the donation module and unfortunately takes tag inputs for the amount options, not line item edits (this should really change) so that line item should also get excluded.

Additionally, there are also "Saturday Lunch" and "Sunday Lunch" line items generated by the SplashInfoCosts module that take tag inputs, so these should also be excluded.

willgearty:
I've added a few more line item types to exclude, and I fixed the templates!

kkbrum:
Great, those changes look good. One more comment:

The line item form lets you add line items that have the same name as an existing line item, but then on the extracosts page, only one line item per unique name shows up. I looked at the classrooms form, and if you try to add a classroom with the same name as a current classroom, it updates that classroom instead of creating a new one. Perhaps this is the behavior we want to adhere to in the line item form? Or an error stating that line items must have unique names?

In a similar vein, perhaps we want to give an error if someone tries to name a line item with the name of one of the "excluded" types. (As creating a line item with the name "Program admission", for example, really does cause various problems, if I recall correctly.)

willgearty:
OK, we now raise an error in both cases (duplicate names or matches an excluded name).

kkbrum (approved):
Spectacular!

Issue #246, opened 2012-01-16
logical "groups" of time blocks


jmoldow:
Currently, the schedule is broken into "groups" of time blocks if two adjacent time blocks are separated by more than 15 minutes (or maybe it's a different number, I don't recall). There should be a setting to allow you to manually group the timeblocks, so that you can get two groups (e.g., morning and afternoon) without having to separate two time blocks by more than 15 minutes.

willgearty:
I think the best first step would be to make the "15 minutes separation" part tag-based. Then, if we really want to be fancy, we could add a "group" field to time blocks and use those if they aren't NULL (and you would probably want that to be integrated with the time block resource page.

PR #3231, opened 2021-03-01
Contiguous blocks tolerance tags

willgearty:
This adds two global tags which allow for admins to adjust how tolerant we are of classes' contiguousness (in minutes) for visual purposes ('availability_group_tolerance') and for functional/scheduling purposes ('timeblock_contiguous_tolerance'). I imagine the former will be more useful, but I guess there are some weird edge cases for the latter, so I included it.

I also made a little fix to the tag settings forms where if you saved the form with a blank answer, it would show as blank after the form was submitted, even though the website would then be using the default for the tag (so now we show that default after successful form submission).

First step towards fixing https://github.com/learning-unlimited/ESP-Website/issues/246

### DIFF
--- a/esp/esp/cal/models.py
+++ b/esp/esp/cal/models.py
@@ -141,9 +141,9 @@ class Event(models.Model):
         return newList
 
     @staticmethod
-    def contiguous(event1, event2):
-        """ Returns true if the second argument is less than 20 minutes apart from the first one. """
-        tol = timedelta(minutes=20)
+    def contiguous(event1, event2, tol = 20):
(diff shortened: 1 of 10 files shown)


kkbrum (commented):
This is so much fun to use! Love it. 

I personally expected to find these tags under program tags though and think it might make more sense. One example of why it might be a program tag instead of global is that the timeblock setup for Splash vs Sprout at Yale is very different and one uses a larger chunk of campus requiring more time between things, etc. so I can see why they might set different values for the programs.

willgearty:
Good point! I've converted them to program tags!

kkbrum (approved):
Splendid!

Issue #2584, opened 2018-06-21
Default teacher availability to none?


willgearty:
Based on our most recent scheduling for Stanford Splash, it appeared that some people were indicating the complete opposite availability with the new layout (#2419). despite the presence of what I think is a pretty clear legend/key.
My best guess is that the teachers clicked the timeslots that they were available for, but since they are by default available for all timeslots, this actually marked them as unavailable for those particular timeslots.
Therefore, perhaps teacher availability should function similar to volunteer registration where all timeslots are marked as unavailable by default and teachers must mark which timeslots they are available?
My one concern with such a change is that we will get less availability in general from teachers, which is presumably why we originally defaulted to completely available?

PR #3232, opened 2021-03-02
Make default availability none for teacher availability form

willgearty:
This changes the default behavior for the teacher availability form from all timeblocks available to all timeblocks not available. I believe this should increase availability accuracy, since the current default behavior was opposite of how our other forms work (e.g. volunteer signup) and how other forms on the internet work (e.g. whenisgood, when2meet).

It should be noted that teachers have always had no availability by default if the availability module is enabled, but previously we just highlighted all of the timeblocks in the form if they hadn't filled it out before. This change only affects the default behavior of the form itself.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/2584.

### DIFF
--- a/esp/esp/program/modules/handlers/availabilitymodule.py
+++ b/esp/esp/program/modules/handlers/availabilitymodule.py
@@ -192,14 +192,6 @@ class AvailabilityModule(ProgramModuleObj):
 
         #   Show new form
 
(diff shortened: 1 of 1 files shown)


kkbrum (approved):
Magnificent!

Issue #1406, opened 2014-10-17
Scheduling check info in scheduler


benjaminjkraft:
@emannes wrote:

> There's a lot of information that's useful for manual scheduling that shows up in the scheduling checks page but isn't in a useful format for people who are trying to decide where/when to put a class.

For example, we could have how many classes (of a subject/grade) in a timeblock listed at the top somewhere -- maybe when you pick up a class it could show classes and capacity in that timeblock for all classes, that category, and overlapping grades.

> This doesn't have to be integrated into the AJAX scheduler--it could be a separate webpage and be projected onto a screen in the office so that everyone can see it.

It would also be useful to have integrated info about other checks, like teachers who like running (e.g., show where else a teacher is teaching so they can be scheduled somewhere near by) and hungry teachers (highlight if you're about to schedule a teacher for both blocks of lunch.

benjaminjkraft:
++, although I think that will be more useful for some checks than others.

Something more along the lines of @emannes's original suggestion: I think with the new scheduling checks page it would be fairly easy to just add an auto-refresh button (with suitable warnings, of course), and then we could have the normal checks page, open to the appropriate check, and with auto-refresh turned on, projected somewhere.  Do people think that would be useful?

PR #3235, opened 2021-03-06
Add scheduling checks to ajax scheduler

willgearty:
This adds a set of scheduling checks to the ajax scheduler that can be enabled by selecting them in the new options tab (only one can be enabled at a time):
image

Once a scheduling check is enabled, it will be used to color the cells in the scheduling matrix. For example, for the "Double-Booked Teachers" check, it will color sections red that overlap in time and have the same teacher(s):
image

This functionality works for looking at sections that are already scheduled and for scheduling sections in new spots (e.g. if the capacity check is enabled, the color of the cell will change as you move your cursor over different cells in the matrix based on the capacities of the classrooms). This means that in addition to using these checks AFTER scheduling, you can also use them DURING scheduling to see if adding/moving a section in a particular spot will cause or remove an issue (e.g. will make a teacher "hungry" or have to "run"). You can see this in this example screenshot where I am moving the B6s1 section:
image

Then, once a section is scheduled, we update the colors of ALL cells to indicate if new issues have arisen. Note: I'm not sure how the performance is for this last part for larger programs, but I imagine it shouldn't be too bad since javascript is pretty efficient, and it's not like it will prevent you from scheduling other sections.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/1406.

### DIFF
--- a/esp/public/media/scripts/ajaxschedulingmodule/ESP/Cell.js
+++ b/esp/public/media/scripts/ajaxschedulingmodule/ESP/Cell.js
@@ -234,7 +234,8 @@ function Cell(el, section, room_id, timeslot_id, matrix) {
                 teacher_loop:
                     for(var teacher of section.teacher_data){
                         for(var sec of teacher.sections){
-                            if(sec == section.id || (sec in this.matrix.sections.scheduleAssignments && this.matrix.sections.scheduleAssignments[sec].room_id == scheduleAssignment.room_id)) continue;
+                            if(sec == section.id || !(sec in this.matrix.sections.scheduleAssignments)) continue;
+                            if(sec in this.matrix.sections.scheduleAssignments && this.matrix.sections.scheduleAssignments[sec].room_id == scheduleAssignment.room_id) continue;
                             for(var ts1 of this.matrix.sections.scheduleAssignments[sec].timeslots){
                                 for(var ts2 of scheduleAssignment.timeslots){
                                     if(ts1==ts2) continue;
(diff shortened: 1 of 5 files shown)


esp/public/media/scripts/ajaxschedulingmodule/ESP/Cell.js
```diff
@@ -96,13 +100,175 @@ function Cell(el, section, room_id, timeslot_id, matrix) {
...
+                                        continue teacher_loop;
+                                    }
+                                }
+                            }
+                        }
+                }
+                if(n_teachers == 0) return grey;
+                return this.cellColors.HSLToRGB(0,100,100 - 10 * n_teachers); // Color red based on number of teachers
+            case "double_booked":
+                // Count how many teachers are teaching other sections at the same time as this one
```
kkbrum, line 214: is this referring to coteachers of a class? I think that could be more clear if so.

esp/public/media/scripts/ajaxschedulingmodule/ESP/Cell.js
```diff
@@ -96,13 +100,175 @@ function Cell(el, section, room_id, timeslot_id, matrix) {
...
+                                    continue teacher_loop;
+                                }
+                            }
+                        }
+                        if(n_teachers == 5) break teacher_loop;
+                    }
+                if(n_teachers == 0) return grey;
+                return this.cellColors.HSLToRGB(0,100,100 - 10 * n_teachers); // Color red based on number of teachers
+            case "running":
+                // Count how many teachers are teaching other sections immediately before or after this one in a different room
```
kkbrum, line 232: ditto

kkbrum (changes requested):
Really cool feature addition.

Not sure what's going on, but for both the "running teachers" and "double-booked teachers", one of my classes turns black and is no longer clickable: 

image

willgearty:
Can you take a screenshot of the console log when that happens @kkbrum?

kkbrum (approved):
A little hard to test with my limited classes, but it looks good and I'm excited to see it in action!

Issue #3028, opened 2020-06-23
Individual teacher schedules and printables teacher schedules are different


willgearty:
The individual teacher schedules show unscheduled classes whereas the printables version does not.

Encountered by @kkbrum on dev3.

PR #3237, opened 2021-03-08
Unify individual and program printable teacher schedules

willgearty:
This changes the individual teacher schedule printable (usually accessed from the userview page) to only show scheduled classes like the teacher schedules program printable.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3028.


kkbrum (approved):
Awesome!

Issue #3108, opened 2020-09-11
Floating resources included in scheduling checks


willgearty:
From @kkbrum while testing #3100 on dev3:

> Looks like floating resources are being included in the “mismatched rooms and meeting times” box, so if a class has floating resources assigned to it, it has extra resource events listed and is claimed to be a mismatch
> The same problem occurs for the “classes not completely scheduled or with gaps”

PR #3238, opened 2021-03-09
Fix scheduling checks affected by floating resources

willgearty:
This fixes the "Mismatched rooms and meeting times" and "Classes not completely scheduled or with gaps" scheduling checks that were broken when sections had been assigned floating resources (e.g. through the teacher check-in module). Now we only look at the classroom assignments, instead of all resource assignments, to determine if there are mismatches or the classes are not completely scheduled.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3108.


kkbrum (approved):
Brilliant!

Issue #3236, opened 2021-03-06
Scheduler availability wrong when only one lunch block in a day


willgearty:
When there is only a single lunch timeblock in a given day, the scheduler shows the timeblock as green (available) when scheduling a section. However, when you then try to schedule the section there, it just silently fails. It probably shouldn't be marked as available, since you wouldn't want to schedule a section over the only lunch timeslot. Note: this behavior appears to work properly when there is more than one lunch timeblock in a given day.

PR #3239, opened 2021-03-09
Fix ajax scheduler highlighting for single-block lunches

willgearty:
Previously, if you had a single lunch on a particular day, the scheduler would show this timeblock as available for single-block sections. However, if you tried to schedule one of these sections in this timeblock, it would just fail silently. I realized this is actually because we weren't doing any lunch constraint checks for single-block classes, which I've fixed. Now, if there is a single lunch block, it is marked as available but not schedulable (green with slanted grey lines) for single-block sections, as should be desired.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3236.


kkbrum (approved):
Wonderful!

Issue #2847, opened 2019-11-20
Barcode checkin for teachers


willgearty:
Request from Stanford:

"Potentially a barcode-type system for teacher check-in too. Caveats would be that 1) we’d need to be more careful about sorting teacher schedules and 2) make sure there’s a schedule/barcode for Sunday, for teachers teaching on both days"

PR #3240, opened 2021-03-12
Teacher barcode check-in

willgearty:
This adds barcodes to teacher schedules (and makes them a bit prettier) and then adds the ability to scan those barcodes with physical scanners or the user's device on the teacher check-in page. I also made the teacher check-in page a little friendlier for teachers that might be teaching multiple classes at once (since the ajax scheduler now allows this via the override).

@milescalabresi has physical scanners, so he could test the functionality for that.

And also, as @kkbrum mentioned during development, this will likely be a major component of the admin webapp, but this PR will need to suffice until then.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/2847.


willgearty:
OK, that should fix the LGTM problems that we actually need to worry about.

kkbrum:
Okay so using the teacher schedules, although I only have 4 data points, the single digit user IDs scan whereas the double digit user IDs don't (perhaps this is correlation and not causation though).  Using the teacher nametags with barcodes on the back, they all scan. Looking at the difference between the barcodes on the schedules vs the nametags, the only difference I see is that the bars on the nametags are thicker?

For user ID 4, username "tteacher@" (for the sake of testing weird characters), I get the following console error when scanning:
image

willgearty:
OK, I've tested the new code with barcodes up to 4 digits and they work fine now. It should also now handle weird usernames.

kkbrum (approved):
Flawless!

Issue #3025, opened 2020-06-23
Graphical Update to Onsite


nrjean:
Personally I think its sorta ugly and could do with a visual update - just so it isn't a list. Make it easier to read and navigate.

PR #3252, opened 2021-04-09
Made onsite buttons

nrjean:
Makes onsite links into buttons - also introduces a class called makeButtonLink which can be called to make a link that is a button. 
Should be useful if anyone is making a new page and wants to make a link, that is you know, a button.

If anyone has any comments on how I could make the onsite page prettyier that would be great.


nrjean:
Fixes #3025 sorta

hwatheod:
There are 2 new functions called makeButtonLink but it's only called once in the template - which one is being called and what is the other one for? (Or somehow both are being called depending on `module` ?)

nrjean:
So the first function in modules/base.py is a global function that works just like makeLink but when it makes the link it makes it a link that a button. It makes just one link per module.

In handlers/onsiteclasslist.py multiple links need to be generated for that module so is has an alternate function which essentially calls makeButtonLink twice for each link that is needed.

@willgearty can probably explain the technical of this better lol

esp/templates/program/modules/onsitecore/mainpage.html
```diff
@@ -18,20 +18,23 @@ <h1>Onsite Portal for {{program.niceName}} </h1>
...
+
 
 {% for module in modules %}
 {% if module.isStep %}
   {% if not module.useTemplate %}
-    <li>{% autoescape off %}{{ module.makeLink }}{% endautoescape %}</li>
+    {% autoescape off %}{{ module.makeButtonLink }}{% endautoescape %}<br>
   {% else %}
-    <li><em>{{ module.module.link_title }} (below)</em></li>
+    <<em>{{ module.module.link_title }} (below)</em><br>
```
hwatheod, line 32: There shouldn't be two `<` signs here

hwatheod:
> So the first function in modules/base.py is a global function that works just like makeLink but when it makes the link it makes it a link that a button. It makes just one link per module.
> 
> In handlers/onsiteclasslist.py multiple links need to be generated for that module so is has an alternate function which essentially calls makeButtonLink twice for each link that is needed.

Oh I see, the first one is used for most of the onsite modules, and the second one in a subclass which overrides the first, and is used for one particular special module.  Got it.

willgearty:
Code looks good (aside from the comments from @hwatheod). However, the page currently looks like this:
image

This doesn't seem much better than the links from before, in my opinion. But, I think if you include the "management.css" file you'll be much closer, since it should make the buttons look somewhat like the ones on the main management page:
image

But you'll probably still want to play around with it to make sure they have the same heights, etc.

nrjean:
fixed a teds changes and also make it look better by organizing the buttons yay @willgearty

willgearty (approved):
Code looks good to me, and it works well on my dev server! What a difference it makes!

PR #3254, opened 2021-04-09
Centers Bigpicture

### LINKED ISSUE #3020: Bigpicture theme content isn't centered
Centering includes the Navbar. Could be debatable but I don't think it should. The page looks fine as long as the nav bar is in view, but as soon as you scroll down (say to read the catalog) the whole webpage is effectively off center.

nrjean:
Does what is say - redefines the width of the container to 1180 pixels which centers the content area of big picture.
Fixes #3020


willgearty (changes requested):
Looks like a different width is defined in theme_compiled.css. I'd suggest changing the LESS file that defines the bigpicture width for the container class instead of hard coding it into the template here.

willgearty:
Yep, looks good on my dev server (and on Cloud), just need the change requested above.

nrjean:
all done

Issue #3230, opened 2021-03-01
Weird characters in names cause email failure


willgearty:
It looks like certain characters in a user's name can cause emails to fail bring sent. Specifically, Columbia reports that a user with an apostrophe in their name caused an email to fail as part of a class student email list.

Reported by Columbia Splash.

hwatheod:
Looked through the server logs and found a similar error for an email sent from Yale Splash on 2021-02-19. The student's name had an accented a (á) in it, and the exception thrown was `BadRequestsError()`.  This is an exception thrown by the SendGrid Python client. The send_to address was of the form `name <email@example.com>` where the name part had the á in it.

hwatheod:
We shouldn't have to do the encoding ourselves. This ticket is probably more relevant: https://code.djangoproject.com/ticket/19107 but we are on Python version 2.7.3 which is later than the 2.6 versions mentioned in the ticket

hwatheod:
I changed my name on dev3 to á in the name, and requested a password reset email, and it sent fine. The á showed up in the name. Maybe it's some issue specific to the comm panel but the comm panel wasn't working for me on dev3.

hwatheod:
When I changed my name to have a comma in it, then the password reset email failed with a "Whoops" error. I see similar errors in the server logs for production sites too.

willgearty:
Thanks, @hwatheod! Looks like we want to be encoding the to/from names whenever we send a request off to SendGrid.

hwatheod:
The problem appears to be in the [_parse_email_address function](https://github.com/sklarsa/django-sendgrid-v5/blob/master/sendgrid_backend/mail.py) in the django-sendgrid-v5 library.

```
In [15]: sg = sendgrid_backend.mail.SendgridBackend()

In [16]: sg._parse_email_address(u'John Doe <john_doe@example.com>')   # name doesn't have a comma
Out[16]: (u'john_doe@example.com', u'John Doe')

In [17]: sg._parse_email_address(u'John, Doe <john_doe@example.com>')  # name has a comma
Out[17]: (u'John', None)
```

hwatheod:
Confirmed this by changing my first name on dev3 to `Ted, comma. \"period` and then requesting a password reset email. The email was received successfully.

PR #3255, opened 2021-04-12
fix some logging issues in send_email_requests()

hwatheod:
While investigating #3230, I found that the logging code itself was throwing exceptions (if the email address contains a non-ascii character), and also sometimes the exception logged was an empty string (I guess because SendGrid didn't implement `__str__` in its exceptions). This fixes both of these issues.

Note that this does *not* fix the issue of #3230 itself. That is still under investigation.


willgearty (approved):
Looks good to me, thanks!

PR #3260, opened 2021-04-15
Fix bug for ranks beyond first choice

padajar:
Current list comprehension overwrites priority value `i` set in the for loop in line 462, changing it to 1. Change list comprehension variable to properly assign all ranks.


willgearty (approved):
Looks good, thanks for finding this and fixing it!

Issue #2082, opened 2016-03-16
Cache needs to be flushed after scheduling classes to see timeslot capacities


bhorowitz:
When scheduling classes, the dashboard doesn't show correct time slot capacities unless the cache is flushed. This is a bit annoying when trying to properly balance the time slot capacities.

mgersh:
I recommend using the "Total capacity in each block by grade" scheduling check for this--scheduling checks are now much quicker to refresh than the dashboard. (Although the caching bug is still a bug, of course.)

Issue #827, opened 2013-11-25
Dashboard attended number wasn't updating


benjaminjkraft:
I think this is a caching bug; when we were checking in students via barcode, the records were getting made, but the number on the dashboard wasn't increasing.  Before we checked the admin panel, this made us think it wasn't working.  My guess is that this is a caching bug.

PR #3263, opened 2021-04-18
Fix caching for dashboard stats

willgearty:
The dashboard has notoriously been bad at showing up-to-date stats for a program. This is because the caching dependencies were never updated as more and more stats were added, meaning it would only update for a small number of database changes (or if the cache was flushed). Now, each component of the dashboard stats has its own cache dependencies, which should both speed up stat calculations overall and make sure the stats are always up-to-date. This should also help encourage similar cache dependency additions when new stats are added in the future.

For the reviewer, I've tested most of the dependencies on my dev server, but it would be good to have someone else do a second check and also to visually check if I've included all necessary cache dependency for each stat category. I didn't have every module enabled, so it's entirely possible I missed an important dependency for some list of teachers or students.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/827 and fixes https://github.com/learning-unlimited/ESP-Website/issues/2082.


willgearty:
I've also added some fixes to a problem that was recently discovered on Yale's site. Basically, we had the DerivedField dependent on ALL records, which meant we were updating the field for ALL sections for ALL programs every time any record was added. On a development server, this isn't a huge deal because we don't have that many sections, but on a production server (especially for an older chapter like Yale), it results in thousands of calculations which causes each record to take nearly a minute to save. The only real solution seems to be to remove the DerivedField entirely (RIP). We only used it in the jsondatamodule for calculating dashboard stats, which is now cached better in this PR, so now we just do the calculations in a for loop there.

kkbrum (approved):
Spectacular! Who knew the dashboard could be up to date!

Issue #3251, opened 2021-04-05
Nametag USC issues


willgearty:
From @kkbrum on #3249:

> If I do a combination AUL for nametag generation, and press "continue", I get "You need to select the TYPE of Name Tag to print. (students,teachers,etc)". If I press "generate tags" without pressing "continue" first, I get a server error

Issue #3248, opened 2021-03-30
"Teachers of a student" filter is broken for multiple printables that use the USC


willgearty:
Several of the printables use the USC, but they weren't updated for #3153, which adds a new form to the context for searching for a student to get the teachers of that student. I previously had to fix this for the nametags page: https://github.com/learning-unlimited/ESP-Website/commit/0a75bc31b584742224433ee87b6378f647766753. Is there a way for us to ensure it always gets included for pages that are using the USC?

kkbrum:
I made some partial fixes on the moderators branch. I added the student search form to the context generated by "prepare_context" that was already being called for all the USC instances. This means people won't have to manually import the studentSearchForm and add it to the context. However, people will need to use ListGenModule.processPost and not process the post dictionary manually, but that was pretty much already the case. 

However, there are a couple of problems:

- [ ] in commit e051835, I added a TODO, because for some reason i wasnt able to import the ListGenModule into the usersearch.py and had to copy the process post function (not ideal). No idea why I can't import it
- [ ] The form for the teachers of a student filter only works for base lists, not for combination lists. I'm not sure why yet.

willgearty:
For the form not working in the combo lists, it looks like it is because the javascript associated with the field uses the ID of the field, which should be unique, so it only works on the first instance of the form on the page:
https://github.com/learning-unlimited/ESP-Website/blob/98501776a014a01eeba8c8105ee2aca2a7d7bee3/esp/esp/db/forms.py#L61-L97

PR #3264, opened 2021-04-21
Fix nametag USC combo selector

willgearty:
This fixes the combo selector for the AUL/USC user selector in the nametag module. In the process, I also fixed the "teachers of a student" form field by removing one instance of the filter accordion and just moving the remaining accordion between the two tabs. This way we don't have any duplicate IDs and everything should just work (TM).

I also made it so that all filters are now cleared, rather than just those in a list, since that's gotten us into trouble multiple times in the past.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3251 and fixes https://github.com/learning-unlimited/ESP-Website/issues/3248.


kkbrum (approved):
Amazing!

Issue #3246, opened 2021-03-26
Sorting options for class list in scheduler


willgearty:
I think it'd be useful, especially for larger programs, if we could sort the list of sections ("Directory") in the ajax scheduler based on section details. Some potential sort factors that come to mind include:

- Category
- ID
- Teacher availability
- Capacity/num students
- Length

willgearty:
And also sort options for moderators.

PR #3266, opened 2021-04-22
Sort classes and moderators in scheduler

willgearty:
This makes classes and moderators sortable in the AJAX scheduler:

1. The class directory now has an option below the search box that allows the user to filter by different fields (ID, category, length, capacity, and teacher availability). This can be in ascending or descending order.
2. The moderator directory is now sortable using sorttable.js. Clicking a column header sorts the table by that column.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3246.

### DIFF
--- a/esp/public/media/default_styles/scheduling.css
+++ b/esp/public/media/default_styles/scheduling.css
@@ -8,6 +8,10 @@ a {
     cursor: pointer;
 }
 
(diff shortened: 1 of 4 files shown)


kkbrum (approved):
Incredible!

Issue #3243, opened 2021-03-22
Scheduler doesn't list sections with floating resources


willgearty:
When we collect the section schedule assignments, we skip any sections that have _only_ floating resource assignments:
https://github.com/learning-unlimited/ESP-Website/blob/453dfef8bd34542412043e3f0edff8ca1608fe51/esp/esp/program/modules/handlers/jsondatamodule.py#L145-L156

This isn't a problem for already scheduled sections, since they also have classroom assignments, but if a floating resource is assigned to a section before it is scheduled, it will not show up in the class directory in the ajax scheduler because it thinks it is rejected:
https://github.com/learning-unlimited/ESP-Website/blob/66e9a2d94c52f9969d28f856232ec3e6fa0c6e55/esp/public/media/scripts/ajaxschedulingmodule/ESP/Sections.js#L171-L174

PR #3270, opened 2021-04-23
Fix sections with floating resources in scheduler

willgearty:
This makes it so that floating resources are never considered when populating the scheduling data for the scheduler. There wasn't really a way to fix the raw SQL query we were making, so I just removed it and replaced it with a normal django query. Hopefully this doesn't cause too much of an increase in loading time for the scheduler for larger programs. I also made it so that we check that there is a scheduled room before we try to put a section in the matrix upon loading the scheduler (because some sections, e.g. lunches, have times but not rooms). Finally, we now check the actual status of the section, not the scheduling status, to exclude rejected and cancelled classes.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3243


kkbrum (approved):
Looks good to me. It does take a noticeable amount of time to load the Splash Fall 13 scheduler on dev3 (maybe around 30 seconds), but I think it already took a noticeable amount of time to begin with, so I think it's probably fine.

Issue #3244, opened 2021-03-23
Option in student lottery to not set permissions


willgearty:
Right now the student lottery always opens new permissions for the winners:
https://github.com/learning-unlimited/ESP-Website/blob/98501776a014a01eeba8c8105ee2aca2a7d7bee3/esp/esp/program/modules/handlers/studentregphasezeromanage.py#L116-L128

This should probably be an option that can be disabled by the admin.

willgearty:
Maybe we should always set the override permission and the Student/All permission should be based on the form option?

PR #3272, opened 2021-04-27
Add option to not set Student/All permission for student lottery winners

willgearty:
This adds a checkbox to both student lottery forms that allow admins to disable the creation of the Student/All permission for the lottery winners. I also made it so that the submission buttons are disabled if the lottery has already been run (with some hover text explaining why they are disabled).

image

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3244.


hwatheod (approved):
LGTM

Issue #3256, opened 2021-04-13
Add ID column to custom form page


milescalabresi:
Using custom forms (customforms) usually requires entering the form's ID number into a tag. Right now, these numbers are only visible in URLs and /admin. We should expose them in a column in the /customforms view.

willgearty:
You want the entire column to have the ID number in every row? Or would it be better to just put it in the page header?

milescalabresi:
Something like this. Not mutually exclusive with putting it in the page header in parentheses too
![image](https://user-images.githubusercontent.com/5506191/116175613-a10f0a00-a6de-11eb-85cb-212d5cb4b234.png)

willgearty:
Ah, that makes a lot more sense. I mistakenly read your original post as adding the ID to the responses page.

Issue #2959, opened 2020-04-18
UI for editing custom forms


willgearty:
It would be great if you could use the custom form creation UI to also _edit_ custom forms.

willgearty:
NOTE: It appears that it may not be possible to add fields after a form has already received responses (this needs confirmation from someone that understands the custom form DB better than I do). If that's the case, we should NOT permit _adding_ fields once a form has received responses. We probably shouldn't be permitting _editing_ of fields after there are responses either (since people responded to the original questions, not whatever they would be after editing).

Nonetheless, I'm still in favor of making it so that forms can be edited before they have received responses.

willgearty:
Or we could just rewrite custom forms (maybe like surveys)

willgearty:
Apparently, there is already an interface to edit a custom form. But it doesn't seem to work when you submit the changes (I tried with a form that already had responses and one that had no responses, neither worked),

Issue #2904, opened 2020-01-23
customforms should automatically set up tags in creation


kkbrum:
I think we could add a question or two to the customforms creation form that would then give it enough information to make any required tags automatically (learn_extraform_id/teach_extraform_id/whatnot). Haven't looked into feasibility but the process of making the tags confuses admins and seems like it could be automated.

Issue #2690, opened 2019-03-03
Render HTML in custom form questions


willgearty:
We should render any html in custom form questions as actual html. It would add a nice way for admins to further customize their forms.

Requested by Northwestern.

PR #3273, opened 2021-04-28
Allow for custom forms to be edited (+ misc custom form fixes)

willgearty:
This fixes the ability to edit custom forms through the form builder UI. Apparently this was originally designed to work (my best guess is a django upgrade broke it at some point), but I don't think it's worked for years (if not ever). I've also added links on the custom form landing page to edit a form, which preloads the form for the admin in the form builder. I've tested adding, removing, and editing fields.

I've also used this PR to fix some other custom form issues that have popped up over the years. It's easier than creating a completely new system for the time being (#1842). These include:

- Fix the handling of long field titles in the response grid (#1103)
- Update the gender field (now has an "Other" option)
- Enforce textField limits (#1896)
- Add custom form IDs to the custom form landing page (#3256)
- Link custom forms to modules and set/delete the appropriate tags (#2904)
- HTML in field labels and help texts is now rendered (#2690)
- Catch errors when a form is submitted via the custom form module (#1102)
- Fix adding/removing of instructions to/from existing forms (#792)
- Allow for changing field type of existing fields
- Fix bug where blank options were added to multi-answer fields in the form builder (e.g. checkboxes, dropdown)
- Make the "Add to Form"/"Update" button responsive (hover/click)
- Separately scrolling preview and settings panels
- Fix teacherinfo and contactinfo fields
- Better error handling
- Show instructions question mark and indicate correct answers

Fixes https://github.com/learning-unlimited/ESP-Website/issues/2959 and fixes https://github.com/learning-unlimited/ESP-Website/issues/1938.
Also fixes https://github.com/learning-unlimited/ESP-Website/issues/3256, fixes https://github.com/learning-unlimited/ESP-Website/issues/1103, fixes https://github.com/learning-unlimited/ESP-Website/issues/1896, fixes https://github.com/learning-unlimited/ESP-Website/issues/2904, fixes https://github.com/learning-unlimited/ESP-Website/issues/2690, fixes https://github.com/learning-unlimited/ESP-Website/issues/1102, and fixes https://github.com/learning-unlimited/ESP-Website/issues/792.

### DIFF
--- a/esp/esp/customforms/DynamicForm.py
+++ b/esp/esp/customforms/DynamicForm.py
@@ -774,7 +774,7 @@ class FormHandler:
                     tl = "teach"
                     module = "TeacherQuizModule"
             elif tags.count() > 1:
-                raise ESPError('Custom form # %i is linked to multiple registration modules for %s' % (self.form.id, prog.name))
+                raise ESPError('Custom form #%i is linked to multiple registration modules for %s' % (self.form.id, prog.name))
             else:
                 tl = ''
                 module = ''
--- a/esp/esp/customforms/DynamicForm.py
+++ b/esp/esp/customforms/DynamicForm.py
@@ -10,6 +10,7 @@ from django.http import HttpResponseRedirect
 from localflavor.us.forms import USStateField, USPhoneNumberField, USStateSelect
 from esp.customforms.forms import NameField, AddressField
 from esp.customforms.DynamicModel import DMH
+from esp.tagdict.models import Tag
 from esp.utils.forms import DummyField
 from esp.users.models import ContactInfo, ESPUser
 from argcache import cache_function
@@ -756,6 +757,24 @@ class FormHandler:
             'perms': self.form.perms,
             'pages': self._getFormMetadata(self.form)
         }
+        # Identify if this form has been linked to a registration module
+        if self.form.link_type == 'Program' and self.form.link_id != -1:
+            try:
+                prog = Program.objects.get(id=self.form.link_id)
+            except Program.DoesNotExist:
+                raise ESPError('No program with ID %i' % (self.form.link_id))
+            tags = Tag.objects.filter(content_type=ContentType.objects.get_for_model(Program), object_id=prog.id, value=self.form.id)
+            if tags.count() == 1:
+                tag = tags[0]
+                if '_extraform_id' in tag.key:
+                    tl = tag.key.split("_")[0]
+                    module = "CustomFormModule"
+                elif tag.key == 'quiz_form_id':
+                    tl = "teach"
+                    module = "TeacherQuizModule"
+            else:
+                raise ESPError('Custom form # %i is linked to multiple registration modules for %s' % (self.form.id, prog.name))
+            metadata.update({'link_tl': tl, 'link_module': module})
         return metadata
 
 
--- a/esp/esp/customforms/DynamicForm.py
+++ b/esp/esp/customforms/DynamicForm.py
@@ -1,6 +1,7 @@
 from esp.customforms.models import Field, Attribute, Section, Page, Form
 from django import forms
 from django.forms.models import fields_for_model
+from django.utils.safestring import mark_safe
 from form_utils.forms import BetterForm
 from collections import OrderedDict
 from formtools.wizard.views import SessionWizardView
@@ -135,7 +136,7 @@ class CustomFormHandler():
 
             for field in section:
                 field_name = 'question_%d' % field['id']
-                field_attrs = {'label': field['label'], 'help_text': field['help_text'], 'required': field['required']}
+                field_attrs = {'label': mark_safe(field['label']), 'help_text': mark_safe(field['help_text']), 'required': field['required']}
 
                 # Setting the 'name' attribute for combo fields
                 """
--- a/esp/esp/customforms/DynamicForm.py
+++ b/esp/esp/customforms/DynamicForm.py
@@ -764,7 +764,7 @@ class FormHandler:
                 prog = Program.objects.get(id=self.form.link_id)
             except Program.DoesNotExist:
                 raise ESPError('No program with ID %i' % (self.form.link_id))
-            tags = Tag.objects.filter(content_type=ContentType.objects.get_for_model(Program), object_id=prog.id, value=self.form.id)
+            tags = Tag.objects.filter(content_type=ContentType.objects.get_for_model(Program), object_id=prog.id, value=self.form.id, key__in=['learn_extraform_id', 'teach_extraform_id', 'quiz_form_id'])
             if tags.count() == 1:
                 tag = tags[0]
                 if '_extraform_id' in tag.key:
@@ -773,8 +773,11 @@ class FormHandler:
                 elif tag.key == 'quiz_form_id':
                     tl = "teach"
                     module = "TeacherQuizModule"
-            else:
+            elif tags.count() > 1:
                 raise ESPError('Custom form # %i is linked to multiple registration modules for %s' % (self.form.id, prog.name))
+            else:
+                tl = ''
+                module = ''
             metadata.update({'link_tl': tl, 'link_module': module})
         return metadata
 
--- a/esp/esp/customforms/DynamicForm.py
+++ b/esp/esp/customforms/DynamicForm.py
@@ -88,7 +88,7 @@ class CustomFormHandler():
         for attr in attrs:
             if attr['attr_type'] == 'options':
                 other_attrs['choices'] = []
-                options_list = attr['value'].split('|')[:-1]
+                options_list = attr['value'].split('|')
                 for option in options_list:
                     other_attrs['choices'].append( (option, option) )
             elif attr['attr_type'] == 'limits':
@@ -200,7 +200,7 @@ class CustomFormHandler():
                     if not field_is_custom:
                         # Add in other classes for validation
                         generic_type = cf_cache.getGenericType(form_field)
-                        if 'widget_attrs' in self._field_types[generic_type] and 'class' in self._field_types[generic_type]['widget_attrs']:
+                        if generic_type in self._field_types and 'widget_attrs' in self._field_types[generic_type] and 'class' in self._field_types[generic_type]['widget_attrs']:
                             form_field.widget.attrs['class'] += self._field_types[generic_type]['widget_attrs']['class']
 
                     # Adding to field list
--- a/esp/esp/customforms/DynamicModel.py
+++ b/esp/esp/customforms/DynamicModel.py
@@ -219,11 +219,12 @@ class DynamicModelHandler:
         with connection.schema_editor() as schema_editor:
             model = self.createDynModel()
             new_field = self._getModelField(field.field_type)
-            new_field.column = self.get_field_name(field)
-            # We need to set a default (if one isn't set already) in case there are already responses to the form
-            if new_field.default == NOT_PROVIDED:
-                new_field.default = ''
-            schema_editor.add_field(model, new_field)
+            if new_field:
+                new_field.column = self.get_field_name(field)
+                # We need to set a default (if one isn't set already) in case there are already responses to the form
+                if new_field.default == NOT_PROVIDED:
+                    new_field.default = ''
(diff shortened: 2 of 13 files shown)


willgearty:
OK, I think that's everything that I can find to fix at the moment!

esp/public/media/scripts/custom_form.js
```diff
@@ -1190,10 +1271,16 @@ var submit=function() {
 	
 	var form={'title':$j('#form_title').html(), 'desc':$j('#form_description').html(), 'anonymous':($j('#id_anonymous').prop('checked') == "checked"), 'pages':[]}, section, elem, page, section_seq, page_seq=0;
 	form['link_type']=$j('#links_id_main').val();
-	if(form['link_type']!='-1' && $j('#links_id_specify').val()=='particular')
+	if(form['link_type']!='-1' && $j('#links_id_specify').val()=='particular'){
 		form['link_id']=$j('#links_id_pick').val();
-	else form['link_id']=-1;	
-	
+        if($j("#links_id_main").val()=="Program"){
```
kkbrum, line 1276: I don't think this was your doing, but when you pick which program in the "links" section, the list of programs has the oldest at the top instead of the newest at the top (which seems to be the convention everywhere else on the site, and in the dropdown for programs in form permissions)

esp/templates/customforms/index.html
```diff
@@ -156,7 +159,19 @@ <h3><a href='#'>Links</a></h3>
...
+					</p>
+
+                    <p>User type -
+					<select id="links_id_tl">
+                        <option value="learn">Students</option>
+                        <option value="teach">Teachers</option>
+					</select>
+					</p>
+
+                    <p>Module -
```
kkbrum, line 171: I assume the lack of entries here means I have neither customforms nor quizzes enabled, but it would be helpful to have help text show up if that's the case saying "For teachers to be able to fill out this form for this program, you need to enable one of the following modules: ... "

image

esp/public/media/scripts/custom_form.js
```diff
@@ -605,37 +669,40 @@ var onSelectElem = function(item) {
...
+            $j('#id_instructions').val('Not required');
+        }
+        
+        //Set 'Required' to a sensible default
+        setRequired(item);	
+            
+        $j('#id_question').val(question_text);
+        $prevField=$currSection.children(":last");
+        if($button.val()!='Add to Form')
+            $button.val('Add to Form').html('Add Field to Form').unbind('click').click(function(){insertField($j('#elem_selector').val(),$prevField)});
```
kkbrum, line 704: When I click the add field to form button, the field options kind of reset, but the field i maybe added doesn't show up in the form preview on the left?

esp/esp/customforms/DynamicForm.py
```diff
@@ -756,6 +758,24 @@ def rebuildData(self):
...
+            if tags.count() == 1:
+                tag = tags[0]
+                if '_extraform_id' in tag.key:
+                    tl = tag.key.split("_")[0]
+                    module = "CustomFormModule"
+                elif tag.key == 'quiz_form_id':
+                    tl = "teach"
+                    module = "TeacherQuizModule"
+            else:
+                raise ESPError('Custom form # %i is linked to multiple registration modules for %s' % (self.form.id, prog.name))
```
kkbrum, line 777: So I'm getting such an error in my terminal, but on the actual custom forms page on the site, nothing appears out of the ordinary. Don't we want to signal to the user that they need to fix something? Not just give an error in the log or terminal?

kkbrum (changes requested):
Definitely a lot of good changes, but I am having trouble with both adding fields and with linking to programs. It's hard for me to test more components without those, so I'll submit these comments for now and come back to it when the comments about the "add field" and the "multiple modules linked error" get addressed.

willgearty:
OK, I _think_ that addresses all of your comments @kkbrum (and other bugs I found).

esp/esp/customforms/views.py
```diff
@@ -84,55 +85,73 @@ def onSubmit(request):
...
+                        Tag.setTag(key='quiz_form_id', value=form.id, target=prog)
                     else:
-                        is_required = False
+                        return ESPError('Module %s does not use a custom form or is not implemented' % (metadata['link_module']))
 
-                    new_field = Field.objects.create(form=form, section=new_section, field_type=field['data']['field_type'],
-                        seq=int(field['data']['seq']), label=field['data']['question_text'], help_text=field['data']['help_text'],
-                        required=is_required
-                        )
+                # Inserting pages
```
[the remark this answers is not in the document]
kkbrum: In a similar vein, these three fields should really not have the little checkbox for "required" since they're not true fields and i presume required vs not required would have no effect here.

esp/esp/users/models/__init__.py
```diff
@@ -1473,7 +1473,7 @@ class TeacherInfo(models.Model, CustomFormsLinkModel):
         ('shirt_type', 'Shirt type'),
```
kkbrum, line 1473: For the teacherinfo field section, I dont understand this comment:

image

My main thing is I don't see any '*'s anywhere, so I'm just confused by it

kkbrum, line 1473: the "shirt size" and "shirt type" fields produce long answer fields, which seems wrong. They should either be short answers, or even radio buttons / dropdowns

kkbrum: I see them now for contact info. However, I don't really believe this comment. Having a custom form doesnt mean that users would not have to update their profile info, so why should we be including them in the custom form?

esp/esp/customforms/DynamicModel.py
```diff
@@ -217,25 +218,33 @@ def get_field_name(self, field):
     def addField(self, field):
```
kkbrum, line 218: So I was able to add literally every field to a form and save it, but when I go to fill it out, I get this:

image

kkbrum: If you need me to add fields one by one to debug if there's a certain field causing the error, but I'd rather not if you can spot the issue without that :P

kkbrum (changes requested):
Looking better and better! I do have an error when trying to fill out my new form, so that needs to get fixed. The many other comments are truly just suggestions as I looked at every field, I won't reject the PR if they arent addressed or anything!

esp/public/media/scripts/custom_form.js
```diff
@@ -605,37 +681,40 @@ var onSelectElem = function(item) {
...
+	if($j('div.field_selected').length){
+        onSelectField($j('div.field_selected'), $j.data($j('div.field_selected')[0], 'data'), item);
+    } else {
+        var currCategory=$j('#cat_selector').val();
+        var $option,$wrap_option,i, question_text=formElements[currCategory][item]['ques'], $button=$j('#button_add');
+        
+        //  Add validation options
+        if (item in formElements['Generic'])
+        {
+            addCorrectnessOptions(item);
```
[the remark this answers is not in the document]
willgearty: Hmm...I don't honestly know much about the inner workings of the "correctness" options, but it appears that only some things are "supported" (by django or us, I have no idea):
https://github.com/learning-unlimited/ESP-Website/blob/ca767d07f60633733145948fee1c1de6fdb0c52b/esp/public/media/scripts/custom_form.js#L632-L634

It looks like we add some custom validation to the field, but I feel like we would have added it to all fields if it was possible? (although maybe that's a bad assumption)
https://github.com/learning-unlimited/ESP-Website/blob/98501776a014a01eeba8c8105ee2aca2a7d7bee3/esp/esp/customforms/DynamicForm.py#L145-L162

esp/esp/customforms/linkfields.py
```diff
@@ -21,7 +21,7 @@
     'phone': {'typeMap': USPhoneNumberField, 'attrs': {'widget': forms.TextInput,}, 'widget_attrs': {'class': 'USPhone '}},
     'email': {'typeMap': forms.EmailField, 'attrs': {'max_length': 30, 'widget': forms.TextInput,}, 'widget_attrs': {'class': 'email '}},
     'state': {'typeMap': USStateField, 'attrs': {'widget': USStateSelect}, 'widget_attrs': {'class': ''}},
-    'gender': {'typeMap': forms.ChoiceField, 'attrs': {'widget': forms.RadioSelect, 'choices': [('F', 'Female'), ('M', 'Male')]}, 'widget_attrs': {'class': 'gender '}, },
+    'gender': {'typeMap': forms.ChoiceField, 'attrs': {'widget': forms.RadioSelect, 'choices': [('F', 'Female'), ('M', 'Male'), ('O', 'Other')]}, 'widget_attrs': {'class': 'gender '}, },
     'radio_yesno': {'typeMap': forms.ChoiceField, 'attrs': {'widget': forms.RadioSelect, 'choices': (('T', 'Yes'), ('F', 'No'))}, 'widget_attrs': {'class': ''}},
     'boolean': {'typeMap': forms.BooleanField, 'attrs': {'widget': forms.CheckboxInput}, 'widget_attrs': {'class': ''}},
     'null_boolean': {'typeMap': forms.NullBooleanField, 'attrs': {'widget': forms.NullBooleanSelect}, 'widget_attrs': {'class': ''}},
```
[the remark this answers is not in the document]
willgearty: https://stackoverflow.com/a/49198682/4660582
Thoughts on how to make this clear?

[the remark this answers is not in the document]
willgearty: Although, we still use a checkbox for this field, so we aren't actually giving respondents the ability to select the "unknown" option. We should really have a dropdown with three options: https://docs.djangoproject.com/en/1.11/_modules/django/forms/widgets/#NullBooleanSelect.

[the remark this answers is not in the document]
willgearty: Oh, we actually do that in the actual form...I'll just update the display of this field in the form preview and hopefully that makes it clear.

esp/esp/customforms/DynamicModel.py
willgearty: Ah, I think what's going on here is that the "correct answers" are supposed to be the indices of the choices, not the text values of the choices? Could you try that and see if it works?

kkbrum: yeah, i did 1-based indices. Switching to 0 based fixed the form. 
Can we do at least one of the following:
- specify that 0 based indexing is desired
- use 1 based indexing here because 0 based in a non-coding setting is unusual
- label the choices with their index (1. ____ 2. ____ etc)
- check that the specified correct answer is valid before creating the form

esp/templates/customforms/index.html
```diff
@@ -31,7 +34,7 @@ <h2 id='form_title'>Form Title</h2>
 			</div>
 			
 			<div id="form_toolbox">
```
kkbrum, line 36: Would it be possible to have the panel on the right move with you as you scroll down a long form? It's quite annoying to edit a form that's longer than the computer screen when you have to scroll up to get to the panel to edit and then scroll down to see the question again.

esp/esp/users/models/__init__.py
willgearty: I think what it is trying to say is that the fields need to be required if you include them in the form, because they are required fields in the profile? But there was a bug that didn't set the required checkbox. I fixed the bug and I changed the text of the comment too.

esp/public/media/scripts/custom_form.js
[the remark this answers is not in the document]
willgearty: I've moved this request to a separate issue.

esp/public/media/scripts/custom_form.js
```diff
@@ -865,9 +865,10 @@ var renderNormalField=function(item, field_options, data){
 			type:"text",
 			size:"30"
 		});
+        $new_elem = $new_elem.add($j("<span class='correct_answer'> Correct answer: " + $j('#textField_correct_answer').val() + "</span>"));
```
kkbrum, line 870: For the text questions, if the admin doesn't specify a correct answer, can we hide the "Correct answer:" in the preview? It's just blank and confusing if there was no correct answer specified

kkbrum (approved):
Monumental!

PR #3278, opened 2021-05-06
Use labels for ajax scheduler scheduling checks

willgearty:
This makes the text for the scheduling checks in the ajax scheduler `<label>`s, so now it selects the respective radio button when you click the text.


hwatheod (approved):
very simple change, looks good.

Issue #3002, opened 2020-05-29
Implement "max sections" option for Two-Phase Lottery management


willgearty:
The idea here is that an admin could specify the maximum number of sections that students should be enrolled in when the two-phase lottery is run.

Requested by Cloud.

PR #3279, opened 2021-05-06
Add enrollment limit options to class lottery

willgearty:
This adds two new options to the class lottery which allow admins to limit the number of timeslots for which students have enrollments and/or the number of sections in which students have been enrolled.

image

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3002.

### DIFF
--- a/esp/esp/program/controllers/lottery.py
+++ b/esp/esp/program/controllers/lottery.py
@@ -83,7 +83,9 @@ class LotteryAssignmentController(object):
         'stats_display': (False, False),
         'directory': (os.getenv("HOME"), False),
         'use_student_apps': (False, 'Whether to use student application ranks'),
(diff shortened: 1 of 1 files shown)


willgearty:
OK, I was able to test this on my dev server (after fixing a bunch of weird stuff on my server like having weird lunch constraints, etc). Both filters seem to work individually and in tandem. An important note: we only check that the number of timeslots/sections isn't already surpassed when we make a new enrollment, not whether a new enrollment will cause the limits to be surpassed. This means that a student could theoretically end up with more timeslots filled than the set limit if the last enrolled section has more than one timeslot (but I think this is better than the alternative where a student ends up with a less full schedule).

hwatheod (approved):
Looks good!

PR #3283, opened 2021-05-07
Master moderator schedule

kkbrum:
Adds two locations to get a copy of the moderator schedule. The first is in an extra column on the room schedules, showing the moderator of each class during each time in that room. The second is a new csv, which mimics the google spreadsheet Rainstorm was making (rooms as rows, times as columns, filled in with moderators).

Requested by Rainstorm.

### DIFF
--- a/esp/esp/program/modules/handlers/programprintables.py
+++ b/esp/esp/program/modules/handlers/programprintables.py
@@ -1837,6 +1837,57 @@ class ProgramPrintables(ProgramModuleObj):
 
     @aux_call
     @needs_admin
+    def moderator_rooms_spr(self, request, tl, one, two, module, extra, prog):
+        """
+        Create a spreadsheet with a row for each room, a column for each timeblock,
+        filled out with moderator names and phone numbers.
+        """
+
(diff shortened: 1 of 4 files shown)


esp/esp/program/modules/handlers/programprintables.py
```diff
@@ -1835,6 +1835,57 @@ def time_possible(time, sections_list):
         response['Content-Disposition'] = 'attachment; filename=ok_times_concise.csv'
         return response
 
+    @aux_call
+    @needs_admin
+    def moderator_rooms_spr(self, request, tl, one, two, module, extra, prog):
```
kkbrum, line 1840: @willgearty I basically tried to combine code from the room schedules and the oktimes_spr, but honestly I was surprised that this code worked and am not 100% crystal clear on some of it so please read this code for me and make sure it's logical :) In particular, I don't know the difference between the event group and timeslots, so I'm not sure if there are cases where the code below wouldn't work.

willgearty (approved):
Looks good to me! Both printables look good!

Issue #2683, opened 2019-02-23
usersearch.html should be explicit of which module you are using


willgearty:
Right now, for all of the modules that use usersearch.html as a way to identify a user (e.g. advanced checkin and some printables, I believe), the page looks like this:
![image](https://user-images.githubusercontent.com/7232514/53291050-0a430d80-3762-11e9-9954-b05d2f3ed76f.png)

Unfortunately, the only way to see which module you are using is by looking at the URL. Yes, theoretically they should know which module they are using because they clicked the link to get here, but I still think we should be displaying something at the top of the page to identify the module. Or maybe alter the text to say "Search for User for/to X".

900miles:
I took a look at this, and I'm not sure how to solve it except if every module sends itself / its name to the webpage to be displayed. Unless there's a way for the page to tell what module it's being used for itself?

willgearty:
It looks like most of the uses of the template are through the `search_for_user` function:
https://github.com/learning-unlimited/ESP-Website/blob/7a31858169ef596c6c2691996ec591d1ae84c516/esp/esp/users/views/usersearch.py#L259-L307

For those cases, we'd just want to add "module" or something as a parameter for the function and adjust all of the calls of the function accordingly (this could either be the module object itself or a string).

In other cases, it looks like some modules render the template themselves, so we'd just want to do whatever we do in `search_for_user` in those views as well, e.g.:
https://github.com/learning-unlimited/ESP-Website/blob/7a31858169ef596c6c2691996ec591d1ae84c516/esp/esp/program/modules/handlers/onsitecheckinmodule.py#L260-L261

PR #3289, opened 2021-05-13
Add module name to usersearch page

willgearty:
Instead of just saying "Search for User" at the top of the page (that can be served by several different modules), we now list the module name at the top:
image

Fixes https://github.com/learning-unlimited/ESP-Website/issues/2683.


hwatheod (approved):
Looks good.

PR #3293, opened 2021-05-14
Fixes Class Seach on Big Picture

### LINKED ISSUE #3258: bigpicture class search styling bug
The user type dropdown has a height of 0 for some reason on the class search page on bigpicture:
image

I think that this must be related to some sort of bigpicture styling "feature"/"bug" because I don't see the same issue on fruitsalad:
image

nrjean:
Fixes #3258
Just adds a little code to override the inherit for this specific filter box.
I think this is easier to do then alternative because the inherit is definitely used (for better or for worse) in other input boxes

### DIFF
--- a/esp/esp/themes/theme_data/bigpicture/less/main.less
+++ b/esp/esp/themes/theme_data/bigpicture/less/main.less
@@ -274,3 +274,6 @@ div.nav-collapse.collapse {
   margin: 0;
   padding: 0;
 }
+select#regtypes {
+    font-size: 13px;
+}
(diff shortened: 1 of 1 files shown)


esp/esp/themes/theme_data/bigpicture/less/main.less
```diff
@@ -274,3 +274,6 @@ div.nav-collapse.collapse {
   margin: 0;
   padding: 0;
 }
+select#regtypes {
+    font-size: 13px;
+}
```
willgearty, line 279:
```suggestion
#reg_types_div {
    font-size: 13px;
}
```

willgearty: Looks like the text that is supposed to be above the <select> is also hidden, so this should catch both

willgearty (changes requested):
Just one suggested change. Otherwise, looks good!

PR #3311, opened 2021-05-17
Fix invalid template syntax in credit card failure.html template

### LINKED ISSUE #3309: TemplateSyntaxError at /learn/Rainstorm/2021_Spring/charge_payment
This is an exception reported in the logs.

Could not parse the remainder: '['support']' from 'settings.DEFAULT_EMAIL_ADDRESSES['support']'

The template is at esp/templates/program/modules/creditcardmodule_stripe/failure.html 
Lines 24 and 49 appear to be the problem - `['support']` is not allowed in the Django template language.

hwatheod:
Fix invalid template syntax in the credit card failure message page.

Fixes #3309.

### DIFF
--- a/esp/esp/program/modules/handlers/creditcardmodule_stripe.py
+++ b/esp/esp/program/modules/handlers/creditcardmodule_stripe.py
@@ -316,7 +316,6 @@ class CreditCardModule_Stripe(ProgramModuleObj):
         if 'error_type' in context:
             #   If we got any sort of error, send an email to the admins and render an error page.
             self.send_error_email(request, context)
-            context['support_email'] = settings.DEFAULT_EMAIL_ADDRESSES['support']
             return render_to_response(self.baseDir() + 'failure.html', request, context)
 
         #   Render the success page, which doesn't do much except direct back to studentreg.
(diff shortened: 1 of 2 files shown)


esp/esp/program/modules/handlers/creditcardmodule_stripe.py
```diff
@@ -316,6 +316,7 @@ def charge_payment(self, request, tl, one, two, module, extra, prog):
         if 'error_type' in context:
             #   If we got any sort of error, send an email to the admins and render an error page.
             self.send_error_email(request, context)
+            context['support_email'] = settings.DEFAULT_EMAIL_ADDRESSES['support']
```
willgearty, line 319: since we use render_to_response(), the context should already include the settings

willgearty (changes requested):
Since we're using our custom render_to_response, the context should already include `settings`, so I think we just need to use `settings.DEFAULT_EMAIL_ADDRESSES.support` in the template (looks like we were mixing up our python and django template languages).

willgearty (approved):
Looks good to me, thanks @hwatheod !

Issue #3304, opened 2021-05-16
Section cancellation form should have section times


willgearty:
![image](https://user-images.githubusercontent.com/7232514/118403873-1878ef00-b636-11eb-8050-5f373af3a1f7.png)

We should put the dates/times for each section so the admin knows which one they want to cancel without having to go to another page or elsewhere on the page to figure it out.

willgearty:
This looks useful (although probably overkill): https://medium.com/@alfarhanzahedi/customizing-modelmultiplechoicefield-in-a-django-form-96e3ae7e1a07

PR #3315, opened 2021-05-17
Add times/dates to the section cancellation form

willgearty:
This adds times/dates to each of the sections in the section cancellation form, so it's easier to identify which ones you want to cancel without going elsewhere on the page/other pages.
image

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3304.


hwatheod (approved):
If i read this correctly, it

Issue #3259, opened 2021-04-14
bigpicture login button missaligned


nrjean:
![image](https://user-images.githubusercontent.com/60372830/114644559-bb3ff580-9ca5-11eb-9492-86ef94a459c7.png)

PR #3319, opened 2021-05-19
Fix spacing and alignment within login/logout bar in bigpicture

willgearty:
Previously looked like this:
image

Now it looks like this:
image

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3259.


nrjean:
Looks good to me

nrjean (approved):

Issue #3280, opened 2021-05-06
Moderator scheduling: Teachers' own classes outside preferred moderator categories won't show unless "Hide sections..." box unchecked


laurenmaclachlan:
Kind of annoying to have to toggle the box when attempting to schedule a moderator consecutively in the same room, would be ideal if the teacher's classes show regardless

![moderator teacher](https://user-images.githubusercontent.com/78887914/117233684-9eed2f80-adf1-11eb-9258-bffe554267a8.png)

PR #3320, opened 2021-05-20
Always show moderating/teaching sections, regardless of subject matching

willgearty:
This makes it so that a section is always shown on the scheduler (specifically, when a moderator is selected) if the moderator is moderating or teaching the section, regardless of whether the subject of the class is preferred by the moderator (if the checkbox is checked).

All I did was change the CSS order so that the mod/teaching styling takes precedence over the hidden styling.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3280.


kkbrum (approved):
Splendid!

Issue #3297, opened 2021-05-15
Moderators for onsite attendance


willgearty:
In the last table on the onsite attendance page we have a list of classes that don't have attendance marked:
![image](https://user-images.githubusercontent.com/7232514/118371108-a094c100-b570-11eb-8573-a66457d78677.png)

For programs with moderators, the moderators are likely the ones that are marking attendance, so we probably want their information in the table as well.

PR #3322, opened 2021-05-20
Add moderators to onsite attendance

willgearty:
This adds moderators to last table on the onsite attendance page:
image

While I was at it, I also added the "recipient_type" to all of the comm panel, etc. links.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3297.


nrjean (approved):
It works yay :D

Issue #3323, opened 2021-05-28
Custom form registration modules should have separate handlers


willgearty:
We should split the teach and learn custom form modules to separate handlers. Once this is done, we can then have:

- Separate isStep methods for each of them, to decide whether a custom form is set up for registration
- Separate requirements for each module, such as @meets_grade for the learn module

PR #3351, opened 2021-06-16
Separate custom form handlers

willgearty:
This splits the custom form handler into separate teach and learn custom form handlers. This allows us to have separate isStep functions, so the module won't be shown on the respective main registration page if the proper tag and form aren't set up. I've also split the original custom form record type to separate student and teacher record types, which also allows us to have separate lists for students and teachers (e.g., for the comm panel/AUL).

The modules migration is set up to update the existing Program Module objects in a site's database. The users migrations is set up to assign existing Record objects to the new record types (as best as we can).

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3323.


kkbrum (approved):
Spectacular!

Issue #3343, opened 2021-06-06
Prevent new programs from having the same name as old programs


willgearty:
It breaks things when you try to go to the URLs for either program (because we can't resolve which one the user wants).

PR #3352, opened 2021-06-17
Prevent non-unique program urls

willgearty:
This prevents chapters from creating a new program with a URL that is already being used by an existing program. This is prevented at the form level (when creating a new program) and at the database level. If such programs with duplicate URLs already exist, they are modified by the migration so that they now have unique URLs.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3343.


kkbrum (approved):
Brilliant!

Issue #3338, opened 2021-06-03
Shirt stats for moderators


willgearty:
We should have shirt stats for moderators on the dashboard.

PR #3353, opened 2021-06-18
Add moderator tshirt stats

willgearty:
Add moderators to the list of tshirt stats if the moderator module is enabled.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3338.


kkbrum (approved):
Amazing!

Issue #3355, opened 2021-06-18
Custom form labels have ":" in preview but not in actual form


willgearty:
This is the preview:
![image](https://user-images.githubusercontent.com/7232514/122613556-a89dc000-d04a-11eb-8bcd-f1ab1d24fd56.png)

However, this is what the actual form looks like:
![image](https://user-images.githubusercontent.com/7232514/122613505-91f76900-d04a-11eb-9a8f-8c7150c0dc89.png)

Notice that all of the questions have a colon at the end in the preview, but most (**but, strangely, not all**) of the them do NOT have a colon in the actual form.

Issue #3354, opened 2021-06-18
Custom form builder doesn't enforce field label length limit


willgearty:
This works as a label:
`Would you be able to attend an in person splash? (in person splash programs are listed  <a href="https://www.learningu.org/current-programs/" target="_blank">here</a>)`

But this causes an error:
`Would you be able to attend an in person splash? (in person splash programs are listed here: <a href="https://www.learningu.org/current-programs/" target="_blank">https://www.learningu.org/current-programs/</a>)`

willgearty:
Upon further investigation, this has nothing to do with the link itself, just that using the link instead of "here" pushes the entire label over 200 characters, which seems to be the limit.

Issue #3339, opened 2021-06-04
Removing custom form fields has undesired behavior


willgearty:
There are three different undesired behaviors/bugs here:

1. If you delete a field, it then automatically selects the section that the field was in.
2. When a section is selected (like due to # 1), it adds an X to all of the fields within it.
![image](https://user-images.githubusercontent.com/7232514/120727488-4799b800-c4a0-11eb-8d71-08ac4b7587de.png)

3. When you delete all of the fields and/or the "page" that the fields are on, you can not add a new field until you add a new page. The interface should either 1) tell you this is why, 2) just automatically add a page when you add a field without a page, or 3) never let you remove the last page.

willgearty:
More issues that have arisen:

1. When you click a section, the title/description details aren't loaded into the side panel (do we even show these in the actual custom form?)
2. When you click a section, then click something else, the hover class isn't removed from the previously selected section.
2. This might be specific to my custom form, but there's an empty section at the top of the first page.
3. You are supposed to be able to select which page/section you are currently adding to by double clicking, but this doesn't work, and, even if it did, there's no visual cue or reaction to this behavior.

PR #3356, opened 2021-06-21
Custom form builder bug fixes

willgearty:
This makes the many following bug fixes to the custom form builder:

1. If you delete a field, it no longer selects the section that the field was in.
2. If you select a section, it no longer adds an X to all of the fields within it.
3. You can no longer delete the last section of a page or the last page of the form. If you attempt to do this, you receive a popup. You CAN delete the last field in a section.
4. When you select a section, the title/description is now loaded. You can also now change these elements, and they properly update in the actual form.
5. When you select a section, then select something else, the former section is unselected.
6. There are no longer empty sections at the top of each page.
7. The current page and current section now have styling to show where new fields or sections will be added. When a field is selected, the current page and section are shifted to those of the selected field. The current page can be changed by double clicking on the desired page/page break.
8. If you change a non-multi-answer field to a multi-answer field, a blank option is added. You can no longer remove the last blank option for a multi-answer field, which prevents situations where you can't add more options.
9. Fields don't shift in location in small but very annoying ways when hovering or selecting fields.
10. When you add a new page, it now has a section with a header/description.
11. You can now edit/remove the default section when making a brand new custom form.
12. We no longer show a ":" after a field label if the label text ends in a punctuation mark (like in the actual form).
13. Enforces the field label length limit in the builder.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3339, fixes https://github.com/learning-unlimited/ESP-Website/issues/3355, and fixes https://github.com/learning-unlimited/ESP-Website/issues/3354.

### DIFF
--- a/esp/public/media/scripts/custom_form.js
+++ b/esp/public/media/scripts/custom_form.js
@@ -825,7 +825,7 @@ var updateField=function() {
     var field_type=$j('#elem_selector').val();
     var curr_field_id=$j.data($currField[0],'data').parent_id;
 	if(curr_field_type=='section' && field_type=='section'){
-        $field_data = $currField.data('data')
+        var $field_data = $currField.data('data')
         $field_data.question_text = $j('#id_question').val()
         $currField.find('h2').html($j('#id_question').val());
         $field_data.help_text = $j('#id_instructions').val()
@@ -851,7 +851,7 @@ var onSelectField=function($elem, field_data, ftype=null) {
     
 	//De-selecting any previously selected field
 	if($j('div.field_selected').length == 0 || !$elem.hasClass('field_selected')){
-        $divs = $j('div.field_selected');
+        var $divs = $j('div.field_selected');
         $divs.click();
         $divs.removeClass('field_hover');
         $divs.children(".wrapper_button").removeClass("wrapper_button_hover");
--- a/esp/public/media/scripts/custom_form.js
+++ b/esp/public/media/scripts/custom_form.js
@@ -1287,7 +1287,10 @@ var renderNormalField=function(item, field_options, data){
             $j(this).children(".wrapper_button").removeClass("wrapper_button_hover");
         }).toggle(function(){onSelectField($j(this), $j.data(this, 'data'));}, function(){deSelectField($j(this));});
 		$outline.appendTo($currPage);
-		$currPage.sortable();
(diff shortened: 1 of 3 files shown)


kkbrum:
This all looks good and can probably be approved. Thoughts/observations:

- Probably shouldnt be able to convert a field to a section, a field to a page, a section to a page. Etc. So disable the options that shouldnt be clicked when you have something selected
- Would be cool to be able to move sections to different pages, and fields to different sections / pages. Also, I think the side panel could have up and down arrows next to update field which would be the way to reorder things without using the preview/drag interface
- If you select a section, but then double click on a different page, the section should probably get unselected? 
- what's the logic behind needing to double click to select a page and not just standard click?
- if i select a field within a section, and then double click on the field, the first section in the page becomes pink but the field remains selected & blue? a lil weird, not sure if this is something intentional

esp/public/media/scripts/custom_form.js
```diff
@@ -865,6 +868,8 @@ var onSelectField=function($elem, field_data, ftype=null) {
 	//Select the current field and category in the field and category selectors
 	if(ftype in formElements['Generic']){
 		$j('#cat_selector').val('Generic');
+        // Don't allow changing a field into a page or section (instructions are ok, though)
+        disabled_fields=['page', 'section'];
```
kkbrum, line 872: So instead of disabling the option in the drop down list for the field type, you can still click section or page, but then once you do that, both the field category and field dropdowns are entirely disabled. THis means you can't undo what you just did and select a different field category or field type without unselecting and reselecting the item in the preview. Is it possible to disable the section and page options within the drop down before you click it instead of disabling the whole dropdown afterwards? or if you cant do that, when someone selects a bad option, then leave the field category box enabled so they can at least switch back to an allowable category? this still wouldnt allow them to switch to instructions without switching the category away and back though, so disabling individual options within the dropdown would still be preferable.

kkbrum:
^ most recent two commits also look good.

willgearty:
OK, the newest three commits do the following:

- If you select a page/section and the selected field/section is not within it, it is deselected.
- Fields can now be moved to any section. Sections can now be moved to any page. Pages can now be reordered.
- You only need to click once to select a page.

kkbrum (approved):
Fancy!

PR #3359, opened 2021-06-25
Add texescape to program names in tex documents

### LINKED ISSUE #3340: Escape underscores in program names
If a program name has a "_" in it (eg. BC_Splash), the student schedule tex file breaks (just add "|texescape" after the program nice name) to fix it. We should check that program names are properly encoded everywhere in case this is causing other issues now / in the future.

kkbrum:
Adds texescape to program names in case they have underscores in them. Fixes #3340.


hwatheod (approved):
We should really `texescape` everything, not just the program name. But that can be for another day.

PR #3365, opened 2021-06-27
Fix rendering of multiple pages for custom forms

### LINKED ISSUE #3360: prev step buttons are broken for customforms
When you click the button, it seems to just submit the form, and if you haven't included required fields, you get an error.

### LINKED ISSUE #3361: Custom form modules don't support multiple pages
Right now, all of the custom form modules (student custom form, teacher custom form, and teacher quiz) only support a single page of a custom form. You can have multiple pages for the form, but they will only show the first page, and will accept answers to only the first page.

willgearty:
This fixes how we render custom forms with multiple pages when they are rendered as part of custom form modules (teacher, student, or teacher quiz). There is already rendering/validation built-in to the formtools stuff we're using, so instead of reinventing the wheel, we just use slightly modified versions of that (like how we render custom forms for the `/view/[id]` view). I also fixed the templates so the next/previous page buttons now work properly. Finally, I removed two old teacher quiz templates that were basically just slightly modified versions of custom_forms.html and also removed the teacher quiz controller which basically did nothing. Overall, this should result in a well-oiled machine for custom forms and should make it very easy to make new custom form modules in the future, if needed.

Fixes #3361 and fixes #3360


kkbrum (approved):
Yay!

Issue #3364, opened 2021-06-27
Changing text field to multiple choice field doesn't remove old attributes


willgearty:
If you change a text field in an existing custom form to a non-text field (e.g., radio button field), it doesn't remove the old 'charlimits' attribute, causing an error when the form is rendered.

PR #3366, opened 2021-06-28
Clean up old attributes when modifying custom form fields

willgearty:
We now clean up unnecessary attributes when we save the modified form.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3364.


kkbrum (approved):
Wow!

Issue #3223, opened 2021-02-21
Easier way to add/remove individual users to/from groups


willgearty:
Currently you have to go into the admin pages to change a user's group affiliations. I think it would be great if it could be done somewhere on the front end, ideally the userview page.

Requested by Yale at SplashCon 2021.

PR #3368, opened 2021-06-29
User groups module

### LINKED ISSUE #2007: Create a Group from a user list
Add an interface for creating a new Group of users from an arbitrary user list, either a list generated by the "Arbitrary User List" functionality, or just a copy/pasted list of usernames.  This is useful when known groups of students (for example, for Stanford, outreach students) need special permissions like extended deadlines, etc.

(Right now, this can only be done either one-by-one per student through the admin interface, or needs someone with server access to run a script to add a whole list.)

willgearty:
This adds a module that lets you add or remove users from a user group. You can also create a new group in the process. While I was at it, I cleaned up the group text module, which had various small bugs.

Fixes #2007 and fixes https://github.com/learning-unlimited/ESP-Website/issues/3223.


esp/esp/program/modules/handlers/usergroupmodule.py
```diff
@@ -0,0 +1,124 @@
...
+            raise ESPError(), "Your query did not match any users"
+
+        group, created = Group.objects.get_or_create(name = group_name)
+
+        if clean:
+            group.user_set.clear()
+
+        group.user_set.add(*users)
+
+        message = "%i users have been added to user group %s. User group %s now has %i users." % (users.count(), group.name, group.name, group.user_set.count())
```
kkbrum, line 117: It might be helpful to do something similar to what we do with checkins, and state how many of the users we tried to add were ALREADY in the group, how many we added, and how many there are total?

kkbrum (approved):
Very fun!

Issue #3358, opened 2021-06-21
Add multiselect to tags for hiding form fields


willgearty:
Right now you need to specify a list of the field names, but most admins won't know how to find them (and it would likely end up breaking the form). Instead, we should try to extract the field names from the forms and display them as a multiselect widget in the tag settings. I'm mostly thinking about the `teacherreg_hide_fields` and the `teacher_profile_hide_fields` tags, but it's possible there are more.

Issue #3357, opened 2021-06-21
Strip values for teacherreg_hide_fields tag


willgearty:
We did this for the `teacher_profile_hide_fields` tag in https://github.com/learning-unlimited/ESP-Website/pull/2958, but we never did it for the `teacherreg_hide_fields` tag. If you include spaces, it currently breaks the teacher class registration page.

PR #3370, opened 2021-07-02
Show options for tags that hide form fields

willgearty:
This shows all of the fields that could be hidden for each form for the tags that let admins hide the fields, hopefully making it much easier for these tags to be used:

image

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3358 and fixes https://github.com/learning-unlimited/ESP-Website/issues/3357.

### DIFF
--- a/esp/esp/program/forms.py
+++ b/esp/esp/program/forms.py
@@ -423,7 +423,23 @@ class TagSettingsForm(BetterForm):
             if tag_info.get('is_setting', False):
                 self.categories.add(tag_info.get('category'))
                 field = tag_info.get('field')
-                if field is not None:
+                # Some field widgets need to be setup manually because we can't do it during compilation
+                if key == 'teacher_profile_hide_fields':
+                    from esp.users.forms.user_profile import TeacherProfileForm
+                    self.fields[key] = forms.MultipleChoiceField(choices=[(field, field) for field in TeacherProfileForm.declared_fields.keys()])
+                elif key == 'student_profile_hide_fields':
+                    from esp.users.forms.user_profile import StudentProfileForm
+                    self.fields[key] = forms.MultipleChoiceField(choices=[(field, field) for field in StudentProfileForm.declared_fields.keys()])
+                elif key == 'volunteer_profile_hide_fields':
+                    from esp.users.forms.user_profile import VolunteerProfileForm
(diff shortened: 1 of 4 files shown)


kkbrum:
So this looks great, and technically functions perfectly. But:

1) it allows admins to set themselves up for failure. You can remove important fields such as "title" or "duration" or "number of sections" from teacher reg, but then when you try to register a class, you'll never be able to succeed: 

image

Thus we should filter out the required fields from the dropdowns. 

2) Perhaps we want to display the label of the field names, ie "Maximum Number of Students" instead of "class_size_max". Right now it looks like most of the coded field names are self-explanatory, but it may not be the most user friendly in the future. But the labels may also be too long to be pretty in a dropdown. Additionally, you can override the label for some of the fields which may be a bit weird, but I think we would just display the current label (that may get changed when you hit submit on the tags page). Up to you if you want to just leave the coded field names as is, or try to make it a bit friendlier.

willgearty:
OK, now we only show fields that are NOT required (e.g., this results in only 6 field options for the teacher reg form tag). I also changed the labels for the teacher reg form tag to the labels used in the form, but we don't have built-in labels for the profile form fields (we specify the labels in the templates instead), so I kept the field names for those tags.

kkbrum (changes requested):
This looks good for teacherreg. Can we do the same thing with using label text instead of the code names for the drop downs for volunteer/guardian/student/educator profiles?

Additionally, this might be for another PR, but it seems like we can deprecate a bunch of tags due to this hide feature. For example, there are tags asking about whether to include shirt size, shirt type, gender fields in various profiles, which are now redundant to the hide fields tag. I'm a fan of getting rid of all these extra tags now.

willgearty:
@kkbrum For the profile forms, we don't actually have labels defined for any of the fields. We just put labels in the templates for the profiles.

I'll look into which tags can be deprecated.

willgearty:
OK, so the tags you are referring to actually SHOW those fields (they are hidden by default). If you want to get rid of those tags, we'd either need to

1. Make them the default values for this tag or
2. Make them not hidden by default

kkbrum:
Okay that's fine about the labeling. I have some more thoughts about the shirt/gender tags that I'm going to put in a new PR.

kkbrum (approved):

Issue #3294, opened 2021-05-14
Add correct answers to more fields in custom forms


willgearty:
See request from @kkbrum here: https://github.com/learning-unlimited/ESP-Website/pull/3273#discussion_r631489832
It's unclear if there aren't more fields with correct answer options due to technical limitations or dev laziness.

willgearty:
Update: it was dev laziness

hwatheod:
Is the purpose of "correct answers" for grading teacher logistics quizzes and the like ?

willgearty:
"Correct answers" are set as field validators, and if a field is required, the form can only be submitted if the correct answer is provided to the field. If you don't require a field, "correct answers" are actually meaningless. They are also not shown (currently) anywhere in the results interface.

PR #3372, opened 2021-07-11
Correct answer option for more custom form field types

willgearty:
This adds the ability to specify correct answers for more types of custom form fields. Also, this makes it much more intuitive to specify correct answers when using radio, dropdown, multiselect, and checkbox fields:

image

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3294.


kkbrum (changes requested):
- [ ] Boolean field doesn't have a correct option
- [ ] If you submit a form that has a field with a correct answer but the field isnt required, the form won't update correctly and there's an error in the log, but no error message on the actual site.
image

- [ ] For the numeric field, when filling out the form, if i put the correct answer, it's still telling me it's incorrect?
- [ ] ^Same for date , time, drop down, and radio fields. perhaps i screwed up my form somehow to make this the case...

willgearty:
For the boolean field, since you need to make a field required in order for it to do the correct answer validation, the only correct option for a checkbox is for it to be checked (if a checkbox is required, it needs to be checked). So it doesn't make much sense to have a correct answer for a boolean field. If there is a boolean question that needs a correct answer, a yes/no question should be used.

I just was able to successfully save a custom form with a non-required field with a correct answer. Maybe the problem you encountered is more specific?

willgearty:
Oh, I just tried saving a non-required numeric field with a correct answer and it failed.

willgearty:
But then I just tried it again and it worked...

willgearty:
OK, I've fixed numeric, date, and time fields (and dropdown and radio already worked for me).

kkbrum:
- [ ] Things were going pretty well until.... so i have this successful form below, which works and i can hit modify successfully. I can also add a field successfully. However, if i now delete a field, i think any one of them, i get an error message (only in the console, the page just silently doesn't modify)

image

image

- [ ] my yes/no correct answer also appears broken (when i fill it out, tells me its still wrong no matter what)

willgearty:
@kkbrum, I've fixed a few things with the recent commits. Could you try again and let me know what error(s) you get?

willgearty:
OK, I think that should actually fix the errors you were (presumably) encountering. Try with a new form and let me know if there still any weirdness @kkbrum.

willgearty:
I also decided to implement a change that makes it a lot easier to pick the correct answer (no more specifying indices) for some fields.

kkbrum:
Welp. Everything submitted fine; when i went to fill out my form I got this. It's either the multiselect or the checkbox group or both that's breaking:
image

willgearty:
Whoops, my mistake! Should be fixed now!

kkbrum (changes requested):
welp. i broke it by including a date field, that is not required, but has a correct answer, and then submitting a form without entering an answer. image

willgearty:
OK, that should fix date and time fields (you will probably need to delete the fields and make new ones or even just make a new form) and a couple other fields that seemed to have validation errors (there's still an issue with the null boolean field where it doesn't validate if you answer "Unknown", but that seems to be a django design choice.

kkbrum (approved):
Magnificent!

Issue #3326, opened 2021-05-29
Update class status on dashboard when it is changed


willgearty:
On the dashboard, you can click the "status" button for a class and there are options to "Approve", "Unreview", or "Reject" the class. When you click these buttons, the class status changes in the database, but this isn't reflected anywhere on the dashboard. We should update the coloring of the class title and should update the text in the popup to reflect the class's new status.

PR #3374, opened 2021-07-13
Update popup status and dashboard styling when changing class status

willgearty:
This makes it so that the styling of class titles and the class statuses in the status popups always match the database (especially if we've just changed the status using the status popup).

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3326.


kkbrum (approved):
Spectacular!

Issue #2875, opened 2020-01-11
Repeated class titles for longer classes are confusing on student schedules


willgearty:
If a class is more than one timeslot long, the title will be repeated on a student's schedule for each additional timeslot followed by "(continued)". This has been reported by numerous chapters to be very confusing to students, who often think they are enrolled in the same class multiple times. This occurs on printed schedules and on the inline schedule on the student registration main page.

SplashCon attendees suggested that a good solution would do the following:

1. Alter the repeated text. Some suggestions include:
    1. remove the title completely and modify the remaining text to be clearer that the other timeslots are a continuation of the class
    2. leave the title, but modify the grammar/etc (e.g. put the parentheses around the entire text, not just "continued")
2. to style the set of timeslots together to show it is all one class. Some suggestions include:
    1. Make the border between the timeslots dashed instead of solid
    2. Add indications on the side of the schedule to notate complete classes (e.g. alternating black and white bars)

Reported at SplashCon 2020.

willgearty:
Related to https://github.com/learning-unlimited/ESP-Website/issues/50. Since we have buttons (in FCFS) for students to see classes for each of the timeslots, we don't want to collapse multi-hour classes into a single row.

Issue #50, opened 2011-11-23
Merge multi-hour class display


aseering:
When we display multi-hour classes in some places (student and room schedules), we show one block per hour.  We should merge them to one row per class.

PR #3376, opened 2021-07-16
Better schedule styling for continued classes

willgearty:
This changes the way we handle multi-block classes/sections in student and room schedules. For student schedules (webapp and student reg mainpage), the sections now span all of the timeblocks over which they occur. For room schedules, by default they are collapsed like they are in the student and teacher schedule printables. There is also now an option to print room schedules with all timeblocks listed ("fully expanded"), which tells whether a room is available or not whenever there isn't a scheduled class. See below for screenshots.

I also checked the student schedule email, student schedule printable (LaTeX and HTML), and the 2-phase lottery pages, and those don't need any changes.

Also also, I duplicated some of the styling changes to the student webapp schedules to the teacher webapp schedules while I was at it.

Finally, I renamed the room schedules template file because why would anyone name it "roomrosters.html" instead of "roomschedules.html"?

Student reg mainpage schedule:
image

Webapp schedule:
image

Fully expanded room schedules:
image

Fully collapsed room schedules:
image

Fixes https://github.com/learning-unlimited/ESP-Website/issues/2875 and fixes https://github.com/learning-unlimited/ESP-Website/issues/50.


kkbrum:
I see you did the multicolumn approach to having a weird conflicted schedule. However, the formatting can get thrown off. Here I have 2 two hour classes and 1 one hour class all within 2 class blocks and it's not very happy.

image

kkbrum (approved):
Ignoring being enrolled for multiple classes in a chunk of times that are different lengths, everything looks great!

PR #3377, opened 2021-07-19
Switch to using required HTML attribute instead of class

willgearty:
As we've discovered through many iterations of this functionality, Django doesn't do a good job of updating the classes associated with a field when the field is modified as part of the `__init__` function (https://github.com/learning-unlimited/ESP-Website/pull/2427, https://github.com/learning-unlimited/ESP-Website/pull/2806, https://github.com/learning-unlimited/ESP-Website/pull/3086). However, as of Django 1.10, it does a very good job of making sure the `required` HTML attribute is included if a field is required (and vice versa). Therefore, when doing things like adding asterisks to required fields (https://github.com/learning-unlimited/ESP-Website/pull/2420), instead of looking for the `required` class which can be misleading, we should look for the `required` attribute which should also be truthful.

To this end, I've updated the code in the profile template to make sure the asterisks will behave properly for existing and new fields. I've also updated our CSS code to make sure we are styling the correct fields. Finally, I updated one of our custom widgets which wasn't including the required attribute (because of a reasonable change django folks made to select widgets). I also fixed the formatting for the text message field.

This might be a little premature, but I declare that this pull request solves, once and for all, our nerve wracking issues with marking required fields in forms.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/2435.

### DIFF
--- a/esp/esp/themes/theme_data/fruitsalad/less/main.less
+++ b/esp/esp/themes/theme_data/fruitsalad/less/main.less
@@ -861,9 +861,9 @@ table.plaintable th {
 #content th.small { font-size: .8em }
 #content label.header { font-weight: bold; }
 #content select, input[type=text], input[type=url], input[type=email], input[type=number], input[type=file], input[type=password], textarea { border: 1px solid #b7b7b7; }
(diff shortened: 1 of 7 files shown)


kkbrum (approved):
Revolutionary!

Issue #3331, opened 2021-06-02
Specify program when linking to /userview


willgearty:
We often link to a user's userview page from various /manage pages (e.g., /manageclass and /ajax_scheduling). However, when you get there, at least if you are managing an older program, the links on the right will be for the user's most recent program. We should modify all of these /userview links to specify the program of interest, that way the links are more relevant.

PR #3378, opened 2021-07-20
Add program parameter to userview links

willgearty:
This adds the `program` GET parameter to any userview links that are on program-specific pages. This way, when we go to the userview page via these links the program we are most likely interested in is already selected in the dropdown.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3331.


kkbrum (approved):
Beautiful!

PR #3380, opened 2021-07-21
Fix /manage/mergeaccounts

### LINKED ISSUE #3379: /manage/mergeaccounts is broken
I didn't even know this exists, but there is an interface to merge accounts. It moves all related fields of a user to another user, sets up a UserForwarder, and optionally deactivates the old user. Unfortunately, a link to this interface (at /manage/mergeaccounts) was never added anywhere, so no one ever knew about it and no one ever maintained it. So now we need to fix it because of Django changes: https://docs.djangoproject.com/en/1.11/ref/models/meta/#migrating-from-the-old-api

willgearty:
This fixes the functionality for /manage/mergeaccounts which broke due to the upgrade to Django 1.11. I also fixed some of the text and styling and added a couple extra options for the merge and a link from /manage/programs.

Fixes #3379.


hwatheod (approved):
I had no idea this functionality existed either. The merge code looks so simple too. I would have thought merging accounts is a difficult process, but I guess Django manages things nicely to make it easy.

PR #3381, opened 2021-07-22
Add scheduler support for timeslots of any length

### LINKED ISSUE #2705: AJAX Scheduler has bad class duration assumptions
While fixing #2696, I noticed that the AJAX scheduler makes some pretty bad assumptions regarding the lengths of sections and timeslots. I'm pretty sure these assumptions include (although I'd be very happy to be proven wrong):

1. Timeslots aren't longer than an hour
2. There isn't more than one timeslot in a single hour
3. `ceil(sec.duration) == len(sec.get_meeting_times)`
4. All timeslots are the same length

An example of this can be seen here, where the code is highlighting which cells are valid for a section to be scheduled in:
https://github.com/learning-unlimited/ESP-Website/blob/77dbf1a95d3d02de9813be93c614d9140bcfdf7b/esp/public/media/scripts/ajaxschedulingmodule/ESP/Matrix.js#L96-L119

It determines which cells are valid for scheduling by counting along until it gets to `section.length`. I was able to abuse this by making timeslots that were 45 minutes long and a class that was 1 hour 55 minutes long. The scheduler was totally fine with letting me schedule this class across the two 45 minute timeslots (although I've now made it such that this will be caught in the scheduling diagnostics with #2704).

willgearty:
Previously, parts of the ajax scheduler code assumed timeslots were once per hour, never longer than an hour, and always the same length across a program. This removes those assumptions and allows for timeslots of any length (possibly varying across a program) and any number per hour.

Given that this now supports timeslots that may be longer than the scheduled classes (e.g., you can now schedule a 2 hour class across two 1.5 hour timeslots, which would result in a large time gap between the end of the class and the end of the second timeslot), this puts a little extra responsibility on the admins that are scheduling the classes. To hopefully help with this, the "length" on the hover and infopanel tooltips is now shown as "X hours and Y minutes".

Fixes #2705.

### DIFF
--- a/esp/public/media/scripts/ajaxschedulingmodule/ESP/Cell.js
+++ b/esp/public/media/scripts/ajaxschedulingmodule/ESP/Cell.js
@@ -323,7 +323,14 @@ function Cell(el, section, room_id, timeslot_id, matrix) {
         tooltip_parts['Teachers'] = this.matrix.sections.getTeachersString(this.section);
         if(has_moderator_module === "True") tooltip_parts[moderator_title + 's'] = this.matrix.sections.getModeratorsString(this.section);
         tooltip_parts['Class size max'] = this.section.class_size_max;
-        tooltip_parts['Length'] = Math.ceil(this.section.length);
+        var length_str = '';
+        if(Math.floor(this.section.length) > 0){
+            length_str += Math.floor(this.section.length);
+            length_str += ' hour';
+            if(Math.floor(this.section.length) > 1) length_str += 's';
+        }
+        if((this.section.length % 1) * 60 > 0) length_str += ' ' + ((this.section.length % 1) * 60) + ' minutes';
+        tooltip_parts['Length'] = length_str;
         tooltip_parts['Grades'] = this.section.grade_min + "-" + this.section.grade_max;
         tooltip_parts['Room Request'] = this.section.requested_room;
         tooltip_parts['Resource Requests'] = this.matrix.sections.getResourceString(this.section);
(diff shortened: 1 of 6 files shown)


kkbrum (changes requested):
Hm well the scheduler and I don't get along very well, but I'm having some undesired behavior.

It's (sometimes?) allowing me to schedule longer classes in fewer timeslots than should be allowed:

image

or even worse:

image

willgearty:
OK, I believe that should be the last of the code that makes horrendous assumptions about timeslot lengths/times. Please try to break it again!

hwatheod:
I guess this isn't the place to address this issue, but how do we ever end up with a non-integer number of minutes for a class length?  No one schedules class times at a sub-minute level. If we are storing class durations as fractional hours, perhaps we should store all class durations as an integer number of minutes instead.

willgearty:
> I guess this isn't the place to address this issue, but how do we ever end up with a non-integer number of minutes for a class length? No one schedules class times at a sub-minute level. If we are storing class durations as fractional hours, perhaps we should store all class durations as an integer number of minutes instead.

Yeah, the problem stems from trying to store all durations as fractions of hours (as a `DecimalField` with only 2 digits), then converting back to minutes. We've had to deal with this rounding issue several times now (#2771, #2704, #2668), and the solution to now has basically been to always round things the same way for other time-based parts of the website. My most recent commit rounds the times to a whole number of minutes (at least visually), and I didn't notice any actual rounding issues when comparing the fit of classes to timeslots (at least with the specific example @kkbrum demonstrated here).

It's not a bad idea to store duration as an integer of minutes...Although it would certainly take a lot of work to make sure everything is updated appropriately across the codebase.

kkbrum (changes requested):
This mostly looks great. But I do have an issue where I can't put a 34 minute class into a 34 minute timeslot. Presumably another rounding issue? But I'm able to put a 17 minute class into a 17 minute timeslot...

image

kkbrum (approved):
Woohoo!

PR #3382, opened 2021-07-30
Add many filter options to classsearch

### LINKED ISSUE #3373: Add more /classsearch options
Some things that could be useful for the /classsearch options:

- duration
- number of sections
- presence of comments to directors
- class format
- presence of planned purchases

willgearty:
This adds the following filters to the classsearch page:

- class duration
- class grade (as a range)
- number of sections (as a range)
- class capacity (as a range; works for all capacity types)
- has a message for directors
- has prerequisites
- has a room request
- has a purchase request
- class style
- class hardness

The complete filter options are now as follows:
image

The range filters look like this:
image

Fixes #3373.

### DIFF
--- a/esp/esp/program/models/__init__.py
+++ b/esp/esp/program/models/__init__.py
@@ -1041,7 +1041,7 @@ class Program(models.Model, CustomFormsLinkModel):
 
         durationList = durationDict.items()
 
-        return durationList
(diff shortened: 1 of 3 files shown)


kkbrum (approved):
Extraordinary!

Issue #3245, opened 2021-03-24
Message panel hidden on ajax scheduler


willgearty:
Looks like at some point the message panel became hidden on the scheduler and doesn't appear even if there is an error message.

willgearty:
Hmm...maybe it's only hidden when a class is selected?

willgearty:
Right, so I _think_ the issue was if you encounter an error trying to schedule a class, the class stays selected, so the error message isn't shown (presumably this also happens for scheduling moderators). If we encounter an error, should we unselect the class/moderator so the message panel is shown? But if the class/moderator is unselected, that seems to imply it was successful? Tagging @kkbrum since she's been testing the scheduler a lot lately.

kkbrum:
Could we put an error message above the class info in the panel?

PR #3383, opened 2021-07-31
Fix scheduler message panel

willgearty:
This fixes the message panel on the ajax scheduler so that it is shown, even if there is a class or moderator selected. I've also made it so that errors are displayed in red and success messages are displayed in blue.

image

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3245.


kkbrum (approved):
Nice!

Issue #3218, opened 2021-02-21
Mass deactivation tool


willgearty:
It would be convenient to have a module that you could use to deactivate a large number of users at once (ideally through the UserSearchController).

Requested by Yale at SplashCon 2021.

Issue #3224, opened 2021-02-21
"Banned" status


willgearty:
It would be useful to be able to give users a "banned" or ("semi-banned") status, either through some sort of comment/tag system or through groups.

Requested by Yale at SplashCon 2021.

willgearty:
I'm not sure how integrated we'd want this status to be with various modules in registration (as it would probably take a bit of work to make sure it's integrated with everything).

kkbrum:
Does this have advantages over just deactivating the account?

willgearty:
Well, one drawback of deactivating someone's account is that they can reactivate it if they try to login in the future (I believe). This would be admin-facing only, so the user couldn't change the status.

Supriyo760:
hey @willgearty 
I've analyzed the current ESPUser model and registration flow, and I can confirm that standard deactivation (is_active=False) is indeed reversible by users if they trigger a re-activation email.
To implement a persistent, admin-only "Banned" or "Semi-banned" status, I propose a two-pronged approach that leverages existing patterns in the codebase:

1) Group-Based Roles: We can add a Banned or Restricted Group. This is easy to check globally using user.hasRole('Banned'). We can then update the core registration modules to block or flag users in these groups during the registration process.
2) Audit Records: To support the "semi-banned" request (and provide history), we can define a new RecordType called user_restriction. This allows admins to document the reason for the restriction, the duration, and which administrator applied it, creating a permanent audit trail on the user's profile.

This approach avoids the need for a database migration on the auth_user table while providing the granular control requested by Yale at SplashCon."
Does this look good to you, or would you like me to refine the technical details?

PR #3384, opened 2021-08-01
Mass deactivation module

willgearty:
This adds a module that lets you use the UserSearchController to pick a set of users and deactivate them all at the same time.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3218.


hwatheod:
I'm wondering what is the use case for this, and if there should be some maximum numbers of users as a fail-safe. Since we don't have a "mass re-activate", if someone makes a mistake with this, it's going to be hard to undo.

willgearty:
I think the use case that Yale described at SplashCon was to deactivate really old (in age) student users or maybe a bunch of duplicate accounts for the same user (related to https://github.com/learning-unlimited/ESP-Website/issues/3224). I think in both cases, the admins just wanted a way to make sure those users didn't receive emails (technically, both cases can be covered by comm panel filters, but it's very easy for admins to forget, which often results in many "unsubscribe me" replies).

It shouldn't be too hard to add a "Mass Reactivate" option to this module if that would be useful, since you are correct @hwatheod that it would be a lot of work to manually undo a mass deactivation of lots of users (although it's possible the USC can't select deactivated users; I'll need to look into that).

willgearty:
Another possible addition could be an "Undo" button afterwards that would allow you to reactivate everyone you just deactivated (although I'm not sure this is more useful than the already existing count and confirmation checkbox on the confirm page).

kkbrum (changes requested):
- [ ] For the commpanel, if you go to filters, you can get a recipient checklist. I don't see that option here. Could we add that? I think it could be useful.
- [ ] In step 2, where it says "You have selected _ active users", could we add a sentence stating what query they made (as we do in the commpanel)? This helps avoid mistakes imo. I also think the number of users could be bolded.
- [ ] If there wasssssss a way to somehow make some sort of record of who has been deactivated through this module and a timestamp, I think it would probably be a very smart idea. No mass reactivation is probably fine, but it does seem smart to at least somehow have a list of users who have been mass deactivated so you can have a record of it. Although I suppose it probably gets sent to the logs? but you might forget when it was done so that's pretty inaccessible to check at a later date.

willgearty:
OK, I've added the recipient checklist and the query info just like in the comm panel (bonus: I added the module names to the top of the checklist page). There isn't really an easy way to add records like you are describing. We'd probably need to add an entire new model for that. It's probably not a perfect substitute, but I did add logging for the list of user IDs, just in case we want to go back and try to revert manually.

kkbrum (approved):
Excellent!

Issue #3277, opened 2021-05-03
Cache emails data for /manage/emails


willgearty:
If your program sends lots of emails, the /manage/emails page will take a long time to load. Given that the data for this page shouldn't be updated often (outside of when emails are in the process of being sent), we should cache most of the data.

willgearty:
Since we allow for admins to specify a custom start date, it'd be hard to cache all of this. Maybe only cache the page for the default date?

willgearty:
Since we allow for admins to specify a custom start date, it'd be hard to cache all of this. Maybe only cache the page for the default date?

PR #3385, opened 2021-08-02
Cache the emails view

willgearty:
This caches `/manage/emails` which shows the status of comm panel emails. The cache is triggered any time any message request of text of email is created/updated. Normally we'd want to try to filter this dependency down to something more manageable, but since we are always including the most recent emails, and any message requests or texts that will be modified will most certainly be very recent, this should be fine. This should also set a separate cache for each `start_date` (which means the cache will unfortunately be reset every day, since the default changes every day).

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3277.

### DIFF
--- a/esp/esp/program/views.py
+++ b/esp/esp/program/views.py
@@ -39,6 +39,8 @@ import traceback
 from operator import __or__ as OR
 from pprint import pprint
 
+from argcache import cache_function
+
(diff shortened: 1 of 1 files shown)


kkbrum (approved):
Seems fine? Maybe add a note to test this on dev3 when we do release testing.

Issue #2907, opened 2020-01-29
Palette adding and removing colors has wierd behavior. (Themes)


nrjean:
There is some really weird behavior with the palette in themes. It's hard to describe exactly what is going on, however here are some behaviors I was able to replicate.

1. Some base colors are impossible to remove
2. Removing more than one color at a time is impossible.

Compound this with the fact that you can only remove the "last color" and this means it is impossible to delete some colors from the website. (Which is a big pain if two colors are very similar to one another)

willgearty:
Can you give a set of instructions to replicate what you're trying to do? What is the expected outcome, and what is the actual outcome?

nrjean:
Mechanics:

1. Go to palette (in customize the themes)
2. Click remove last color more than once - or - click add another color more than once
3. Click update palette
This will result in no change to the palette

For the second issue of some colors being unremovable from the palette I don't really have a good way of replicating it, however - if you click remove last color enough times, you'll encounter a color that just won't go away

willgearty:
I think I've figured out what's going on here. It's a combination of the following:

1. We always include any LESS variable colors from the theme in the palette:
https://github.com/learning-unlimited/ESP-Website/blob/e5cd5acfc62fa22cc4f69d78a81b16c785134c86/esp/esp/themes/controllers.py#L568-L574

2. We sort the palette before it is displayed:
https://github.com/learning-unlimited/ESP-Website/blob/e5cd5acfc62fa22cc4f69d78a81b16c785134c86/esp/esp/themes/controllers.py#L577

3. As @nrjean stated above, you can only ever remove the last color.
4. If you remove ALL of the colors in the palette, it doesn't update the palette:
https://github.com/learning-unlimited/ESP-Website/blob/e5cd5acfc62fa22cc4f69d78a81b16c785134c86/esp/esp/themes/views.py#L217-L218

Therefore, in order to remove a custom color, you will need to remove a lot of colors from the palette (most of which will return because they are LESS variables for the theme).

I'm thinking of two separate changes to fix this:

1. For the palette editing section, separate the LESS variable palette from the custom palette. Alternatively, always put the custom colors AFTER the LESS variable colors (and maybe don't remove duplicates?)
2. Allow for removing any color from the palette, not just the last one.

Issue #2906, opened 2020-01-29
Big Picture Theme - Some theme settings seem to be hardcoded.


nrjean:
In the big picture theme the font for both headers and navbar is hard-coded and doesn't respond to theme editor changes. This is true for both font and color. Navbar does respond to settings regarding links however.

Issue #1342, opened 2014-10-05
Themes have hardcoded fonts


pricem:
Our themes don't do a good job specifying fonts.  For example, the "Georgia" font is hard-coded in several places in the main.less file for the circles theme.  This poses a problem for any chapter that would like to change the font on their site: The font control on /themes/customize (which selects the @baseFontFamily LESS variable) only affects some of the text on the site.  I think this is the text not covered by the theme, which gets Bootstrap default formatting (and hence the adjustable font).  Everything else gets the hard-coded font specified by the theme (in main.less).

We need to adjust the templates and LESS code in the circles theme to use one font everywhere, or add any auxiliary fonts as variables to be adjusted in the theme editor.

Reported by Janet from Duke in fall 2014.

PR #3388, opened 2021-08-04
Fix theme palette management

willgearty:
This fixes the management of a custom theme palette on the `/themes/customize` page. Since we can't change what the theme colors are, I've separated them out and now we only save the "Custom Palette" that is defined by the user:
image

If you somehow specify a color that is already in the theme palette or a duplicate of a color already in your custom palette, it's automatically removed. Also, you can now remove any colors from your palette (not just the last one) by clicking on them then clicking the "Remove" button:
image

I also updated the colorpicker (v. 1 -> v. 2), which is a little more visually appealing. While I was at it, I updated the font size and family pickers so you can actually see what the different font sizes and families look like when you are picking them:
image

Fixes https://github.com/learning-unlimited/ESP-Website/issues/2907.

### DIFF
--- a/esp/esp/themes/controllers.py
+++ b/esp/esp/themes/controllers.py
@@ -504,7 +504,7 @@ class ThemeController(object):
         if vars is None:
             vars = self.get_current_params()
         if palette is None:
-            palette = self.get_palette()
+            palette = self.get_palette()['custom']
 
         vars_orig = self.find_less_variables(theme_name, flat=True)
         for key in vars.keys():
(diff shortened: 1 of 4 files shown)


kkbrum (changes requested):
This all looks good, but while you're here...

- [x] Can you fix the divs so the customize theme page isnt cut off at the bottom in fruitsalad
- [ ] If you try to save a custom theme with a name that includes slashes, it breaks. Can we fail with an informative error at least or escape things so it doesnt fail?
- [ ] This is probably a known larger thing, but if I change the font, it changes in only very few places?

willgearty:
I've fixed the first two items, but the latter is definitely outside of the scope of this PR. It's also already been documented in https://github.com/learning-unlimited/ESP-Website/issues/2906 and https://github.com/learning-unlimited/ESP-Website/issues/1342. I've added them both to SR14.

kkbrum (approved):
Extraordinary!

Issue #3324, opened 2021-05-29
Not all management modules have a button to return to /main


willgearty:
Some management modules have a button that takes you back to the main management page:
![image](https://user-images.githubusercontent.com/7232514/120075714-06e90b80-c068-11eb-906f-afcb80b42e8a.png)

using this include:
`{% include "program/modules/admincore/returnlink.html" %}`

But some don't (e.g., deadlines, volunteer management). We should make them all have the button.

PR #3389, opened 2021-08-05
Add "Return to main program management page" links to manage modules

willgearty:
I _think_ this covers all of the management modules that didn't previously have links, but it's entirely possible I missed one or more.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3324.

### DIFF
--- a/esp/templates/program/modules/accountingmodule/accounting.html
+++ b/esp/templates/program/modules/accountingmodule/accounting.html
@@ -92,4 +92,5 @@ No grant
     <tr><td><strong>Owed</strong></td><td><strong>${{ due|floatformat:"-2" }}</strong></td></tr>
 </table>
 {% endif %}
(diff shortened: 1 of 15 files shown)


kkbrum (changes requested):
Missing links (some of these might not be possible, but some definitely are):
- [x] manage/1/2/commpanel
- [x] manage/1/2/studentschedules
- [ ] manage/1/2/catalog
- [ ] manage/1/2/ajax_scheduling
- [ ] manage/1/2/lunch_constraints has one in the middle of the page-- can we move it down?

willgearty:
OK, I've added the return link to all of those except for the scheduler. I wasn't really sure where it should go. Under the matrix, maybe?

kkbrum (approved):
Spiffy! I think we can let the scheduling page slide since it would probably be more in the way than helpful.

Issue #3242, opened 2021-03-20
Adding a coteacher doesn't check if teacher is available for scheduled time


willgearty:
When we attempt to add a coteacher to a class, we check whether the teacher has any conflicts with the class:
https://github.com/learning-unlimited/ESP-Website/blob/06d9835a4227c797b8a6678fa690fa816abdbf2e/esp/esp/program/modules/handlers/teacherclassregmodule.py#L592-L593

For a ClassSubject, this involves checking whether they have any conflicting scheduled classes and whether they have enough remaining availability to satisfy the total length of the class:
https://github.com/learning-unlimited/ESP-Website/blob/2dd7c8f3a406b8f5c87947b9c3c0be3be8558156/esp/esp/program/models/class_.py#L1753-L1784

However, one thing I realized is we don't make sure they are actually available at the times for which the class is actually scheduled!

I just tested this, and as long as a teacher has enough availability, you can add them as a coteacher, even if they aren't available for the class's scheduled time!

PR #3390, opened 2021-08-06
Check if a teacher is available for scheduled times before adding them as a coteacher

willgearty:
This makes sure that a teacher is available for all of the scheduled times of a class before adding them as a coteacher (through either the teach or manage version of the coteacher page). If the teacher is not available, we display their name and list the times that they are not currently available (and the manage version links to their edit availability page).

Speaking of which, I pulled out the coteacher logic into a static method, since it was basically duplicated in the handlers for the teacher class reg and admin class modules. I don't expect there to be another part of the codebase that will need to use the static method, but it will certainly make it easier for fixes and changes in the future.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3242.


kkbrum (approved):
Woot woot!

Issue #2880, opened 2020-01-12
Deadlines page is cluttered and overwhelming


willgearty:
Multiple admins at SplashCon 2020 noted that the deadlines page is very cluttered and overwhelming.  A potential solution was suggested that:

- Lists each user type/group as a single table
- Each row would reflect a permission that is set for that user type.
    - The easy way to do this would be to just include any permission (and the implied permissions), which could result in duplicates.
    - The trickier way to do this would be to collapse duplicate permissions/implied permissions; the tricky aspect of this figuring which permission you modify when the user wants to edit them
- Each permission would have the opening and closing listed (and a button to edit them; or maybe a calendar that you can edit inline, then click submit at the bottom of the page)
- Each type would then have a section to add new permissions

![image](https://user-images.githubusercontent.com/7232514/72212656-f0133a80-34ad-11ea-875c-5dda0cfc3613.png)

PR #3391, opened 2021-08-12
Redesign of the deadline management page

willgearty:
This is a nearly complete redesign of the layout of the deadline management page. This attempts to solve the following problems with the old layout:

- Didn't list closed deadlines that never had permissions (e.g., volunteer registration)
- Deadlines were listed seemingly randomly (both for deadlines of the same type and for the same role)
- Recursive/implied deadlines were confusing
- Couldn't create a deadline with the same permission type of a preexisting deadline (needed to do this through the admin pages)

So, the new redesign does the following to try to alleviate these issues:

- For all of the core roles (Students, Teachers, and Volunteers), all possible deadlines are listed. For other custom roles, all deadlines with permissions are listed.
- Deadlines are listed by their defined order in the models and are grouped by role.
- If a deadline is implied (included in a recursive deadline), it's status and start/end dates are shown, but no changes are allowed.
- Users can edit an individual permission or merely open or close a permission for a role (note, a permission can not be closed if it is implied open by another permission). This provides two levels of management: coarse open/close details for new admins and fine details for experienced admins.
- The "Create a Deadline" form hasn't changed, but it will now allow you to create any number of deadlines of the same permission type.

image

image

image

Fixes https://github.com/learning-unlimited/ESP-Website/issues/2880.


kkbrum (changes requested):
This looks wonderful. 

Some rambling thoughts:

- [x] I can't see a way to delete a deadline row? I know you can just close or open or change the dates such that the extra deadlines are not active, but I still think deleting a row should be possible (when there's another existing deadline for the field). Perhaps just if you empty both the start and end date boxes & there's another deadline for that field, it removes the deadline?
- [ ] I don't love that you still have to go down to the bottom to add a deadline, so I was thinking maybe we could put a little plus button on the right of the open / close buttons to add the deadline (then we could remove the create new deadlines section entirely). However, perhaps we do want to discourage creating new deadlines as opposed to editting the current ones, in which case maybe we do want to leave it more inaccessible at the bottom?
- [x] I think the ordering of the deadlines could still be improved some. For students, I think it should be chronological based on the typical flow of student reg, so i think phase zero should move way up, as should registration mainpage, and set profile info. I think extra costs could move down to right above pay for program, since the two are intricately related. I also think access to fin aid should go right below pay for program so that it's still with the access ones, but is close to the related payment things.
- [x] SImilarly, for teachers: review students' apps should go down to after teachers have created classes. Profile info and registration mainpage should go way up.
- [ ] Deadlines that relate to modules that are disabled should probably be either hidden entirely, OR just instead of being red/green, should be light gray? The only reason hiding them might not be good is for the phase zero crap where you enable and disable the module but still want to have deadlines prepared. But, these deadlines obviously do nothing while the module is disabled, so i think they should be gray. They might still feature the corresponding open/close button though.

willgearty:
OK, I've reordered the permissions and I've added delete buttons when a permission isn't implied and isn't the last row of a permission type.

I agree that we don't want to really encourage the adding of lots of permissions (although probably less of a problem now that they can be deleted from the page as well). Also, currently if you open a permission type that is closed because of an implication, it does create a new permission for you, and I think that is really the only case of creating permissions that we want to encourage for inexperienced users/chapters?

With respect to disabled modules, there isn't a super clean way for us to keep track of which modules use which permissions. We would probably need a manually curated dictionary or something, which would likely become out of date as more permissions are created. I agree that it would be nice to hide things that aren't relevant, but I just can't think of an elegant way to make it happen.

kkbrum (approved):
Wonderful!

PR #3392, opened 2021-08-13
Add moderator check

### LINKED ISSUE #3325: Add new ajax scheduling checks to scheduling checks module
Four new checks were added to the ajax scheduler as part of the moderator addition:

1. Number of moderators (per class)
2. Number of teachers (per class)
3. Number of moderators and teachers (per class)
4. Moderators with mismatched categories

Ideally, we'd also be able to have these checks as part of the scheduling checks module. The 4th one can probably just be ported over. However, the other three probably need to be modified slightly to be more useful in the module (e.g., "classes without a moderator").

nrjean:
Fixes #3325 

Introduces a scheduling check to look for moderators assigned to different categories than they selected.


kkbrum (approved):
Nice!

Issue #2546, opened 2018-04-08
Many non-unique ids on resources page


willgearty:
Doesn't seem to be causing any problems at the moment, but there are lots of non-unique HTML element ids on the resources page which should probably be replaced with unique ids (and we'll need to adjust any other code that depends on those former ids).

willgearty:
Looks like this is due to the automatic ids for fields in the forms (many of which have fields with the same names).

willgearty:
You can append/prepend strings to the auto-ids: https://docs.djangoproject.com/en/1.11/ref/forms/api/#configuring-form-elements-html-id-attributes-and-label-tags

PR #3393, opened 2021-08-13
Make resource form fields have unique ids

willgearty:
This makes it so we don't get little warnings in the browser console telling us that there are form fields with non-unique ids.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/2546.


kkbrum (approved):
Great!

Issue #3395, opened 2021-08-15
sorttable.js stops working on attendance page in some situations


hwatheod:
Under some conditions (still investigating), sorttable.js on the section attendance page stops working. When you click on the arrows in "First Name" or "Last Name", it doesn't sort anything and you see this error in the console log:

<img width="487" alt="sorttable.js error" src="https://user-images.githubusercontent.com/1224184/129487710-42b6a426-95d9-46b6-94c6-e3f111515397.png">

Possibly related to #3394.

willgearty:
Do you know which section this was for @hwatheod?

willgearty:
I've also noticed that sometimes the up/down arrows can appear in the cells of the first row of the table instead of just the header cells.

hwatheod:
Able to reproduce this on dev3 as follows:

1. Go to https://dev3.learningu.org/teach/Sploosh/current/section_attendance?secid=1182
2. Make sure both students (Draco Malfoy and Luna Lovegood) are not marked as attending. If necessary, unmark them and refresh the page.
3. By default, it is sorted by last name. So Luna should be listed first, and Draco second.
4. Click the arrow to sort by first name. Now Draco should be listed first, and Luna second.
5. Mark Luna as attending.
6. Click the arrow to sort by last name. It doesn't sort and you get the error above.

hwatheod:
> Do you know which section this was for @hwatheod?

This wasn't on the production site. I found this on dev3 while investigating #3394. 

On another project I worked on, I had problems with sorttable.js getting out of sync in certain situations if the table gets updated. I am suspecting the cause of #3394 is that sorttable put the check marks in the wrong places after sorting.

willgearty:
Ah, I think this error has something to do with the `<tfoot>` in the table. In the new version of sorttable.js, it makes the headers in the footer clickable for sorting columns:
https://github.com/learning-unlimited/ESP-Website/blob/450099579b87e24496cd74853ff7a77dafafed4a/esp/public/media/scripts/sorttable.js#L81-L83

If I comment out these lines, it seems to fix the issue you outlined above.

PR #3396, opened 2021-08-15
Keep tfoot "headers" from being included in sorttable.js

willgearty:
This fixes the behavior reported in https://github.com/learning-unlimited/ESP-Website/issues/3395 where the attendance tables would no longer be able to be sorted by specific columns. It appears to have been caused by interaction with the `<tfoot>`, so we just make sure the sorttable.js script doesn't use those "headers".

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3395.


hwatheod (approved):
Verified on my dev server that the issue was resolved.

PR #3397, opened 2021-08-15
More sorttable fixes

willgearty:
This fixes a bug where a sorttable with a single row with a `<th>` tag in it assumes that all of the `<td>`s in that row are also `<th>`s, causing them to have the arrows and be sortable. I'm honestly not sure what the purpose of the code in sorttable.js is to handle this single row case, but this seems like a better overall fix anyways (where we just always check that the header is a `<th>`).

While I was at it, I added some comments to sorttable.js to point out what local modifications I've made here and in the past. I also added the "nosort" class to a bunch of `<th>` tags that we don't want to be handled.

### DIFF
--- a/esp/public/media/scripts/sorttable.js
+++ b/esp/public/media/scripts/sorttable.js
@@ -43,7 +43,8 @@ var sorttable = {
 
 	addOnClick: function(info,cells) {
 		for (var i=0,ie=cells.length; i<ie; ++i) {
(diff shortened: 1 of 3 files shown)


hwatheod:
I think your description is missing some words " assumes that all of the s in that row are also s "

willgearty:
Fixed, silly markdown!

hwatheod (approved):

Issue #2728, opened 2019-04-22
Printable or big board of high enrollment classes by time block


willgearty:
It would be really cool to have a way to visualize (in a printable or big board-like format) which classes have the highest enrollment for each time block across the entire program. Maybe something like a stacked bar graph or something?

willgearty:
https://www.highcharts.com/demo/column-stacked

willgearty:
One concern from @kkbrum: do we care about the small classes? If so, we might want a better way to display them so they don't just get lost at the bottom of the stack.

PR #3399, opened 2021-08-18
Student big board upgrades

willgearty:
This makes two changes to the student big board (/bigboard):

1. Split the "signed up for classes" line into separate "set class lottery preferences" and "enrolled in classes" lines (for the example below I changed the arguments for `make_graph_data()`)
image

2. Changed the popular class information from lists to bar charts (through time), adding another axis of information to the visualization. The charts cover stars, first priorities, and enrollments. Each chart is only shown if we have data for that information type.
image

Fixes https://github.com/learning-unlimited/ESP-Website/issues/2728.


kkbrum (approved):
Fabulous!

Issue #3119, opened 2020-09-26
Text message number shouldn't be required to confirm student registration


hwatheod:
At the bottom of the student registration confirmation page, there is a box that says "Enter your cell phone number in the box below if you would like to receive a text message reminder in the week before Splash."  This suggests that you don't have to enter a phone number.  However, if you click "Confirm Registration" without entering a phone number, it doesn't allow you to confirm, and prompts you to "Please fill out this field."

<img width="588" alt="Screen Shot 2020-09-26 at 12 32 47 PM" src="https://user-images.githubusercontent.com/1224184/94349162-ea876180-fff6-11ea-8f34-d1766ab582de.png">

Tested on dev3.learningu.org.

willgearty:
I think that text message module can actually be deprecated, because we now have a text message check box in the profile form.

PR #3400, opened 2021-08-19
Remove TextMessageModule

willgearty:
This module has been deprecated for some time, since there are now questions about a student's cell phone number and whether they would like to receive text messages in the contact info form (as of 10+ years ago, lol). In one of the recent releases, I think this module accidentally got set to be enabled by default. Instead of just changing the default behavior, it should just be removed entirely to reduce any student and admin confusion.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3119.


kkbrum (approved):
Seems fine

Issue #3327, opened 2021-05-29
Volunteer timeslot import should have a confirmation page


willgearty:
When you import volunteer timeslots, it just imports them, without giving any sort of confirmation page, unlike when you import resources or classrooms.

PR #3402, opened 2021-08-21
Add confirm page for volunteer shift import

willgearty:
This adds a confirmation page (like those for importing resources) for when importing volunteer shifts from another program.

image

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3327.


kkbrum (approved):
Spiffy!

PR #3404, opened 2021-08-26
Add test coverage stats

### LINKED ISSUE #989: Add coverage stats for our tests
We should integrate a tool to give us stats on our test coverage so that we can determine what areas need more tests written. It looks like it's easy enough to run Django tests with coverage.py: https://docs.djangoproject.com/en/1.4/topics/testing/#integration-with-coverage-py.

willgearty:
This modifies a github action to assess the codebase coverage of our tests using the [coverage.py](http://coverage.readthedocs.io/) package. It then uploads the results to codecov.io. By default the coverage is posted as a comment to every PR, but I turned that off.

Fixes #989.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -50,12 +50,9 @@ jobs:
       env:
         TRAVIS_JOB: ${{ matrix.travis_job }}
       run: deploy/travis/script
-  codecov:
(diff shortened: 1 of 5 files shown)


kkbrum (approved):
Yay!

Issue #3406, opened 2021-08-27
Theme customizer has non-hex colors (by default?)


willgearty:
It appears that some of the colors on the theme customization page are not hex codes, but are something like @white (presumably a css or less variable). These get placed in the template and then cause the color selectors to break. When the color selectors (e.g., for the background color of the page) break, they default to black (#000000), and then when you save or test everything, the background color changes to black. We ideally would replace these css colors before displaying the page. I'm not really sure how this used to work in the past (maybe these settings weren't editable?).

Reported by @kkbrum while testing SR13-fixes on dev3.

PR #3407, opened 2021-08-27
Theme editor fixes

willgearty:
This fixes the following:

1. When loading one of the default theme customizations ("Default" or "Rupaa"), various theme colors were not loaded properly. When you then saved, this caused lots of things to turn black (including buttons and the background).
2. The default font for both default theme customizations was Copperplate Gothic (ewww).
3. Many fonts secretly had different font families as backups (now they all just have serif/sans-serif as backups).
4. Fonts were sorted alphabetically, which is nice if you know what you are looking for, but if you want to compare different fonts it's not terribly useful. Now they are sorted alphabetically within category (e.g., serif).
5. There wasn't any spacing between the color labels and the color pickers (and there were stray spaces).

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3406.

### DIFF
--- a/esp/esp/themes/controllers.py
+++ b/esp/esp/themes/controllers.py
@@ -533,6 +533,11 @@ class ThemeController(object):
         for match in re.findall(r'@(\w+):\s*(.*?);', data):
             vars[match[0]] = match[1]
 
+        #   Substitute LESS variables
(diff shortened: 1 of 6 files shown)


kkbrum (approved):
Looks great!

PR #3411, opened 2021-09-08
Upgrade to Python 3

### LINKED ISSUE #2576: Upgrade to Python 3
I realized we don't have an issue to track this!  Python 2 hits end of support on Jan 1, 2020, and Django 1.11 will be the last series to support it.  Also, python 3 is great and has lots of nice things in it.  This issue is to track our upgrade path.

### LINKED ISSUE #2820: Upgrade pillow to 7.1.0?
Apparently there's a low severity security vulnerability in all versions < 6.2.0. We're on 3.3.3, which means upgrading involves a LOT of changes. @hwatheod made the last upgrade to pillow, so maybe he's the best to figure out if we need to address any changes? My understanding is that it's only used for the imagefield in the teacher bios?

Changelog is here: https://github.com/python-pillow/Pillow/blob/master/CHANGES.rst

willgearty:
This is a very much work-in-progress PR to migrate from Python 2 to Python 3.

TODO:

- [x] Run modernize
- [x] Remove selenium package (now built-in to django)
- [x] Replace pydns with py3dns
- [x] Update django-argcache to python3
- [x] Encode before hashing
- [x] Upgrade xlwt (might need to replace with openpyxl)
- [x] Fix StringIO imports
- [x] Fix migrations (remove byte strings) ([python docs](https://docs.djangoproject.com/en/1.11/topics/migrations/#supporting-python-2-and-3))
- [x] Upgrade other dependencies (Twilio, flake9, stripe, pygments, pillow)
- [x] Replace \_\_unicode\_\_() with \_\_str\_\_() ([django docs](https://docs.djangoproject.com/en/1.11/topics/python3/#str-and-unicode-methods))
- [ ] Replace uses of `cmp()`, fix uses of `sort()` and `sorted()` ([python docs](https://docs.python.org/3.0/whatsnew/3.0.html#ordering-comparisons)) ([more python docs about `cmp` to `key`](https://docs.python.org/2/howto/sorting.html#the-old-way-using-the-cmp-parameter))
- [ ] Fix strings vs byte strings ([pythons docs](https://portingguide.readthedocs.io/en/latest/strings.html))
- [ ] Verify integer division: Make sure `/` (Python 2) has been changed to `//` (Python 3) in all the right places (I did a first pass on this in #3412 but should be double-checked)
- [ ] Many more things...

Fixes #2576. Fixes #2820.


esp/esp/themes/controllers.py
```diff
@@ -378,7 +378,7 @@ def get_file_summaries(self, dir):
                 result += self.get_file_summaries(full_filename)
             else:
                 file_data = open(full_filename, 'rb').read()
-                result.append((full_filename, os.path.getsize(full_filename), hashlib.sha1(file_data).hexdigest()))
+                result.append((full_filename, os.path.getsize(full_filename), hashlib.sha1(file_data.encode('UTF-8')).hexdigest()))
```
hwatheod, line 381: This file_data is a sequence of bytes (because it's read with 'rb' ), not a string. It should not be encoded with UTF-8. Python 3 will give an error.  You can just pass the file_data directly to the sha1 function.

esp/esp/users/models/__init__.py
```diff
@@ -2030,14 +2030,14 @@ def create_from_Q(item_model, q_filter, description = ''):
         dumped_filter = pickle.dumps(q_filter)
 
         # Deal with multiple instances
-        query_q = Q(item_model = str(item_model), q_filter = dumped_filter, sha1_hash = hashlib.sha1(dumped_filter).hexdigest())
+        query_q = Q(item_model = str(item_model), q_filter = dumped_filter, sha1_hash = hashlib.sha1(dumped_filter.encode('UTF-8')).hexdigest())
```
hwatheod, line 2033: pickle.dumps produces a sequence of bytes, not a string, and should not be encoded with UTF-8.  Same with all the other instances where you hash dumped_filter (I've only commented on this one).

hwatheod (changes requested):

hwatheod:
As a general comment: the whole text vs bytes issue is one of the most annoying parts of porting python 2 to 3, and can't be automated. See e.g. https://portingguide.readthedocs.io/en/latest/strings.html

hwatheod:
The latest failure seems to be a text/bytes issue with migrations, https://docs.djangoproject.com/en/1.11/topics/migrations/#supporting-python-2-and-3 and https://docs.djangoproject.com/en/1.11/topics/python3/

hwatheod (approved):

willgearty:
Hmm... That approval might be a little premature @hwatheod

hwatheod:
> Hmm... That approval might be a little premature @hwatheod

Yeah, that's not what I meant to do. I thought I was just approving the specific changes I requested. I haven't found a way to undo the approval yet.

esp/esp/qsd/tests.py
```diff
@@ -138,7 +138,7 @@ def testPageCorrectness(self):
             qsd_rec_new.save()
             response = self.client.get(self.url)
             self.assertEqual(response.status_code, 200)
-            self.assertTrue('Testing 456' in response.content)
+            self.assertTrue('Testing 456' in str(response.content))
```
hwatheod, line 141: In python 3, `str` doesn't convert from bytes to string in the way you think it would.  For example:

```
>>> u = b'xyz'
>>> print(str(u))
b'xyz'
>>> len(str(u))
6
```
The best thing would be to figure out some way to make `response.content` return a string rather than bytes in the first place so you wouldn't have to do this conversion in the first place. If there's no way to do that, then you can do `response.content.decode('UTF-8')` to do the conversion.

hwatheod: More generally: any time you convert from string to bytes or vica versa, you have to specify an encoding.

hwatheod:
Started a PR #3412 branched off this one with some Python 3 changes unrelated to strings/bytes

hwatheod:
Baffled by this failing test in Python 3:
```
 ======================================================================
FAIL: testNavbarBehavior (esp.web.tests.NavbarTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/runner/work/ESP-Website/ESP-Website/esp/esp/web/tests.py", line 108, in testNavbarBehavior
    self.assertTrue(self.get_navbar_titles('/') == ['NavBar1'], 'New navbar not showing up: got %s, expected %s' % (self.get_navbar_titles('/'), ['NavBar1']))
AssertionError: False is not true : New navbar not showing up: got [], expected ['NavBar1']
```
Can't see anything in the code that would be different between Python 2 and 3.

hwatheod:
Okay, I see it now, it's because of the `str(response.content)` in `get_navbar_titles`

All remaining test failures appear to be related to string/bytes, or sorting.

willgearty:
I can fix my str() changes tomorrow

esp/esp/qsd/tests.py
willgearty: Looks like the `response.content` is always bytes: https://docs.djangoproject.com/en/3.2/topics/testing/tools/#django.test.Response.content.

hwatheod:
Sorting related changes are in #3413

hwatheod:
Failing tests:
(1)
```
======================================================================
FAIL: testGradesStats (esp.program.modules.tests.jsondatamodule.JSONDataModuleTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/vagrant/devsite/esp/esp/program/modules/tests/jsondatamodule.py", line 112, in testGradesStats
    self.assertContains(self.stats_response, json_str)
  File "/home/vagrant/.local/lib/python3.8/site-packages/django/test/testcases.py", line 393, in assertContains
    self.assertTrue(real_count != 0, msg_prefix + "Couldn't find %s in response" % text_repr)
AssertionError: False is not true : Couldn't find '{"data": [{"grade": 7, "num_subjects": 10, "num_sections": 10, "num_students": 0}, {"grade": 8, "num_subjects": 10, "num_sections": 10, "num_students": 0}, {"grade": 9, "num_subjects": 10, "num_sections": 10, "num_students": 0}, {"grade": 10, "num_subjects": 10, "num_sections": 10, "num_students": 10}, {"grade": 11, "num_subjects": 10, "num_sections": 10, "num_students": 0}, {"grade": 12, "num_subjects": 10, "num_sections": 10, "num_students": 0}], "id": "grades"}' in response
```

Above is failing because of dictionary ordering change between Python 2 and 3. The "id" needs to be moved to the first position in the expected response.

(2)
```
======================================================================
FAIL: testNavbarBehavior (esp.web.tests.NavbarTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/vagrant/devsite/esp/esp/web/tests.py", line 113, in testNavbarBehavior
    self.assertTrue(self.get_navbar_titles('/') == ['NavBar1A'], 'Changes to navbar not showing up: got %s, expected %s' % (self.get_navbar_titles('/'), ['NavBar1A']))
AssertionError: False is not true : Changes to navbar not showing up: got ['NavBar1'], expected ['NavBar1A']

----------------------------------------------------------------------
```
I can't figure this second one out.

hwatheod:
Changing all None comparisons to use `is` or `is not` in #3414

hwatheod:
Script to run tests locally (from the VM):
```
export GITHUB_ACTIONS=0
export PYTHON_VERSION=$2
export TRAVIS_JOB=$1
export PATH=$HOME/.local/bin/:$PATH
cd $HOME/devsite
echo "Executing $TRAVIS_JOB with python version $PYTHON_VERSION"
deploy/travis/before_install && deploy/travis/before_script && deploy/travis/install && deploy/travis/script
```

For example: 
`script_name test 3.7`
`script_name lint 3.7`

hwatheod:
I spent a lot of time reading about pickling and trying many things (most notably the protocol=0 option for pickle which produces pickles in Python 3 that are compatible with Python 2), but things still don't work yet. 

However, I would like to draw attention to this old Django ticket entitled [Django 1.10 is unable to unpickle models pickled in Django pre-1.10](https://code.djangoproject.com/ticket/27187).  It ends with "The consensus ... seemed to be that cross-version pickling support wasn't something we wanted to try to do. I guess you could ask on the mailing list whether we should revisit that."

So whenever we upgrade Django, there is no guarantee that any of the pickled objects we are storing could actually be unpickled by a future version of Django. The problem is that pickled objects store the full names of Python classes, and sometimes these classes are internal Django classes that get dropped/renamed from one version to the next. 

I will keep trying to look for a solution, but to be honest, I am not super optimistic right now.

willgearty:
Thanks for looking in it @hwatheod! For these tests, aren't we using a fresh DB, though? So there shouldn't be any old pickled objects?
Either way, I've got the travis tests working localling on my computer, so I'll start taking a look at the other failing tests.

hwatheod:
> Thanks for looking in it @hwatheod! For these tests, aren't we using a fresh DB, though? So there shouldn't be any old pickled objects?

The failling tests are not related to old pickled objects. The reason I mentioned them is because if right now, we are really storing a bunch of objects we can't unpickle anyway, perhaps one option is to simply drop all the old pickled objects and start fresh with using `BinaryField` instead of `TextField`. That will be so much easier under Python 3, without having to mess with a migration. The main problem I ran into yesterday is that when you try to `save()` something which is a `TextField`, Django under Python 3 seems like it always tries to decode it as `UTF-8` before saving, and that doesn't work with pickled objects (even the ones created during the test), since they are actually binary.

willgearty:
Ah, that makes more sense, and will definitely be a headache to deal with otherwise...
Is the only field that stores a pickled object `PersistentQueryFilter`? If so, it looks like the only other model that uses this would be `MessageRequest` (as I mentioned last night). I think it's a required field, so we'd have to delete all of the old MessageRequests to make it work? (which means no email lookup in the comm panel)

hwatheod:
Long term it would be better to store such objects in JSON instead of pickle so that it will be portable across versions of Django.

hwatheod:
Okay, after some more research, I learned more about pickles and found a reliable way to read Python 2 pickled objects in Python 3: https://rebeccabilbro.github.io/convert-py2-pickles-to-py3/

I propose the following solution, which will allow us to preserve as many pickled objects as possible:

1. Change all fields representing pickled objects from `TextField` to `BinaryField`.
2. Do the following migration. For each pickled object,
    - attempt to unpickle it.  
    - If the unpickling fails, then give up on that object because it means the object was written by a version of Django prior to  1.11 so we can't recover it anyway. Probably we just drop that row entirely.
    - if the unpickling succeeds, the re-pickle it using the Python 3 pickle, and store it in the new `BinaryField`.

This should get us from Python 2 to 3 smoothly, but we will face the same problem again when we do another Django version migration. At that point, maybe we should try to convert objects to JSON (and maybe later versions of Django have functionality to do that) so we don't have to drop pickled objects for each new version of Django.

Does this sound reasonable?

willgearty:
That sounds like a good plan to me.

willgearty:
Looks like this is the cause of the last three errors: https://docs.djangoproject.com/en/1.11/topics/python3/#str-and-unicode-methods. I'll work through the codebase to update all of our __unicode__ methods.

hwatheod (commented):
Python 2 tests are failing immediately in `esp/esp/users/models/__init__.py` but this can be fixed by updating line 927 as follows (adding a `str` around the right side expression)

```
_new_method.__name__    = str('is%s' % str(user_class))
``` 

However, after making this fix, there are now a whole bunch of failures with this stack trace (in Python 2):

```
Traceback (most recent call last):
  File "/home/vagrant/devsite/esp/esp/web/tests.py", line 170, in setUp
    super(NoVaryOnCookieTest, self).setUp()
  File "/home/vagrant/devsite/esp/esp/program/tests.py", line 694, in setUp
    new_class.add_section(duration=settings['timeslot_length']/60.0)
  File "/home/vagrant/devsite/esp/esp/program/models/class_.py", line 1557, in add_section
    new_section.save()
  File "/home/vagrant/venv/lib/python2.7/site-packages/django/db/models/base.py", line 808, in save
    force_update=force_update, update_fields=update_fields)
  File "/home/vagrant/venv/lib/python2.7/site-packages/django/db/models/base.py", line 848, in save_base
    update_fields=update_fields, raw=raw, using=using,
  File "/home/vagrant/venv/lib/python2.7/site-packages/django/dispatch/dispatcher.py", line 193, in send
    for receiver in self._live_receivers(sender)
  File "/home/vagrant/venv/lib/python2.7/site-packages/argcache/argcache.py", line 461, in delete_cb
    new_key_set = selector(instance)
TypeError: 'unicode' object is not callable
```

I think it's because of this code in argcache (argcache.py) lines 448-453:

```
        if isinstance(selector, str):
            # Special-case this
            selector_str = selector
            selector = lambda instance: {selector_str: instance}
            # Make the token
            token = self.get_or_create_token((selector_str,))
```
After the str/unicode changes, the type of `selector` is now `unicode` and not `str` so the if statement no longer executes, which causes a problem in line 461 shown in the stack trace.

hwatheod:
Fix for the above issue is in https://github.com/learning-unlimited/django-argcache/pull/1

Some additional fixes in #3416

Please merge the first one (django-argcache) before #3416 or else the Python 2 tests will still fail.

esp/esp/program/modules/forms/teacherreg.py
```diff
@@ -91,7 +90,6 @@ class TeacherClassRegForm(FormWithRequiredCss):
     class_size_max = forms.ChoiceField( label='Maximum Number of Students',
                                         choices=[(0, 0)],
                                         widget=BlankSelectWidget(),
-                                        validators=[validators.MinValueValidator(1)],
```
hwatheod, line 92: Why are we removing this validator?

esp/esp/program/tests.py
```diff
@@ -341,8 +342,6 @@ def makeprogram(self):
                 'teacher_reg_end':   '3001-01-01 00:00:00',
                 'student_reg_start': '2000-01-01 00:00:00',
                 'student_reg_end':   '3001-01-01 00:00:00',
-                'publish_start':     '2000-01-01 00:00:00',
-                'publish_end':       '3001-01-01 00:00:00',
```
hwatheod, line 342: Why remove these fields?

esp/esp/program/modules/forms/teacherreg.py
willgearty: `ChoiceField`s coerce the selection to a string, which causes an error during validation because you can't compare strings and integers. In this case, since it uses a select widget, you shouldn't be able to submit a non-positive value anyways, so I think it is ok to remove?

esp/esp/program/tests.py
willgearty: They don't exist anymore in the form.

willgearty:
I believe the final errors are caused by these bits of code:
https://github.com/learning-unlimited/ESP-Website/blob/5eeecff1a6cce3f40b641487ea4fac29407ebc0a/esp/esp/users/models/__init__.py#L2009-L2016
https://github.com/learning-unlimited/ESP-Website/blob/5eeecff1a6cce3f40b641487ea4fac29407ebc0a/esp/esp/users/models/__init__.py#L2023-L2030

The `q_filter` field is a `TextField`. In Python 2, `pickle.dumps()` returned a `str`, which would work with a `TextField`. However, in Python 3, `pickle.dumps()` returns a `bytes` object, so Django tries to coerce it to a `str`, but that fails because `pickle.dumps()` objects can only be converted to `str` objects using `pickle.loads()`.

I think ultimately we would want the `q_filter` field to be a `BinaryField`. The [Django 1.11 docs](https://docs.djangoproject.com/en/1.11/ref/models/fields/#binaryfield) say you can't perform filters on `BinaryField`s, but I tested this by just changing the initial migration and the models, and it does seem to work.

esp/esp/program/tests.py
hwatheod: Ok. Can we get rid of the remaining references in `esp/esp/program/tests.py` and also all the commented out reference in `esp/esp/program/forms.py` lines 155-174?

hwatheod:
Yes, the final errors are pickle issues which I will work on fixing with the plan I wrote many comments ago. So far I have figured out how to switch between running Python 2 and 3 on `fab runserver` which will be needed to test any such fix.

Also, these may not be the "final" errors. Once these errors are fixed, they may unmask more errors :)

willgearty:
Ah, yes, there've been so many comments and LGTM comments that I totally forgot you already had a plan! I've got the pickled errors temporarily fixed for my travis builds, so I can start preemptively fixing the new errors.

willgearty:
I've run into an issue with `MessageRequest.special_headers` being changed to a `BinaryField.` In https://github.com/learning-unlimited/ESP-Website/pull/3129 I made it so that we include the `MessageRequest` as a `MessageVariable`, which means the `MessageRequest` object gets pickled here:
https://github.com/learning-unlimited/ESP-Website/blob/5eeecff1a6cce3f40b641487ea4fac29407ebc0a/esp/esp/dbmail/models.py#L480
However, you can't pickle a `BinaryField` because it's converted to a `memoryview` by Django:
https://github.com/django/django/blob/2da029d8540ab0b2e9edcba25c4d46c52853197f/django/db/models/fields/__init__.py#L2303-L2307

willgearty:
Here are some comments for each of the currently pickled models:
**PersistentQueryFilter.q_filter**
Looks like this can have some fairly complex pickled objects in it, and I'm guessing it will be a pain (if not impossible) to convert them. It looks like the only time we ever use a preexisting `PersistentQueryFilter` object is when we use the user search controller (and only for the confirmation or checklist parts). All other calls can just create new objects. I think we can just delete all of the old objects and start from scratch with this field as a JSONField.

**MessageRequest.special_headers**
As far as I can tell, we only ever use this to store a pickled dictionary. I imagine this should be easy to unpickle and repickle (or JSONify).
https://github.com/learning-unlimited/ESP-Website/blob/f161031b61bcf0e68d750bc0bea98c9fd66a5e51/esp/esp/program/modules/handlers/commmodule.py#L191-L192

**MessageVars.pickled_provider**
This includes pickled instances of `Program`, `User`, and `request` objects, which means it'll probably be tricky to convert them. Fortunately, I don't think we ever use these once an email request has been processed, so we can probably just delete the old ones.

**ResourceType.attributes_pickled**
This is always set as just a |-delimited list. I'm not sure why we call the field attributes_pickled when it never even seems to be pickled (or why we also have an un-pickled cached version of the attributes). As far as I can tell, we never call `_attributes_cached` or `attributes` (which would call `_attributes_cached`). I have a feeling that the vast majority of existing `ResourceType` objects will just have normal strings for this field (not pickled) and we can convert the small minority (if any) that have pickled attributes to a |-separated string (and then remove any of the code related to pickling).

**Answer.value**
I think these are almost exclusively pickled strings, so they should hopefully be easy to convert?

milescalabresi:
Resolved by #3616; closing

PR #3418, opened 2021-09-29
Fix pdf survey responses styling

### LINKED ISSUE #3316: Survey review_pdf responses go off page
If there are too many responses for a single survey question, the responses go off the page in the pdf version of the responses:
image

kkbrum:
Fixes #3316 so that long responses no longer get cut off in the survey responses pdf.

Unfortunately to fix this issue, I had to redo the styling of the whole pdf since minipages do not support spanning across pages. I think the restyling looks good, but am open to feedback!

image


willgearty (approved):
Looks great to me, thanks for taking care of this!

Issue #2436, opened 2017-09-05
Teacher trainings alter teacher availability


willgearty:
When a teacher signs up for a teacher training event, it adds an extra timeslot to their availability.

For example (in this case, we use an arbitrary date/time for our office hours):

![image](https://user-images.githubusercontent.com/7232514/30077635-520b853e-9231-11e7-9d50-d3ab726bbdec.png)

willgearty:
We should just filter this list to only class blocks when we set up the email context.

PR #3420, opened 2021-10-03
Filter teacher availability to only class time blocks

willgearty:
Before, teacher availability included events of any `EventType`. Now, we only use "Class Time Block" events (which are the events for the actual program). This means that we no longer will include training events, etc as part of a teacher's availability. I believe this will actually fix some weird edge cases that we've encountered in the past where teachers could register more sections of a class than they should have been able to register (I remember lots of weird instances of this at Stanford).

Fixes https://github.com/learning-unlimited/ESP-Website/issues/2436.


kkbrum (approved):
Nice!

PR #3422, opened 2021-10-04
Do more work at the database level

willgearty:
I've been noticing that loading /manage/emails on some production sites can take quite a long time, and sometimes can even fail (or worse, crash the server). Therefore, this attempts to move a lot of the work for the view to the database level, which should generally be more efficient. Testing this on dev3 indicates that this causes slightly more SQL time, slightly less CPU time (this is good), and overall slightly more total time (not noticeable). I think the best benefit to this, though, is the reduction in memory usage (especially on the production server), since we are no longer iterating over lists of potentially tens of thousands of `TextOfEmail`s. It would probably be useful to also test this on one of the larger sites, like Yale or Stanford, to make sure this actually helps.

### DIFF
--- a/esp/esp/program/views.py
+++ b/esp/esp/program/views.py
@@ -761,7 +761,7 @@ def get_email_data(start_date):
             req.num_rec = CommModule.approx_num_of_recipients(req.recipients, req.get_sendto_fn())
         req.num_sent = toes.filter(sent__isnull=False).count()
         if req.num_rec == req.num_sent:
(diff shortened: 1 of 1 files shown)


hwatheod (approved):
Tested on Yale site and performance is much improved (doesn't run out of memory anymore) although still slow.

willgearty:
Here's a screenshot of htop with the old behavior/code:
image

Issue #3269, opened 2021-04-23
Add suggested/common email/arbitrary user list selections


milescalabresi:
The arbitrary user list (AUL) interface, which is used for the email comm panel, is very powerful but offers a level of flexibility not always the most helpful. I propose we create some "commonly used lists," such as the following. These (and others?) could have their own buttons on the first tab for the majority of the time that most programs need them. The usual tabs (basic list, combination list) would still exist for users who know what they're doing and want to contact a different subset of individuals.
- all teachers who have not graduated (teacher/volunteer recruitment)
- students within the age ranges of the program + all K-12 educators (student recruitment)
- all teachers teaching at least one approved class (current teachers)
- students enrolled in at least one class (current students)
- all volunteers for this program (current volunteers)

willgearty:
This should be really easy now that we can use GET parameters with the USC.

PR #3424, opened 2021-10-08
Various improvements to the user search controller interface

willgearty:
This makes the following improvements to the user search controller (comm panel, arbitrary user list, etc):

- The default min and max grade filters correspond to the min and max grades of the program
- There is now a list of commonly used searches (Fixes https://github.com/learning-unlimited/ESP-Website/issues/3269)
- Fixes a bug that would cause an infinite loop in the javascript if you specified "recipient_type" in the GET parameters
- Makes the labels for the student email options (students/guardians/emergency contacts) associated with the checkboxes, which enables clicking the text to toggle the checkbox


kkbrum (approved):
Wonderful!

Issue #3403, opened 2021-08-23
Use GET parameter(s) for USC printables


willgearty:
Now that we can specify the user type as a GET parameter, we should do that for any of the printables that use the USC and aren’t for students

From testing SR13.

PR #3427, opened 2021-10-08
Use GET parameters for USC printable links

willgearty:
This uses GET parameters for the links to the printables that use the UserSearchController to make it slightly easier for admins to output these printables with the correct sets of users. As part of this, the GET parameters can now specify any of the AND/OR/NOT checkboxes to be checked.

While I was at it, I fixed a bug introduced in #3424 where the USC links always directed to the commpanel, instead of staying in the module.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3403.

### DIFF
--- a/esp/templates/program/modules/programprintables/options.html
+++ b/esp/templates/program/modules/programprintables/options.html
@@ -65,7 +65,7 @@ Please select from options below.
     <a href="./studentschedules/png/?recipient_type=Student&base_list=enrolled">PNG (1 schedule max)</a>
 </li>
 <li><a href="./flatstudentschedules?recipient_type=Student&base_list=enrolled" title="Student Schedules">Flat Student Schedules</a> (by students)</li>
-<li><a href="./teacherschedules?recipient_type=Teacher&base_list=approved" title="Teacher Schedules">Teacher Schedules</a> (by teachers)</li>
+<li><a href="./teacherschedules?recipient_type=Teacher&base_list=class_approved" title="Teacher Schedules">Teacher Schedules</a> (by teachers)</li>
 {% if program|hasModule:"TeacherModeratorModule" %}
     <li><a href="./teachermoderatorschedules?combo_base_list=Teacher:class_approved&checkbox_or_assigned_moderator#tab_combo" title="Teacher and Moderator Schedules">Combined Teacher and {{ program.getModeratorTitle }} Schedules</a> (by teachers)</li>
     <li><a href="./moderatorschedules?recipient_type=Teacher&base_list=assigned_moderator" title="Moderator Schedules">{{ program.getModeratorTitle }} Schedules</a> (by teachers)</li>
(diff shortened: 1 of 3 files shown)


esp/templates/users/usersearch/list_selector.html
```diff
@@ -7,12 +7,12 @@
...
-        <li><a href="./commpanel?recipient_type=Volunteer&base_list=all_Volunteer">All volunteers in the database</a></li>
-        <li><a href="./commpanel?recipient_type=Volunteer&base_list=volunteer_all">All volunteers signed up for this program</a></li>
+        <li><a href="?recipient_type=Student&base_list=all_Student&grade_min={{ program.grade_min }}&grade_max={{ program.grade_max }}">All students within the age range of the program</a></li>
+        <li><a href="?recipient_type=Student&base_list=enrolled">All students enrolled in at least one class</a></li>
+        <li><a href="?recipient_type=Teacher&base_list=all_Teacher&gradyear_min={% now 'Y' %}">All teachers who have not graduated</a></li>
+        <li><a href="?recipient_type=Teacher&base_list=class_approved">All teachers teaching at least one approved class</a></li>
+        <li><a href="?recipient_type=Volunteer&base_list=all_Volunteer">All volunteers in the database</a></li>
+        <li><a href="?recipient_type=Volunteer&base_list=volunteer_all">All volunteers signed up for this program</a></li>
```
kkbrum, line 15: Hm I know this is more for the PR that I already approved, and seems like it would be a bit of extra work, but I think we should probably hide the common searches that are teacher/volunteer based when on the studentschedules page, and similarly for other pages

esp/templates/program/modules/programprintables/options.html
```diff
@@ -87,23 +87,23 @@ <h3>Student Lists</h3>
...
-<li><a href="./studentchecklist" title="Student Checklist">Student Checklist</a> (by students)</li>
-<li><a href="./student_financial_spreadsheet">Student Financial Spreadsheet</a> (used by the check-scanning app)</li>
-<li><a href="./studentsbyname" title="Student List">Students by Name</a> (by students)</li>
-<li><a href="./emergencycontacts" title="Student List">Students' Emergency Contact Info</a> (by students)</li>
+<li><a href="./studentchecklist?recipient_type=Student&base_list=enrolled" title="Student Checklist">Student Checklist</a> (by students)</li>
+<li><a href="./student_financial_spreadsheet?recipient_type=Student&base_list=enrolled">Student Financial Spreadsheet</a> (used by the check-scanning app)</li>
+<li><a href="./studentsbyname?recipient_type=Student&base_list=enrolled" title="Student List">Students by Name</a> (by students)</li>
+<li><a href="./emergencycontacts?recipient_type=Student&base_list=enrolled" title="Student List">Students' Emergency Contact Info</a> (by students)</li>
 {% if not li_types|length_is:0 %}
 <li>Students who have ordered/reserved line items:
```
kkbrum, line 108: I think these could use default GET parameters too, although a bit more complicated

willgearty: I'm just gonna do "All students in database", because I guess there might be students that ordered something that aren't enrolled?

esp/templates/users/usersearch/list_selector.html
willgearty: It would be really complicated to show specific ones based on the printable, so I decided to just not show them at all on the printable pages.

willgearty:
OK, I've addressed those two comments @kkbrum (see comments above). I've also made it so that the USC page tells you for which printable you are selecting users.

esp/esp/program/modules/handlers/programprintables.py
```diff
@@ -666,13 +668,13 @@ def teacherlist(self, request, tl, one, two, module, extra, prog):
     @needs_admin
     def teachermoderatorlist(self, request, tl, one, two, module, extra, prog):
         """ default list of teachers; function left in for compatibility """
-        return self.teachersbyFOO(request, tl, one, two, module, extra, prog, teaching=True, moderating=True)
+        return self.teachersbyFOO(request, tl, one, two, module, extra, prog, teaching=True, moderating=True, display_name = 'Teacher and Moderator List')
```
kkbrum, line 671: need to {{ prog.getModeratorTitle }}

kkbrum: same for all the moderator related printables below

kkbrum (changes requested):

PR #3429, opened 2021-10-15
Create more categories for tag settings

### LINKED ISSUE #3333: More expandable sections on program tag settings page
There are now a lot of "help text" tags on the manage program tag settings page, particularly for teacher registration settings. I think these could use a new expandable section or cleaned up in some way so as to not have such a long list show up just from opening one section.

kkbrum:
Fixes #3333. 

- [ ] I'm not sure how to quickly do this, but I think we should reorder the categories too. They are currently in alphabetical order by the short category name, but I think it should be general, teacher, class, moderator, student, volunteer

image


willgearty (approved):
Works great! It's so nice splitting up the previously HUGE teacher category!

PR #3442, opened 2021-11-10
Convert errors and successes to banners

### LINKED ISSUE #3434: Make warnings bootstrap banners
Alert/warning/danger messages (usually in red font) should be turned into bootstrap banners:

```html
<div class="alert alert-danger">
  <span class="glyphicon glyphicon-info-sign" aria-hidden="true"></span>
  DANGER WILL ROBINSON
</div>
```

Other messages (likes those in green font) could even be turned into happy banners.

kkbrum:
Change many errors and successes to banners. I checked all the ones I could, but some errors are hard to get to so I didn't do those. There are definitely more messages that should be banners, but less easily searchable and will need to get updated as folks notice them. Fixes #3434

### DIFF
--- a/esp/templates/program/modules/admincore/lunch_constraints.html
+++ b/esp/templates/program/modules/admincore/lunch_constraints.html
@@ -9,9 +9,21 @@
 
 {% block content %}
     <h1>{{program.niceName}} Lunch Constraints</h1>
-    <br />
     <p>Please select the timeslots you would like to be used for lunch.</p>
     <p>You can also use this page to setup lunch scheduling constraints, which will enforce that each student have at least one lunch block per day.</p>
+
+    {% if POST and saved %}
+        <div class="alert alert-success">
+        <span class="glyphicon glyphicon-info-sign" aria-hidden="true"></span>
+            Settings successfully saved.
+        </div>
+    {% endif %}
+    {% if POST and not saved %}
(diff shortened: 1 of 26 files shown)


esp/templates/program/modules/admincore/lunch_constraints.html
```diff
@@ -23,10 +23,16 @@ <h1>{{program.niceName}} Lunch Constraints</h1>
...
+            <span class="glyphicon glyphicon-info-sign" aria-hidden="true"></span>
+                Settings successfully saved.
+            </div>
     {% endif %}
     {% if POST and not saved %}
-        <h3 style="font-weight:bold; color:red;">There was an error with saving your settings. Please try again.</h3><br /><br />
+        <div class="alert alert-danger">
+            <span class="glyphicon glyphicon-info-sign" aria-hidden="true"></span>
+                There was an error with saving your settings. Please try again.
+        </div><br />
```
willgearty, line 35: I think these should go before/above the form.

willgearty (commented):
Everything looks good to me (I tested about 60% of the pages, and the others look like they should work just as well). Only one comment which has nothing to do with the changes you made, but since you're here, you might as well make the change.

willgearty (approved):
Perfect, thanks!

PR #3443, opened 2021-11-10
Various user interfaces for program records

### LINKED ISSUE #3220: Registration module(s) to show status of arbitrary record(s)
Sometimes students and/or teachers need to submit a variety of forms (especially with many universities having new requirements lately), but these forms don't have associated modules (e.g. attended training) or the modules won't work for the specific use case (e.g. medical or liability forms). Often, these forms are via email, through 3rd party services (eww), or via a customform. I'm envisioning two components:

1. A tag or settings interface that would allow admins to specify (at the very least as a comma-separated list) which records should be included for the module.
2. A user-facing module that would have a list of those records with an "X" or checkmark for each like the checkboxes interface for student/teacher registration.

There would need to be versions of both of these for student and teacher registration.

Requested by Yale at SplashCon 2021.

### LINKED ISSUE #3221: Module to set arbitrary records for arbitrary users
Sometimes, there are forms or other actions that can not be processed by the website (either ever or just at the moment) due to technical limitations or university requirements. We would want a way to be able to mass-record that these forms have been received (e.g. medical or liability) or that these actions have been performed (e.g. attended training). We need a module that allows admins to select an arbitrary list of users (USC) and then select a record (or, ideally, a dynamic set of arbitrary records) to be set for those users.

Then, the status of these records could be shown in user-facing modules (https://github.com/learning-unlimited/ESP-Website/issues/3220) or admin-facing modules (https://github.com/learning-unlimited/ESP-Website/issues/3222).

Requested by Yale at SplashCon 2021.

### LINKED ISSUE #3222: Show status of (arbitrary) record(s) on attendance page
Yale requested that record status be shown on the attendance page (like the checked-in column). I think the first step for this would be to make a template filter that can pull the status of an arbitrary record for an arbitrary user (e.g. `{{ user|record_status:"attended" }}`). This would provide an easy avenue for programs to at least make template overrides for the attendance page to show any arbitrary records as additional columns. To make this easier for chapters, we could then make a tag (comma-separated list) that would specify which records should be added as columns to the attendance page (and maybe to the student attendance/records printable).

Requested by Yale at SplashCon 2021.

willgearty:
This adds the following functionality for program records that are not (and/or can not be) handled via program modules for a given program:

- Records can now be bulk set for an arbitrary user list with a new module ("User Records Module") (Fixes #3221)
- A set of records can be set as required for student or teacher registration using tags (presumably ones that are not handled by enabled modules)
- These sets of records are shown in student/teacher registration as checkboxes (Fixes #3220)
- These sets of records are also shown on the userview page, the student checklist printable, and the teacher list printable (Fixes #3222?)

While I was at it, I also fixed several bugs related to the registration progress bar interface, the order of modules on the student/teacher registration mainpage, and the grade min/max defaults for the USC.

image
image
image


kkbrum (approved):
Nice!

PR #3449, opened 2021-11-16
Unit test that loads all views

### LINKED ISSUE #1900: We should have a test that loads every program module view

### LINKED ISSUE #3438: Confirmation email controller is broken
The confirmation email controller is broken due to the deprecated use of `Context`:
https://github.com/learning-unlimited/ESP-Website/blob/d111eec695ad1a2c1f95e574cb995eca84493d62/esp/esp/program/controllers/confirmation.py#L57-L60
I believe just removing the `Context` wrapper should fix the problem (which can be reproduced using the confirm registration button module).

Reported by Berkeley.

willgearty:
This adds a unit test that loads every single view (`main_call` and `aux_call`) from every single program module. Unfortunately, we aren't terribly consistent in how our views work, so I had to brute force a whole bunch of different ways of loading the views (e.g., GET variables, POST data, various `extra` arguments). After writing the test, I then fixed any of the views that it identified as broken (e.g., Fixes #3438). I also changed some views to redirect to the main registration page if the incorrect data was provided (e.g., `waitlist_subscribe` when there isn't a waitlist).

While I was at it, I realized that the two credit card modules displayed an error if they weren't configured. Now, if they aren't configured, they aren't shown as steps in registration.

Fixes #1900.

### DIFF
--- a/esp/esp/program/controllers/confirmation.py
+++ b/esp/esp/program/controllers/confirmation.py
@@ -37,7 +37,7 @@ from esp.program.models import Program, ClassSection, ClassSubject
 from esp.users.models import ESPUser, Record
 from esp.program.modules.module_ext import DBReceipt
 
-from django.template import Template, Context
+from django.template import Template
 from django.template.loader import select_template
 from esp.dbmail.models import send_mail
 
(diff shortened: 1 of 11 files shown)


kkbrum (approved):
Yay testing!

PR #3450, opened 2021-11-17
Remove user from list of coteachers on teacher reg mainpage

### LINKED ISSUE #3261: Teacher list on teacherreg includes signed in user
In the inclusion template, we call request.user to make sure we don't print the actual user's name as one of the teachers:
https://github.com/learning-unlimited/ESP-Website/blob/98501776a014a01eeba8c8105ee2aca2a7d7bee3/esp/templates/inclusion/program/class_teacher_list_row.html#L112

But this doesn't work because the inclusion template doesn't have access to request.user (or any of the other context for the rest of the page). For other inclusion tags we provide a user, so maybe we want to do that with class_teacher_list_row:
https://github.com/learning-unlimited/ESP-Website/blob/98501776a014a01eeba8c8105ee2aca2a7d7bee3/esp/esp/program/templatetags/class_render.py#L61-L62

willgearty:
This makes it so teachers no longer show up as their own coteachers on the teacher registration mainpage:
image

Fixes #3261.


esp/templates/inclusion/program/class_teacher_list_row.html
```diff
@@ -108,8 +108,8 @@ <h3>Cancellation Request for {{ cls|truncatewords:6 }}</h3>
 
 {% if crmi.allow_coteach %}
 <tr>
-  <td class="clsleft"><b>Teachers (<a href="mailto:{{ cls.emailcode }}-teachers@{{ email_host_sender }}">email</a>): </b></td>
-  <td class="clsmiddle" colspan="5">{% for t in cls.get_teachers %}{% ifnotequal t request.user %}{{ t.first_name }} {{ t.last_name }}{% if not forloop.last %}, {% endif %}{% endifnotequal %}{% endfor %}</td>
+  <td class="clsleft"><b>Coteachers (<a href="mailto:{{ cls.emailcode }}-teachers@{{ email_host_sender }}">email</a>): </b></td>
+  <td class="clsmiddle" colspan="5">{% for t in cls.get_teachers %}{% ifnotequal t user %}{{ t.first_name }} {{ t.last_name }}{% if not forloop.last %}, {% endif %}{% endifnotequal %}{% endfor %}</td>
```
kkbrum, line 112: The multiple if's in this line generate a slight problem. If it's not the last teacher, it puts a comma, but then when it goes to the next teacher and it's equal to the user, it passes. So if the user happens to be last in the list then you end up with a trailing comma

kkbrum (changes requested):

willgearty:
I've added a template filter that removes the teacher from the list, so now there shouldn't be any extra commas. I did the same thing for the teacher webapp sectioninfo page (and cleaned up the moderators section to match the teacher registration mainpage).

kkbrum (approved):
Nice!

Issue #1777, opened 2015-10-04
if you are already logged in, the login view's next param doesn't work


benjaminjkraft:
You just always get redirected to the homepage.  We should probably be checking it in the first conditional of `esp.users.views.login_checked`.

PR #3453, opened 2021-11-17
Improve login redirect

willgearty:
This makes the following improvements to the login redirect logic:

- Admins can now set custom landing pages for students and teachers with global tags (`teacher_home_page` and `student_home_page`) (this only works if a user isn't already being redirected to some other page)
- If a user is already logged in, they will be redirected properly to the "next" param (Fixes https://github.com/learning-unlimited/ESP-Website/issues/1777)


kkbrum (approved):
Awesome!

Issue #1769, opened 2015-09-11
Schedule generation should return the link as a download, rather than into the browser


benjaminjkraft:
Apparently Chrome's PDF viewer will choke on huge PDF files, and the user probably wants to save it anyway.  So we should just pop up a download window rather than returning it in-browser.  This could also be done for other printables.

PR #3454, opened 2021-11-19
Download student schedules pdf to file

willgearty:
This makes it so that the student schedules pdf is downloaded to the user's disk as opposed to being loaded within the browser. For larger pdfs, Chrome (and possibly other browsers) might have some trouble loading them, so this should help.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/1769.

This might also help with https://github.com/learning-unlimited/ESP-Website/issues/1198, but I haven't tested that at all.


kkbrum (approved):
Wow!

PR #3459, opened 2021-12-01
Don't render tex on survey results

### LINKED ISSUE #3302: LaTeX makes the survey results sad
If there is a LaTeX in a class title, we try to render it in the survey results page, but this seems to fail:
image
(for the record, it seemed to work for a second, then switched to the above)

The easiest solution would probably just be to not trying to render the LaTeX.

Reported by Cloud.

kkbrum:
Fixes #3302.


willgearty (approved):
Looks good, thanks!

Issue #3419, opened 2021-09-30
Anonymous option for contact form


willgearty:
It would good to have a way for someone to reach out to a program anonymously (about concerns, etc). I think we could just add this as an option to the existing contact form (although we would want to make sure it is truly anonymous, even if they are signed in).

Suggested by Rainstorm

PR #3462, opened 2021-12-08
Add anonymous option to the contact form

willgearty:
This adds an anonymous option to the contact form:
image

If the box is checked, the email address field is no longer required. It will not put any personal information in the email, it will set the given email address to the bcc (if the copy me option is checked), and it will set the "from" email address as the same as the "to" email address.

It seems to work on my dev server, but it never actually sent emails, so it might be good to test this on dev3 (or even on cloud since they requested it).

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3419.

### DIFF
--- a/esp/esp/web/forms/contact_form.py
+++ b/esp/esp/web/forms/contact_form.py
@@ -56,11 +56,14 @@ hear_about = (
     )
 
 class ContactForm(forms.Form):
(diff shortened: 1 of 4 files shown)


kkbrum (approved):
Huzzah! Tested off of dev3 and things look good!

PR #3463, opened 2021-12-15
Add checkbox interface for admins

### LINKED ISSUE #3451: Check for credit card set up upon program creation
Since the credit card set up is a global thing, I think that when an admin submits a program and checks off credit card payments, we could then check if credit cards are set up and if not, have a page that informs them to email websupport to set it up (or provide whatever instructions we want). 

I think this is particularly important due to #3449, which is a good thing because it doesn't show students errors, but also removes a major admin checkpoint of "is student reg ready to go" when an admin tests student reg and doesn't get an error.

### LINKED ISSUE #3452: List of things admins need to do to set up program?
Similar to #3451 but on a larger scale: We could take the inputs to the program settings form, and turn the things the admins SAID they wanted into a checklist at the top of the manage page, saying "You said you wanted the following things but have not set them up: ... (eg custom form for teachers, additional purchases, etc)". On an even larger scale, this list could also include BASIC registration steps, such as creating timeslots and creating classrooms. I think the lists would probably only contain the items that have NOT been completed, in a warning/danger box. That way it's not in the way for programs who have completed them.

willgearty:
This adds a checkbox interface for the steps that are required for ensuring a program is completely setup (depending on which modules have been enabled). When an admin module exists that requires setup, it should be included in this list by having `isStep` return `True` (which is the default for program modules). By default the link will direct the user to the `main_call`, but this can be overriden with the `setup_path` attribute (and the text can be overriden with the `setup_title` attribute). The `checked` attribute is based on the response of the `isCompleted` method (so this should be coded to check for whether the setup is complete)..

For cases where a module doesn't exist for a particular part of setup, a step (required or optional) can be manually added in `admincore.py`.

image

Fixes #3452 and fixes #3451


kkbrum (approved):
Fabulous!

PR #3467, opened 2021-12-27
Show default values in the tag settings ui

### LINKED ISSUE #3440: Show tag default values on tag settings page
Right now it's not clear what the defaults are for tags (especially when you are copying a program, so it's not clear from the tag settings page whether a value is the default or has been copied over).

willgearty:
This displays the default value for each tag on the tag settings UI (to make it easy for an admin to reset a tag to its default value):
image

Fixes #3440.

### DIFF
--- a/esp/templates/program/modules/admincore/tags.html
+++ b/esp/templates/program/modules/admincore/tags.html
@@ -65,6 +65,8 @@
         <br/>
         <span> {{ field.help_text }} </span>
       {% endif %}
+      <br/>
+      <span style="font-style:italic;">Default: </span>{% if field.field.default != None %}<span>{{ field.field.default }}</span>{% endif %}
       {% for error in field.errors %}
       <br/>
        <span class='form_error'> {{ error }} </span>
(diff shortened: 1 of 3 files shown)


esp/templates/program/modules/admincore/tags.html
```diff
@@ -65,6 +65,8 @@ <h1>{% if program %}Tag Settings for {{ program.niceName }} (ID: {{ program.id }
         <br/>
         <span> {{ field.help_text }} </span>
       {% endif %}
+      <br/>
+      <span style="font-style:italic;">Default: </span>{% if field.field.default != None %}<span>{{ field.field.default }}</span>{% endif %}
```
kkbrum, line 69: I think it might be cleaner to only include this line where there is a default (that isn't ""), and then maybe add a line to the intro to the page about defaults being listed when they exist. But I also think it's pretty clear that the non-listed defaults would be those where they don't exist based on context.

kkbrum (approved):

esp/templates/program/modules/admincore/tags.html
willgearty: I'm worried people will think the default is whatever the tag is currently set to (which might not be blank) if the default isn't explicitly listed (even if it is blank).

PR #3469, opened 2021-12-30
Add button to scheduler to print matrix

### LINKED ISSUE #829: Printing the grid of classes
For use day-of, MIT likes to print out the big schedule grid with all of the classes on it.  (It's often printed at some giant size; this year's was 24"x50".)  Currently, this is done by manually editing some HTML on the Ajax Scheduler, and then doing a bunch of somewhat hackish file conversion.  It would be nice if there were just a printable for it.  The view of the Ajax Scheduler is great, it should just be more designed for printing.

willgearty:
This adds a button to the ajax scheduler that lets you print just the schedule matrix:
image

image

Fixes #829.


kkbrum:
I've got something weird going on. Here's what my scheduler looks like:

image


Here's what the printout looks like:

image

Note that the third and final timeblock has been duplicated(?) and the room name formatting is bizarre

willgearty:
OK, can you give it a go now, @kkbrum?

kkbrum (approved):

PR #3477, opened 2022-01-19
Add module that allows students to get completion certificate

### LINKED ISSUE #3344: Student module for completion certificate
Currently, a completion certificate needs to be generated for a student by an admin upon request through the printables module. We could add a student registration module that provides a completion certificate to a student. The module would only appear in the list of modules if they were enrolled in classes and/or attended classes/the program (this could be tag controlled).

willgearty:
This adds a simple optional module to student registration that, when enabled, will activate after a program ends, allowing students to get a completion certificate for the program. While I was at it, I cleaned up the completion certificate a bit.

Fixes #3344.


kkbrum (commented):
Do we want to only allow students to get certificates if they have been marked as attended for the program? Similarly only show courses for which they were marked attended? Or make these both tag options?

willgearty:
I guess if we really want to be doing this correctly, we should have a tag with the following options:

- all: all students get access to the default certificate once the program has ended
- program attendance: if a student attended the program, they get access to the default certificate
- class attendance: a student gets access to a certificate with ONLY the classes they attended

willgearty:
Alright, let me know what you think of the new tag @kkbrum

willgearty:
Alright, let me know what you think of the new tag behavior @kkbrum

kkbrum (approved):
Looks good to me

PR #3479, opened 2022-01-28
Hipparchia config

milescalabresi:
Changing some of the setup and update scripts for Ubuntu 20.04+ while maintaining backwards compatibility with the version on diogenes.


esp/make_virtualenv.sh
```diff
@@ -71,5 +71,5 @@ sudo pip2 install "virtualenv>=1.10"
 if id "vagrant" >/dev/null 2>&1; then
     virtualenv "$VENVDIR"
 else
-    sudo -u www-data virtualenv "$VENVDIR"
+    virtualenv "$VENVDIR"
```
willgearty, line 74: Do we need this if/else any more?

willgearty (changes requested):
Just one comment, otherwise everything looks good to me!

esp/make_virtualenv.sh
milescalabresi: I suppose not. Good catch

willgearty (approved):

PR #3480, opened 2022-01-29
Deprecate splash info module

### LINKED ISSUE #1320: Allow required line item types with options
Right now, when a student visits studentreg, all required line item types are charged to their account. This is not compatible with line item with options. Furthermore, required line items are explicitly excluded from the optional costs module. And even if such a line item were included in the module, it wouldn't be possible to enforce its required status, since all line items with options have a button to remove the current selection.

This is something that we should support. For example, Yale is currently trying to use line items with options to force students to select lunch options. With the current behavior, it will always be possible for students to skip answering the questions.

If a line item type has options and is marked as required, it should not be charged upon visiting studentreg. Instead, it should appear in the extra costs module, and the question should be required, with no reset button for that field.

### LINKED ISSUE #2727: Costs are hard coded in splashinfomodule
Presumably this is because it was designed for/by a particular chapter, but the information for this module should be flexible to account for different costs of the program, different sibling discounts, and different lunch prices.

willgearty:
This deprecates (and removes) the splash info module in favor of using the extra costs module for both sibling discounts and required line items such as lunch choices:

image

This solves the issue with lots of hard-coded values in the splash info module (e.g., sibling discount was always $20, even though it was listed as changeable). The extra costs module now shows all line items that aren't admission or financial aid related (whether they are optional or required). Required line items will be forced to be selected, and the module will be required during student registration if there are any required line items. I also fixed up the form such that errors are now shown properly.

Fixes #2727 and fixes #1320.


kkbrum (approved):
This looks good. It would probably be nice to have a way to see the sibling names more easily-- I added an issue (#3488) for it though so it can be added outside this PR

PR #3484, opened 2022-02-10
Deprecate paid records

### LINKED ISSUE #2185: Records for "paid for a program" are defunct
We don't use the record type "paid for a program" to keep track of credit card (or other payments received in advance) anymore, but they're still viewable at /admin/users/record/?event__exact=paid and I'm told that onsite check-in and registration modules will still create these records. We should probably either change this to keep track of transfers in a way that's easier to search, filter, and sort than transfers, or just remove the record type entirely to avoid confusion (and maybe improve transfer searching tools).

willgearty:
This removes all uses of the "Paid for a program" record in favor of using accounting transfers. I've also added a migration that converts all records to transfers (and back if undone).

~~Note: the migration does not currently work because of database weirdness during migrations.~~

Fixes #2185.


willgearty:
OK, now we include a message about where the transfers came from and then if the migration is undone, we delete those transfers (and only those transfers, not all payment transfers like before). I also left the paid record option in so we can use that option when undoing the migration.

I'm not able to reproduce the weirdness you mentioned. Like you said, it should only be looping through users that have "paid" records. To test this, I made three paid records for users, then migrated, and then the three records were gone and the three proper transfers were created. Could you try to reproduce the issue that you encountered with the teachers?

kkbrum (approved):
I couldn't reproduce the erroneous transfers this time either. Looks good.

Issue #1664, opened 2015-04-07
Combination list selector with NOT alone doesn't work correctly


benjaminjkraft:
@hwatheod wrote:

> Getting a report that in the combination list selector, if you click only the NOT box for some set of users while both AND and OR remain unchecked, then you get strange results / no results.  Can we adjust the interface so that it won't allow you choose NOT without choosing either AND or OR?

PR #3485, opened 2022-02-11
Prevent combination list with only NOT

willgearty:
When a NOT checkbox is selected for a combination list, we now automatically select AND as well (if neither AND nor OR is already selected).

While I was at it, I cleaned up some of the styling of the checkbox interface:
image

Fixes https://github.com/learning-unlimited/ESP-Website/issues/1664.


kkbrum (approved):
I like it! So dynamic!

Issue #1405, opened 2014-10-17
Atomic "swap classes" operation in scheduler


benjaminjkraft:
It would be useful to have a nice way to swap two classes in an atomic way.  For example, one could drag one class onto the other, and they would swap if possible, or both stay scheduled if not.

PR #3487, opened 2022-02-20
Swap sections in the ajax scheduler

willgearty:
This allows you to swap two sections in the ajax scheduler by selecting one and then holding CTRL/CMD and clicking on the other one. If one is longer than the other, it will attempt to swap any other sections necessary to perform the full swap (after a confirmation is provided from the user). If the swap is not possible (due to teacher availability or temporal fitting), an error message is provided.

~~The swapping appears to work very cleanly for almost all cases (e.g., same room, same times, different lengths, etc.) EXCEPT when you try to swap a single section with multiple sections that partially overlap in time. I'm not sure what the issue is, but I have a hunch it has to do with availability caching, but I'm going to wait until this has been reviewed to see if there are any other edge cases that need to be fixed. Also, sometimes the sections are unscheduled after swapping, but they should always be rescheduled automatically once the site double-checks the change log (every 5 seconds).~~

~~In hindsight, I'm wondering if I maybe should have performed the swap as a single back-end action, instead of as 3 discrete actions (unschedule, reschedule, schedule). If the issues described above persist, I might need to change to this behavior instead.~~

I also fixed a bug that prevented scheduled sections from being rescheduled to an overlapping time (e.g., from 10-12 to 9-11) even if there was the proper availability.

Finally, I've also cherry-picked (with some modification) a couple of commits from the MIT branch that looked useful (but are not necessarily relevant to this PR).

Fixes https://github.com/learning-unlimited/ESP-Website/issues/1405.

### DIFF
--- a/esp/public/media/scripts/ajaxschedulingmodule/ESP/Sections.js
+++ b/esp/public/media/scripts/ajaxschedulingmodule/ESP/Sections.js
@@ -326,8 +326,9 @@ function Sections(sections_data, section_details_data, categories_data, teacher_
      * The second has all the timeslots where teachers are available but already teaching
      *
      * @param section: The section to check availability.
+     * @param ignore_sections: An optional array of sections to ignore.
      */
-    this.getAvailableTimeslots = function(section) {
+    this.getAvailableTimeslots = function(section, ignore_sections=[]) {
         var availableTimeslots = [];
         var already_teaching = [];
         if(this.matrix.sectionInfoPanel.override){
@@ -343,7 +344,7 @@ function Sections(sections_data, section_details_data, categories_data, teacher_
                 }.bind(this));
                 $j.each(teacher.sections, function(index, section_id) {
                     var assignment = this.scheduleAssignments[section_id];
-                    if(assignment && section_id != section.id) {
+                    if(assignment && section_id != section.id && !ignore_sections.includes(this.getById(section_id))) {
                         $j.each(assignment.timeslots, function(index, timeslot_id) {
                             var availability_index = teacher_availabilities.indexOf(timeslot_id);
                             if(availability_index >= 0) {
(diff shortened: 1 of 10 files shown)


esp/public/media/scripts/ajaxschedulingmodule/ESP/Sections.js
```diff
@@ -541,6 +544,136 @@ function Sections(sections_data, section_details_data, categories_data, teacher_
...
+        var new_assignments1 = [];
+        var start_slot = this.matrix.timeslots.get_by_id(old_timeslots2[0]);
+        for(var sec of swapping_sections1){
+            if(sec){
+                if(new_assignments1.length == 0 || sec !== new_assignments1[new_assignments1.length - 1].section){
+                    var new_timeslots = this.matrix.timeslots.get_timeslots_to_schedule_section(sec, start_slot.id);
+                    var valid = this.matrix.validateAssignment(sec, room2, new_timeslots, ignore_sections);
+                    if(!valid.valid){
+                        console.log(valid.reason);
+                        this.matrix.messagePanel.addMessage("Error: can not swap those classes.", color = "red");
```
kkbrum, line 615: It would be more helpful to give an indication as to why the classes can not be swapped

kkbrum (changes requested):
Can you add a note about the swap feature to the upper right hand info box on the scheduler? I think the slight lag to the unschedule & rescheduling process is fine, but I would include a note about it in the info box as well. Like "note: this will take a few seconds to complete the action" just so people know it's not behaving weirdly.

The behavior of checking the "override availability" box is a bit weird to me with this. I'm not really sure what it's doing; sometimes it says it can't swap the classes, sometimes it unschedules one of the classes. I believe that if the reason the classes cannot be swapped is teacher availability of one of the classes, the override button should allow the swap.

willgearty:
Random thought: are there repercussions of unscheduling one set of the classes? (student/teacher registration-wise)

kkbrum:
Idea: Why not have it so that if you click an unscheduled class from the class list, you can replace a scheduled class with it by doing ctrl click? Seems like this is a logical extension and would make scheduling even faster

kkbrum:
The more seamless swapping looks great! And I have not found any more weirdness with overriding availabilities.

willgearty:
That should be everything, @kkbrum!

kkbrum (approved):
Wow this is so seamless and quick and easy! Now when you accidentally schedule section 2 before section 1, it'll be so easy to fix!!

Issue #3275, opened 2021-04-30
Clarifying the "update personal info page" on user account creation


reneesu99:
Currently the second page of account creation has users go to the "update personal info" page. This presents a user flow problem since it has user's re-enter information they already entered on the first page and also uses confusing "update" wording when the user is still unregistered. 

Currently:
![image](https://user-images.githubusercontent.com/65254283/116634218-b2dbf180-a918-11eb-81b2-e10cfbf31c61.png)

Changes:
during registration **hide the "first name, last name, and email registration"**
during registration **have "update" button changed to "submit"**

![image](https://user-images.githubusercontent.com/65254283/116634479-6a710380-a919-11eb-89ec-5b959ef305a8.png)

PR #3494, opened 2022-03-09
Fix registration/profile forms for new users

willgearty:
This makes minor changes to the registration and profile forms for new users, including:

- Change the wording when a user hasn't submitted a profile before
- Minor grammar fixes

I decided not to hide the fields as requested in https://github.com/learning-unlimited/ESP-Website/issues/3275 because:

1. There's potentially an edge case where a user might want to edit these after the previous registration form
2. The fields are already filled in, so the user doesn't need to fill them in again
3. I couldn't figure out an easy way to identify within the view the specific case where a user has come from the new user registration form (we could hide the fields whenever the user doesn't have a profile, but this seems to make #1 a more likely edge case)

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3275.


kkbrum (approved):
Nifty!

PR #3499, opened 2022-03-23
Add actual glyphicons to theme icon dropdown

willgearty:
This adds the glyphicons to the theme icon dropdown so admins know what the icons look like when they are picking them:

image


kkbrum (approved):
Visionary!

PR #3500, opened 2022-03-23
Add class category ajax scheduling "check"

### LINKED ISSUE #3498: Capacity mismatch check should distinguish between too small/too big
In the ajax scheduler, we have a capacity mismatch scheduling check which indicates with green/red whether a class's capacity matches the room's capacity. Right now, it only shows green/red for match/mismatch, but it would probably make more sense to color it one color for too big (more of an issue, so maybe red) and one color for too small (less of an issue, so maybe blue?).

willgearty:
This adds a scheduling "check" to the ajax scheduler that colors classes by their category. This can be used to check the distribution of different class categories across time/space (like the category check in the scheduling checks module). The coloration is flexible to the total number of categories and evenly splits up the full rainbow palette across the categories (i.e., the more categories there are, the closer in color they will be).

While I was at it, I also fixed the capacity mismatch check to distinguish between "class too small" (blue, because not that bad) and "class too big" (red, because very bad). This fixes #3498.


kkbrum (approved):
Colorful!

Issue #3497, opened 2022-03-13
Hide administration pages


milescalabresi:
Before Will and Katherine made /manage interfaces for the modules, the /admin pages interface was necessary for anyone who wanted to modify modules, costs, and many other program configurations. Now it is not, and gives webmins the power to mess things up. I think we should remove the "Administration pages" from the admin toolbar on the left and add a warning (landing page, header, or popup) that the user should probably go to /manage/current/main instead unless they really know what they're doing.

willgearty:
I'm hesitant to remove it entirely because I know people more familiar with the website are used to it being there...but at the same time, it does seem to cause more confusion for less experienced folks.

At the very least, we should always include the real link for dev servers.

milescalabresi:
Yes agreed for dev servers. Per our conversation, these items would benefit from having safer "front-end" UIs. We can pare down this list and/or discuss more as needed, and these definitely vary in priority
- Cal
  - events - can't change description of teacher trainings
- Dbmail
  - dbreceipts
  - plain redirects
- Programs
  - class categories
  - class flag types
  - phase zero records (add a student to the lottery)
  - program module objs (high priority)
  - registration types (low priority)
- Redirects
- Users
  - Grade Change requests (but you get an email); maybe it should be on the userview page
  - Permissions (for individual users)
- Utils
  - printers (low priority)
  - Template Overrides for schedules

PR #3510, opened 2022-03-31
Add UI for program modules

willgearty:
This page lets you change the sequence and requiredness of program modules. This is the first step towards #3497.

image


kkbrum (changes requested):
This looks very nice!

I know you left warnings, but I think this page currently has too much flexibility for admins to not understand what a module does and put it in a bizarre spot. 

It would probably be too hard / not flexible enough to enforce certain constraints (ie. not allowing certain modules to switch requiredness or go before/after a different module), but I'm wondering if we can add a comment when you hover over each module with some tips. Like "this should only go after __" and "this module does ___" (like the core student reg module, who knows what that means and where it should go in the order? presumably first? should it be required?). I think some constraints/tips for each module would be helpful. 

Also, in a similar vein, I'm wondering if we can add a "restore to defaults" button so that when someone inevitably changes a ton and can't get anything to work anymore, they can start over with the default sequence. Probably a default button for sequence and another one for requiredness. 

Finally, Yale uses the "program module objects" admin page to also change required labels (ie to say that credit card payment is "required by ___" even though the module is not required). So could we incorporate the required labels field into this management page? Since this is very uncommon, perhaps there could be a little button next to each module with a comment glyph that would open a field below that would allow you to change the label for that module? Just so that this rare occasion doesn't clutter the page by having the field open for each module. Then if the required label has anything in it, that field would stay visible (so admins dont forget it's there and to change it if necessary).

willgearty:
OK, how's that all look @kkbrum?

kkbrum (changes requested):
The requirement label text boxes are too wide in fruit salad:

image

Otherwise, everything looks great!

kkbrum (approved):

PR #3515, opened 2022-04-04
Fix phase zero lottery for complex grade caps

### LINKED ISSUE #3514: Phase zero lottery doesn't work for complex tag values
The `program_size_by_grade` tag is used to determine how many students of each grade can be accepted in the student lottery (phase zero). Generally, this tag is set at the individual grade level, e.g., `{"9": 80, "10": 80, "11": 80, "12": 80}`. However, it can also be more complex, with each capacity spanning multiple grades, e.g., `{"7-8": 1000, "9": 300, "10-12": 1500}` (this is the example given in the tag help text). If a tag whose capacities span multiple grades, the phase zero management page loads, seemingly indicating everything is fine, but then the lottery fails to run. The issue is here:
https://github.com/learning-unlimited/ESP-Website/blob/187d10f204c5a52f49f35f5a7021265c218cdba5/esp/esp/program/modules/handlers/studentregphasezeromanage.py#L72-L74
I think the solution is to not modify the grade caps like we do in line 74. Instead, we should keep the multiple grades and check which dictionary key each student's grade is `in`.

willgearty:
This fixes the student lottery to work with complex grade capacities (e.g., `{ "7-9": 3, "10-12": 4}`). I've also cleaned up and clarified the lottery error and success messages to be clearer to the admin user.

Fixes #3514.

### DIFF
--- a/esp/esp/program/modules/handlers/studentregphasezeromanage.py
+++ b/esp/esp/program/modules/handlers/studentregphasezeromanage.py
@@ -59,15 +59,14 @@ class StudentRegPhaseZeroManage(ProgramModuleObj):
         }
 
     def lottery(self, prog, role, post):
-        messages = []
+        messages = {'success':[], 'error': []}
         winners, _ = Group.objects.get_or_create(name=role)
         students = []
 
         if post.get('mode') == 'default':
             # Get grade caps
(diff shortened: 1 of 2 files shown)


kkbrum (approved):
Woah!

PR #3520, opened 2022-04-12
Add a user interface for permissions

willgearty:
This adds a user interface for permissions for individual users to the "Deadline Management" page (now "Deadline and Permissions Management"). I've made a clear distinction between deadlines (for user ROLES) and permissions (for individual USERS), even though technically they are the same thing in the database. My thought is that this will hopefully reduce confusion around the terms "deadlines" and "permissions" (which we have tended to use interchangeably up to now). The interface works just like the deadlines interface (opening/closing/adding/deleting/editing), and the permission form has an autocomplete field for users.

Another step towards #3497


kkbrum:
I'm getting this error when I try to assign a deadline to a custom group

image

willgearty:
OK, that should fix that error @kkbrum (and any further errors are now included within the form instead of in the error banner)

kkbrum (approved):
🤯

Issue #3199, opened 2021-01-28
manage program settings page text cut off


kkbrum:
In fruitsalad, on the manage program settings page, some of the module questions are cut off:

![image](https://user-images.githubusercontent.com/25392075/106081002-28e87680-60e6-11eb-8898-d73d5e0e1059.png)

willgearty:
This appears to be fixed? I'm not able to reproduce the problem on my dev server nor on Yale's site.

PR #3521, opened 2022-04-17
Add user interface for DBreceipts

willgearty:
This adds a user interface to edit the DBReceipts for a program to the program settings page:
image

Another step towards #3497.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3199.


kkbrum (changes requested):
The help text paragraphs are cut off on the right in fruitsalad 

image


This is the error i got when trying to confirm reg when i have not yet confirmed reg (all default dbreceipts)

image

I also get an error when trying to confirm or cancel reg if i've already confirmed it, but I'm not sure if my program is just broken or if it's a real error. I cant test on another program while the above error exists though.

image

willgearty:
That should fix the first two issues @kkbrum. The cancelreg should show the "cancel" receipt. Did you perhaps copy the confirmation text to the cancel receipt during testing? Also, it only actually cancels classes if this box is checked:
image

kkbrum (changes requested):
Could you add "registration receipts" to the list on the program management page describing what the program settings page does? It currently says: "Modify Settings for This Program: Emails, program caps and modules, class categories, and flag types"

Whether I have text in the cancel receipt or not, and whether I check the cancel classes option or not, it still redirects me to the confirm reg receipt instead of the cancel receipt or core reg...

kkbrum (approved):

PR #3525, opened 2022-05-04
Prettify teacher class info

### LINKED ISSUE #3455: Class list on teacher mainpage is UGLY
We've added so many buttons, so there's lots of functionality, but now it just doesn't look very good IMHO. I think we could either make some of the buttons/details hidden with a "expand for options" link or have another page that you need to click to to access some of the functionality. It also probably wouldn't hurt to use bootstrap buttons (that might even be an easy way to clean things up, with a button dropdown or something like the class search/printables pages).

This is pretty gross:
image

kkbrum:
This fixes #3455 by consolidating many buttons into a "Manage options" dropdown list and a "Student list options" dropdown list. I only looked at this in fruitsalad theme!

### DIFF
--- a/esp/templates/inclusion/program/class_teacher_list_row.html
+++ b/esp/templates/inclusion/program/class_teacher_list_row.html
@@ -43,10 +43,10 @@ Popup div for class cancellation request
   </td>
 <td>
     <div class="btn-group" id="manage_links">
-        <a class="btn dropdown-toggle" data-toggle="dropdown" href="#">
+        <button class="btn btn-primary dropdown-toggle" data-toggle="dropdown">
             Manage options
             <span class="caret"></span>
-        </a>
+        </button>
         <ul class="dropdown-menu">
           <li><a href="/teach/{{program.getUrlBase}}/editclass/{{cls.id }}" title="View/edit class" target="_blank">View/Edit</a></li>
     <li><a href="/teach/{{program.getUrlBase}}/coteachers?clsid={{cls.id}}" title="Add/remove coteachers" target="_blank">Modify Coteachers</a></li>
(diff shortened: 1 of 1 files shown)


esp/templates/inclusion/program/class_teacher_list_row.html
```diff
@@ -72,58 +41,92 @@ <h3>Cancellation Request for {{ cls|truncatewords:6 }}</h3>
...
-    <a href="/teach/{{ program.getUrlBase }}/survey/review?classsubject_id={{ cls.id }}"
-       title="Review Results from Student Survey" class="abutton">
-      Student Survey Results
-    </a>
+<td>
+    <div class="btn-group" id="manage_links">
+        <a class="btn dropdown-toggle" data-toggle="dropdown" href="#">
+            Manage options
+            <span class="caret"></span>
+        </a>
```
willgearty, line 49:
```suggestion
        <button class="btn btn-primary dropdown-toggle" data-toggle="dropdown">
            Manage options
            <span class="caret"></span>
        </button>
```

esp/templates/inclusion/program/class_teacher_list_row.html
```diff
@@ -72,58 +41,92 @@ <h3>Cancellation Request for {{ cls|truncatewords:6 }}</h3>
...
+        </ul>
+    </div>
+</td>
+    {% if cls.isAccepted %}
+<td>
+    <div class="btn-group" id="student_links">
+        <a class="btn dropdown-toggle" data-toggle="dropdown" href="#">
+            Student list options
+            <span class="caret"></span>
+        </a>
```
willgearty, line 81:
```suggestion
        <button class="btn btn-primary dropdown-toggle" data-toggle="dropdown">
            Student list options
            <span class="caret"></span>
        </button>
```

willgearty (changes requested):
This looks great and is certainly much cleaner than what we had before! Not sure if this is a bigpicture thing, but the `<a>` tags behaved a little weirdly (e.g., they became white when hovering), so I suggest changing them to `<button>`s since they aren't linking to anything anyways. In general, `<button>`s behave better with the `btn` class than `<a>`s and are definitely preferred when you don't need to link to anything.

willgearty (approved):

PR #3526, opened 2022-05-06
Add teacher bio link

### LINKED ISSUE #1730: Teacher bio more prominent and easy to find
Teacher bios are a prominent google result for many teachers. It would be nice if it were easier to find.

At minimum, the teacher bio should be available from the [profile page](https://esp.mit.edu/myesp/accountmanage.html) and ideally from the edit profile page as well (since teachers are periodically shown this page when registering classes).

kkbrum:
Adds link on the account manage page to edit teacher biography. Fixes #1730


willgearty (approved):
Looks good to me, thanks!

PR #3531, opened 2022-05-11
Add interfaces to add and remove students to/from the student lottery

willgearty:
This adds an interface on the Phase Zero Management page to add students to and remove students from the student lottery:
image

This is another step towards #3497.


kkbrum (commented):
This looks great and works perfectly. My only thought is that maybe the options to add/remove students from the lottery should be before the options to run the lottery?

kkbrum (approved):

PR #3533, opened 2022-05-15
Sort custom forms by the programs/courses they are associated with

### LINKED ISSUE #3522: Show custom form program affiliation on /customforms/
Right now, it can be hard to know which custom form is assigned to which program (or not assigned at all). Instead of sorting by form ID on /customforms/, I'd suggest grouping the forms by which program they are affiliated with (and we could even list the module that they are used for). If this is too tricky, we could at least have a column that lists the program.

willgearty:
image

Fixes #3522.

### DIFF
--- a/esp/esp/customforms/views.py
+++ b/esp/esp/customforms/views.py
@@ -25,7 +25,7 @@ def test_func(user):
 
 @user_passes_test(test_func)
 def landing(request):
-    forms = Form.objects.all().order_by('-link_type', '-link_id')
+    forms = Form.objects.all().order_by('-link_type', '-link_id', '-id')
     if not request.user.isAdministrator():
         forms = forms.filter(created_by=request.user)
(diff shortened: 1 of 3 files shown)


esp/esp/customforms/views.py
```diff
@@ -25,9 +25,12 @@ def test_func(user):
 
 @user_passes_test(test_func)
 def landing(request):
-    forms = Form.objects.all().order_by('-id')
+    forms = Form.objects.all().order_by('-link_type', '-link_id')
```
kkbrum, line 28: I think this says to sort by whether program/other, then program id, but I think we should then also sort by form id. Because in the other forms section, it looks very odd to me that the forms aren't very sorted

kkbrum (changes requested):

kkbrum (approved):

PR #3535, opened 2022-05-25
Fix dashboard stats and better distinguish between "reject" and "cancel"

### LINKED ISSUE #3213: Dashboard counting approved classes even if no approved sections left
I have an approved class with two sections, both of which have been cancelled. 

The big board & statistics pages skip over this class when counting things such as teachers teaching an approved class, approved classes, etc. However, the dashboard still counts this class for such things. I think it should skip over it like the big board & statistics pages do. 

Perhaps the best solution is that when the last section of a class gets cancelled, the class status should switch to cancelled as well.

willgearty:
This started off as a quick fix for #3213, but quickly devolved into a larger PR to deal with the longstanding confusion between the "reject" and "cancel" class/section statuses. These have always had different backend meanings, but I feel like they've used fairly interchangeably on the frontend. This changes this behavior such that classes/sections may only be "rejected" if they have not been scheduled. Once a class/section is scheduled, it can only be "cancelled". This change has been implemented on the /manageclass, /classsearch, and /dashboard pages.

Further, as @kkbrum requested in #3213, when a section's status is changed, we now check if all sibling sections of the section's parent class have the same status. If they do, we then change the class to this status as well.

Fixes #3213.


kkbrum:
I believe(?) there may be a weird edge case as follows: 
1. cancel the one section of an accepted class
2. it wants to change the whole class to cancelled but says it has no scheduled sections so needs to be rejected in the form instead
3. the form still says cancelled though and if you use it, it tells you that there are scheduled sections so you need to use the cancellation form, which lands you back to point number 2.
4. however, in all of this, if you then refresh the dashboard, the whole class DID get cancelled like it should have, and if you reload a new copy of the manage class page, the whole class is cancelled.
5. so i think when you cancel the only section of a class, the manage page needs to refresh itself?

kkbrum:
edge case:
- you have one section of a class, and it is cancelled (the class is thus also cancelled)
- you try to approve the class, and the there's no error message or anything, but when the page refreshes, everything is still cancelled
- i know we agreed on this behavior, and it makes sense for when you approve a class and there is 1 unreviewed section and 1 cancelled section. But when there is just 1 section, i think its confusing. So i propose two possible options:
- 1. if you change the status of either the class or the section, but there is only 1 section, then just apply that switch to both the section and class (even if there is a rejected section or whatnot)
- 2. if there is only one section, then grey out the class status options with hover text saying to use the section form instead. since the status of the section will automatically change the status of the class since there's only the one section, everything will match nicely
- I think option 2 is cleaner coding wise

kkbrum:
okay so your fix for redirecting after cancelling a section worked in terms of the page being clean, but now it doesnt cancel the class like it should have (even if you reload in the dashboard or a new manage page)

kkbrum:
The only thing that still seems weird to me is- 
i rejected a class and it now has 2 rejected sections. 
Then I accept one of the two sections.
The class remains rejected. But what does it actually mean if a class is rejected and has an accepted section? Isn't that misleading when you look at the dashboard and it says the class is rejected but then in actuality it has an accepted section? 
I'm wondering whether it shouldn't let you turn the status of a section to a "greater" value than the class status and give you a pop up saying that you would need to upgrade the class status first in order to upgrade any sections?

willgearty:
OK, the new changes adds an alert message that pops up if you try to approve a section of an unapproved class. I also fixed a very small bug where if you changed the status dropdown of a class or section from cancelled to unreviewed, then changed it back to cancelled (before submitting anything), you got the "already scheduled" alert.

kkbrum (approved):
Beautiful!

PR #3536, opened 2022-05-27
Rename documents

### LINKED ISSUE #2226: Allow renaming uploaded class documents
This would be convenient if we could.

kkbrum:
Fixes #2226


esp/esp/web/forms/fileupload_form.py
```diff
@@ -47,3 +47,6 @@ class FileUploadForm_Admin(forms.Form):
 
     def set_choices(self, new_choices):
         self.fields['target_obj'] = forms.ChoiceField(choices = new_choices)
+
+class FileRenameForm(forms.Form):
+    title = forms.CharField(widget = forms.TextInput({'size': 25}), label = 'Rename')
```
willgearty, line 52:
```suggestion
    title = forms.CharField(widget = forms.TextInput({'size': 25}), label = 'Rename')

```

willgearty (changes requested):
Overall, this works great! However, there seem to be some styling things that could use some work (tested on bigpicture).
1. On the manage page, when the title of a document is more than one line long, there is a weird indenting:
image

2. On both the manage page and teacher page, the rename button isn't lined up with the input field:
image

willgearty (approved):

PR #3538, opened 2022-06-01
Remove link to admin pages

### LINKED ISSUE #3497: Hide administration pages

willgearty:
This removes the link to the admin pages for all themes (unless you are on a dev server). I also cleaned up some old code, made it so the class search field only shows if the class search module is enabled, and added a scary banner to the top of the admin pages.

Note, I'm not sure you can actually test the link removal functionality on a dev server, because I tried setting `DEBUG = False` in my settings, and it caused all sorts of issues.

Fixes #3497.

### DIFF
--- a/esp/esp/themes/theme_data/barebones/templates/sidebar/admin.html
+++ b/esp/esp/themes/theme_data/barebones/templates/sidebar/admin.html
@@ -1,34 +1,6 @@
-<!-- <li class="hidden nav-header admin">Admin</li>
-
-{# TODO(benkraft): Allow configuring these too. #}
-<li class="admin hidden"><a href="/admin/">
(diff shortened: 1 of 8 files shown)


kkbrum (approved):
the class search disappearing if not enabled and the banner look good to me! didn't do anything about testing the link to the admin pages going away.

PR #3549, opened 2022-08-10
Add qsd blocks to student profile form

### LINKED ISSUE #1763: Profile studentinfo text should be QSD
The text in the yellow box under "Student information" is not editable.  In particular, it links to a QSD page that may or may not exist.  Duke has a local change to remove that link, but we should likely just make the entire assembly QSD.

kkbrum:
Fixes #1763


willgearty (approved):
Great, thanks!

Issue #3448, opened 2021-11-12
Create arbitrary record types


kkbrum:
With #3443, there will be many nice interfaces to set/monitor records for record types that already exist, but I feel like some programs may want record types that don't already exist, so it would be cool if admins could create new record types.

PR #3550, opened 2022-08-12
Custom record types

willgearty:
This adds the ability for admins to create custom record types. This makes the new program record interfaces (https://github.com/learning-unlimited/ESP-Website/pull/3443) much more powerful because any arbitrary required action can now be documented for teacher/student registration.

I added several migrations which ensure that existing records are migrated to having record types (instead of string "events"). All of the built-in record types are added to the database (even if they weren't being used for existing records), and any of these built-in record types are not allowed to be edited or deleted. I believe I've updated all of the code that used records accordingly, although it is possible I've missed something.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3448.


kkbrum:
if you set a custom record as required for student/teacher reg for a program, then delete the record type, then go to reg, you get an error. Deleting a record should probably delete any associated tags as well

willgearty:
I took a slightly different approach and just skipped any RecordTypes that don't exist anymore.

kkbrum (approved):

PR #3559, opened 2022-09-28
Big picture hardcoded

### LINKED ISSUE #2906: Big Picture Theme - Some theme settings seem to be hardcoded.

kkbrum:
Fixes #2906. Every customizable parameter should now have some effect throughout the big picture theme. Almost everything should be customizable in an intuitive way. Some things aren't perfectly customizable given the current customization parameters, but more parameters could be added later if desired.

### DIFF
--- a/esp/esp/themes/theme_data/bigpicture/less/main.less
+++ b/esp/esp/themes/theme_data/bigpicture/less/main.less
@@ -5,14 +5,12 @@
 #main {
     h1 {
         color: #444;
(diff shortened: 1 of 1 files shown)


willgearty (approved):
Everything works great! I didn't think it would ever be possible to have a different font for the big picture theme!

PR #3580, opened 2022-11-18
Student self checkin

### LINKED ISSUE #2892: Self check-in
Theoretically, a student might have already completed all check-in tasks before arriving at a program, for example:

1. Paid online
2. Filled out waiver online (formstack/DocuSign module yet to be implemented)
3. Using webapp for schedule/map (don't need printed schedule/map)

In this case, these students should be able to check themselves in (maybe scan a qr code or use a code on a site or something), probably through the webapp

willgearty:
This adds the infrastructure for students to check themselves in if they have already completed all required registration actions. This includes (partially copied from #2892):

- A tag that specifies whether self checkin requires a unique code or not (or whether it is disabled, the default)
- If a unique code is NOT required, then students just need to go to a page with a checklist of required uncompleted tasks. If the whole checklist is done (in which case there is no checklist), they can click a button and they are checked-in and are taken to the webapp
- If a unique code is REQUIRED, then students will need a code for the same page. This code is a short hash code is unique to a student and the particular program. They need to enter this code into the form to check themselves in.
- These hash codes can be included on the backs of student nametags along with the existing barcodes.

Requested by Stanford.

Fixes #2892.

### DIFF
--- a/esp/esp/program/modules/forms/admincore.py
+++ b/esp/esp/program/modules/forms/admincore.py
@@ -1,6 +1,7 @@
 from decimal import Decimal
 from django import forms
 from django.contrib import admin
+from django.template import Template, Context
 from django.template.loader import select_template
 from django.utils.safestring import mark_safe
 from form_utils.forms import BetterForm, BetterModelForm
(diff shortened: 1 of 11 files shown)


kkbrum (approved):
It all works so well!!

PR #3592, opened 2023-01-13
Change extracost selections after paying

### LINKED ISSUE #3570: Allow purchasing additional items after paying
Now that we can pay via credit card for partial amounts (e.g., after receiving partial financial aid, https://github.com/learning-unlimited/ESP-Website/pull/3157), it is now possible to add items to your "cart" after paying and then paying again. Therefore, we should remove the restrictions that prevent students from going to the donation and extracosts modules after paying. We could even change the /payonline page such that you could increase your donation amount after paying (e.g., after really enjoying the program) (I bet @bhorowitz would really like that).

willgearty:
This adds the ability (enabled by default) for students to return to the extracosts page after they have paid via credit card and choose different extracost options. This could be a 1 for 1 change (where there is no balance change), a positive change in balance (where the student would then need to pay via credit card again), or a negative change in balance (which we treat as a temporary/permanent donation).

The students can return to this page as many times as they would like between and after paying (provided the tag is enabled, which is the default). As changes are made, the (new!) subtotal (and balance if the student has paid already) at the bottom of the form updates automatically.

image

Fixes #3570.

### DIFF
--- a/esp/esp/program/modules/forms/lineitems.py
+++ b/esp/esp/program/modules/forms/lineitems.py
@@ -16,6 +16,9 @@ class LineItemForm(forms.ModelForm):
         }
         help_texts = {
             'required': 'Should this line item be automatically added for every student?',
+            'for_finaid': 'Should financial aid cover this line item? Note that if this is checked, all quantities will be covered up to the max quantity. \
+                           If you would like only one instance of this line item to be covered by financial aid, you should set the max quantity to 1 and make \
+                           a duplicate line item with a different name and higher max quantity that is not covered by financial aid.'
         }
         widgets = {
             'amount_dec': forms.NumberInput(attrs={'placeholder': '(cost)'}),
--- a/esp/esp/program/modules/forms/splashinfo.py
+++ b/esp/esp/program/modules/forms/splashinfo.py
@@ -34,9 +34,10 @@ Learning Unlimited, Inc.
 
 from django import forms
 from esp.program.models import Program
+from esp.utils.widgets import RadioSelectWithData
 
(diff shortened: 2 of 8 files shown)


kkbrum (changes requested):
This looks awesome! Thanks for getting all of this to work!

There is one edge case: If someone received financial aid, and then later adds an item that is covered by financial aid, the financial aid amount in the subtotal does not change. If they were given a percentage for financial aid, they will automatically receive more aid for the covered item-- this ends up reflecting correctly in the credit card form, but not in the subtotal calculated here. So I think there are two options- 1) recalculate the financial aid amount as people click on things or 2) put a little asterisk  next to the financial aid in the subtotal and then at the bottom (or the top along with the other message), add a note saying that the financial aid amount may change once different options are selected. Technically option 1 is more informative for the user (since they otherwise don't typically know which items are covered by FA and which arent), but it's obviously more tedious so I'm also fine with option 2.

Also, at least in fruitsalad, there isn't any space below the form before the end of the page-- a space would be nice. I didn't see an unclosed div anywhere causing the cut off page, but perhaps there is something lurking...

willgearty:
I've made the following changes:

- Financial aid now updates dynamically in the extra costs form
- Reordered the subtotal lines
- More help text for the line items form to try to prevent cases of financial aid covering lots of quantities of items

kkbrum (changes requested):
This is great! One small thing: I don't love how if you have no financial aid, it removes the financial aid line heading but not the -0.00 on that line. Can we just remove the line entirely? 
image

kkbrum (approved):
🥳 🎉 😄

PR #3605, opened 2023-02-01
Fix bulk account module issues

### LINKED ISSUE #3576: Bulk create accounts module issues
1. The module links to the admin pages for creating new groups. This should be updated to link to the User Group module.
2. The module fails when you try to make a set of users with a prefix that has already been used. Either we should fail gracefully with a useful error or we should attempt to continue the numbering of the previous set of users.

willgearty:
This makes the following changes to the bulk account creation module:

- Changed the link to the admin pages to the user group module page
- Now return an error if a prefix has been used before (this is checked by just looking for a user with the username {prefix}1)

Fixes #3576


kkbrum (changes requested):
Looks good. While we're here, can you add a return link back to some manage page from the "success" page for bulk creating accounts? It feels weird that there's nowhere to go from that page.

kkbrum (approved):

PR #3606, opened 2023-02-01
Support any record type name

### LINKED ISSUE #3586: Record type names with capital letters break the record interfaces
If you include capital letters in the name of a custom record type, the record type will not be shown in the registration interfaces because we use `.lower()`:
https://github.com/learning-unlimited/ESP-Website/blob/f9c11c8c45fc7e5e4f58a86041e8bd70963516a5/esp/esp/program/modules/handlers/studentregcore.py#L271-L274

willgearty:
I fixed the handling of record types in student/teacher registration to now allow any string to be used as a name for a record type (before, only snake_case was really supported).

Fixes #3586


kkbrum (changes requested):
This looks great. Not really related, can you add a link from the manage user records page to the records creation page?

kkbrum (approved):

PR #3608, opened 2023-02-10
Self checkin fixes

### LINKED ISSUE #3587: Students enrolled in zero classes can't self checkin
From Stanford:

"If a student is in the program (Splash Fall 2022 Winner) but has enrolled in zero classes, their selfcheckin will display the message "You must complete the following registration requirements before you can check yourself in:" but with nothing after it."

willgearty:
This makes it so that the "Sign up for classes" link shows up (and links to the main studentreg page) on the student selfcheckin page if the class registration module is required and the student hasn't registered for any classes yet:
image

This also adds to option (on by default) to require payment before self checkin (as requested by Stanford). If neither of the credit card modules are enabled, the student is instructed to see an admin team member.

Fixes #3587.

### DIFF
--- a/esp/esp/program/modules/base.py
+++ b/esp/esp/program/modules/base.py
@@ -282,6 +282,9 @@ class ProgramModuleObj(models.Model):
 
         return mark_safe(link)
 
+    def makeSelfCheckinLink(self):
+        return self.makeLink()
+
     def get_setup_title(self):
         if hasattr(self, 'setup_title') and self.setup_title is not None and str(self.setup_title).strip() != '':
             return self.setup_title
--- a/esp/esp/program/modules/handlers/studentclassregmodule.py
+++ b/esp/esp/program/modules/handlers/studentclassregmodule.py
@@ -209,8 +209,12 @@ class StudentClassRegModule(ProgramModuleObj):
         return (len(user.getSectionsFromProgram(self.program)[:1]) > 0)
 
     def makeSelfCheckinLink(self):
+        if self.deadline_met():
+            text = self.module.link_title
+        else:
+            text = "Class changes is currently closed, please contact the admin team to register for classes"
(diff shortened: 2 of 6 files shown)


kkbrum:
This looks good.

I do think it's a bit odd that students have to go out of the webapp just to return later. I know we don't allow changing classes via webapp unless a student is checked in, but if they can add classes on the site, presumably they should be able to add classes on the webapp?

Given that we are telling them to add classes before checking in, perhaps we should change the webapp rules and allow changing classes via webapp whether checked in or not, and just not show room numbers until a student is checked in. Or, we could only allow this functionality if they came directly from the self checkin page. And, if a student came from the self checkin page to the schedule page, then we would add a "return to self checkin" link to the schedule page (so they dont have to retype the link or hit the back button).

willgearty:
@kkbrum, thanks for the comments! I've made it so that if class registration is open on the main student registration page, students can also use the webapp to register for classes. I've also made it so the link now directs to the webapp schedule page if you need to register for classes to self checkin. I'm not sure how to keep track of students that were sent to the webapp pages via the self checkin page (especially since they will then be adding/removing classes potentially multiple times), so I didn't add a "return to self checkin" button.

kkbrum (commented):
When the deadline to change classes is closed, the message for the greyed out class changes on the webapp is still "Not checked in yet!" but the checkin page says to add classes and links there. So that's pretty confusing to students-- could we change the error message to be "class changes are currently closed, please contact the admin team to make any changes" or something? and maybe on the self checkin page, if class change deadline is closed, change the message to "see the admin team to check in and enroll in classes" or something?

willgearty:
Let me know if that adequately addresses your concern @kkbrum!

willgearty:
Note that I've also added the payment requirement option to this PR @kkbrum.

kkbrum (changes requested):
The change I suggested is required for the tag to work properly, but otherwise, looks good!

PR #3615, opened 2023-02-15
Removed time estimate and added a link to /manage/emails. 

### LINKED ISSUE #3503: Update or remove estimate of time to send emails
Our current email estimate 1.5 seconds * number of emails is very inaccurate. We should update it (ideally with some kind of ping to Sendgrid's API) or just remove it entirely.

laraao:
# Path of page in question is: 
`/manage/Splash/<program name>/commfinal` which appears after a user has sent an email to users in the database.

# Changed:
`The email is in the process of being sent. To avoid being marked as a spam source, it will take approximately 1.5 seconds to email everyone on the list, so please be patient!`

# New Change:
Instead of listing an approximate time, users will instead be directed to monitor emails page at `/manage/emails`.

# Screenshots:
# # Old:
3503-old-crop
# # New:
3503-new-crop

Fixes #3503


willgearty:
Now that we don't display a time, we can probably also get rid of the code that was used to figure out that time:
https://github.com/learning-unlimited/ESP-Website/blob/main/esp/esp/program/modules/handlers/commmodule.py#L203-L210

kkbrum (approved):

PR #3616, opened 2023-02-17
Migration to Python3

### LINKED ISSUE #2576: Upgrade to Python 3

milescalabresi:
Migrate the codebase to Python 3 (currently 3.7, aiming for higher once we bump Django and other packages)

Fixes #2576.


hwatheod:
Regarding Python 3.7, its end-of-life is only a few months way ([2023-06-27](https://www.python.org/downloads/)).

milescalabresi:
Good catch, Ted. I think we want to avoid upgrading Django in this PR (the main limiting library), but we should make a separate issue to bump Django as soon as possible. Then we can bump to a current version of Python. Only potential reason I can see to do it sooner is that this tool https://docs.python.org/3.10/library/functools.html#functools.cmp_to_key buys us a lot of ground with the `cmp` issue

milescalabresi:
For the record, once the tests are fixed up and everything, you can test locally by running `fab manage:test`

hwatheod:
Using the raw string is the right idea (so we don't need pickle at all), but you also need to get rid of the pickle when the value is read later (line 544 of the same file).

milescalabresi:
Thanks @hwatheod. Do you think it's worth trying to use something like this https://opensource.com/article/20/11/django-rest-framework-serializers? Or just go with raw strings?

hwatheod:
I think our objects are simple enough that it's overkill to use such serializers. 

In this particular case, pickle is easy to remove because we pickle and then immediately unpickled on the other side. This is more-or-less the way pickle is intended to be used.

The real difficulty comes with the cases where pickled objects are stored in the database (which you are not supposed to do, but we did). That's where #3411 left off. It's worth reviewing all the discussion there on pickle again before tackling that case.

milescalabresi:
I see. Maybe the django serializers could be used there? An alternative that would be a lot of work but is is cleaner at the end is to migrate the database and store the objects individually... I'll have to read those cases more closely

hwatheod:
I would double check the documentation of the serializers to see if they are really intended for long term storage in a database, so we don't run into the same problems we did for pickle.

Pickle is intended as a short-term form of storage only, and even between one version of django and the next, pickles stored in a database can become unreadable because the pickle hard codes the names of classes into the serialized form, and django sometimes changes the names of classes from one version to the next. In the discussion in #3411 I mentioned I have found some pickled objects in our DBs from older versions of Django that can no longer be read by 1.11.

milescalabresi:
To explain the previous commit a bit, Python 2 used the slash symbol `/` for integer division, unless one argument was already a float or Decimal. Python 3 assumes floating point division. Thus to "go back," modernize means changing single slashes to double slashes. However, it misses that "unless," so we have to look for obvious cases where one operand is a float. I used the decimal point and the `float()` constructor as signs. Terminal commands below, in case they help in the future.

```
find ./esp -name '*.py' | xargs git grep -n '\.[0-9].*[^\/\:]\/\/[^\/].*' # decimal point before division
find ./esp -name '*.py' | xargs git grep -n 'float.*[^\/\:]\/\/[^\/].*' # `float` constructor before division
find ./esp -name '*.py' | xargs git grep -n '[^\/\:]\/\/[^\/].*float' # `float` constructor after division
find ./esp -name '*.py' | xargs git grep -n '[^\/\:]\/\/[^\/].*\.[0-9]' # decimal point after division
```

kkbrum:
Given how much testing is required here, could one of you push this to dev5 to speed things up a bit? @milescalabresi @willgearty

docs/dev/vagrant.rst
```diff
@@ -141,48 +141,160 @@ Problems
...
+1. Make a copy of `esp/esp/local_settings.py` somewhere with a different name (e.g., on your desktop as "old_local_settings.py"). The `local_settings.py` file will get overwritten by the end of this process and you will want to restore the settings from your previous VM setup.
 
 2. From within the "devsite" folder, run ``fab dumpdb``. This action will save your database as a dump file in the `devsite` folder called `devsite_django.sql`. You can also specify a filename if you would like with ``fab dumpdb:filename``.
 
-3. Run ``vagrant destroy`` (note: this destroys your virtual machine. Only do it once you are sure your database has been backed up and you are ready to continue).
+3. Run ``vagrant destroy`` (note: this destroys your virtual machine. Only do it once you are sure your database has been backed up and you are ready to continue). 
 
 4. Now follow the `VM installation instructions above <#installation>`_, starting at ``vagrant up``.
```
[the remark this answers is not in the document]
kkbrum: (actually, i think steps 4-5 are duplicates of steps 6-8, where it specifies the branch in step 6)

docs/dev/vagrant.rst
```diff
@@ -141,48 +141,160 @@ Problems
...
+If you run into trouble, clear your SSH keys in ``~/.vagrant.d/`` and ``devsite/.vagrant``.
```
kkbrum, line 289: Is this only if you have trouble with the vagrant up step, or is it also if you have trouble with the fab setup step?

vagrant up was fine, but I'm having trouble with fab setup. 

I got this error:

image


so then i tried to drop the tablespace encryption but it didn't work:

image

docs/dev/vagrant.rst
```diff
@@ -141,48 +141,160 @@ Problems
...
 
 Upgrading your personal dev VM
 ------------------------------
 
 If the base VM has been changed (see above), you will want to upgrade your development server. However, upgrading Ubuntu within a virtual machine can cause problems with your database. Therefore, you'll need to export your database, create a new virtual machine, then import your database:
 
-1. Make a copy of `esp/esp/local_settings.py` somewhere with a different name (e.g. on your desktop as "old_local_settings.py"). The original `local_settings.py` file will get overwitten by the end of this process and you will want to restore some of the settings from your previous VM setup.
+1. Make a copy of `esp/esp/local_settings.py` somewhere with a different name (e.g., on your desktop as "old_local_settings.py"). The `local_settings.py` file will get overwritten by the end of this process and you will want to restore the settings from your previous VM setup.
 
 2. From within the "devsite" folder, run ``fab dumpdb``. This action will save your database as a dump file in the `devsite` folder called `devsite_django.sql`. You can also specify a filename if you would like with ``fab dumpdb:filename``.
```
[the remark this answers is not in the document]
milescalabresi: 0. Suggestions accepted
1. Ah, I will add `vagrant up` that to step 2.
2. Yeah, I should have clarified this procedure is for upgrading the VM not upgrading the VM and the branch simultaneously as we are doing. In general, I think it's a bad idea to do both at once

docs/dev/vagrant.rst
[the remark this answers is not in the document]
milescalabresi: 1. Noted. I will add that
2-3. Yep, fixed

docs/dev/vagrant.rst
milescalabresi: Clearing keys only help with Vagrant. To drop the tablespace, you'd want to do `sudo -u postgres psql -c ...`. I'm surprised the tablespace already exists the first time you ran fab setup (unless you did it once before and it died mid-stream?) We can troubleshoot together

hwatheod:
I've been out of the loop on this PR for a while but it's great that you find a solution to the pickle migration problem. And also, `cucumber` is a great name for the util for working with pickles.

willgearty:
When running `fab setup` to make my new VM, I had to do the following undocumented things:

1. install `setuptools`
2. hit enter twice to confirm various steps

willgearty:
I also ran into the same error that @kkbrum encountered while trying to run `fab setup`:

> ERROR:  tablespace "encrypted" already exists

docs/dev/vagrant.rst
milescalabresi: Try again? I updated the  VM

milescalabresi:
I made a new, clean VM for Ubuntu 24.04, so hoping that fixes things. Try again, reviewers!

willgearty:
That seems to have worked @milescalabresi. I've a dev server up and running. Although I will note that it seems to be running significantly slower than before. I realize that all of the caching needs to be done again, but even loading the same page or submitting a form takes a looooong time.

kkbrum:
It worked for me as well. Mine seems a bit slower than before as well, but not as significantly as it seems to be for Will.

esp/esp/program/modules/handlers/programprintables.py
```diff
@@ -1327,9 +1327,9 @@ def get_student_schedules(request, students, prog, extra='', onsite=False):
             return render_to_response(basedir+'studentschedule.html', request, context)
         elif file_type == 'pdf':
             if len(students) > 1:
-                response = HttpResponse(content_type='application/pdf')
+                response = HttpResponse(content=render_to_latex(basedir+'studentschedule.tex', context, 'pdf'),
+                                        content_type='application/pdf')
                 response['Content-Disposition'] = 'attachment; filename="studentschedules.pdf"'
-                response.write(render_to_latex(basedir+'studentschedule.tex', context, 'pdf'))
```
milescalabresi, line 1331: @willgearty what is this supposed to do? I'm not sure I understand

willgearty: Not sure why, but on Python 3 the student schedule download was broken. This fixes it.

milescalabresi: can you help me understand what this change does? This is the file that causes the database errors that are making the PR tests fail

willgearty: That's bizarre. There shouldn't be anything different here than what was happening before on Python 2. We render the schedules as a pdf, then set the response as a file download. That's basically it.

milescalabresi: yeah I don't understand it. Maybe best to discuss in real time.... Or maybe this was actually mitigating an error in another file.  Stack seems to indicate these requests are notoriously asynchronous/hard to debug 🤷 😢

willgearty (approved):

willgearty:
🤩

Issue #3617, opened 2023-02-17
Clarify time slots on availability page when grouping is false


milescalabresi:
Setting  "Availability group timeslots" global teacher registration tag to false makes all of the timeslots appear in one group/block under the first day on the "indicate your availability" page for teachers, even if the slots are spread across multiple days. This issue is to modify that tag's behavior to clarify the availability page. Options include splitting over multiple days (but respecting the tag by grouping each day into a single block), changing the header over the single group/block so it shows the multiple days in one line, or inserting day "headers" into the single block. Reported by Chicago

PR #3620, opened 2023-02-22
Avail groups

willgearty:
This fixes the availability form interface for various timeslot grouping edge cases, such as:

1. The "Availability group timeslots" global tag is set to False (so all timeslots are in a single group)
2. A group of timeslots spans multiple days

While I was at it, I also made some minor visual changes to give the availability form a bit more breathing room:
image

I also updated the resources page to have the same formatting (so now admins will know roughly what the groupings will look like to teachers):
image

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3617.

### DIFF
--- a/esp/esp/program/modules/handlers/resourcemodule.py
+++ b/esp/esp/program/modules/handlers/resourcemodule.py
@@ -633,7 +633,10 @@ class ResourceModule(ProgramModuleObj):
 
         #   Group contiguous blocks of time for the program
         time_options = self.program.getTimeSlots(types=['Class Time Block','Open Class Time Block'])
-        time_groups = Event.group_contiguous(list(time_options), int(Tag.getProgramTag('availability_group_tolerance', program = prog)))
+        if not Tag.getBooleanTag('availability_group_timeslots'):
+            time_groups = [list(time_options)]
+        else:
+            time_groups = Event.group_contiguous(list(time_options), int(Tag.getProgramTag('availability_group_tolerance', program = prog)))
 
         #   Retrieve remaining context information
         context['timeslots'] = [{'selections': group} for group in time_groups]
--- a/esp/public/media/default_styles/availability.css
+++ b/esp/public/media/default_styles/availability.css
@@ -4,6 +4,10 @@
   align-items: center;
 }
 
+.container1 {
+  flex-grow: 2;
+}
+
 .side {
   width: 280px;
 }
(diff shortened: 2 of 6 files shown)


willgearty:
This now also supports custom timeslot groups, which you can specify in the new/edit timeslot form:
image

And here is what they look like below that:
image

esp/esp/program/modules/forms/resources.py
```diff
@@ -19,6 +19,7 @@ class TimeslotForm(forms.Form):
     hours = forms.IntegerField(widget=forms.TextInput(attrs={'size':'6'}))
     minutes = forms.IntegerField(widget=forms.TextInput(attrs={'size':'6'}))
     openclass = forms.BooleanField(required=False, label='Open Class Time Block', help_text="Check this if the time block should be used for open classes only. If in doubt, don't check this.")
+    group = forms.IntegerField(required=False, label='Group', help_text="All timeslots with this value will always be included in the same timeslot group.")
```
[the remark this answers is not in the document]
kkbrum: I would also add a sentence to the effect of "Any timeslots with no value indicated will be grouped together based on the 'Availability group tolerance' tag" and link to the tag settings page (and is it possible for the link to open the teacher registration settings div automatically?).

willgearty:
Thanks @kkbrum! I've made the suggested changes!

esp/esp/program/modules/handlers/resourcemodule.py
```diff
@@ -105,7 +105,7 @@ def resources_timeslot(self, request, tl, one, two, module, extra, prog):
 
             elif data['command'] == 'addedit':
                 #   add/edit timeslot
-                form = TimeslotForm(data, auto_id="timeslot_%s")
+                form = TimeslotForm(data, auto_id="timeslot_%s", program = prog)
```
kkbrum, line 108: I now get an error when I try to add or edit a timeslot:

image

image

esp/esp/program/modules/forms/resources.py
kkbrum, line 1: In the spirit of keeping the distinction between timeslots, time blocks, and groups clear, can we also change all the text that says "time block" when it is actually a "timeslot"? 
- in the description for "name"
- "open class time block" field and in the description
- in the "Type" displayed in the table of existing timeslots (currently it is "Class Time Block")
- Also in the template above the form, the text that says "So, for example, a 9 - 10 AM time block..."

kkbrum (changes requested):

PR #3624, opened 2023-03-08
Add theme color customizability

### LINKED ISSUE #1171: Background color doesn't have same effect across themes
On circles theme, background color changes the color in the sidebars AND the color of the content area.

On the fruitsalad theme, it only changes the sidebar color, but the content area is always white.

I assume the former, not the latter, is intended, though I'm not sure. We could also separate these into two different settings.

kkbrum:
Fixes #1171. 

More broadly, I went through all the customizable color variables in the theme editor and tried to make sure they did something in the current theme that more or less made sense. In some cases, I removed some variables from the theme editor depending on the theme and also altered some of the help text so it made sense given the current theme. Note that there really should be more of an overhaul of the help text on the theme editor page-- I did not address this fully in this PR. I'll make an issue about this.


esp/public/media/theme_editor/less/variables.less
```diff
@@ -240,6 +241,18 @@
 @navbarInverseBrandColor:                @navbarInverseLinkColor;
 
 
+// Sidebar
```
willgearty, line 244: Are none of these used for any of the themes? Is that why they are now commented out?

esp/templates/themes/editor.html
```diff
@@ -121,6 +121,15 @@ <h3 class="accordion-toggle" data-toggle="collapse" href="#scaffolding" rel="pop
...
+                    {% if theme_name != "bigpicture" and theme_name != "barebones" %}
+                        <div class="control-group">
+                            <label class="control-label" for="contentbg">Content Background:</label>
+                            <div class="controls">
+                                <input type="text" id="contentbg" name="contentBackground" class="color" value="{{contentBackground}}"/>
+                            </div>
+                        </div>
+                    {% endif %}
+
                         <div class="control-group">
```
willgearty, line 133: What are the "well" and "hero unit" backgrounds? I couldn't find anywhere that those were used in the themes I checked. Should those also be hidden for some/all themes?

willgearty (commented):
This looks and works great overall! Just a couple comments/questions...

esp/public/media/theme_editor/less/variables.less
kkbrum: lines 4-8 of this file have the note
 "// Note: Variables that are customized by theme editor
//       have been commented out.  The default values
//       are replicated in variables_custom.less and the
//       selected values will be included in the final
//       LESS output that is compiled to CSS."
so I commented these out since they are customizable

esp/templates/themes/editor.html
kkbrum: "well" is used for all themes on the themes landing page "http://localhost:8000/themes/" at the very least, I'm not sure if it's used elsewhere. 

Hero unit is used on the front page of barebones, but based on a github search, I don't think it's used elsewhere. I'll hide it for other themes-- thanks for catching that!

willgearty: Hmm...do we really need folks to be able to customize the themes landing page appearance?

kkbrum: Hmm, I agree that it's probably not necessary. I'll remove that option to clean things up.

willgearty (approved):
Great, thanks @kkbrum!

PR #3632, opened 2023-03-22
Implement Class Filters for the Catalog

### LINKED ISSUE #564: Filtering the catalog
Suggestion (paraphrased) from a student: In the lottery studentreg, fillslot, and catalog views, it would be nice to be able to filter by things like class length.

We've discussed similar things for grade ranges, but it would be cool to have filters for class length.  I know there are caching issues, but maybe write it in javascript and have the filtering client-side?

AlpacaFur:
This PR implements client-side filters for the class catalog that lets users filter the classes in the catalog by various conditions. 

The Grade filter was already implemented, but this PR adds **Duration**, **Status**, and **Difficulty** filters.

Resolves #564 

image

### DIFF
--- a/docs/admin/program_modules.rst
+++ b/docs/admin/program_modules.rst
@@ -378,7 +378,7 @@ This module allows teachers to register and view classes.  They can upload files
 The class creation/editing form requires that you have set up time slots for the program (see ResourceModule) in order to establish the possible lengths of classes.  It can be customized using the following Tags:
 
 * teacherreg_difficulty_label - This controls the name of the 'Difficulty' field on the class creation/editing form.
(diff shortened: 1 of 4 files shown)


willgearty:
Looks like you'll need to fix one (or more) of the catalog tests now, @AlpacaFur.

kkbrum (commented):
When I select options from the dropdowns, nothing happens. Here's what my console shows:

image

AlpacaFur:
Got so wrapped up in the possibility of non-default values for the difficulty tag that I forgot to go back and test when the tag isn't present 😔. Should be fixed now!

kkbrum (approved):

PR #3637, opened 2023-03-29
Consider index.html Redirects for Navbar

### LINKED ISSUE #3596: bigpicture nav not highlighted when link redirects
It's very common for programs to assign links to the nav bar that ultimately redirect to other pages. For example, /learn and /teach always redirect to /learn/index.html and /teach/index.html, respectively. The nav bar in many themes is designed to highlight the "selected" nav link, but this is broken if the ultimate URL does not match the URL listed for a nav bar item. See here for an example: https://dev5.learningu.org/teach/Splash/index.html (the "Euro Splash" link was clicked to get to this page).

Note: I only observed this with bigpicture, but this may be (and probably is) an issue with other themes

AlpacaFur:
This PR refactors existing nav-highlighting logic and considers the case where a link is the same as the page minus "index.html".

For example if the user is on `/learn/Splash/index.html` and there's a navbar entry with a href of `/learn/Splash/` or `/learn/Splash`, these will now be considered matches for highlighting purposes.

Resolves #3596

### DIFF
--- a/esp/public/media/scripts/nav.js
+++ b/esp/public/media/scripts/nav.js
@@ -13,8 +13,26 @@ $j('#side-dashboard a').prop('href', function() {
 
 //Select the appropriate nav-list <li> that contains an <a> which points to the current 
 //page, and make it active.
-$j('ul.nav li a[href="'+window.location.pathname+'"]').parent().addClass('active');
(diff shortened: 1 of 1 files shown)


kkbrum (approved):
Looks great! 🥳

Issue #3636, opened 2023-03-28
Some modules should not be able to be required


willgearty:
Some modules should not be able to made required in the modules UI. For example, right now making the confirm registration or credit card modules required breaks the site in various ways. There might also be modules that should never be optional.

Issue #3635, opened 2023-03-28
Can't make a module not required if all modules are required


willgearty:
If all modules are required, there is no "Not required" section in the modules UI, preventing admins from making any of the modules not required.

Issue #3634, opened 2023-03-27
Only steps are shown in modules UI


willgearty:
There are modules that are encountered by students during registration but are not shown as registration steps (e.g., the phase zero page). Right now, we filter these out for the modules UI:
https://github.com/learning-unlimited/ESP-Website/blob/db9865e4b543d13a884de0b83c79114f586eb412/esp/esp/program/modules/handlers/admincore.py#L558-L559

This then means that the order of these modules can not be changed via this UI. Should we add them back? Or should the phase zero page, for example, be added manually? What other modules are not `isStep`?

Reported by Yale and Columbia.

PR #3638, opened 2023-03-31
Mod UI fixes

willgearty:
This fixes the following issues with the modules UI:

1. The phase zero module is now shown on the page and there is now a way to manually include other non-isStep modules (Fixes https://github.com/learning-unlimited/ESP-Website/issues/3634).
2. There are always Required and Not Required sections, even when there are no modules in those sections (Fixes https://github.com/learning-unlimited/ESP-Website/issues/3635).
3. We now enforce various module setting restrictions (e.g., Credit Card Module can never be required) and list these restrictions in the Big Red Box (TM) (Fixes https://github.com/learning-unlimited/ESP-Website/issues/3636).
 
image

### DIFF
--- a/esp/esp/program/modules/base.py
+++ b/esp/esp/program/modules/base.py
@@ -398,6 +398,9 @@ class ProgramModuleObj(models.Model):
     def isStep(self):
         return True
 
+    def inModulesList(self):
+        return self.isStep()
+
     @classmethod
     def module_properties(cls):
         """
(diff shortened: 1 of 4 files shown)


hwatheod:
Why does the two-phase lottery module always have to be required?

willgearty:
Isn't it like the phase zero lottery where it shows a special page that prevents you from doing any optional steps?

kkbrum (changes requested):
Looks good other than the tiny wording changes I suggested!

Issue #3573, opened 2022-11-08
choosable=2 modules are hard to enable


willgearty:
Now that admins don't have easy access to /admin, it has become difficult to enable modules that are hidden by default (choosable=2). @milescalabresi thoughts on how to improve this?

milescalabresi:
Hmm, are we talking about enabling them after the creation of a program or also at creation time?

willgearty:
Both?

PR #3640, opened 2023-04-12
Copy over choosable=2 modules from template programs

willgearty:
This copies over any modules from a template program that are "choosable=2" (i.e., manually enabled). This makes it so we aren't encouraging the use of these modules, but we are copying them over in case chapters have decided to use them.

While I was at it, I also cleaned up the new program form with a pretty blue box with information about all of these seemingly random things we copy over from template programs.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3573.

### DIFF
--- a/esp/esp/program/modules/handlers/financialaidappmodule.py
+++ b/esp/esp/program/modules/handlers/financialaidappmodule.py
@@ -34,6 +34,7 @@ Learning Unlimited, Inc.
 """
 from esp.program.modules.base import ProgramModuleObj, needs_teacher, needs_student, needs_admin, usercheck_usetl, meets_deadline, main_call, aux_call
 from esp.program.modules import module_ext
+from esp.accounting.controllers import ProgramAccountingController
 from esp.utils.web import render_to_response, secure_required
 from esp.middleware      import ESPError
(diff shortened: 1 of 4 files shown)


kkbrum (changes requested):
This works well and I like the new alert box. When I made the new program, although it did enable the module from the imported program, it wasn't listed on the "double check everything" page with the long list of modules right before I clicked the final magic button. It should probably be listed there? Not sure why it wasn't.

willgearty:
OK, I've gone with a new approach that modifies the newprogram/program settings javascript to keep any custom modules. Now, instead of only enabling the modules that are included via the selected questions, it is the act of checking/unchecking questions that enables/disables only those specific modules. This means that these custom modules should now be shown on the "double check everything" page.

This also fixes a bug where the "charge for items" checkbox wouldn't uncheck if your program had an admission fee.

willgearty:
I'm also adding a small change to the financial aid module which makes it hidden if there are no costs to cover. I could have made a separate PR, but it's such a small change and somewhat related to the changes already implemented here. Fixes https://github.com/learning-unlimited/ESP-Website/issues/3641.

kkbrum (approved):

PR #3642, opened 2023-04-19
Added Grade Checking For All Student Modules

### LINKED ISSUE #1032: Many Student Program Modules Don't Check for Grade
We should consider adding `@meets_grade` to all student program modules. Or make a `@needs_student_in_grade` decorator and replace all program module uses of `@needs_student` with it.

laraao:
# Change Log
- The decorator `@needs_student_in_grade` combines existing `@needs_student` with `@meets_grade`.
- All occurrences of `@needs_student` have been replaced with `@needs_student_in_grade`. 
- All occurrences of  `@meets_grade` have been removed.

# Testing
- I have not been able to test the changes locally, but I'm not sure if the code is broken or my local website configuration is broken!

Fixes #1032

### DIFF
--- a/esp/esp/program/modules/base.py
+++ b/esp/esp/program/modules/base.py
@@ -557,6 +557,35 @@ def needs_onsite_no_switchback(method):
     _checkAdmin.has_auth_check = True
     return _checkAdmin
 
+def needs_student_in_grade(method):
+    def _check_student_and_grade(moduleObj, request, tl, *args, **kwargs):
+        """Check if student and is in correct grade."""
+        errorpage = 'errors/program/wronggrade.html'
+        if not_logged_in(request):
+            return _login_redirect(request)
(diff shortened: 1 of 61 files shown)


willgearty (changes requested):
The new decorator looks and works great, thanks @laraao!

I did notice that there are quite a few other uses of `@needs_student` and `@meets_grade` that need to be replaced with the new decorator. A quick search with Notepad++ (or your desired text editor) should reveal them. Once those are swapped over, I think this should be ready to merge!

willgearty:
I've updated this for all handlers and I even cleaned up a bunch of unused imports in the most recent commit (for review purposes, you may want to only look at the first three commits). @kkbrum could you please review this to make sure it still works?

kkbrum (approved):

Issue #466, opened 2012-11-10
AJAX scheduler legend


jmoldow:
The AJAX scheduler should have a legend that explains what bolding, underlining, italicizing, and various highlighting colors mean.

PR #3644, opened 2023-05-05
Legend for ajax scheduler

### LINKED ISSUE #461: Improve teacher availability highlighting in AJAX
From MIT: "having the timeblocks where teachers available light up even if there is already a class in the rooms you can see at that time, or maybe in the bar at the top where you can see all the times."

Availabilities are currently only highlighted in blocks for empty rooms. If a teacher is free in timeblock n, but you are currently looking at a section of the grid that has classes assigned in timeblock n to all classrooms that are in the visible window, you might be tricked into believing that the teacher is not free in timeblock n. It gives the appearance of more restricted availability.

willgearty:
This adds a popup legend to the ajax scheduler. We no longer do any sort of text styling, so this only is a legend for the cell background and border colors.

While I was at it, I also fixed various bugs related to moderator assignment and scheduling classes with moderators.

image

Fixes https://github.com/learning-unlimited/ESP-Website/issues/466. Fixes #461.

### DIFF
--- a/esp/esp/program/modules/handlers/ajaxschedulingmodule.py
+++ b/esp/esp/program/modules/handlers/ajaxschedulingmodule.py
@@ -43,12 +43,13 @@ from esp.cal.models              import Event
 from esp.users.models            import User, ESPUser, UserAvailability
 from esp.middleware              import ESPError
 from esp.resources.models        import Resource, ResourceRequest, ResourceType, ResourceAssignment
+from esp.tagdict.models          import Tag
 from datetime                    import timedelta, time
 import json
 from collections                 import defaultdict
 from argcache                    import cache_function
 from uuid                        import uuid4 as get_uuid
-from esp.utils.decorators         import json_response
+from esp.utils.decorators        import json_response
 import calendar, time, datetime
 
 class AJAXSchedulingModule(ProgramModuleObj):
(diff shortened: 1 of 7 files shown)


kkbrum (commented):
A couple thoughts that may or may not be coherent and also may contradict the opinions I gave along the way:
- If a classroom has no resources and I hover over it, it shows a bulletpoint with nothing beside it- maybe delete the list of resources if there are none? (I know this is not super relevant)
- For classes, the timeslot headers change colors, but not for moderators-- do you want to also have headers highlighted when we select a moderator?
- For adding moderators, if they are available the first timeslot of a multi-timeslot class, it colors the whole class green and allows you to assign them, even though they aren't available for the whole class. Should we do something similar to classes where the timeslot they are available gets the slashed lines?
- If you assign a moderator to a class, but then unschedule the class, I feel like the moderator should get removed from the class? Or maybe there should be a popup asking whether to unschedule the moderator as well? Or we could have two buttons, one to unschedule class and one to unschedule both? One issue I have with moderators being assigned to unscheduled classes is that the "available block" count for the moderator doesn't take that into account, so it's a bit odd accounting-wise
- When a moderator is unavailable but all teachers are available, I feel like we should denote that differently than when a teacher is unavailable, since it would let the admin know that if they remove the moderator then it would work. One idea I have is that the blocks where it is ONLY a moderator that is unavailable would be grey with blue slashes through them or something like that.

kkbrum (approved):
Looks good. I've moved the two larger thoughts into new issues in case someone ever wants to address them.

PR #3645, opened 2023-05-05
Add partition name handling to dump dev DB function while upgrading VM

milescalabresi:
New function saves existing DB so that the dev VM can be upgraded without starting over. Additional infrastructure added to allow jumping between multiple VMs with multiple partition names


willgearty:
For the record, the dumpdb function is already on main (see the diff of this PR), so this PR only adds the different partition name handling.

fabfile.py
```diff
@@ -72,6 +69,17 @@
...
+# Name of the encrypted volume group in the Vagrant VM based on which VM is loaded
+ubuntu_version = run("lsb_release -r | awk '{print $2}'")
+try:
+    env.encvg = {"22.04": "ubuntu--vg-vgvagrant--keep_1",
+                 "20.04": "vgvagrant-keep_1",
+                 "12.04": "ubuntu--12--vg-keep_1"}[ubuntu_version]
+except KeyError:
+    raise ValueError("Unrecognized version of Ubuntu: " + ubuntu_version +
+                     "Web Team needs to update fabfile.py to add the "
+                     "partition name for this VM")
```
willgearty, line 81:
```suggestion
    raise ValueError("Unrecognized version of Ubuntu: " + ubuntu_version +
                     ". Web Team needs to update fabfile.py to add the "
                     "partition name for this VM.")
```

docs/dev/vagrant.rst
```diff
@@ -175,20 +175,14 @@ Upgrading your personal dev VM
...
 
-2. From within the "devsite" folder, run ``git checkout dumpdb``. This branch has the proper code to interact with your current database and create the dump file.
+2. From within the "devsite" folder, run ``fab dumpdb``. This action will save your database as a dump file in the `devsite` folder called `devsite_django.sql`. You can also specify a filename if you would like with ``fab dumpdb:filename``.
 
-3. Run ``vagrant up`` to start the virtual machine.
+3. Run ``vagrant destroy`` (note: this destroys your virtual machine. Only do it once you are sure your database has been backed up and you are ready to continue).
 
-4. Run ``fab dumpdb``. This will save your database as a dump file in the "devsite" folder ("devsite_django.sql"). You can also specify a filename if you would like with ``fab dumpdb:filename``.
+4. Now follow the `VM installation instructions above <https://github.com/learning-unlimited/ESP-Website/blob/main/docs/dev/vagrant.rst#installation>`_, starting at ``vagrant up``.
```
willgearty, line 184: We should reference the section of the file without the full path in case the user isn't on the main branch. See https://gist.github.com/TobiasPalludan/5948533e97d260a04ea47e44ea7cc6b0.

docs/dev/vagrant.rst
```diff
@@ -175,20 +175,14 @@ Upgrading your personal dev VM
 
 If the base VM has been changed (see above), you will want to upgrade your development server. However, upgrading Ubuntu within a virtual machine can cause problems with your database. Therefore, you'll need to export your database, create a new virtual machine, then import your database:
 
-1. Make a copy of `esp/esp/local_settings.py` somewhere with a different name (e.g. on your desktop as "old_local_settings.py"). This file will get overriden by the end of this process and you will want to restore some of the settings from your previous VM setup.
+1. Make a copy of `esp/esp/local_settings.py` somewhere with a different name (e.g. on your desktop as "old_local_settings.py"). The `local_settings.py` file will get overwitten by the end of this process and you will want to restore some of the settings from your previous VM setup.
```
willgearty, line 178:
```suggestion
1. Make a copy of `esp/esp/local_settings.py` somewhere with a different name (e.g. on your desktop as "old_local_settings.py"). The original `local_settings.py` file will get overwitten by the end of this process and you will want to restore some of the settings from your previous VM setup.
```

willgearty (changes requested):
Just some minor comments

docs/dev/vagrant.rst
milescalabresi: The link will still work even if you're not on the main branch. Are you worried that the files might differ?

willgearty: Exactly

PR #3647, opened 2023-05-17
Lineitems printable

### LINKED ISSUE #3597: Printable for all line items
Right now there are two ways to get information about line item purchases (I believe):

1. A printable for each line item (but the printables don't even list the choices the students made)
image

2. The arbitrary user list, where you can filter to each choice (but this gives no information about quantity, etc)
image

We should have a single printable that summarizes all of the line items for all students (e.g., quantities, selections, amounts, etc).

kkbrum:
Fixes #3597.

Adds option types to the purchased items printable and also fixes the paid column. If someone purchases >1 of something, they will appear multiple times in the list. I think the purchased items printable is most suited for the receipts section, since it is meant to provide info about the purchases and not so much about the students.

In the students section on the printables options page, I added links to filter by option of line item in addition to just line item. I haven't added anything to those printables about quantity, etc, since those are just a filtered version of the student list. The purchased items printable should have all the useful info about that though.

In the process, I've changed it so that filtering to students who purchased a line item does not automatically restrict to registered students. If folks want, they can combine the list with students who are enrolled in a class.


esp/templates/program/modules/programprintables/paid_list.html
willgearty, line 1: I think this printable only lists optional items at the moment. Can we change it to all items?
https://github.com/learning-unlimited/ESP-Website/blob/173b9af7e299fd09078bfb1879eaab63ae8e724d/esp/esp/program/modules/handlers/programprintables.py#L103-L109

willgearty: Here too:
https://github.com/learning-unlimited/ESP-Website/blob/173b9af7e299fd09078bfb1879eaab63ae8e724d/esp/esp/program/modules/handlers/programprintables.py#L87

esp/templates/program/modules/programprintables/options.html
```diff
@@ -113,7 +114,8 @@ <h3>Student Lists</h3>
 <li>Students who have ordered/reserved line items:
     <ul>
     {% for li_type in li_types %}
```
willgearty, line 116: Looks like this is optional items only as well. Can we update that too?

willgearty (changes requested):
These changes look great! I just had some comments about switching from optional-only to all line item types (outside of admission/finaid/etc).

willgearty (approved):
Looks good, thanks!

Issue #3607, opened 2023-02-04
Show error message if any tag form fields have errors


willgearty:
If there is an error in the tag settings form, the error is only shown alongside the specific field that caused the error. This means that it appears that there is no error at face value to the admin user (since all of the field groups are collapsed). If there is an error, we should an alert of some sort at the top of the page indicating so.

Reported by Yale.

milescalabresi:
I don't fully understand the behavior. Was that tag with the invalid value saved via /admin instead of the tags settings page? If not, how did it get there?

willgearty:
Yes, I have a feeling Yale has old instructions that say to go to the admin pages (despite the warnings on the page now) and set that tag manually.

willgearty:
Another possibility is that the format of the tag value has changed in the last year or two and the (now improperly formatted) tag has just been copied over for each new program that Yale has made.

PR #3648, opened 2023-05-19
Show any/all tag form errors at the top of the page

willgearty:
If there are any errors when submitting the tag settings form, they are now collectively shown at the top of the page (because otherwise admins can't tell that there are any errors since all of the categories are collapsed).

image

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3607.

### DIFF
--- a/esp/templates/program/modules/admincore/tags.html
+++ b/esp/templates/program/modules/admincore/tags.html
@@ -15,6 +15,7 @@
 
 {% block content %}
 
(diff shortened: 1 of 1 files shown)


kkbrum (approved):
I haven't been able to find a way to submit the form with any errors since it prevents you from submitting errors, but the code looks good and it looks like you tested it successfully, so I think it's fine!

Issue #839, opened 2013-12-02
Grade ranges for individual sections


pricem:
Clark has requested that teachers be able to specify different grade ranges for different sections of a class so that they don't have to register the class multiple times.  For example, you could teach one section for high schoolers and one for middle schoolers.

milescalabresi:
Probably makes the most sense to make two subjects. As @willgearty has pointed out, it is quite easy to import the subject you just registered but change the grade (probably just as easy as any section-level mechanism we would implement). However, there is room for improvement: the import text is confusing since it says "teach a class from a previous program." So let's improve that and if the issue is brought up again, we can recommend this easy process as the preferred way to do what the teacher is asking for.

PR #3650, opened 2023-05-19
Fix "import class" text

willgearty:
Just a minor update to the import class text to be clearer that the class can also be from the current program

Fixes https://github.com/learning-unlimited/ESP-Website/issues/839

### DIFF
--- a/esp/templates/program/modules/teacherclassregmodule/listclasses.html
+++ b/esp/templates/program/modules/teacherclassregmodule/listclasses.html
@@ -130,7 +130,7 @@ Popup div for class cancellation request
     </a>
     <br />
     <a class="create-class-link" href="/teach/{{ one }}/{{ two }}/copyclasses" title="Teach your existing class for {{ settings.ORGANIZATION_SHORT_NAME }}!">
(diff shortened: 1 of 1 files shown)


esp/templates/program/modules/teacherclassregmodule/listclasses.html
```diff
@@ -130,7 +130,7 @@ <h3>Cancellation Request for {{ cls|truncatewords:6 }}</h3>
     </a>
     <br />
     <a class="create-class-link" href="/teach/{{ one }}/{{ two }}/copyclasses" title="Teach your existing class for {{ settings.ORGANIZATION_SHORT_NAME }}!">
-      Teach a class from a previous program for {{ program.niceName }}!
+      Copy a previously registered class (including from a previous program) for {{ program.niceName }}!
```
kkbrum, line 133:
```suggestion
      Copy a registered class (including from a previous program) for {{ program.niceName }}!
```

I think it's clearer without the "previously"

kkbrum (changes requested):

kkbrum (approved):

Issue #3658, opened 2023-07-07
Custom form permissions duplicate item


kkbrum:
When making a custom form, in the "permissions" section, "teachers who have filled out a profile" shows up twice

willgearty:
Interestingly, there is no "Students who have filled out a profile" option...perhaps related?

willgearty:
Yep, it's due to the student and teacher profile modules being in the same handler.

PR #3670, opened 2023-08-04
Fix permissions in custom forms

willgearty:
This fixes the lists of permissions in the custom form builder. I also fixed the text next to the checkbox so when you click the text it activates/deactivates the checkbox.

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3658.

### DIFF
--- a/esp/esp/customforms/views.py
+++ b/esp/esp/customforms/views.py
@@ -72,16 +72,7 @@ def getPerms(request):
             except ValueError:
                 return HttpResponse(status=400)
             prog = Program.objects.get(pk=prog_id)
-            perms = {'teachers': [], 'students': []}
-            for module in prog.getModules(None):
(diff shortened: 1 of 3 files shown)


kkbrum (changes requested):
Your changes look good, but do these restrictions actually work? It let me save the form with restricted permissions, but then if I morph into a user that does not meet the restrictions, they can still fill out the form. And, if I go back to edit the form, the permission restrictions are unselected. It either never saved, or does not pre-fill correctly, it seems.

willgearty:
I think that's outside the scope of this PR.

willgearty:
I think the permissions are only checked if you fill the form out outside of a program module (e.g., /customforms/view/[id]). I was able to trigger it and get the correct error message:
image

kkbrum (approved):

PR #3671, opened 2023-08-16
Change moderator titles throughout dashboard stats

### LINKED ISSUE #3667: Moderator label doesn’t get applied on the dashboard stats

kkbrum:
Fixes #3667


willgearty (changes requested):
This is great, thank you @kkbrum! Two comments/requests:

1. Some of these changes are within cached functions. Can we update the cache dependencies to include this tag? I think you should be able to make it dependent only this specific program tag (I think there's an example somewhere).
2. The class popup from clicking "status" on the dashboard appears to be hard coded to say "Moderator". Can we update that to also use the specified title?

willgearty:
Looks good to me, thanks @kkbrum!

willgearty (approved):
Looks good to me, thanks @kkbrum!

PR #3672, opened 2023-08-18
Fix form submission handling

### LINKED ISSUE #3664: Volunteer registration hangs if no times submitted
From @milescalabresi: For volunteer registration, if you don't select any times, there is a dialog box that asks you to confirm you mean to leave them all blank, but if you hit submit, that box hangs

### LINKED ISSUE #3668: 7-12 dialog box appears twice when importing class
For class import from a previous program, if you select 7-12, the dialog box appears twice. Noticed by @milescalabresi

willgearty:
This fixes the way that we handle form submissions (with respect to the previous fix to prevent double button presses). I have only tested this behavior on the class registration form, but it seems to work well there (including with the popup when the grade range is too wide). During review it would probably be good to check a few other forms to make sure they all still work, too.

Fixes #3668. Fixes #3664.

### DIFF
--- a/esp/templates/elements/html
+++ b/esp/templates/elements/html
@@ -79,7 +79,12 @@
     <!-- prevent forms from being submitted twice -->
     <script text="text/javascript">
       $j(function() {
-        $j('[type=submit]:not(.stripe-button-el)').click(function(){
+        $j('[type=submit]:not(.stripe-button-el)').on('click', function(event){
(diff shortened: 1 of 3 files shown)


willgearty:
This might also fix #3664, but I haven't checked.

willgearty:
This now fixes #3664 (which was a different issue).

kkbrum (approved):

Issue #3674, opened 2023-08-23
Add ability to remove/reset optional variables in fruitsalad


kkbrum:
If you add a bunch of optional variables in the fruitsalad theme (under the fruitsalad tabs expandable div on the theme customization page), these cannot be removed without resetting the whole theme. I'd like to be able to remove them, resetting them to their default color (which is a shade or two darker than the tab itself). Unfortunately the javascript daunts me. The relevant code is here:
https://github.com/learning-unlimited/ESP-Website/blob/03fefcda53c90b57c11f79bcf1d42a33a75a19e5/esp/public/media/theme_editor/theme_editor.js#L74-L92

PR #3675, opened 2023-08-23
Add ability to remove/reset optional variables in fruitsalad

Vayras:
This is my attempt to resolve the #3674 issue

### DIFF
--- a/esp/public/media/theme_editor/theme_editor.js
+++ b/esp/public/media/theme_editor/theme_editor.js
@@ -54,40 +54,47 @@ $j(document).ready(function(){
     showColor();
     showBasePalette();
     showCustomPalette();
-    $j('#addToPalette').click(function(){
-        var newColor = $j('<input>');
-        newColor.addClass('palette');
-        newColor.attr({
-            name: 'palette',
-            type: 'text',
-            value: 'black'
-        });
-        $j('#palette_custom_div').append(newColor);
-        $j('#palette_custom_div').append(" ");
-        showCustomPalette();
-    });
-
+    
     $j('[rel=popover]').each(function(){
         $j(this).popover({placement:'left', animation:false});
     });
     
-    //  Show optional variable selector if there are optional variables available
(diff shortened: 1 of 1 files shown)


esp/public/media/theme_editor/theme_editor.js
```diff
@@ -54,40 +54,47 @@ $j(document).ready(function(){
...
-        newColor.addClass('palette');
-        newColor.attr({
-            name: 'palette',
-            type: 'text',
-            value: 'black'
-        });
-        $j('#palette_custom_div').append(newColor);
-        $j('#palette_custom_div').append(" ");
-        showCustomPalette();
-    });
```
willgearty, line 68: Is there a particular reason that you removed this? Removing this breaks the ability to make a custom palette.

esp/public/media/theme_editor/theme_editor.js
```diff
@@ -54,40 +54,47 @@ $j(document).ready(function(){
...
+                name: select_val,
+                style: 'display: none;'
+            });
 
-        //  Insert a new color selector for the desired variable
-        $j(event.target).parent().parent().parent().append($j('<div class="control-group">')
+        // Create the "Reset Color" button
+        var resetButton = $j('<button class="reset-color">Reset Color</button>').click(function() {
+            colorInput.spectrum("set", "#000000"); // Reset color to default
+        });
```
willgearty, line 87: Adding this button is great, but I think we really want a way to remove the optional variables, not to just reset them to black.

willgearty, line 87: By default, buttons are "type=submit", so whenever you click a "Reset Color", it submits the form, discarding any changes that have been made.

willgearty, line 87: Ideally we would also have similar buttons next to the optional variables that exist on page load.

willgearty (changes requested):
Thanks for submitting this pull request @Vayras! I've left some comments, please let me know if you have any questions.

willgearty:
Superseded by #3675.

PR #3678, opened 2023-09-08
Add and fix moderator scheduling checks

### LINKED ISSUE #3665: Moderator scheduling checks
There should be a scheduling check for moderators moderating when not available.

Also, for the current moderator check (which is about categories), it does not use the moderator label.

willgearty:
This adds a moderator availability scheduling check to the ajax scheduler and the scheduling checks page. It also fixes the moderator title for all checks in both of these places.

I also adjusted the coloration of the cells for the teacher/moderator availability checks to be based on the proportion of teachers that are unavailable (1/1 unavailable is the same color as 5/5 unavailable, which are both much darker than 1/5 unavailable).

Note, I also fixed the tab spacing in one of the files (hence why the diff looks like I changed a lot).

Fixes #3665.

### DIFF
--- a/esp/esp/program/modules/handlers/schedulingcheckmodule.py
+++ b/esp/esp/program/modules/handlers/schedulingcheckmodule.py
@@ -149,6 +149,10 @@ class SchedulingCheckRunner:
 
     # Update this to add a scheduling check.
     def all_diagnostics(self):
+        if self.p.hasModule("TeacherModeratorModule"):
+            two_classes_name = 'Teachers/' + self.p.getModeratorTitle().capitalize() + 's handling two classes at once'
+        else:
+            two_classes_name = 'Teachers teaching two classes at once'
         diags = [
             #Block Diagnostics
             ('lunch_blocks_setup', 'Lunch blocks'),
@@ -166,7 +170,7 @@ class SchedulingCheckRunner:
             ('unapproved_scheduled_classes', 'Classes which are scheduled but not approved'),
             #Teacher Diagnostics
             ('teachers_unavailable', "Teachers teaching when they aren't available"),
-            ('teachers_teaching_two_classes_same_time', 'Teachers teaching two classes at once'),
+            ('teachers_teaching_two_classes_same_time', two_classes_name),
             ('teachers_who_like_running', 'Teachers who like running'),
             ('hungry_teachers', 'Hungry teachers'),
             ('inflexible_teachers', 'Teachers with limited flexibility'),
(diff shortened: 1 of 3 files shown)


esp/esp/program/modules/handlers/schedulingcheckmodule.py
```diff
@@ -142,104 +142,106 @@ def _getLunchByDay(self):
...
+            ('capacity_by_category', 'Total capacity in each block by category'),
+            ('classes_by_grade', 'Number of classes in each block by grade'),
+            ('capacity_by_grade', 'Total capacity in each block by grade'),
+            ('admins_teaching_per_timeblock', 'Admins teaching per timeslot'),
+            ('multiple_classes_same_resource_same_time', 'Double-booked resources')
+        ]
+        if self.p.hasModule("TeacherModeratorModule"):
+            diags.extend([
+                ('unavailable_moderators', self.p.getModeratorTitle() + "s helping when they aren't available"),
+                ('mismatched_moderators', self.p.getModeratorTitle() + 's with category mismatches'),
```
kkbrum, line 184: Could you capitalize the first letter for both of these checks?

esp/templates/program/modules/ajaxschedulingmodule/ajax_scheduling.html
```diff
@@ -233,6 +233,8 @@ <h3>Class Filters</h3>
                         <br />
                         <input type="radio" name="scheduling-checks" value="num_both" id="scheduling-checks-num_both"></input><label for="scheduling-checks-num_both"> Number of Teachers and {{ program.getModeratorTitle }}s</label>
                         <br />
+                        <input type="radio" name="scheduling-checks" value="mod_unavailable" id="scheduling-checks-mod_unavailable"></input><label for="scheduling-checks-mod_unavailable"> Unavailable {{ program.getModeratorTitle }}s</label>
+                        <br />
                         <input type="radio" name="scheduling-checks" value="mod_cats" id="scheduling-checks-mod_cats"></input><label for="scheduling-checks-mod_cats"> {{ program.getModeratorTitle }}s with Mismatched Categories</label>
```
kkbrum, line 238: can we capitalize the first letter of the moderator title in these final four checks as well?

kkbrum (changes requested):
I think this looks good so far. There is an edge case of when someone is moderating and teaching at the same time. The checks show that there is a teacher teaching when they aren't available, but the moderator check does not show an issue. I actually think neither of these should show an issue and that it should show the issue under double booked teachers (it doesnt). Perhaps we also want to add a double booked moderator check, or just include them in the double booked teacher check (my preference, i think).

willgearty:
Thanks for the great comments, @kkbrum! I believe I've addressed all of them. Teacher and moderator availability checks now skip the other availabilities, while the double-booked check now covers both.

kkbrum (approved):

PR #3686, opened 2023-10-04
Add ability to enable/disable contact form

### LINKED ISSUE #2303: Allow removing contact form
In order to avoid spam, we should have an option to remove the contact form in favor of emailing the chapter directly.  This can be tag-controlled so chapters can choose whether to use it.

*Original description:*

We're in the 21st century, we should have a captcha in our contact form to prevent spam from non-humans.
And yes, this is coming from experience (somehow we are getting 5 forms submitted every minute by some chinese bot, even after overriding the template).

I'd imagine it would be fairly easy to use an optional tag to implement one (at least through Google, it requires a domain-specific key).

willgearty:
This adds a tag that allows admins to enable the contact form (the form is now disabled by default). If the form is disabled, the URL now redirects to /contact.html (before POSTs are processed). Since this page didn't exist before (most chapters make their own QSD page), I've now added a default page for this URL. If chapters already had QSD set for this page, it should be used instead. While I was at it, I cleaned up my previous implementation of the default FAQ page (#3174). Both of these pages now properly show various theme and admin toolbar bits that were previously missing.

Fixes #2303 (finally).

### DIFF
--- a/esp/templates/contact_qsd.html
+++ b/esp/templates/contact_qsd.html
@@ -6,12 +6,13 @@
 {% block content %}
 
 {% load render_qsd %}
+{% load getTag %}
 
 {% inline_qsd_block "contact" %}
 
 <p>
-For general communication, please email us at <b><a href="mailto:{{ settings.DEFAULT_EMAIL_ADDRESSES.default }}">{{ settings.DEFAULT_EMAIL_ADDRESSES.default }}</a></b>.
 Before contacting us with a question or concern, please check our <b><a href="/faq.html">FAQ page</a></b>.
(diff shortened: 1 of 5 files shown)


esp/templates/contact_qsd.html
```diff
@@ -0,0 +1,18 @@
...
+{% block content_title %}Contact Us{% endblock %}
+
+{% block content %}
+
+{% load render_qsd %}
+
+{% inline_qsd_block "contact" %}
+
+<p>
+For general communication, please email us at <b><a href="mailto:{{ settings.DEFAULT_EMAIL_ADDRESSES.default }}">{{ settings.DEFAULT_EMAIL_ADDRESSES.default }}</a></b>.
```
kkbrum, line 13: if the contact form is enabled, should we have a link to it on this default page??

hwatheod (approved):
Thank you for doing this.

PR #3688, opened 2023-10-06
Add validation for JSON-formatted tags

### LINKED ISSUE #2084: Validate JSON tags on save
If a tag needs to have valid JSON data, we may want to try validating that on save.  There are a bunch that, when misconfigured, can break the site in not-entirely-obvious ways.

willgearty:
This adds validation to all existing tags that require a JSON-formatted string (if they are set through the tag settings interface).

Fixes #2084.

### DIFF
--- a/esp/esp/tagdict/__init__.py
+++ b/esp/esp/tagdict/__init__.py
@@ -2,10 +2,22 @@ from collections import OrderedDict
 from django import forms
 from django.forms import widgets
 from django.core.validators import RegexValidator
+from django.core.exceptions import ValidationError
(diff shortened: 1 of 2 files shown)


kkbrum (commented):
The red alert box at the top of the tag settings pages that says "Please be aware that these settings can be very sensitive and there are currently no warnings or blocks in place to prevent you from providing incorrect values" isn't technically correct now-- should we change "no warnings" to "few warnings"?

hwatheod (approved):
Ideally there would be more specific validation for each tag, but this is good for now.

PR #3691, opened 2023-10-18
Fix cats and flags links on program settings page

### LINKED ISSUE #3689: Links to editing categories and flags are wrong on manage program settings page
Links to editing categories and flags from manage program settings are both out of date.

The relevant code is here: https://github.com/learning-unlimited/ESP-Website/blob/e71c3b4196f5a2ab7f982b33b3495bec09b1c67d/esp/esp/program/models/__init__.py#L263 and https://github.com/learning-unlimited/ESP-Website/blob/e71c3b4196f5a2ab7f982b33b3495bec09b1c67d/esp/esp/program/models/__init__.py#L267.

The correct link is /manage/catsflagsrecs/ for both, i believe.

kkbrum:
Fixes #3689

### DIFF
--- a/esp/esp/program/models/__init__.py
+++ b/esp/esp/program/models/__init__.py
@@ -260,11 +260,11 @@ class Program(models.Model, CustomFormsLinkModel):
                          'the documentation</a> for details.')
     class_categories = models.ManyToManyField('ClassCategories',
                                               blank=True,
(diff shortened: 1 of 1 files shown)


willgearty (changes requested):
I believe the additional /categories or /flagtypes is meant to open the specific collapsed div, so it might be good to add those back in (if they still work).

PR #3692, opened 2023-10-27
Hide empty categories in the catalog

### LINKED ISSUE #3661: Empty categories shown when filtering class catalog
From @willgearty: "When filtering classes, if a category is empty it is still shown (with a big white gap between it and the next category). Is this the desired behavior? Or would we prefer to hide empty categories? I suppose this shouldn't happen to often (I filtered a dev5 catalog to only 1hour 20 min classes and they are all in a separate category from the shorter classes)"

willgearty:
This hides any empty categories in the catalog (more relevant now thanks to the filtering implemented in https://github.com/learning-unlimited/ESP-Website/pull/3632) provided the new tag is set to True (the default). If a category is hidden, the corresponding category link at the top is disabled and faded.

I didn't check it, but I believe this should also work for the `/fillslot` student registration pages.

Fixes #3661.

### DIFF
--- a/esp/public/media/scripts/program/modules/catalog.js
+++ b/esp/public/media/scripts/program/modules/catalog.js
@@ -1,0 +1,169 @@
+// Inject difficulty data from the tag to support custom difficulties
+const DIFFICULTIES = INJECTED_DIFFICULTIES !== "None" 
+    ? JSON.parse(INJECTED_DIFFICULTIES)
+    : [
+        ["*", "This Shouldn't Appear!"],
+        ["**", "This Shouldn't Appear!"],
+        ["***", "This Shouldn't Appear!"],
+        ["****", "This Shouldn't Appear!"]
+    ];
+
+const MODIFIED_COLOR = "#0066ff22";
+
+const FILTER_IDS = ["grade_filter", "difficulty_filter", "status_filter", "duration_filter"];
+
+/**
+    Converts from hours to a formatted duration (eg. "0.05" -> "5 mins", "1.5" -> "1 hour 30 mins")
+    @param {string} numString - a duration as a float in a string (eg. "0.05")
+    @returns {string} the formatted time (eg. "1 hour 30 mins")
+*/
+function floatToFormattedTime(numString) {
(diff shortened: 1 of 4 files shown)


esp/public/media/scripts/program/modules/catalog.js
```diff
@@ -0,0 +1,169 @@
...
+    ];
+
+const MODIFIED_COLOR = "#0066ff22";
+
+const FILTER_IDS = ["grade_filter", "difficulty_filter", "status_filter", "duration_filter"];
+
+/**
+    Converts from hours to a formatted duration (eg. "0.05" -> "5 mins", "1.5" -> "1 hour 30 mins")
+    @param {string} numString - a duration as a float in a string (eg. "0.05")
+    @returns {string} the formatted time (eg. "1 hour 30 mins")
```
AlpacaFur, line 18:
```suggestion
    Converts from hours to a formatted duration (eg. "0.05" -> "3 mins", "1.5" -> "1 hour 30 mins")
    @param {string} numString - a duration as a float in a string (eg. "0.05")
    @returns {string} the formatted time (eg. "1 hour 30 mins")
```

Just realized I had made a small typo here before :)

esp/templates/program/modules/studentclassregmodule/catalog.html
```diff
@@ -108,166 +116,6 @@ <h2 style="text-align: center;">
...
-function configure_addbuttons()
-{
-    {% if register_from_catalog %}
-    //  Registration from the catalog is allowed
-    $j("input.addbutton").show();
-    {% endif %}
-}
-$j(document).ready(configure_addbuttons);
-</script>
+<script type="text/javascript" src="/media/scripts/program/modules/catalog.js"></script>
```
AlpacaFur, line 119: Good call, I should have done this in my PR because this is prob too much JS to inline.

AlpacaFur (approved):
Looks good to me! Didn't have enough time to run locally but all the changes that touched the code I worked on make sense 👍

kkbrum (approved):

PR #3696, opened 2023-11-08
Reset and remove optional fruitsalad variables

### LINKED ISSUE #3674: Add ability to remove/reset optional variables in fruitsalad

willgearty:
This adds the following functionality to the theme editor:

- Added a "Reset Color" button next to each color chooser

And when optional variables are included (I believe this is just for the fruitsalad theme atm):

- Added a "Remove Variable" button next to the new "Reset Color" buttons which removes the entire variable from the form
- Made it so you can't add a variable that already exists in the form

I've checked and the addition and removal of variables is saved whenever you press the "Test" submission button.

image

Fixes #3674. Supersedes #3675.

### DIFF
--- a/esp/esp/themes/controllers.py
+++ b/esp/esp/themes/controllers.py
@@ -227,39 +227,11 @@ class ThemeController(object):
 
         return results
 
-    def compile_css(self, theme_name, variable_data, output_filename):
+    def compile_less(self, less_data):
         #   Hack to make things work on Windows systems
         INCLUDE_PATH_SEP = ':'
         if os.name == 'nt':
             INCLUDE_PATH_SEP = ';'
-
-        #   Load LESS files in order of search path
-        less_data = ''
-        for filename in self.get_less_names(theme_name):
-            less_file = open(filename)
-            logger.debug('Including LESS source %s', filename)
-            less_data += '\n' + less_file.read()
-            less_file.close()
-
-        if themes_settings.THEME_DEBUG:
-            # TODO(benkraft): should this and its friend below get removed now?
-            # I think they're the last users of THEME_DEBUG.
-            tf1 = open('debug_1.less', 'w')
-            tf1.write(less_data)
-            tf1.close()
-
-        #   Make icon image path load from the CDN by default
-        if 'iconSpritePath' not in variable_data:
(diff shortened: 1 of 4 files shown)


kkbrum (changes requested):
This is great! When I add optional accent variables, could the first color that shows up be the default color? I get black for all of them instead of the default. Also, if I reset the button or tab colors, I also get black. I'd rather it reset to the default, which I don't believe is black.

willgearty:
@kkbrum I'm not quite sure what you mean by the "default" color. Do you mean the first color in the palette (in which case the default for all of them would be the same)? Or do you mean the color that the variable is normally set to (in which case the default may be different for each variable)?

kkbrum:
I think it should start as the color that the variable is automatically set to if it is not included as an optional variable. So it's different for each one. Then you can see the current color when you add it as a variable so you know you're working on the right thing and have a good baseline for the color. And then can reset it to that as well.

willgearty:
OK, I've made it so that when you add a new variable, the initial color is the default color. And, whenever you reset one of these variables (plus the other variables, including the rounding_radius), it resets to whatever the default color/value is.

kkbrum (commented):
Thanks! I like this a lot. I have two comments below, which you can address if you want to or if it seems easy, but I'm okay with approving the PR as it is.
1. It would be cool if when you have already added an optional variable, it no longer showed up in the dropdown list of variables that can be added.
2. The reset to default for the accent colors currently gives the right defaults if the fruitsalad tab colors have not been changed from their defaults. I think it would be slightly better if the reset to default for the accent gives the default "%" increase applied to the currently configured tab colors, which may be different than the default tab colors.

willgearty:
OK, the recent commit does #1 (great suggestion), but I don't think I'll implement #2 because I want the "Reset Color" buttons to all do the same thing (reset to the theme default).

kkbrum (approved):

PR #3551, opened 2022-08-12
Add contact info to templates

kkbrum:
Fixes #1121

PR #3697, opened 2023-11-17
Make contact info more editable/hidable

### LINKED ISSUE #2156: Made the list of links in the fruitsalad top bar editable
Some chapters don't want to have an FAQ.  We could just handle this like the FB link.

### LINKED ISSUE #3611: Make contact info hideable
#3551 added the chapter's contact info to all themes. However, while this is customizable on fruit salad (I think?), it is not customizable on other themes (e.g., bigpicture). It would be great if chapters could specify which (if any) of the contact info is shown in these new areas.

Requested by Stanford.

willgearty:
This takes the contact info sections that were added in #3551 and adds a bunch more flexibility to make them fully customizable. This includes being able to hide everything (except for floaty, but who even uses that theme), adding and removing custom contact info links, and adding and removing the FAQ and facebook links (to eventually move to just the custom links). I skipped the barebones theme because no one uses it AND because it's supposed to be _barebones_.

While I was at it, I fixed the styling of /themes/setup and fixed the caching of the logo images on /themes/.

Fixes #3611. Fixes #2156.


willgearty:
Thanks @kkbrum. Those two fields are no longer required and adding a contact link will now add the default link.

kkbrum (approved):

Issue #3625, opened 2023-03-08
Theme editor help text


kkbrum:
Some of the help texts on the theme editor page don't really make sense given the selected theme. There should be more helpful text on the editor page that is relevant to the current theme so that admins can more easily anticipate the effect of each of the customizable variables.

willgearty:
bibpicture and fruitsalad were done in https://github.com/learning-unlimited/ESP-Website/pull/3705

PR #3705, opened 2024-01-05
Theme help

kkbrum:
This fixes some things (some of the custom variables weren't being used fully in all the themes) and adds help text that is specific to the themes. This completes #3625 for big picture and fruitsalad and starts it for circles.

### DIFF
--- a/esp/esp/themes/theme_data/bigpicture/less/variables.less
+++ b/esp/esp/themes/theme_data/bigpicture/less/variables.less
@@ -1,2 +1,2 @@
-@identityBlue: #1C1953;
-@highlightOrange: #FF8000;
+@accent1: #FF8000;
+@accent2: #1C1953;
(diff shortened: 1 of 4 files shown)


esp/esp/themes/theme_data/bigpicture/less/variables.less
```diff
@@ -1,3 +1,2 @@
 @identityBlue: #1C1953;
-@highlightOrange: #FF8000;
-@pale: #f8f8f8;
\ No newline at end of file
+@highlightOrange: #FF8000;
```
willgearty, line 2: Can we rename these to something a little more descriptive? (then we won't need as much help text below)

willgearty (changes requested):
Looks great, just one suggestion!

esp/esp/themes/theme_data/bigpicture/less/variables.less
willgearty: In hindsight, we might not be able to reduce the help text too much, but I think changing the names is still useful.

willgearty (approved):

PR #3707, opened 2024-01-08
Adding "how did you hear about us" to user reg flow

matho-o:
Added new question to user reg flow. New model ESPUserData created to store response.


matho-o (commented):
tested on local

PR #3719, opened 2024-02-14
Improve theme customizability, increase relevancy of theme editor to current theme

### LINKED ISSUE #3625: Theme editor help text

kkbrum:
Fixes #3625. Removes some unused variables, renames some for clarity, makes button types have consistently editable colors across themes, makes theme editor more theme-specific by removing irrelevant variables and adding some text. A few other little things.


willgearty (changes requested):
This all looks and works great, @kkbrum, thanks! Just one comment that needs addressing.

PR #3720, opened 2024-02-16
Sign up text color and padding in fruitsalad

### LINKED ISSUE #3716: "Sign up" text in the top right of fruitsalad is no longer visible when logged out
I think I accidentally made it use the same accent color as the background of that button...

kkbrum:
Fixes #3716


willgearty (approved):
Works great, thanks!

PR #3721, opened 2024-02-16
Unsubscribe links and domain-only "From" email addresses

### LINKED ISSUE #1094: Require that "From" be a valid email in the comm panel
Some improvements can be made to the CommPanel form, so that mistakes aren't made:
- Modify the help text for the "From" and "Reply-To" fields, so that it is clear that they must be valid e-mail addresses.
- Add a validator to check that the "From" and "Reply-To" fields, if not blank, must be e-mail addresses.

### LINKED ISSUE #1719: Automatic unsubscribe link on comm panel emails
It would be good to have an unsubscribe/deactivate link at the bottom of all comm panel emails; while we usually put it in manually, we sometimes forget.  Ideally, rather than just linking to /myesp/disableaccount, we would in fact have per-user generated links that would not require logging in, since people have trouble remembering their username and password and such.  This wouldn't work for mailman lists, of course, but it would at least make it a bit easier for people to unsubscribe (and therefore make us less likely to get marked as spam).

### LINKED ISSUE #3709: Emails in "from" field must match "learningu.org"
Per [new email rules](https://sendgrid.com/en-us/blog/gmail-yahoo-sender-requirements#5-dont-impersonate-gmail-from-headers), Gmail (and likely Yahoo) will not deliver email that don't match a domain we register (more details [here](https://support.google.com/a/answer/10032169?sjid=6612418846590111173-NA#alignment)). It's fine to use any address in the "reply-to" field, though. This issue is to add a validation check to the comm panel form to check if the "from" field ends in .learningu.org, and if not, throw an error to the admin telling them to use an LU address and maybe suggesting they use the reply-to field.

Note: this request is time-sensitive as Gmail will start enforcing these rules soon:
> If a sender does not meet the requirements by February 2024, they will start to see temporary errors occurring on a small percentage of their non-compliant mail to Google recipients. In April 2024, a small percentage of the mail will be rejected and that percentage will gradually increase over time.

willgearty:
This does the following:

- Can now generate unsubscribe links for users (including adding these to commpanel emails)
- Upon clicking these links, users can unsubscribe without logging in (within 7 days of being generated)
- Email clients can also use a "oneclick" version to instantly unsubscribe a user with a single POST
- The "From" email address for comm panel emails must now contain the LU domain or any LU subdomain (or a chapter-specific alias); there is a nice big red banner to announce this and there are frontend and backend checks to enforce this
- To help chapters, we now use the info@subdomain.learningu.org email redirect (which is by default set to their "default" email address); this is now automatically populated in the "From" field
- The "director_email" field for a program now must match our new "From" constraints. This is enforced at the database level and also in the new program and program settings forms.
- I've set up a migration that will replace any existing "director_email" fields that don't match with info@subdomain.learningu.org. This migration also sets up the info@subdomain.learningu.org email redirect.
- `send_mail()` also requires that the "from_email" matches these constraints and throws an error if this is not the case (just in case it somehow got through the above restrictions).
- We now include a plaintext version of HTML emails to support all email clients (even plaintext-only ones)
- Various minor fixes to emails that were being sent from domains/subdomains not associated with our DMARC records (like spoofing users when we sent the director an interview notification).

Fixes #1719 and fixes #3709 and fixes #1094.


willgearty:
This is finally ready to be reviewed @kkbrum @milescalabresi!

esp/templates/users/unsubscribe.html
```diff
@@ -0,0 +1,32 @@
...
+{% if deactivated %}
+<p>
+Your account, <b>{{ user.username }}</b>, has been properly deactivated and you have been unsubscribed from emails from {{ settings.INSTITUTION_NAME }} {{ settings.ORGANIZATION_SHORT_NAME }}.
+</p>
+
+<p>
+Thank you for your interest in {{ settings.ORGANIZATION_SHORT_NAME }}!  If you have any questions, please <a href="/contact/">contact us</a>.
+</p>
+{% else %}
+<p>We're sorry to see you go. If you are sure you would like to deactivate your account and unsubscribe from emails from {{ settings.INSTITUTION_NAME }} {{ settings.ORGANIZATION_SHORT_NAME }}, please click the confirmation button below:</p>
```
kkbrum, line 23: For the purpose where you are logged into the wrong account but the token is valid, I think we should clarify the "your account" bit here. I think putting the username is parentheses after the "your account" would be sufficient. It clarifies it on the success page, but that's after the fact...

kkbrum (changes requested):
The migrations and such look good. The template tag unsubscribe link works well at least when the token is valid. For the other forms, I'll have to test a bit more on dev5 when we test the stable release. For now, just one minor request clarifying the username.

kkbrum (approved):

Issue #3601, opened 2023-01-15
Back to top button on course catalog


willgearty:
It would great to have a "go to top" button on the normal catalog (like the onsite catalog):
https://github.com/learning-unlimited/ESP-Website/blob/da33e29481f7a9bac144988fe132cbf1213231bb/esp/templates/program/modules/studentonsite/catalog.html#L11-L26

Requested by Northeastern at SplashCon.

willgearty:
Fixed by #3722

PR #3722, opened 2024-02-21
added scroll to top button in catalog

alexh555:
resolves issue #3601

### DIFF
--- a/esp/templates/program/modules/studentclassregmodule/catalog.html
+++ b/esp/templates/program/modules/studentclassregmodule/catalog.html
@@ -7,7 +7,9 @@
 {% block stylesheets %}
 {{ block.super }}
 <link rel="stylesheet" type="text/css" href="/media/styles/catalog.css" />
-
+<link href="https://fonts.googleapis.com/icon?family=Material+Icons"
+      rel="stylesheet">
+      
 <style>
 a.disabled {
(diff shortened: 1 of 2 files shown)


esp/templates/program/modules/studentclassregmodule/catalog.html
```diff
@@ -29,6 +31,26 @@
 const register_from_catalog = {{ register_from_catalog|yesno:"true,false" }};
 const hide_empty_cats = {{ "hide_empty_categories"|getBooleanTag|yesno:"true,false" }};
 </script>
+
+<script>
```
willgearty, line 35: It might be better for future us if you move this javascript into /media/scripts/program/modules/catalog.js. Fun fact: if you have a separate javascript file, when you make changes to it you only need to hard refresh your page to see the effects (no shutting down and restarting your server)!

kkbrum (changes requested):
I think it would be good to address Will's comment, but the functionality works great!

kkbrum (dismissed):

kkbrum (changes requested):
The change you made works, but that code was actually cleaner the way you had it. It's the script in lines 34-51 of the html file that it would be best to move to the javascript file!

kkbrum (approved):
🥳

Issue #3713, opened 2024-01-17
Extra costs double text when no extra costs


willgearty:
When there are no extra costs to select, it still says that there is a "total cost of selected items", then lists program admission separately (even though it's already included).
![image](https://github.com/learning-unlimited/ESP-Website/assets/7232514/e96f1dbe-b780-400b-a081-6da108cfa2a6)

Solution would be to just not show "total cost of selected items" line when there are no extra costs to select.

PR #3734, opened 2024-03-20
Fix extracosts form when there are no extra costs

willgearty:
Fixes https://github.com/learning-unlimited/ESP-Website/issues/3713.


kkbrum (changes requested):
This looks good when there are no items. However, I feel like the issue still exists when there are items. 

image

I feel like the "Total cost of selected items" should exclude the program admission cost

willgearty:
OK, that should fix the totals when there isn't a sibling discount.

kkbrum (approved):

Issue #3728, opened 2024-03-09
Redirect tables cut off when text is too long


willgearty:
![image](https://github.com/learning-unlimited/ESP-Website/assets/7232514/841de016-18f0-43f6-b5e5-98d0bfcbe64f)

willgearty:
I think the tables just need "table-layout: fixed;"

willgearty:
Another example: https://princeton.learningu.org/manage/redirects/

MysteryManav:
@willgearty i would like to fix this issue, can you specify where the table that needs to be modified is ?

willgearty:
@MysteryManav the tables are in this file: https://github.com/learning-unlimited/ESP-Website/blob/main/esp%2Ftemplates%2Fprogram%2Fredirects.html.

PR #3735, opened 2024-03-20
fixes: Redirect tables cut off when text is too long

MysteryManav:
fixes: #3728

### DIFF
--- a/esp/public/media/default_styles/forms.css
+++ b/esp/public/media/default_styles/forms.css
@@ -176,6 +176,8 @@ input.fancybutton:active
    border: 1px solid #CCC;
    border-width: 1px;
    border-collapse: collapse;
+   table-layout: fixed;
+   overflow-wrap: break-word;
 }
 
 #program_form td {


willgearty:
This is great, thanks @MysteryManav! Unfortunately, it looks like this has revealed a different bug with the email redirects table:
image

I think just changing the `colspan` to 4 here should do the trick:
https://github.com/learning-unlimited/ESP-Website/blob/b40480728e2eae0d9c42876a2a975df8973e6479/esp/templates/program/redirects.html#L136

willgearty (changes requested):
This is great, thanks @MysteryManav! Unfortunately, it looks like this has revealed a different bug with the email redirects table:
image

I think just changing the `colspan` to 4 here should do the trick:
https://github.com/learning-unlimited/ESP-Website/blob/b40480728e2eae0d9c42876a2a975df8973e6479/esp/templates/program/redirects.html#L136

willgearty:
@MysteryManav do you have any plans to address my review?

PR #3738, opened 2024-03-22
Modernize color_code CRMI field

### LINKED ISSUE #3711: Color code needs help text in teacher registration settings
From @kkbrum:
For the teacher registration settings, this needs help text given that I don’t know what it is?
image

willgearty:
This does the following:

- Adds help text to the color_code class registration module info field
- Makes the widget for this field into a color picker (like the theme editor) and adds regex validation to the field to ensure values are hex color codes
- Adds a migration to clean up any old values for this field (and update them to the new format)

Fixes #3711.


kkbrum (approved):
Spectacular!

PR #3743, opened 2024-03-29
Add ability to reorder nav structure and contact links

### LINKED ISSUE #3740: Reorder nav bar without deleting
Right now you need to delete nav bar items (in the theme editor) in order to reorder them, but I could envision dragging and dropping the items to reorder them like you can do in the custom form builder (and maybe in other places, I can't remember at the moment). I imagine this functionality would also be useful for the new contact info links as well.

Requested at SplashCon 2021.

willgearty:
This adds the ability to reorder the nav structure (moving tabs and moving links within/between tabs) and the contact links on the theme configuration page.

While I was at it, I cleaned up the interface to make it less cluttered and more user friendly:

image

Fixes #3740.

### DIFF
--- a/esp/esp/utils/widgets.py
+++ b/esp/esp/utils/widgets.py
@@ -197,7 +197,7 @@ function {{ name }}_delete_link(event)
 
 function {{ name }}_add_link(obj, data)
 {
-    obj.append($j("<li />"));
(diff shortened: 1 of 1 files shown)


kkbrum (approved):
This is lovely!!

PR #3752, opened 2024-04-24
Allow named mailboxes in comm panel

### LINKED ISSUE #3750: Comm panel doesn't allow name for sender email address
Ideally we would allow the use of mailboxes ([specifying a name and email address](https://superuser.com/questions/1625686/what-is-the-email-address-format-name-email-called)) for the comm panel to alleviate the following situation:
image

Reported by Stanford Splash.

willgearty:
This allows admins to use named "mailboxes" (e.g., "Will's Server \<info@test.learningu.org\>") in addition to normal email addresses ("info@test.learningu.org") in the comm panel. I've changed the default "From" email address to make this clearer (in addition to adding text to the Big Red Box TM).

Fixes #3750.

### DIFF
--- a/esp/esp/dbmail/models.py
+++ b/esp/esp/dbmail/models.py
@@ -68,7 +68,7 @@ def send_mail(subject, message, from_email, recipient_list, fail_silently=False,
     from_email = from_email.strip()
     # the from_email must match one of our DMARC domains/subdomains
     # or the email may be rejected by email clients
-    if not re.match(r"(^.+@%s>?$)|(^.+@(\w+\.)?learningu\.org>?$)" % settings.SITE_INFO[1].replace(".", "\."), from_email):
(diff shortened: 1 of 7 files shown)


hwatheod (approved):
If I read this correctly, the user name is still required to be `info`, right?

willgearty:
Thanks for taking a look @hwatheod! For some standard emails (e.g., grade change confirmation email) we default to info@domain and this is not customizable. For other emails we use the director email address which is now info@domain by default, but can be customized using the program settings. For comm panel emails, we also default to a info@domain mailbox, but this can be changed to any @domain email address or mailbox.

kkbrum (commented):
In the name of the pull request, you say you can use a mailbox as director email address as well, but I'm not seeing that in reality? (And the help text for the field doesn't specify that.) Is that something you wanted? 

image

willgearty:
@kkbrum thanks for spotting that. I've reverted the changes to the director_email field. As far as I can tell, a name is always added whenever we use that field, so we can just have it be a normal email address (no mailboxes).

kkbrum (approved):

PR #3706, opened 2024-01-05
Make registration receipt prettier

willgearty:
This makes the student registration receipt prettier (like the PDF student schedules). The receipt can be seen by clicking the "confirm registration" button. The "pretext" can be edited by changing the confirm receipt on the settings page (and the entire receipt can be overriden by a single template override or by program-specific template overrides, like before).

- [x] Integrate with themed pages
- [x] Make only part of the template configurable on the receipts page (use a block or insert context?)
- [x] A button to use javascript to print only the receipt like in https://github.com/learning-unlimited/ESP-Website/pull/3469
- [x] A button to go back to the registration page
- [x] Student schedule should be formatted like in https://github.com/learning-unlimited/ESP-Website/pull/3376
- [x] complete dates for the program (although this is already in the schedule)
- [x] the date that (class) registration closes? ("You have until X/Y/Z to change your classes")
- [x] items purchased and/or amount owed?

Fixes #1131.

PR #3755, opened 2024-04-26
Prettify the confirmation email

### LINKED ISSUE #3753: Make registration email receipt better
Now that the confirmation page has been overhauled in #3706, the logical next step is to make many of the same changes to the email receipt.

willgearty:
This makes the student registration confirmation email prettier, copying most of the changes from #3706. The receipt setting now modifies the pretext of the email (which is now an HTML email). The rest of the email includes various bits of information, including the student's schedule. Because the email now includes the schedule, I've changed the behavior of the confirmation button to now always send a new copy of confirmation email.

Fixes #3753.

### DIFF
--- a/esp/esp/program/controllers/confirmation.py
+++ b/esp/esp/program/controllers/confirmation.py
@@ -36,13 +36,14 @@ Learning Unlimited, Inc.
 from esp.program.models import Program, ClassSection, ClassSubject
 from esp.users.models import ESPUser, Record, RecordType
 from esp.program.modules.module_ext import DBReceipt
+from esp.program.modules.forms.admincore import get_template_source
 
 from django.template import Template, Context
 from django.template.loader import select_template
 from esp.dbmail.models import send_mail
 
 class ConfirmationEmailController(object):
-    def send_confirmation_email(self, user, program, repeat=False, override=False):
+    def send_confirmation_email(self, user, program, repeat=False, override=False, context = {}):
         options = program.studentclassregmoduleinfo
         ## Get or create a userbit indicating whether or not email's been sent.
         try:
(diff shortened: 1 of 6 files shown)


kkbrum (changes requested):
This is very minor, but I don't love that there are three different date formats being used (this is more relevant to the original PR but oh well):

image

More importantly, there is a program tag which controls whether confirmation emails are actually sent:

image

I would mention this setting in the help text for the confirmation email registration receipt field since it's confusing if you set a confemail but forget that you have disabled them (it is actually off by default. perhaps it should be true by default). To be consistent with all the changes and terminology on the page, I would also fix the help text for this setting to not mention needing to set a DB receipt and instead mention the option to change the default confemail in the registration receipts section

willgearty:
I think that covers all of it @kkbrum! Thanks for the helpful comments!

esp/templates/program/receipts/default.html
```diff
@@ -58,9 +58,9 @@ <h1>Registration Receipt for {{ user.first_name }} {{ user.last_name }}</h1>
 Program Information: <br />
 <ul>
 <li>Name: {{ program.niceName }}</li>
-<li>Date Range: {{ program.date_range }}</li>
+<li>Date{% if program.datetime_range.1 %} Range{% endif %}: {{ program.datetime_range.0|date:"D M j Y" }}{% if program.datetime_range.1 %} - {{ program.datetime_range.1|date:"D M j Y" }}{% endif %}</li>
```
kkbrum, line 61: I'm still seeing a range even though I have a one-day program. Is it possible this is checking for multiple times instead of just for multiple days? 

image

kkbrum (changes requested):

kkbrum (approved):
Looks great, thanks :)

PR #3759, opened 2024-06-05
Clean up TextOfEmail message texts

### LINKED ISSUE #3539: Improve Email Storage in Database
Storing each individual's email text is very space-inefficient that balloons our DBs. For example, as of creating this ticket, email text occupies approximately 90% of Stanford's database and 95% of Yale's. It would be good to refactor our database setup to store the often-repeated text of emails in a more compressed format and link to the individual information that populates the tags within individual emails in a separate column or table.

Suggestion:
* one table for each batch of emails that gets sent; that table contains the template for the email, the recipients, and the other current metadata (timestamps, etc.)
* separate table whose columns are a foreign key to the previous table, each recipient's userid, and tag values for each of the tags available to admins. The fields would be the values of the tags for the recipient at the time the email is sent.

If we want to recreate the textofemail view for admins, we can do so from these tables, or we can simply show the metadata list in a searchable format, which should answer most questions admins would usually have.

Keywords: e-mail e-mails db databases schema

willgearty:
This cleans up (i.e., deletes) all of the message texts for existing TextOfEmail objects. For large databases (e.g., Yale and Stanford), this should be a sizeable DB size reduction. This is done as a migration which also attaches MessageRequest objects to TextOfEmail objects. ~~This allows us to then repopulate the message texts in the future (via the admin interface). This (and an option to re-clear the message text field) are now included as dropdown options for the TextOfEmail admin page.~~

Fixes #3539.


milescalabresi (approved):
Code looks good. I tested on a dev server and tested the following:
- [x] the `TextOfEmail` message content is gone
- [x] the `MessageRequest` object is linked

PR #3768, opened 2024-08-28
Remove links to dvi and ps catalogs

### LINKED ISSUE #3766: remove links to dvi and ps catalog printables
dvi and ps formats both yield errors for the catalog printable. I think these links should just be removed.

image

kkbrum:
Fixes #3766


hwatheod (approved):
Looks good! PDF should be supported everywhere now. Is there some place we also need to get rid of code that produces the dvi and ps files ?

willgearty:
I don't think that code exists anymore, so only these frontend links need to be removed.

PR #3771, opened 2024-09-18
Student schedule fixes 

milescalabresi:
Isolating this commit to figure out the database error


esp/esp/program/modules/handlers/programprintables.py
```diff
@@ -1326,8 +1326,7 @@ def get_student_schedules(request, students, prog, extra='', onsite=False):
             return render_to_response(basedir+'studentschedule.html', request, context)
         elif file_type == 'pdf':
             if len(students) > 1:
-                response = HttpResponse(content=render_to_latex(basedir+'studentschedule.tex', context, 'pdf'),
-                                        content_type='application/pdf')
+                response = render_to_latex(basedir+'studentschedule.tex', context, 'pdf')
```
willgearty, line 1329: Oops, looks like I was wrapping what was already an HttpResponse (the results of [render_to_latex](https://github.com/learning-unlimited/ESP-Website/blob/5c9f4b7cd5c2a9cde528f1edd13bcbc2cf3af078/esp/esp/utils/latex.py#L92)) in HttpResponse. Weirdly, this still worked on a dev server, but I guess this caused some DB errors during the tests?

willgearty: For the record, the original code was doing something similar...

milescalabresi: Yes, it seems like it!

PR #3772, opened 2024-09-20
Clean up and centralize style code for roster-like printables

### LINKED ISSUE #3763: Attendance sheet printable has misaligned time and room text
The time and room are slightly out of line (see the second roster to see it clearly). Also, the formatting of the time and room isn't pretty anyways (see the first one)

image

kkbrum:
Fixes #3763.

Additionally, the same code occurred in many printables, so I pulled the css out into a common style file. I also found several printables that are never used as far as I can tell:
- userlist.html
- list_html.html
- studentreceipts.html
- studentreceipt.html (only used in studentreceipts.html, which is never used)
- adminclasstime.html (only used in the adminbinder function, which is never used, so I deleted that function from programprintables.py handler as well)


willgearty:
Looks like the `list_html.html` template is actually used here:
https://github.com/learning-unlimited/ESP-Website/blob/fdf0a0b653f9e51d439101973fbf68859a899a2b/esp/esp/program/modules/handlers/listgenmodule.py#L366-L371

willgearty (changes requested):
Looks like the `list_html.html` template is actually sneakily used here:
https://github.com/learning-unlimited/ESP-Website/blob/fdf0a0b653f9e51d439101973fbf68859a899a2b/esp/esp/program/modules/handlers/listgenmodule.py#L366-L371

Looks like `studentreceipts.html` (and `studentreceipt.html`) was for a receipt for some long dead SAT Prep thing we/ESP used to do, so that's fine to remove.

Looks like `userlist.html` was superseded by [`userchecklist.html`](https://github.com/learning-unlimited/ESP-Website/blob/main/esp/templates/users/userchecklist.html) and never deleted, so that's fine to remove.

And I can't find any record of anything ever linking to the `/adminbinder` view, so it may have been one of those URL hacking things that ESP used to do (a la secret scripts). But I'm totally fine with it being removed, since no one uses it now, most of the templates for it are already gone, and it appears that all of the functionality is covered by other printables.

willgearty (approved):
Thanks, looks great!

PR #3779, opened 2024-10-09
Fixes for new program and program settings forms

### LINKED ISSUE #3775: Financial account names are not updated when program name is changed
When a new program is created, a financial account is set up with it's name based on the name of the program:
https://github.com/learning-unlimited/ESP-Website/blob/5c9f4b7cd5c2a9cde528f1edd13bcbc2cf3af078/esp/esp/program/setup.py#L91
https://github.com/learning-unlimited/ESP-Website/blob/5c9f4b7cd5c2a9cde528f1edd13bcbc2cf3af078/esp/esp/accounting/controllers.py#L87-L92

However, when we edit a program, we don't change the name of the account.
https://github.com/learning-unlimited/ESP-Website/blob/5c9f4b7cd5c2a9cde528f1edd13bcbc2cf3af078/esp/esp/program/modules/forms/admincore.py#L80-L88

So, let's say you accidentally create a program with the wrong name (let's say, hypothetically, you use next year for the name). You then change the name of the program to the correct name, but the account name remains. If you then go to make a program next year, it will then try to use that same name for the account, but we require account names to be unique, so it throws a database error:
https://github.com/learning-unlimited/ESP-Website/blob/5c9f4b7cd5c2a9cde528f1edd13bcbc2cf3af078/esp/esp/accounting/models.py#L251-L252

As far as I can tell, the name of these accounts don't matter (they just need to be unique). When we pull this default account, we just look for the account associated with the program that has the smallest ID:
https://github.com/learning-unlimited/ESP-Website/blob/5c9f4b7cd5c2a9cde528f1edd13bcbc2cf3af078/esp/esp/accounting/controllers.py#L158-L162

So, when we change the name of a program, we should update the name of the account accordingly.

Reported by Stanford.

### LINKED ISSUE #3776: New program and program settings forms allow for duplicate program names
Despite what this error message says, we only ever check if the URL of the program is the same as any other program, not the name:
https://github.com/learning-unlimited/ESP-Website/blob/5c9f4b7cd5c2a9cde528f1edd13bcbc2cf3af078/esp/esp/program/views.py#L618-L623

We don't check either of these things when we are editing a program with the program settings form:
https://github.com/learning-unlimited/ESP-Website/blob/5c9f4b7cd5c2a9cde528f1edd13bcbc2cf3af078/esp/esp/program/modules/handlers/admincore.py#L148-L153

Note, at the database level we require unique URLs for programs, so even without the check Django throws an error message.

willgearty:
- Prevent program with duplicate name/URL (Fixes #3776)
- Update account name when program name is changed (Fixes #3775)
- Show form errors in form instead of as a Whoops page

### DIFF
--- a/esp/esp/program/forms.py
+++ b/esp/esp/program/forms.py
@@ -41,6 +41,7 @@ import re
 import unicodedata
 
 from django.conf import settings
+from esp.middleware import ESPError
 from esp.users.models import StudentInfo, K12School, RecordType
 from esp.program.models import Program, ProgramModule, ClassFlag, ClassFlagType, ClassCategories
 from esp.dbmail.models import PlainRedirect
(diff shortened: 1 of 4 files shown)


willgearty:
This revealed another issue where if you submit the new program form and get one of these new errors, then go backwards (with the back button or the "go back" link on the error page, the submit button is then disabled.

willgearty:
OK, instead of throwing a Whoops error page, we should be validating the fields individually and then re-rendering the form with the validation errors.

willgearty:
@kkbrum could you please test this out before I merge it into the python3 branch?

kkbrum (approved):

PR #3793, opened 2024-12-18
Commpanel fixes

### LINKED ISSUE #3757: Fix bug in selecting teachers of specific student
Email (or usersearch more generally) to teachers of a specific student does not work due to changes in #3445. The traceback is as follows.
```
esp/esp/users/controllers/usersearch.py", line 234, in query_from_criteria
    if criteria.get('target_user', '').strip():
AttributeError: 'ESPUser' object has no attribute 'strip'
```

Fix is to identify what the previous code (`criteria['target_user']`) was checking for and restore that behavior, ideally while maintaining the performance of the new `criteria.get()` call.

Reported by Stanford

willgearty:
Fixes #3757. Also fixes a bug I introduced in e7347fb40cbd17d54e6dff3202e10f38bedeadf7 where the long list of email list options wouldn't show up if you clicked "Next" as soon as loading the page.


milescalabresi (approved):
able to email teachers of a student without an error

PR #3794, opened 2024-12-18
Fixes the bottom of the catalog

### LINKED ISSUE #3787: improper HTML in catalog template
Looks like there might be some improper HTML in the catalog template. The very bottom of the page has some weirdness:

dev4 (https://dev4.learningu.org/learn/Splash/2024_Winter/catalog):
image

Yale (https://yale.learningu.org/learn/Splash/2024_Fall/catalog):
image

Looks like, at least for dev4, the bottom color doesn't go around the bottom, and for both dev4 and Yale the copyright/privacy policy is styled incorrectly. Here's what it should look like:

Yale (https://yale.learningu.org/):
image

willgearty:
Fixes #3787 on fruitsalad, but I haven't tested it on other themes yet.

For the record, I understand why this fixes the problem in #3787 (by removing a stray `</div>`), but I don't understand why it was a stray `</div>` to begin with. As far as I can tell, this code should be there to close the `<div>` that is created above it.

### DIFF
--- a/esp/esp/program/modules/handlers/studentclassregmodule.py
+++ b/esp/esp/program/modules/handlers/studentclassregmodule.py
@@ -543,8 +543,6 @@ class StudentClassRegModule(ProgramModuleObj):
                 class_blobs.append(category_header_str % (class_category_id, class_category_id, cls.category.category))
             class_blobs.append(render_class_direct(cls))
             class_blobs.append('<br />')
-        if categories_sort:
-            class_blobs.append('</div>')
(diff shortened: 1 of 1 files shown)


willgearty:
I don't really have an easy way to check the raw HTML outside of inspecting the rendered page (at which point the browser has already "fixed" it up). I agree that proper rendered HTML is the goal, so as long as this doesn't break other tests, I think it's good to go.

kkbrum (approved):

PR #3814, opened 2025-01-19
Droplets theme

willgearty:
This is the start of a new theme, based largely on bigpicture, called "droplets", named after the dropdown nav menus. It includes the following improvements:

- Fixed navbar that stays at the top of the screen even when scrolling
- Dropdown navigation menus
- Admin toolbar and login info/form integrated into navbar
- Supports glyphicons for dropdown options
- Increases width of main content to full width of screen
- Responsive to device screen size (include the navbar turning into a collapsed vertical menu)

Still to be done:

- [ ] Integrate bootstrap variables? (@kkbrum is more familiar with this than I am)
- [ ] Make and integrate optional variables to alter navbar appearance?
- [ ] Check how lots of pages look with full screen width

While making this theme responsive, I discovered that all we needed to do to make other themes responsive was to compile an additional bootstrap file. Since this is now included in theme_compiled.css, I've removed the line pulling bootstrap from the CDN which apparently wasn't doing anything.

image


willgearty:
I've moved the contact info to another dropdown, but moving it to the footer and changing the footer to be static instead of fixed may be a better design.

willgearty:
Also, maybe we want to make the dropdown menus like the ones here: https://sustainability.stanford.edu/. This way we would still have the text link to pages and the dropdown would be a separate button.

flora-concise:
## 1. Box too wide

We might need a maximum width limit. This seems to be a common problem across several pages. In combination with this, we could make it so that when there is not much content in the table/form, its maximum width would be accordingly narrower.
- https://dev5.learningu.org/teach/teachers/fchen25/bio.edit.html

image

- https://dev5.learningu.org/manage/Splash/2024_Spring/dashboard

image

## 2. Bubble hint outside of page
- Could we put the text inside the dropdown panels instead?

image

## 3. On the mobile/portrait version of the navigation header/dropdown, it is unclear which pieces of text are clickable.
Note that the same issue exists on the desktop/laptop version as well, although it is a bit less confusing.

image

## 4. On mobile, the explanation on the individual buttons would be barely discernible.
https://dev5.learningu.org/manage/Splash/2024_Spring/main

image

willgearty:
@flora-concise what would you recommend as a fix for # 3?

flora-concise:
@willgearty Thank you for the updates.

For # 3, I personally think it would help if you uniformly make the dropdown parents non-linkable, or put somewhere the suggestion that chapter admins leave the "learn" and "teach" as well as other such dropdown **_parents_** unlinked, but instead put the link in the dropdown as "Learn home" and "Teach home", for example.

Frankly, I found the distance between the parent node texts and the downward arrows/triangles to be a little confusing, but I think this would be less of an issue on an actual mobile device instead of a portrait-sized PC browser.

Please let me know if you would like some more suggestions on this topic, or if you'd like me to clarify something.

willgearty:
@flora-concise Thanks for giving some more insight, it's always much easier to make improvements when we have a better idea about the current user experience! I actually modelled this menu style off of the recently redone Syracuse University website: https://www.syracuse.edu/. Could you check it out and let me know if you have the same issues there? And if not, what aspects of the two may cause the discrepancy?

Ultimately, I imagine we may do as you suggest and leave this up to the admins to decide. We can recommend they not add links to the top menu, but I do think we want to allow that flexibility for chapters that want it, since it works this way for many of the existing themes (also, a huge push recently with the websites has been to give more editing power to admins).

willgearty:
Some notes to myself and @kkbrum:

- I think it would be cool if the navbar font size could be changed separately from the typography font size.
- We should add a custom toggle for the header image to not be shown on pages that aren't the homepage (and/or an option to not have the header at all).

flora-concise:
- [ ] I wonder if we could change the "Theme Palette" heading text under Palette to something like "Splash defaults - permanent" to indicate that these cannot be changed; that is, if admins want to customize their own colors, they should use Custom Palette.
image

- [ ] We currently have "Theme Properties" as the last heading, but looking at the content in the dropdown, I think perhaps "Create a save file" or something similar would be more intuitive.

- [ ] Both the theme editor help popup ("New ...? Click here") and the theme-saving section refer to a "Test" button that has been renamed. Do we want to edit the text?
image
image

Issue #3803, opened 2025-01-17
Comm panel email templates


willgearty:
Lots of chapters use email templates in various ways, such as:

- Keeping a local text template, filling it in offline, then copying and pasting into the comm panel (I did this when I was at Stanford Splash)
- Pulling up an old email in the comm panel, editing it, then sending again
- Custom code to insert the comm panel text into a template (Yale does this)

We should implement email templates in a more official manner. This would likely involve the following incremental steps:

- [x] Pull in the custom changes from Yale's branch to support template for emails. By default, the template would be blank. (see #1898)
- [ ] Implement a way for chapters to then supply multiple templates (and perhaps seeding this with templates of our own e.g., https://github.com/sendgrid/email-templates)
- [ ] Improve on the UI to make the selection of templates more visual like proposed in #2885
(a full-fledged template builder seems out-of-scope, but is perhaps a pipe dream)

milescalabresi:
First box done by #3817

Issue #3796, opened 2024-12-24
Incorporate chapter-prod customizations


milescalabresi:
Maintaining separate production branches is a pain. Let's merge the following.

- [ ] https://github.com/learning-unlimited/ESP-Website/compare/stable-release-15...stanford-sr15
- [ ] https://github.com/learning-unlimited/ESP-Website/compare/stable-release-15...chicago-sr15
- [ ] https://github.com/learning-unlimited/ESP-Website/compare/stable-release-15...yale-sr15 (should be taken care of by #1898)

willgearty:
Fixing this should also fix #2555

PR #3817, opened 2025-01-26
Create a default email template

milescalabresi:
Supplants #1898. This is a first step towards more configurable email templating. Partially addresses #3803 and #3796.


kkbrum:
Is this ready for testing or are you still tweaking it? I see you requested my review but then pushed a commit after, so just double checking.

milescalabresi:
Ready for review! that additional push was a flourish

esp/esp/program/modules/handlers/commmodule.py
```diff
@@ -124,9 +125,13 @@ def commprev(self, request, tl, one, two, module, extra, prog):
 
         contextdict = {'user'   : ActionHandler(firstuser, firstuser),
                        'program': ActionHandler(self.program, firstuser),
-                       'request': ActionHandler(MessageRequest(), firstuser)}
+                       'request': ActionHandler(MessageRequest(), firstuser),
+                       'EMAIL_HOST_SENDER': settings.EMAIL_HOST_SENDER}
 
-        renderedtext = Template(body).render(DjangoContext(contextdict))
+        # Use whichever template the user selected or the default (just an unsubscribe slug) if 'None'
```
kkbrum, line 131: How does someone select which template to use? I'm not seeing where the admin would have the chance to select the minimal template.

kkbrum (commented):
This is nice! Do you want to make a note on the first comm panel page about the default email template? Or maybe on the preview email page? So an admin knows why the text is showing up on the preview and where they might go to change it if desired? Unless we don't want that...

esp/esp/program/modules/handlers/commmodule.py
milescalabresi: One doesn't... yet. This is a stepping stone to a selectable set. That comment is probably bad form, but my hope is to get another PR out before the next release that actually allows selection from multiple. I can change it if you think I should

milescalabresi:
There's nowhere they can go right now to change it (planning to add that in a future PR). And the place to go to know why it's there is of course the release notes, or websupport@, or #websupport, or, or, or...

kkbrum (approved):
Okay; this is fine by me then!

PR #3827, opened 2025-04-11
Add modern favicon files

willgearty:
I made these new favicon files using favicon.io: https://favicon.io/favicon-converter/. This is still a work in progress, since we should also:

1. generate these other files when new favicon images are uploaded
2. list them in the `<head>`


willgearty:
And should the favicon have a white background or transparent background?

willgearty:
Looks like [Pillow](https://pypi.org/project/pillow/) or the [favicons](https://github.com/thatmattlove/favicons) package would work for generating these dynamically. https://realfavicongenerator.net also has an API that can be used to generate favicons, [as used in the pkgdown R package](https://github.com/r-lib/pkgdown/blob/fe04924b3df129bea60ac871614b01d87bcae147/R/build-favicons.R#L68C26-L68C58).

prem-2006:
Hi @willgearty I would like to work on this as I’m interested in working on improving favicon generation and updating the <head> entries. Could you assign this to me.

willgearty:
Looks like @antcybersec has already made an initial pull request for this.

PR #3832, opened 2025-04-23
fix and improve incoming email relay

### LINKED ISSUE #125: Make internal (not Mailman) mailing lists act more like real lists
Mailing lists generated through the website should act more like mailing lists; most notably, make the mail still appear to have come from the list when it gets to the recipients, rather than rewriting the address.  Probably other changes too.

### LINKED ISSUE #1662: Duplicate emails sent to students in class
Sometimes, when mail is sent to students via the (non-mailman) class mailing lists, students receive 1-2 dozen duplicates of the message.

If I remember correctly, it is either connected with sending the mail via BCC, or sending to multiple lists at once, or both.

Reported by Stanford and Yale.

### LINKED ISSUE #2393: Plain redirects should protect recipient info
Under the current implementation, senders can see recipient info if, for example, they use the wrong case (Splash-directors@ instead of splash-directors@). In that case (and others?) Mailer Daemon returns a "failure to send" message, which includes the destination of the redirect. That seems like a big privacy concern.

### LINKED ISSUE #3172: bcc'ing class lists causes email recursion
Looks like if you email a class list as a "bcc", the list itself is then included in the recipients, causing a recursive mess that can result in dozens of the same email being sent and will likely crash the server. You can tell there is recursion going on because the appended part of the subject is duplicated. It's also possible this happens with "cc", but I haven't confirmed that yet.

Reported by Cloud.

### LINKED ISSUE #3760: Add forwarding headers to emails
See "add forwarding headers" section of https://apps.google.com/supportwidget/articlehome?hl=en&article_url=https%3A%2F%2Fsupport.google.com%2Fa%2Fanswer%2F175365%3Fhl%3Den&assistant_id=generic-unu&product_context=175365&product_name=UnuFlow&trigger_context=a and add the appropriate headers to https://github.com/learning-unlimited/ESP-Website/blob/main/esp/mailgates/mailgate.py#L87.

### LINKED ISSUE #3977: Tests for email relay
From copilot on [this MR](https://github.com/learning-unlimited/ESP-Website/pull/3832#discussion_r2734657187):

>This substantial refactoring of email handling logic (lines 87-186) introduces complex new recipient 
>resolution and sender validation logic without accompanying tests. Given that the codebase has extensive 
>test coverage in other areas, the lack of tests for this critical email forwarding functionality represents 
>a significant gap. Consider adding unit tests to verify: 
>1) correct filtering of learningu.org aliases, 
>2) PlainRedirect and ESPUser lookups, 
>3) sender validation logic, 
>4) edge cases like empty recipient lists, malformed from addresses, and users without accounts.

milescalabresi:
replace sha with hashlib for python3 to fix issue with emailing class list aliases

Fixes https://github.com/learning-unlimited/ESP-Website/issues/3760
Fixes https://github.com/learning-unlimited/ESP-Website/issues/3172
Fixes https://github.com/learning-unlimited/ESP-Website/issues/1662
Fixes https://github.com/learning-unlimited/ESP-Website/issues/125
Fixes https://github.com/learning-unlimited/ESP-Website/issues/2393
Fixes https://github.com/learning-unlimited/ESP-Website/issues/3977


willgearty:
Some additional ideas based on discussion:

1. We should verify sender email addresses in mailgate.py before sending along emails:
- [ ] https://pypi.org/project/dkimpy/
- [ ] https://pypi.org/project/checkdmarc/
- [ ] https://pypi.org/project/pyspf/
2. We should reject any emails from

- [x] a sender email address that isn't associated with an account on [subdomain].learningu.org or learningu.org

milescalabresi:
We discussed a better way to do it.

1. Catch sender's message and grab the data fields (`To`, `From`, `Subject`, `Body`, etc.)
2. If the sender's email is *not* associated with an account on the site:
  a. Do not forward the email
  b. Send the user a bounce message (rate-limit once per day/week/something using something similar to `dbmail.MessageRequests` to track)
3. If the sender's email is associated with an account on the site:
  a. If there is more than one such account, do some logic to identify the "right" account (options include: prefer admin > teacher > volunteer > student; choose earliest account created; send a bounce and instruct them to merge accounts), say `sender@learningu.org`
  b. Use SendGrid to send an email to each recipient of the original message (`To`, `Cc`, `Bcc`) individually from `sender@site.learningu.org`. We should be able to import and use the [`send_mail`](https://github.com/learning-unlimited/ESP-Website/blob/49f209341a93a948d77a9bd4ea8f8201efbb284a/esp/esp/dbmail/models.py#L71) function

esp/mailgates/mailgate.py
```diff
@@ -76,47 +68,35 @@ def send_mail(message):
...
-            send_mail(str(message))
-        else:
-            # send an email for each recipient
-            for recipient in instance.recipients:
-                del(message['To'])
-                message['To'] = recipient
-                send_mail(str(message))
+        # Catch sender's message and grab the data fields (To, From, Subject, Body, etc.)
+        data = dict()
+        for field in ['to', 'from', 'cc', 'bcc', 'subject', 'body']:
```
willgearty: What about attachments?

esp/mailgates/mailgate.py
```diff
@@ -76,47 +68,35 @@ def send_mail(message):
...
+            data[field] = message[field].split(',')
```
willgearty: I think you need to use `instance.recipients` to get the 'to' field

esp/mailgates/mailgate.py
milescalabresi: Good catch; added keyword

esp/mailgates/mailgate.py
milescalabresi: Another good catch; added TODO

hwatheod:
Is the purpose of these changes to address the spam abuse reports? Will all emails be sent through SendGrid now?

milescalabresi:
Yes, the vision is send all emails through SendGrid. That's one benefit but
not the only one

willgearty:
For the class codes, you may want to use something like the original code, that way it can also include a subject prefix if it is set:
https://github.com/learning-unlimited/ESP-Website/blob/373c94502161c1eb5ed147903c2ea3523fa6d6dc/esp/mailgates/mailgate.py#L95-L98

esp/mailgates/mailgate.py
```diff
@@ -74,63 +81,124 @@ def send_mail(message):
...
-            # send an email for each recipient
-            for recipient in instance.recipients:
-                del(message['To'])
-                message['To'] = recipient
-                send_mail(str(message))
-
+            if len(data['from']) != 1:
+                raise AttributeError(f"More than one sender: `{data['from']}`")
+            email_address = data['from'][0]
+            if '<' in email_address and '>' email_address:
```
willgearty:
```suggestion
            if '<' in email_address and '>' in email_address:
```

esp/mailgates/mailgate.py
```diff
@@ -74,63 +81,124 @@ def send_mail(message):
...
+                email_address = email_address.split('<')[1].split('>')[0]
+            if email_address.endswith(settings.EMAIL_HOST_SENDER):
+                users = ESPUser.objects.filter(username__iexact=email_address.split('@')[0]).order_by('date_joined') # sort oldest to newest
+            else:
+                users = ESPUser.objects.filter(email__iexact=email_address).order_by('date_joined') # sort oldest to newest
+            if len(users) == 0:
+                logger.warning('Received email from {}, which is not associated with a user'.format(data['from']))
```
willgearty: I'm seeing this warning for every single email I send from my gmail:
`WARNING: Received email from ['willgearty@dev5.learningu.org'], which is not associated with a user`

esp/mailgates/mailgate.py
```diff
@@ -74,63 +81,124 @@ def send_mail(message):
...
+            # use SendGrid to send an email to each recipient of the original message (To, Cc, Bcc) individually from
+            # the sender's site email
+            logger.info('Sending email as {}'.format(sender))
+            if isinstance(data['to'], str):
+                data['to'] = [data['to']]
+            for recipient in data['to']:
+                logger.debug(f"Sending to `{recipient}`")
+                send_mail(subject=data['subject'], message=data['body'],
+                         from_email='{}@{}'.format(sender, settings.EMAIL_HOST_SENDER),
+                          recipient_list=[recipient], attachments=data['attachments'], fail_silently=False)
```
willgearty: Should we pass along the CC field? I can imagine you might want to CC someone else on your email that the students could then reply to? But maybe that's a security risk?

.gitignore
```diff
@@ -36,7 +36,7 @@ local_fabfile.py
 *.orig
 *.swp
 .espsettings
-
+.github/copilot-instructions.md
```
willgearty, line 42: 😏

esp/mailgates/mailgate.py
```diff
@@ -74,63 +81,124 @@ def send_mail(message):
...
-        # get a new message id
-        message['Message-ID'] = '<%s@%s>' % (sha.new(str(random.random())).hexdigest(),
-                                             host)
-
-        if handler.cc_all:
-            # send one mass-email
-            message['To'] = ', '.join(instance.recipients)
-            send_mail(str(message))
+       # If the sender's email is not associated with an account on the site,
+       # do not forward the email
```
willgearty: What about @learningu.org email addresses?

esp/mailgates/mailgate.py
willgearty: I had to fix this on the server to get the emails to start processing.

esp/mailgates/mailgate.py
```diff
@@ -74,63 +81,124 @@ def send_mail(message):
         instance.process(user, *match.groups(), **match.groupdict())
 
         if not instance.send:
+            logger.info("Instance did not send")
             continue
```
willgearty, line 275: I think this is triggered when the handler doesn't match (and/or returns zero recipients):
https://github.com/learning-unlimited/ESP-Website/blob/a62374888c6a9122efaef7bdc15c103e1714ef94/esp/esp/dbmail/receivers/classlist.py#L42-L43
Probably don't need to log this on production.

willgearty (changes requested):
Very glad to see this finally ~wrapped up! I've left a handful of comments and questions. Also, did you change anything server-side that needs to be documented/stored on GitHub anywhere? Maybe somewhere in [here](https://github.com/learning-unlimited/ESP-Website/tree/main/esp/mailgates/exim_config)? Or in the [scripts repo](https://github.com/learning-unlimited/lu-scripts)?

esp/mailgates/mailgate.py
milescalabresi: Should we allow this? It would mean you don't get the privacy of the aliasing, but I'm open to the discussion

esp/mailgates/mailgate.py
milescalabresi: Yes. When I tested, it seemed like it got folded into the `recipient`. Are you seeing something different?

esp/mailgates/mailgate.py
milescalabresi: Agreed, should be `debug`

esp/mailgates/mailgate.py
willgearty: Oh, that's true. Hmm...I guess it's fine the way it is. We can always make a dummy account on a website to then send an email if need be.

esp/mailgates/mailgate.py
willgearty: Would also be good to have the message be more descriptive (although I don't think we even need to log this at all)

esp/mailgates/mailgate.py
willgearty: (but it still sends the email)

esp/mailgates/mailgate.py
willgearty: Hmmm...maybe that was happening. I was thinking that it would be nice to have the CC still listed as CC so that the email address is on the received email, along with the to and from, but not sure if that will work.

milescalabresi: What's the envisioned use case? This setup currently sends an individual email to each to/cc/bcc recipient. Are you saying that if you were cc'd on the sent email, it should send a single email where you are on the cc line with no "to" recipients?

willgearty: I was thinking that someone that was cc'd should then also be cc'd on each of the individual emails sent, but I think that's not necessary.

esp/mailgates/mailgate.py
milescalabresi: Fixed by [f13fba9](https://github.com/learning-unlimited/ESP-Website/pull/3832/commits/f13fba92074057cda1200ade7eda84009925f778)

esp/mailgates/mailgate.py
```diff
@@ -74,63 +81,124 @@ def send_mail(message):
...
+            email_address = data['from'][0]
+            if '<' in email_address and '>' in email_address:
+                email_address = email_address.split('<')[1].split('>')[0]
+            if email_address.endswith(settings.EMAIL_HOST_SENDER):
+                users = ESPUser.objects.filter(username__iexact=email_address.split('@')[0]).order_by('date_joined') # sort oldest to newest
+            else:
+                users = ESPUser.objects.filter(email__iexact=email_address).order_by('date_joined') # sort oldest to newest
+            if len(users) == 0:
+                logger.warning('Received email from {}, which is not associated with a user'.format(data['from']))
+                continue
```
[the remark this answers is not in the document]
milescalabresi: This is intentional to avoid using our SendGrid credits in case we get a lot of spammy emails

esp/mailgates/mailgate.py
```diff
@@ -74,63 +81,124 @@ def send_mail(message):
...
+        # if the instance has a `recipients` attribute, then it is a class list (such as `a123-teachers@`)
+        if hasattr(instance, 'recipients'):
+            data['to'] = []
+            aliases = []
+            for recipient in instance.recipients:
+                # If the recipient has an email address that does not end with @anysite.learningu.org, keep them
+                # TODO: it would be better not to hardcode `.learningu.org` in case we ever change the name, but we
+                # only store `thissite.learningu.org` in settings, and we want `anysite.learningu.org` while still
+                # allowing user@learningu.org because those resolve to enterprise Gmail addresses
+                if recipient.endswith('.learningu.org'):
```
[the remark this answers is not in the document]
milescalabresi: Nope, it will not match such email addresses.

esp/mailgates/mailgate.py
```diff
@@ -74,63 +81,124 @@ def send_mail(message):
...
+            # Split out individual email addresses if any of the redirects is a list
+            redirects = list(itertools.chain.from_iterable(map(lambda x: x.destination.split(','), redirects)))
+            users = ESPUser.objects.annotate(username_lower=Lower("username"
+                    )).filter(username_lower__in=[x.lower() for x in aliases])
+            # Grab the emails from the users
+            users = [x.email for x in users]
+            # Theoretically at least one of these should be empty, but now doesn't seem like the time
+            # If the redirect resolve to anything@anysite.learningu.org, kill it
+            for address in redirects + users:
+                if not address.endswith('.learningu.org'): # TODO: again, would be nice not to hardcode
```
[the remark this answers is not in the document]
milescalabresi: Disagree. The `endswith` function will not allow matching malicious domains

willgearty (approved):
Looks good to me! Just address the Copilot comments where applicable, and then lets get this merged and on production!

docs/dev/email.rst
```diff
@@ -8,18 +8,18 @@ This document describes how email, both incoming and outgoing, makes its way whe
 Outgoing email
 --------------
 
-Outgoing email originates in the comm panel.  The comm panel is outside the scope of this document, but after an admin writes an email and selects recipients, the website creates a signle ``MessageRequest`` object with that data -- the email template, the list of students to which it will be sent, and some metadata like the sender.  (The ``MessageRequest`` model, and the others described below, are in ``esp/esp/dbmail/models.py``.)  
+User-generated outgoing email originates in the comm panel.  The comm panel is outside the scope of this document, but after an admin writes an email and selects recipients, the website creates a signle ``MessageRequest`` object with that data -- the email template, the list of students to which it will be sent, and some metadata like the sender.  (The ``MessageRequest`` model, and the others described below, are in ``esp/esp/dbmail/models.py``.)
```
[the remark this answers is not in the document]
milescalabresi: Fixed by [8337896](https://github.com/learning-unlimited/ESP-Website/pull/3832/commits/833789625a19ca989bb9801c632422e93948ac02)

esp/esp/dbmail/models.py
```diff
@@ -109,7 +109,10 @@ def send_mail(subject, message, from_email, recipient_list, fail_silently=False,
 
     #   Get whatever type of email connection Django provides.
     #   Normally this will be SMTP, but it also has an in-memory backend for testing.
-    connection = get_connection(fail_silently=fail_silently, return_path=return_path)
+    connection = get_connection(fail_silently=fail_silently)
```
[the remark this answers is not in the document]
milescalabresi: This is intentional -- as you said, it only set if if it has the attribute.

esp/mailgates/mailgate.py
```diff
@@ -49,14 +48,22 @@
...
-    p.write(message)
+
+def extract_attachments(msg):
+    attachments = []
+    for part in msg.iter_attachments():
+        filename = part.get_filename()
+        content = part.get_payload(decode=True)
+        mimetype = part.get_content_type()
+        if content:
+            attachments.append((filename, content, mimetype))
```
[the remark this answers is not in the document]
milescalabresi: Fixed in 9ccad44f9

esp/mailgates/mailgate.py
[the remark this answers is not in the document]
milescalabresi: Fixed in https://github.com/learning-unlimited/ESP-Website/pull/3832/commits/65b8ee1d03e5699db9673c48471fcf155c549ad4

esp/mailgates/mailgate.py
```diff
@@ -74,63 +81,124 @@ def send_mail(message):
...
+            logger.info('Sending email as {}'.format(sender))
+            if isinstance(data['to'], str):
+                data['to'] = [data['to']]
+            for recipient in data['to']:
+                logger.debug(f"Sending to `{recipient}`")
+                send_mail(subject=data['subject'], message=data['body'],
+                         from_email='{}@{}'.format(sender, settings.EMAIL_HOST_SENDER),
+                          recipient_list=[recipient], attachments=data['attachments'], fail_silently=False)
+            del sender, recipient, users
         sys.exit(0)
```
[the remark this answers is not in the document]
milescalabresi: Made separate issue #3977

esp/mailgates/mailgate.py
```diff
@@ -5,7 +5,8 @@
 
 from __future__ import absolute_import
 from __future__ import print_function
-import sys, os, email, re, smtplib, socket, sha, random
+import sys, os, base64, email, hashlib, re, smtplib, socket, random
```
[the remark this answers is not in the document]
milescalabresi: The description has become misleading in light of [this comment](https://github.com/learning-unlimited/ESP-Website/pull/3832#issuecomment-2903076526)

esp/mailgates/mailgate.py
```diff
@@ -74,63 +81,124 @@ def send_mail(message):
...
+            # if the above filtering leaves the 'to' list empty, abort
+            if len(data['to']) == 0:
+                continue
+        # if the instance has a `message['to']` attribute, then it is a single address (such as `plain_redirect@` or
+        # `username@`)
+        elif hasattr(instance, 'message'):
+            data['to'] = instance.message['to']
+        else:
+            raise TypeError("Unknown receiver type for `{}`".format(instance))
+        data['from'] = message['from'].split(',') or ''
```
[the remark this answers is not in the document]
milescalabresi: Suggestion accepted

esp/mailgates/mailgate.py
```diff
@@ -74,63 +81,124 @@ def send_mail(message):
...
+            elif len(users) > 1:
+                for group_name in ['Administrator', 'Teacher', 'Volunteer', 'Student', 'Educator']:
+                    group_users = [x for x in users if len(x.groups.filter(name=group_name)) > 0]
+                    if len(group_users) > 0:
+                        sender = group_users[0] # choose the first (oldest) account if there is still more than
+                        # one; it won't matter because they all go to the same email by construction
+                        break
+                else: # if the users aren't in any of the standard groups above, ...
+                    sender = users[0] # ... then just pick the oldest account created by selecting users[0] as above
+                logger.debug(f"Group selection: {group_name} -> {group_users}")
```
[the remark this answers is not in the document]
milescalabresi: The list is hard-coded, so `group_name` will always be defined, but 54eb60804 fixes the issue with the `else` clause

esp/mailgates/mailgate.py
```diff
@@ -74,63 +81,124 @@ def send_mail(message):
...
         if handler.subject_prefix:
-            subject = '[%s] %s' % (handler.subject_prefix, subject)
-        message['Subject'] = subject
-
-        if handler.from_email:
-            del(message['from'])
-            message['From'] = handler.from_email
+            data['subject'] = '[{}] {}'.format(handler.subject_prefix, data['subject'])
+        # Put the email body between HTML tags
+        data['body'] = '<html>{}</html>'.format(message.get_body(preferencelist=('html', 'plain')).get_content())
```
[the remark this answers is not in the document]
milescalabresi: Fixed by 029c7b648

esp/mailgates/mailgate.py
```diff
@@ -74,63 +81,124 @@ def send_mail(message):
...
+            # the sender's site email
+            logger.info('Sending email as {}'.format(sender))
+            if isinstance(data['to'], str):
+                data['to'] = [data['to']]
+            for recipient in data['to']:
+                logger.debug(f"Sending to `{recipient}`")
+                send_mail(subject=data['subject'], message=data['body'],
+                         from_email='{}@{}'.format(sender, settings.EMAIL_HOST_SENDER),
+                          recipient_list=[recipient], attachments=data['attachments'], fail_silently=False)
+            del sender, recipient, users
```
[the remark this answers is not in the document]
milescalabresi: Suggestion accepted in [9a6a9cc](https://github.com/learning-unlimited/ESP-Website/pull/3832/commits/9a6a9cc7b97d090e32844153d3ddbc3ce4aa2141)

esp/mailgates/mailgate.py
```diff
@@ -74,63 +81,124 @@ def send_mail(message):
...
-        message['Message-ID'] = '<%s@%s>' % (sha.new(str(random.random())).hexdigest(),
-                                             host)
-
-        if handler.cc_all:
-            # send one mass-email
-            message['To'] = ', '.join(instance.recipients)
-            send_mail(str(message))
+       # If the sender's email is not associated with an account on the site,
+       # do not forward the email
+        if not data['from']:
```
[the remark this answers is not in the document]
milescalabresi: suggestion accepted in [87c3e3e](https://github.com/learning-unlimited/ESP-Website/pull/3832/commits/87c3e3edf785614aa9b707892b74151cce812a97)

esp/mailgates/mailgate.py
```diff
@@ -74,63 +81,124 @@ def send_mail(message):
...
+                if recipient.endswith('.learningu.org'):
+                    aliases.append(recipient)
+                elif '@' in recipient:
+                    data['to'].append(recipient)
+                else:
+                    logger.warning('Email address without `@` symbol: `{}`'.format(recipient))
+            redirects = PlainRedirect.objects.annotate(original_lower=Lower("original"
+                        )).filter(original_lower__in=[x.split('@')[0].lower() for x in aliases])
+            # Split out individual email addresses if any of the redirects is a list
+            redirects = list(itertools.chain.from_iterable(map(lambda x: x.destination.split(','), redirects)))
```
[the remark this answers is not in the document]
milescalabresi: suggestion accepted in [09f1202](https://github.com/learning-unlimited/ESP-Website/pull/3832/commits/09f1202eb25c5787d9e73dd8295d0af9ddb1884a)

esp/mailgates/mailgate.py
```diff
@@ -74,63 +81,124 @@ def send_mail(message):
...
+            users = ESPUser.objects.annotate(username_lower=Lower("username"
+                    )).filter(username_lower__in=[x.lower() for x in aliases])
```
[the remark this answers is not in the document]
milescalabresi: Suggestion accepted in [09f1202](https://github.com/learning-unlimited/ESP-Website/pull/3832/commits/09f1202eb25c5787d9e73dd8295d0af9ddb1884a)

esp/mailgates/mailgate.py
```diff
@@ -74,63 +81,124 @@ def send_mail(message):
...
+            if len(users) == 0:
+                logger.warning('Received email from {}, which is not associated with a user'.format(data['from']))
+                continue
+            elif len(users) == 1:
+                sender = users[0]
+            # If there is more than one associated account, choose one by prefering admin > teacher > volunteer >
+            # student then choosing the earliest account created. Then send as before using the unique account.
+            elif len(users) > 1:
+                for group_name in ['Administrator', 'Teacher', 'Volunteer', 'Student', 'Educator']:
+                    group_users = [x for x in users if len(x.groups.filter(name=group_name)) > 0]
```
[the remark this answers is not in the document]
milescalabresi: @Copilot we expect `users` to be quite small (under 10) in general, while the model `ESPUser` is very large (tens of thousands). Does that change your comment?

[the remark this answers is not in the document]
milescalabresi: Response:

> 
> You're absolutely right—thank you for the clarification! If the set of users is consistently under 10, then my concern about performance doesn't really apply here. A direct filter like `ESPUser.objects.filter(id__in=ids)` will be efficient for small id lists, even though `ESPUser` itself is large. The database handles small IN queries very well.
> 
> Given that constraint, your current approach is solid and there's no need to over-engineer this. No changes needed unless you encounter specific performance issues in practice.

esp/mailgates/mailgate.py
```diff
@@ -74,63 +81,124 @@ def send_mail(message):
...
+                    if len(group_users) > 0:
+                        sender = group_users[0] # choose the first (oldest) account if there is still more than
+                        # one; it won't matter because they all go to the same email by construction
+                        break
+                else: # if the users aren't in any of the standard groups above, ...
+                    sender = users[0] # ... then just pick the oldest account created by selecting users[0] as above
+                logger.debug(f"Group selection: {group_name} -> {group_users}")
+            else:
+                logger.error('Negative number of possible senders in supposed list `{}`. Skipping....'.format(users))
+                continue
```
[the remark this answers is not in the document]
milescalabresi: Fixed in [02ecb2e](https://github.com/learning-unlimited/ESP-Website/pull/3832/commits/02ecb2e87c55c4b075b74e146a882be14f4455c3)

esp/mailgates/mailgate.py
```diff
@@ -27,141 +43,358 @@
...
+    return '%s %s' % (marker, subject) if subject else marker
+
+
+def _rewrite_headers(message, handler_row, instance, list_address, sender):
+    """Header rewriting for group broadcasts (ClassList, SectionList, PlainList).
+
+    Handlers that set `preserve_headers` (e.g. UserEmail) skip this entirely.
+    """
+    del message['to']
+    del message['cc']
```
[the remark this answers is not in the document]
willgearty: False positive: 
https://docs.python.org/3/library/email.message.html#email.message.EmailMessage.__delitem__
> Delete all occurrences of the field with name name from the message’s headers. No exception is raised if the named field isn’t present in the headers.

willgearty:
## Summary

This outgrew its description (a SHA→hashlib fix) into a large rework of inbound mail handling.

### Security fixes

**Open relay.** The mailgate routed on the message's `To`/`Cc`/`Bcc` headers rather than the envelope recipient, forwarding to any non-`learningu.org` address found there. Confirmed on dev5: a message whose only site involvement was a stripped `Bcc` was relayed to an external address through our SendGrid credentials. Routing now keys on `LOCAL_PART`, so recipients only ever come from an `EmailList` handler.

**Student aliases exposed.** `UserEmail` previously forwarded only for teachers; the rewrite dropped that. Restored and scoped: teacher/admin aliases accept mail from any registered sender, everyone else's only from teachers/admins.

**No sender check.** Mail from addresses with no site account was forwarded; now dropped, with no bounce (bouncing confirms which addresses exist and can be aimed at third parties via a forged `From`).

**Leak in `send_mail`.** `extra_headers={}` is mutated to add per-user `List-Unsubscribe` headers, so in `dbmail_cron` one recipient's one-click unsubscribe link carried into later messages. Fixed, with regression tests.

### Restored

`EmailList` dispatch, which the rewrite removed — silently breaking class and section lists (`S1234-teachers@` resolved to nothing and was dropped). `Message-ID` regeneration and subject prefixing came back with it.

### New

- `From:` rewritten to the sender's `username@<site>` alias — fixes DMARC failures and keeps personal addresses out of forwarded copies. Fixes #3760.
- Per-address `Delivered-To` loop detection. Fixes #3172. Fixes #1662.
- Group broadcasts get `Reply-To: <list>`, plus the sender's alias if they aren't a member. Fixes #125.
- Bounces to registered senders only, rate limited. Fixes #2393.
- `MAILGATE_EMAIL_BACKEND` replaces hardcoded SendGrid SMTP, so esp.mit.edu and other non-SendGrid sites can use this.
- `docs/dev/email.rst` rewritten to match.

### Tests

New `esp/dbmail/test_mailgate.py` (21 tests, locmem backend). Updated other tests in `esp/dbmail/` as appropriate. Fixes #3977.

All Copilot threads are either stale or addressed.

PR #3839, opened 2025-05-30
Fix meal ticket logic

### LINKED ISSUE #3767: meal ticket printables yield wrong number of lines
Not that I think anyone uses this, but the printable for meal tickets gives up to 7 lines even if you click on 3 lines, for example

image

This is what I got from using "3 lines"

image

willgearty:
Fixes #3767.

### DIFF
--- a/esp/esp/program/modules/handlers/programprintables.py
+++ b/esp/esp/program/modules/handlers/programprintables.py
@@ -71,6 +71,8 @@ import copy
 import csv
 import json
 
+from numpy import array_split
+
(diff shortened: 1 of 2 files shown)


kkbrum (changes requested):
The lines logic seems to work correctly now; thanks! While we're touching this printable, could you also make the comma go away if the name is blank? 
2025-06-12 20_36_21-Window

willgearty:
Done! I also fixed the list of default colors while I was at it.

kkbrum (approved):

PR #3844, opened 2025-06-06
Improve custom form module errors

### LINKED ISSUE #3715: Student/teacher custom form errors should be updated
From @kkbrum:

> Now that we have more user friendly ways of setting this tag, perhaps we should have a more helpful message here if the current user is an admin (ie, give them a few instructions and lead them to the custom forms page). In particular, they really don’t need the “learn_extraform_id” wording anymore since they shouldn’t be going to the admin pages and setting that

image

### LINKED ISSUE #3807: Change "links" in custom forms
Change the name "links" in custom forms since it can be misleading on first reading.

willgearty:
This improves the error messages that appear when trying to access a custom form module (student, teacher, or logistics quiz) that doesn't have a linked custom form. If the user is an admin, the message links to /customforms and instructs the admin to make a form and link it to the module. I also updated the header in the custom form creator from "Links" to "Link to a Program or Course" to better align with these instructions.

Fixes #3715
Fixes #3807

### DIFF
--- a/esp/esp/program/modules/handlers/studentcustomformmodule.py
+++ b/esp/esp/program/modules/handlers/studentcustomformmodule.py
@@ -109,7 +109,11 @@ class StudentCustomFormModule(ProgramModuleObj):
         if custom_form_id:
             cf = Form.objects.get(id=int(custom_form_id))
         else:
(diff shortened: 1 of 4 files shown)


kkbrum (commented):
This looks great. This is out of scope, but when testing this, it seems like the custom form modules don't show up in the new module sequence/requiredness editor even when enabled? I had to go to the admin pages to make it required in order to get this error. Perhaps I'm missing something??

PR #3845, opened 2025-06-06
Fix unenroll module

### LINKED ISSUE #3834: Does the unenroll students onsite page actually work?
Stanford reports that /onsite/one/two/unenroll_students may not work (i.e., the students are not unenrolled from the classes)

willgearty:
This fixes the onsite unenroll module. Based on some testing, I think there was some database funkiness that was resulting in the wrong list of student registrations being selected. When the module then "unenrolled" the students, it was expiring the wrong registrations, so it didn't end up looking like anything changed.

I should also note that this module hasn't been touched in 9 years and it could use a bit of love. For example, the individual timeslots don't specify how many students/sections they have, and the results page gives some raw python code to run to undo things. I haven't decided yet if I'm going to work on improving that here or in a separate PR.

EDIT: I've made the following UI changes:

- Added a button that will undo the unenrolling (in place of the janky raw code)
- Made the numbers of students/sections more obvious
- Put some text in bootstrap alert banners
- Made the two columns the same width

Fixes #3834.

### DIFF
--- a/esp/templates/program/modules/unenrollmodule/select.html
+++ b/esp/templates/program/modules/unenrollmodule/select.html
@@ -13,7 +13,10 @@
 
 <h1>Select Students to Unenroll</h1>
 
-<p>Use the form below to unenroll students who have not checked into the program. Before using this, make sure that the latest check-ins have been recorded in the website.</p>
+<div class="alert alert-info">
+  <span class="glyphicon glyphicon-info-sign" aria-hidden="true"></span>
+  Use the form below to unenroll students who have not checked into the program. Before using this, make sure that the latest check-ins have been recorded in the website.
+</div>
 
 <p>
   <span id="message">Loading enrollment numbers...</span>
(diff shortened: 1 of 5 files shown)


esp/templates/program/modules/unenrollmodule/select.html
```diff
@@ -24,8 +27,8 @@ <h1>Select Students to Unenroll</h1>
   <input type="hidden" name="selected_enrollments" value="" />
   <table cellpadding="4" cellspacing="0" align="center" width="300">
     <tr>
-      <th>Kick students whose first class starts at...</th>
-      <th>from sections starting at...</th>
+      <th width="50%">Kick students whose first class starts at...</th>
+      <th width="50%">from sections starting at...</th>
```
kkbrum: I propose to change "sections" here to "classes". The "class" notation on the left and the "section" notation on the right make it seem like that is a key difference in the columns, when it really isn't the point (if I understand correctly)

willgearty:
Done! I changed it in the above javascript-generated text as well.

kkbrum (approved):

PR #3888, opened 2025-08-27
Refactored @media min and max width to use @navbarCollapseWidth for i…

### LINKED ISSUE #3856: Droplets login visual bug
When on a narrow-ish screen (but not tablet narrow), the login in the top navbar gets shifted down a line, making the :

Image

samalarussa:
Refactored @media min and max width to use @navbarCollapseWidth for in Droplets theme, and made that a variable in the theme settings. Fixes #3856 via a .less refactoring. Note that fixing this bug now requires changing a theme setting, since navbar length is variable based on a number of factors.

### DIFF
--- a/esp/templates/themes/editor.html
+++ b/esp/templates/themes/editor.html
@@ -466,7 +466,7 @@ var palette_list = [{% for color in palette.base %} "{{color}}", {% endfor %}{% 
 						{% if theme_name == "droplets" %}
 							<div>
 
-                            <p>navbarCollapseWidth is the minimum width at which the navbar will collapse into a hamburger menu (behaves like the mobile site). Increase this value if the navbar goes onto two lines when the window is resized smaller.</p>
+                            <p>navbarCollapseWidth is the width at which the navbar will collapse into a hamburger menu (behaves like the mobile site). Increase this value if the navbar goes onto two lines when the window is resized smaller.</p>
 
                             </div>
 						{% endif %}


samalarussa:
May be worth creating a future issue to make the "@media" tag dynamic based on navbar size, but this got increasingly complicated and may be addressed by a boostrap update.

esp/public/media/theme_editor/less/variables.less
```diff
@@ -208,8 +208,8 @@
 
 // Navbar
 // -------------------------
-@navbarCollapseWidth:             979px;
-@navbarCollapseDesktopWidth:      @navbarCollapseWidth + 1;
+//@navbarCollapseWidth:             979px;
+//@navbarCollapseDesktopWidth:      @navbarCollapseWidth + 1;
```
willgearty: We can probably keep this one uncommented and remove it from variables_custom.less, since I don't think we'll ever want to modify both variables.

kkbrum (commented):
So this is a guess & check game for the admin, right? I think I would add some help text in the theme editor next to the advanced variables to explain a bit more about the purpose of changing the collapse width.

esp/templates/themes/editor.html
```diff
@@ -463,6 +463,13 @@ <h5>These are additional parameters specific to the currently selected theme.</h
 
                             </div>
                         {% endif %}
+						{% if theme_name == "droplets" %}
+							<div>
+
+                            <p>navbarCollapseWidth is the minimum width at which the navbar will collapse into a hamburger menu (behaves like the mobile site). Increase this value if the navbar goes onto two lines when the window is resized smaller.</p>
```
kkbrum:
```suggestion
                            <p>navbarCollapseWidth is the width at which the navbar will collapse into a hamburger menu (behaves like the mobile site). Increase this value if the navbar goes onto two lines when the window is resized smaller.</p>
```

kkbrum (approved):
I would remove the word "minimum" in the help text; It can be seen as both a minimum and a maximum, and it hurts my brain a bit, so I would leave it as an un-qualified width. Otherwise, this looks great to me!

PR #3897, opened 2025-09-05
Update credit card vs confirm reg order 

### LINKED ISSUE #3710: Module Management red box text and constraints
As of right now, we say:
https://github.com/learning-unlimited/ESP-Website/blob/7c1ed01bdd65398eda176333ec0d92aa5282f7e2/esp/templates/program/modules/admincore/modules.html#L37-L38

But it looks like we actually always make the credit card module second to last (seq 10000) and the confirm module last (seq 99999). I imagine I meant to set the credit card seq to 100000, but just misread the confirm module's seq value. However, maybe it makes sense to have the confirm module always last and the credit card module always second to last? Either way, we'll need to update the behavior and/or the text here.

kkbrum:
Fixes #3710.

### DIFF
--- a/esp/esp/program/modules/handlers/admincore.py
+++ b/esp/esp/program/modules/handlers/admincore.py
@@ -574,15 +574,16 @@ class AdminCore(ProgramModuleObj, CoreModule):
                 pmo.seq = 0
                 pmo.required = True
                 pmo.save()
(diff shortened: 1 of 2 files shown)


willgearty (approved):
Looks good and works as described!

PR #3898, opened 2025-09-05
Refactored schedule check refresh input box to use an id instead of a…

### LINKED ISSUE #3866: Refresh cycle textbox input is poorly formatted on the scheduling checks page
(at least in droplets theme)

Image

samalarussa:
… name attribute to establish specificity. Fixes #3866


willgearty:
I think this page uses React.js, so I think the `name` attribute is important for that.

willgearty (approved):
Works now, thanks!

PR #3899, opened 2025-09-05
Render emails within `iframe`s

### LINKED ISSUE #3896: Email templates with styling break commprev and /manage/emails
When an email has a full template, including a `<body>` tag and styling for that `<body>` tag, it causes the entirety of the droplets page to match that styling. This was probably always an issue, but it's become an even bigger issue now that droplets doesn't have a separate background color for the content vs. the rest of the page. Looks like the best solution is probably to put the emails within `iframe`s as discussed here: https://stackoverflow.com/a/781145/4660582

/commprev (I've highlighted the text so you can see it):
Image

/manage/emails (haven't highlighted the text here to make it clear what the problem is):
Image

willgearty:
This renders the richtext of emails within `iframe`s to isolate their stylings and prevent them from altering the main page while viewing the emails. This has been done on the communications panel preview page, the email status page, and the public email view page.

Fixes #3896

### DIFF
--- a/esp/templates/admin/emails.html
+++ b/esp/templates/admin/emails.html
@@ -96,7 +96,10 @@ tr.email-details {
 $j(function() {
     $j("tr.email-header").click(function(event) {
         var $target = $j(event.target).parent("tr.email-header");
-        $target.next("tr").toggle();
+        var $row = $target.next("tr");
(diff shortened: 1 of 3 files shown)


kkbrum (approved):
Looks good!

PR #3902, opened 2025-09-10
Fix handling of disabled teacher text buttons in teacher checkin

### LINKED ISSUE #3864: Undoing a teacher checkin enables the text button, even if texting is not set up
If you check in a teacher, it disables the texting button. If texting is not set up, this changes nothing, since the button was already disabled. However, if you undo this check-in, then it enables the texting button, even if texting is not set up.

willgearty:
This fixes a bug where the teacher text button on the checkin page would always be reenabled after undo a check-in, even when Twilio is not properly set up.

Fixes #3864.


kkbrum (approved):
Looks great!

PR #3903, opened 2025-09-10
Redid appearance of Survey page to match program management landing page style

### LINKED ISSUE #3859: Survey management landing page needs some love
/manage/one/two/surveys needs some loving to make it a little prettier and more user-friendly

samalarussa:
Redid appearance of survey page. Fixes #3859

### DIFF
--- a/esp/templates/program/modules/surveymanagement/main.html
+++ b/esp/templates/program/modules/surveymanagement/main.html
@@ -2,6 +2,15 @@
 
 {% block title %}{{program.niceName}} Survey Administration{% endblock %}
 
+{% block stylesheets %}
+    {{ block.super }}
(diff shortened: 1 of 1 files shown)


kkbrum (changes requested):
The page formatting looks great. To be an overachiever here, could you disable the buttons for options that would lead to whoops errors? In particular, if there is no survey, then everything except for "create a new survey" should be greyed out. If there is no "overall rating" question, then the "list favorite classes" should be greyed out. It would be especially nice if when you hover over the disabled button, it gives you a hint saying "you need to create a survey first!" or "you need to add an overall rating question to a survey first!".

kkbrum (approved):

PR #3905, opened 2025-09-19
Added editable textbox to the footer in all themes

### LINKED ISSUE #3878: Ability to add footer text
From Cornell Splash on Slack:

> In the bigpicture theme, is there a way to add a default text line to the bottom of every page without manually typing the same line into each page? There’s an editable text box on the header that does this, but is there a way to add one to the bottom of the page too?

samalarussa:
Future versions should wrap this in boolean logic and a checkbox in theme settings to avoid unneccesary padding. Fixes #3878


willgearty:
Would also be great to have this in every theme! droplets might be the only theme with a proper "footer", so you might need to get creative.

See for example the text that I added for Cornell (bigpicture): https://cornell.learningu.org/
Their template override is here: https://cornell.learningu.org/admin/utils/templateoverride/143/change/

kkbrum (approved):
Looks great!

PR #3907, opened 2025-09-24
Fixed Header Height Miscalculation

samalarussa:
Fixed .less calculation error leading to a header miscalculation. Error was introduced when adding navbar width as an editable variable.


willgearty (approved):
Works as described, thanks for fixing this up!

PR #3952, opened 2025-11-14
Add records and payment to barcode checkin

### LINKED ISSUE #3949: Add check-in options to barcode check-in
The "advanced check-in" page has several options for check-in, including "paid", which marks sets a student's balance to zero:

Image

It would be nice if the barcode check-in page had these same options (such that these actions could be performed in bulk).

willgearty:
This adds the records and payment from standard student check-in to the barcode check-in:

image

Fixes #3949


esp/templates/program/modules/onsitecheckinmodule/barcodecheckin.html
```diff
@@ -51,42 +56,61 @@ <h1>Barcode Checkin &mdash; For {{ program.niceName }}</h1>
...
-    {% endif %}
-    {% endcomment %}
+    <div class="alert alert-info">
+        <span class="glyphicon glyphicon-info-sign" aria-hidden="true"></span>
+        <span>
+        Checked in {{ results.new|length }} students.  {{ results.not_found|length }} students were not found, {{ results.existing|length }} students were already checked in, and {{ results.not_student|length }} IDs did not correspond to students.
+        </span>
+        {% if results.paid %}
+        <span>
+        {{ results.paid|length }} student(s) paid.
```
kkbrum: Would we want the payment and waivers blurbs to only count students who had not previously been fully paid or turned in waivers? Right now it will still count them even if they had already done everything. Given that the "attended" feature checks whether they had already been checked in, I feel like the payment and waivers should use parallel logic?

willgearty:
How's that look, @kkbrum?

PR #3959, opened 2026-01-07
Sanitize generated usernames for unregistered volunteers

### LINKED ISSUE #3954: Volunteer form makes invalid usernames for unregistered users
If a user is unregistered and signs up as a volunteer, we automatically make an account for them (I think this was originally to reduce the barrier for signing up as a volunteer):
https://github.com/learning-unlimited/ESP-Website/blob/fa8f2627366fd23a170cfee20178b36512c796fc/esp/esp/program/modules/forms/volunteer.py#L185-L186

However, if a user has a space in their last name (e.g., two last names), the username will also have a space in it 😬:
https://github.com/learning-unlimited/ESP-Website/blob/fa8f2627366fd23a170cfee20178b36512c796fc/esp/esp/users/models/__init__.py#L974-L975

This then can cause issues when parsing the username (e.g., this can permanently hold up the email processing cron job).

willgearty:
When an unregistered user signs up to be a volunteer, we automatically make a username for them. This ensures that that username is valid before saving it.

Fixes #3954.


willgearty:
I suppose we could also add a migration to fix any existing usernames that are invalid?

kkbrum:
Hmm... I have the program setting "require volunteers to be signed in" as off (the default), and yet when I go to http://localhost:8000/volunteer/Splash/2026_Winter/signup, it tells me I have to be logged in. Were you able to sign up as a volunteer without registering? If so, how?

kkbrum:
(turns out my deadline was closed, but the error redirects you to login)

kkbrum (approved):
looks good!

PR #3961, opened 2026-01-28
Added toggle button to hide incomplete finaid requests

### LINKED ISSUE #3950: (Option to) hide empty financial aid applications
If a student opens the financial aid application, but then leaves the page, we save a copy of their form for them in case they come back to fill in the rest later. However, if they never come back, we just have an uncomplete form submission sitting around and which is currently shown in the financial aid approval module. I think we probably want to hide these in the approval module (or have an easy button to click to hide them).

samalarussa:
Fixes #3950


kkbrum (approved):
looks good to me!

PR #3965, opened 2026-02-06
Add support for ARM architecture (silicon mac) dev servers

kkbrum:
1) adds instructions for developers to switch to the ARM version of the box I created. 
2) updates instructions for creating such a box, based on my experience.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -11,7 +11,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   config.vm.box = 'ubuntu-24.04'
   config.vm.box_url = 'https://learningu-static.s3.amazonaws.com/ubuntu-24.04.box'
   # config.vm.box = 'ubuntu-24.04_mac_arm_host.box'
-  # config.vm.box_url = 'https://learningu-static.s3.us-east-1.amazonaws.com/ubuntu-24.04_mac_arm_host.box'
+  # config.vm.box_url = 'https://learningu-static.s3.amazonaws.com/ubuntu-24.04_mac_arm_host.box'
   config.vm.hostname = 'ludev'
   config.ssh.forward_agent = true
   config.ssh.forward_x11 = true


docs/dev/vagrant.rst
```diff
@@ -52,6 +52,8 @@ Using a shell (such as Git Bash, which comes installed with Git), navigate to th
 If you already have a GitHub account with SSH keys set up, you may want to use ``git clone git@github.com:learning-unlimited/ESP-Website.git devsite`` to make it easy to push new code.
 If you already had vagrant installed, consider clearing your keys in ``~/.vagrant.d``.
 
+If you have a Mac with the ARM architecture (a silicon chip), open "Vagrantfile" in the devsite directory. Comment out lines 10 and 11 (by adding a "#" to the beginning of the lines) and uncomment lines 13 and 14. Save the file.
+
```
willgearty: I would specify the content of the lines instead of the line numbers in case we add things to the Vagrantfile.

docs/dev/vagrant.rst
```diff
@@ -219,17 +224,21 @@ Follow the following steps to upgrade the base VM for everyone to use.
...
+	iv. Run the virtual machine using the VirtualBox "Start" button, *not* by typing ``vagrant up`` in a terminal. If you are prompted, the username may be either "ubuntu" with no password or "vagrant" with password "vagrant". If there is an popup prompting you to try or install Ubuntu, choose the "Try" option. You may need to hit the escape key as soon as the virtual machine starts booting up, and then select boot manager, and then select ubuntu.
 
-	v. Once the desktop comes up, open a terminal window (should be in "Applications" in the bottom left corner). Run the following commands to get the names of the volume group (VG) and logical volume (LV)::
+	v. Once the desktop comes up, open a terminal window (should be in "Applications" in the bottom left corner). Run the following commands to get (and write down) the names of the volume group (VG) and logical volume (LV)::
 
 		sudo apt update && sudo apt -y upgrade
 		sudo apt install lvm2
 		sudo lvs
 
+	   You may need to then exit the terminal, restart the virtual machine, hit escape as it starts booting up, select boot manager, then select the CD-ROM, then select "try to install ubuntu".
```
willgearty: If you do this, do you need to then do something before "vi"? "vi" seems to imply you are in a terminal.

Vagrantfile
```diff
@@ -6,9 +6,12 @@ VAGRANTFILE_API_VERSION = '2'
 
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
 
-  # Build off the basic Ubuntu 64-bit VM.
+  # Build off the basic Ubuntu 64-bit VM. If you have a silicon Mac, comment out lines 10 and 11, and uncomment lines 13 and 14.
```
willgearty: Same note about the lines here, maybe say "the following two lines" instead?

Vagrantfile
```diff
@@ -6,9 +6,12 @@ VAGRANTFILE_API_VERSION = '2'
...
   config.vm.box = 'ubuntu-24.04'
   config.vm.box_url = 'https://learningu-static.s3.amazonaws.com/ubuntu-24.04.box'
+  # Build off the Ubuntu 64-bit VM for the ARM architecture (Mac silicon chip).
+  # config.vm.box = 'ubuntu-24.04_macos_host.box'
+  # config.vm.box_url = 'https://learningu-static.s3.us-east-1.amazonaws.com/ubuntu-24.04_macos_host.box'
```
willgearty: I can rename this to have "arm" in the filename if that would be better?

willgearty (changes requested):
Just a few comments here. I think it's more important to get @milescalabresi to review it since he wrote most of these docs.

Vagrantfile
kkbrum: up to you; doesn't matter to me!

willgearty: Renamed to ubuntu-24.04_mac_arm_host.box

Vagrantfile
```diff
@@ -6,12 +6,12 @@ VAGRANTFILE_API_VERSION = '2'
...
-  # Build off the basic Ubuntu 64-bit VM. If you have a silicon Mac, comment out lines 10 and 11, and uncomment lines 13 and 14.
+  # Build off the basic Ubuntu 64-bit VM by default. 
+  # If you have a silicon Mac (ARM architecture), comment out the following two lines and uncomment the two subsequent lines. 
   config.vm.box = 'ubuntu-24.04'
   config.vm.box_url = 'https://learningu-static.s3.amazonaws.com/ubuntu-24.04.box'
-  # Build off the Ubuntu 64-bit VM for the ARM architecture (Mac silicon chip).
-  # config.vm.box = 'ubuntu-24.04_macos_host.box'
-  # config.vm.box_url = 'https://learningu-static.s3.us-east-1.amazonaws.com/ubuntu-24.04_macos_host.box'
+  # config.vm.box = 'ubuntu-24.04_mac_arm_host.box'
+  # config.vm.box_url = 'https://learningu-static.s3.us-east-1.amazonaws.com/ubuntu-24.04_mac_arm_host.box'
```
willgearty: Looks like you can remove the "us-east-1" part and it still works.

hwatheod:
I got this error after running `fab setup`

```
[vagrant@127.0.0.1:2222] out: sudo: npm: command not found
[vagrant@127.0.0.1:2222] out: 

Fatal error: run() received nonzero return code 1 while executing!
```

I tried installing npm on the VM by ssh-ing into the VM (`vagrant ssh`) and then running [these commands](https://nodejs.org/en/download/current) on the VM. But it didn't help, still getting the same error.

kkbrum:
@willgearty it seems like this might be worth merging without Miles' review, since some of the new folks need the ARM box and are having trouble finding it. I'm not sure how to address the failed tests though. Thoughts?

willgearty:
I think I fixed it on main, so merging, we'll see if it works now.

willgearty:
Oh, actually, the tests shouldn't run anymore because you only changed docs and the Vagrantfile. Happy to merge. @milescalabresi can review them later.

PR #4000, opened 2026-02-19
Add tabindex to class flag create/edit form elements

### LINKED ISSUE #1358: Add HTML tabindex on class flag create/edit form
Right now, if you tab from the comment box, you don't get to the save/update button.  Putting `tabindex=foo` on the elements is apparently the way to fix this.

Oval17:
Resolves #1358 
Kindly Review @willgearty .
Thanks


kkbrum (changes requested):
After tabbing to the "save" button, it should tab to the "add flag" button. Right now it tabs to "save" as requested in the issue, but then it tabs to the home page logo. It previously tabbed to the "add flag" button as expected.

esp/templates/program/modules/classflagmodule/flag_detail.html
```diff
@@ -12,14 +12,14 @@
                     <td class="lastmod"> {{flag.modified_by.name}} on {{flag.modified_time.date}} at {{flag.modified_time.time}} </td>
                 </tr>
                 <tr>
+                    <td>
+                        <textarea id="flag-comment-{{flag.id}}" name="comment" rows="4" cols="60">{{flag.comment}}</textarea>
```
Oval17, line 17: corrected the DOM structure itself , instead of adding of tabindexes , works fine now, can be tested now!!

Oval17:
> After tabbing to the "save" button, it should tab to the "add flag" button. Right now it tabs to "save" as requested in the issue, but then it tabs to the home page logo. It previously tabbed to the "add flag" button as expected.

have corrected the DOM structure itself kindly see now!

kkbrum (changes requested):
While the tab order is good, the "comment" label and the comment box have switched sides now, and that doesn't make all that much sense. 

Image

My suggestion would be to switch the sides back, but put the "save" button below the table, or in its own row of the table, since it actually has little to do with saving the comment as opposed to saving the flag. 

Once a flag has been saved, I think the "update" and "remove" buttons can also be below the table or in a final row of the table. 

Image

The "add flag" button is not really part of the current flag, but to add a whole separate flag, so I would add some spacing beneath the flag table before the "add flag" button. This is especially relevant if you put the save, update, and remove buttons beneath the table.

Oval17:
Given about 10px gap for add flag and fixed comment box issue.
Thanks

kkbrum (approved):

PR #4012, opened 2026-02-19
Fix UnboundLocalError in LoginView.form_valid()

### LINKED ISSUE #3979: UnboundLocalError in esp/esp/users/views/__init__.py
Something is wrong with this code here because the `reply` variable on line 146 is referenced before it's being set. I'm not quite sure what the intended logic is.

https://github.com/learning-unlimited/ESP-Website/blob/38f914bd53336c999e07c6d408c7feefe041e04a/esp/esp/users/views/__init__.py#L142-L147

Maximus-08:
Description
Fixes an UnboundLocalError where the reply variable was referenced before assignment in form_valid().

As discussed in the issue, this logic was a remnant of the legacy function-based login view. In the current class-based LoginView, form_valid() is only invoked upon successful authentication. Therefore, we are guaranteed to be in the "success" path, and checking for [status_code == 302] is redundant.

Changes:
Changed the redundant elif reply.status_code == 302: to a standard else: block.
Added a code comment to explain the logic for future maintainability and clarity.
Fixes #3979

### DIFF
--- a/esp/esp/users/views/__init__.py
+++ b/esp/esp/users/views/__init__.py
@@ -143,7 +143,9 @@ class CustomLoginView(LoginView):
             reply = HttpMetaRedirect('/myesp/profile')
         elif next_url in mask_locations:
             reply = mask_redirect(user, next_url)
-        elif reply.status_code == 302:
+        else:
+            # form_valid() is only called on successful authentication, so we
+            # are always on the success path here — no need to check status_code.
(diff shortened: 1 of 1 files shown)


Maximus-08:
Hey @willgearty Please take a look at this when you're free

Maximus-08:
Hey @willgearty I rebased on top of the merged commits, can you please re-run the worklows

Issue #4032, opened 2026-02-20
perf: Fix N+1 database query patterns across the codebase


Arnav1709:
## Summary

Systematic identification and elimination of N+1 database query patterns across the ESP-Website codebase. Each fix is submitted as an independent PR targeting a specific module or function.

## Problem

Multiple views and utility functions execute database queries inside Python loops — the classic N+1 pattern. For pages handling hundreds of students, teachers, or sections, this results in thousands of redundant queries per request, causing slow page loads and unnecessary database load.

## Scope of Changes

| File | Function | Old (queries) | New (queries) | Reduction |
|------|----------|--------------|--------------|-----------|
| `usersearch.py` | hour computation | ~1000 (100 users) | 1 | 99.9% |
| [programprintables.py](cci:7://file:///home/arnav/Desktop/ESP-Website/esp/esp/program/modules/handlers/programprintables.py:0:0-0:0) | tag validation | N per tag | 1 | ~99% |
| [statistics.py](cci:7://file:///home/arnav/Desktop/ESP-Website/esp/esp/program/statistics.py:0:0-0:0) | [hours()](cci:1://file:///home/arnav/Desktop/ESP-Website/esp/esp/program/statistics.py:270:0-348:73) | ~8000 (500 students) | 2/program | 99.9% |
| [jsondatamodule.py](cci:7://file:///home/arnav/Desktop/ESP-Website/esp/esp/program/modules/handlers/jsondatamodule.py:0:0-0:0) | [sections_admin](cci:1://file:///home/arnav/Desktop/ESP-Website/esp/esp/program/modules/handlers/jsondatamodule.py:329:4-395:59) | N per section | 0 (prefetched) | 100% |
| [statistics.py](cci:7://file:///home/arnav/Desktop/ESP-Website/esp/esp/program/statistics.py:0:0-0:0) | [demographics()](cci:1://file:///home/arnav/Desktop/ESP-Website/esp/esp/program/statistics.py:91:0-142:80) finaid | ~3000 (500 students) | 3 | 99.9% |
| [jsondatamodule.py](cci:7://file:///home/arnav/Desktop/ESP-Website/esp/esp/program/modules/handlers/jsondatamodule.py:0:0-0:0) | 5 availability endpoints | 2N per teacher | 2/endpoint | ~99% |
[...]

PR #4034, opened 2026-02-20
perf: Query optimization in usersearch

Arnav1709:
## Summary
Fixes #4032 
Fixes a triple-nested N+1 query pattern in `usersearch.py` that fires when filtering students by enrolled hours.

## Problem

The hour-filtering logic in `query_from_criteria()` built a `user_hours` dict using a dict comprehension that, for every matching user, called `getEnrolledSections(program)` and then `section.meeting_times.count()` per section:

```python
user_hours = {
    user.id: sum([section.meeting_times.count()
                  for section in user.getEnrolledSections(program)])
    for user in ESPUser.objects.filter(current_Q)
}
```

For 100 users with an average of 5 sections each, this generates:

- 100 `getEnrolledSections()` queries
- 500 `meeting_times.count()` queries
- **~600 total queries**

## Fix

Replaced the entire dict comprehension and manual Python filtering with a single annotated query:

```python
annotated_users = ESPUser.objects.filter(current_Q).annotate(
    num_hours=Count(
        'studentregistration__section__meeting_times',
        filter=Q(
            studentregistration__section__parent_class__parent_program=program,
            studentregistration__relationship__name='Enrolled',
        ),
        distinct=True,
    )
)
```

The `hours_min` / `hours_max` filtering is then done via `.filter(num_hours__lt=...)` directly in SQL instead of Python loops.

## Impact

| Metric                              | Before                  | After                     |
| ----------------------------------- | ----------------------- | ------------------------- |
| Queries (100 users, 5 sections avg) | ~600                    | 1                         |
| Filtering                           | Python loop             | SQL                       |
| Memory                              | All user objects loaded | Only excluded IDs fetched |

## Files Changed

- `esp/esp/users/controllers/usersearch.py` — `query_from_criteria()` method


hwatheod (approved):
Looks good!

PR #4042, opened 2026-02-20
perf: replace demographics sections loop with SQL aggregation

Arnav1709:
## Summary

Replaces per-program Python loops with a single cross-program SQL aggregate query in the `demographics()` function for computing class/section counts and hours.

Fixes #4032

## Problem

The `demographics()` function ran 3 queries per program inside a loop, plus iterated over all sections in Python to sum hours:

```python
for program in programs:
    program.classes().select_related().filter(...).count()     # query 1
    program.sections().select_related().filter(...).count()    # query 2
    for section in program.sections().select_related().filter(...):  # query 3 + Python loop
        result_dict['num_class_hours'] += section.duration
        result_dict['num_student_class_hours'] += section.duration * section.parent_class.class_size_max
```

Additional issues:

- Bare `select_related()` follows all FK relationships (wasteful)
- The sections queryset is evaluated twice (once for `.count()`, once for iteration)
- All section + parent class objects are loaded into Python just to sum two fields

## Fix

Replaced everything with a single SQL aggregate query using `Count`, `Sum`, and `F` expressions:

```python
CS.objects.filter(
    parent_class__parent_program__in=programs,
    status=ClassStatus.ACCEPTED,
).aggregate(
    num_classes=Count('parent_class', distinct=True),
    num_sections=Count('id'),
    num_class_hours=Sum('duration'),
    num_student_class_hours=Sum(
        F('duration') * F('parent_class__class_size_max'),
        output_field=DecimalField(),
    ),
)
```

All four values are computed in one SQL query with no Python iteration and no object instantiation.

## Impact

| Metric                | Before                        | After       |
| --------------------- | ----------------------------- | ----------- |
| Queries (N programs)  | 3 × N + section iteration     | 1           |
| Python objects loaded | All sections + parent classes | None        |
| Computation           | Python loop sums              | SQL `SUM()` |

## Files Changed

- `esp/esp/program/statistics.py` — `demographics()` function, vitals section

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,29 @@
+name: "Lint"
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+concurrency:
+  group: lint-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+    - uses: actions/checkout@v4
+    - uses: actions/setup-python@v5
+      with:
+        python-version: '3.10'
+    - name: Install flake8
+      run: pip install flake8
+    - name: Run lint
+      run: deploy/lint
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,4 +1,4 @@
-name: "Lint and Unit Tests"
+name: "Unit Tests"
 
 on:
   push:
@@ -28,11 +28,16 @@ concurrency:
 permissions:
   contents: read
 
+
 jobs:
-  build:
-    name: ${{ matrix.ci_job }} (Python ${{ matrix.python-version }})
+  test:
+    name: Test (Python ${{ matrix.python-version }})
     runs-on: ubuntu-24.04
     timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [3.7]
     services:
       postgres:
         image: postgres:10.8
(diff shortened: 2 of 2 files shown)


esp/esp/program/statistics.py
```diff
@@ -91,17 +91,29 @@ def zipcodes(form, programs, students, profiles, result_dict={}):
...
-        result_dict['num_classes'] += program.classes().select_related().filter(status=ClassStatus.ACCEPTED).count()
-        result_dict['num_sections'] += program.sections().select_related().filter(status=ClassStatus.ACCEPTED).count()
-        for section in program.sections().select_related().filter(status=ClassStatus.ACCEPTED):
-            result_dict['num_class_hours'] += section.duration
-            result_dict['num_student_class_hours'] += section.duration*section.parent_class.class_size_max
-    result_dict['num_class_hours'] = int(result_dict['num_class_hours'])
-    result_dict['num_student_class_hours'] = int(result_dict['num_student_class_hours'])
+    #   Get aggregate 'vitals' info via cross-program SQL aggregation.
+    from django.db.models import Count, Sum, F, Value, DecimalField
+    from esp.program.models import ClassSection as CS
```
hwatheod: Please move imports to the top with the other imports.

hwatheod: Also, don't abbreviate ClassSection as CS.  It's less readable and could get confused with ClassSubject.

hwatheod:
Make fixes as requested, otherwise looks good.

esp/esp/program/statistics.py
Arnav1709: Done with both the changes. Thank you for suggestions. I will keep it in mind from next time

Arnav1709:
Done with all the changes.

Issue #4054, opened 2026-02-20
[Bug]: Add missing `transaction.atomic` to multi-step database operations


Arnav1709:
## Summary

Several functions across the codebase perform multiple database writes (`.save()`, `.delete()`, `bulk_create()`, `.update()`) without wrapping them in `transaction.atomic`. If any intermediate step fails, the database is left in an inconsistent state. This umbrella issue tracks adding proper transaction management to all identified locations.

## Background

Django's `transaction.atomic` ensures that a group of database operations either all succeed or all roll back. The codebase already uses `transaction.atomic` correctly in ~20 places (e.g., `submit_payment`, `record_payment_from_identifier`, `charge_payment`), but several critical paths are missing it.

## Affected Areas

| PR  | File                                                  | Functions                                                                                                 | Risk                                |
| --- | ----------------------------------------------------- | --------------------------------------------------------------------------------------------------------- | ----------------------------------- |
| #1  | `esp/program/views.py`                                | `lsr_submit`, `lsr_submit_HSSP`                                                                           | Partial student schedule on failure |
| #2  | `esp/accounting/controllers.py`                       | `clear_all_data`, `ensure_required_transfers`, `apply_preferences`, `set_preference`, `set_finaid_params` | Orphan financial records            |
| #3  | `esp/application/models.py`                           | `StudentClassApp.admit`                                                                                   | Partial admission state             |
| #4  | `esp/users/controllers/merge.py`                      | `merge_users`                                                                                             | Partially-merged users              |
| #5  | `esp/program/controllers/lottery.py`                  | `save_assignments`, `clear_saved_assignments`                                                             | Partial lottery results             |
| #6  | `esp/program/controllers/resources.py`                | `delete_classroom`                                                                                        | Orphan resources                    |
| #7  | `esp/program/models/class_.py`                        | `cancel`, `preregister_student`                                                                           | Inconsistent class/student state    |
| #8  | `esp/program/modules/handlers/onsitecheckinmodule.py` | `checkin`, `barcodecheckin`                                                                               | Partial check-in records            |
| #9  | `esp/users/models/forwarder.py`                       | `updateTarget`                                                                                            | Broken forwarder chain              |

## Design Decisions

- **Email sends excluded from transactions**: `cancel()` sends emails before DB mutations — those stay outside to avoid holding DB locks during I/O
- **External APIs excluded**: `save_assignments()` calls mailman after the transaction block
[...]

PR #4056, opened 2026-02-20
fix: add transaction.atomic to accounting controller methods

Arnav1709:
**Branch:** `fix/transaction-accounting`
**Closes:** Part of #4054 

## Summary

Wraps 5 methods in `esp/accounting/controllers.py` with `transaction.atomic` to ensure atomicity of financial operations.

## Changes

| Method                                                     | Change                            | Why                                                                                        |
| ---------------------------------------------------------- | --------------------------------- | ------------------------------------------------------------------------------------------ |
| `ProgramAccountingController.clear_all_data`               | `with transaction.atomic()` block | Deletes grants, transfers, line items, and accounts — partial delete leaves orphan records |
| `IndividualAccountingController.ensure_required_transfers` | `@transaction.atomic` decorator   | Creates/updates multiple Transfer objects in a loop                                        |
| `IndividualAccountingController.apply_preferences`         | `@transaction.atomic` decorator   | Deletes existing transfers then creates new ones                                           |
| `IndividualAccountingController.set_preference`            | `@transaction.atomic` decorator   | Deletes old transfers, creates replacement                                                 |
| `IndividualAccountingController.set_finaid_params`         | `@transaction.atomic` decorator   | Creates/modifies financial aid grant                                                       |

## Risk Assessment

**Low** — `transaction` was already imported in this file. `submit_payment` and `record_payment_from_identifier` already use `@transaction.atomic` correctly, so this aligns with existing patterns.


esp/esp/accounting/controllers.py
```diff
@@ -83,10 +83,11 @@ def __init__(self, program, *args, **kwargs):
 
     def clear_all_data(self):
         #   Clear all financial data for the program
-        FinancialAidGrant.objects.filter(request__program=self.program).delete()
-        self.all_transfers().delete()
-        self.get_lineitemtypes().delete()
-        self.all_accounts().delete()
+        with transaction.atomic():
```
willgearty: Since this covers the entire function, you should just use the decorator on the function, right?

willgearty (changes requested):
Looks good, just one question.

esp/esp/accounting/controllers.py
Arnav1709: You're right, since it covers the entire function body, using the @transaction.atomic decorator is more consistent with the other methods. Thank you for reviewing. I have updated the code

PR #4058, opened 2026-02-20
ci: split lint and test into separate jobs

### LINKED ISSUE #4057: [Feature]: Split lint and test into separate CI jobs
### Problem Statement

The current CI workflow uses a single `build` job with a matrix strategy to run both `lint` and `test`. However, the lint job doesn't need PostgreSQL, Memcached, or any system packages — it just runs `flake8` to check code style.

Because both jobs share the same configuration, the lint job:
- Starts PostgreSQL container (unnecessary)
- Starts Memcached container (unnecessary)
- Waits for service healthchecks (wasted time)
- Runs on `ubuntu-24.04` with heavy setup (overkill)

This adds ~60-90 seconds of unnecessary overhead to every lint run.

## Proposed Solution

Split into two separate jobs:

**Lint job (lightweight):**
- Uses `ubuntu-latest`
- Uses Python 3.10 via `actions/setup-python`
- No services needed
- Only installs flake8
- 10 minute timeout

**Test job (full setup):**
- Uses `ubuntu-24.04`
- Uses Python 3.7 via deadsnakes
- PostgreSQL and Memcached services
- Full dependency installation
- 30 minute timeout

Jobs run in parallel for faster overall feedback.

## Alternatives Considered

- **Keep matrix but conditionally start services** — Not supported by GitHub Actions
- **Use job-level `if` conditions** — Still starts the runner with all config
- **Keep as-is** — Works but wastes compute resources

## Additional Context

- Lint will complete in ~30 seconds vs ~90+ seconds currently
- Total wall-clock time reduced due to parallel execution
- Clearer failure messages ("lint failed" vs "test failed")
- Affected file: `.github/workflows/tests.yml`

Arnav1709:
## Description

Splits the CI workflow into two separate jobs: a lightweight `lint` job and a full `test` job. Currently, both run in a shared `build` job with PostgreSQL and Memcached services, even though lint doesn't need any of these.

**Before:** Single `build` job with matrix → lint waits for unnecessary service startup
**After:** Separate `lint` and `test` jobs → lint runs fast, both execute in parallel

## Related Issue

Closes #4057

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [x] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

## Checklist

- [x] Self-reviewed the code

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -28,14 +28,16 @@ concurrency:
 permissions:
   contents: read
 
-env:
-  PYTHON_VERSION: '3.7'
 
 jobs:
   test:
-    name: Test (Python 3.7)
+    name: Test (Python ${{ matrix.python-version }})
     runs-on: ubuntu-24.04
     timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [3.7]
     services:
       postgres:
         image: postgres:10.8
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -15,8 +15,22 @@ permissions:
...
+    timeout-minutes: 10
+    steps:
+    - uses: actions/checkout@v4
+    - uses: actions/setup-python@v5
+      with:
+        python-version: '3.10'
+    - name: Install flake8
+      run: pip install flake8
+    - name: Run lint
+      run: deploy/lint
```
willgearty: At this point, should we just move this to a separate file?

.github/workflows/tests.yml
```diff
@@ -34,49 +48,34 @@ jobs:
...
+      PYTHON_VERSION: 3.7
 
     steps:
     - uses: actions/checkout@v4
-    - name: Set up Python ${{ matrix.python-version }}
+    - name: Set up Python 3.7
       uses: deadsnakes/action@v3.2.0
       with:
-        python-version: ${{ matrix.python-version }}
+        python-version: 3.7
```
willgearty: Could you use the environment variable for instances of 3.7 so we only have to change the version number in one place in the future?

willgearty (changes requested):
Looks good, just a couple requests. Thanks!

.github/workflows/tests.yml
Arnav1709: Done. Moved lint to .github/workflows/lint.yml with its own concurrency group.

.github/workflows/tests.yml
Arnav1709: I have done the requested changes

Arnav1709: The job name field doesn't support ${{ env }} context in GitHub Actions (only matrix context is available at the job level), so the version is hardcoded in the display name only. All functional references (setup-python, cache keys, etc.) use ${{ env.PYTHON_VERSION }} as requested.

willgearty: I think I would just go back to using
```
matrix:
  python-version: [3.7]
```
I know it seems weird to use a matrix just for a single version, but it's possible we'll want to test on multiple versions in the version.

Arnav1709: done

willgearty (approved):

PR #4059, opened 2026-02-20
Add per-program editable link title to ProgramModuleObj (#3849)

### LINKED ISSUE #3849: Make reg module link title editable
There should be a way to make the link title (the text that shows up in the registration checklist) editable with the following:

1. The title should be program-specific (this likely will require a new model field on the program reg module obj class)
2. If no title is set, it should default to the module's link title (which should no longer be editable in the admin pages)

imran-techvoyager:
## Description

Adds a per-program `link_title` override on `ProgramModuleObj`.

If unset, behavior falls back to the module’s default `link_title`, preserving existing behavior.

Changes:
- Added nullable `link_title` field to `ProgramModuleObj`
- Added `get_link_title()` helper for clean fallback logic
- Updated `makeLink()` and `get_setup_title()` to use override
- Exposed field in `ProgramModuleObjAdmin`
- Added migration

## Related Issue

Closes #3849

## Type of Change

- [x] New feature

## Testing

- Manually verified:
  - Default behavior unchanged when override blank
  - Custom title renders correctly in registration checklist

### DIFF
--- a/esp/templates/program/modules/admincore/modules.html
+++ b/esp/templates/program/modules/admincore/modules.html
@@ -22,7 +22,7 @@
 
 <h1>Module Management for {{ program.niceName }}</h1>
 
-<p>Welcome to the module management page for {{ program.niceName }}. This page lists the modules that students and teachers will encounter during registration. You can drag individual modules to reorder them and change whether they are required or not. You can also click on a module name to edit the requirement label that will be shown in parentheses next to the module title in the student/teacher registration to-do list, or the link title override that will be shown instead of the module's default title. To add or remove modules, you will need to enable/disable them in the <a href="/manage/{{ program.getUrlBase }}/settings/program">program settings</a>.</p>
+<p>Welcome to the module management page for {{ program.niceName }}. This page lists the modules that students and teachers will encounter during registration. You can drag individual modules to reorder them and change whether they are required or not. You can also click on a module name to edit the Registration Step Title that will be shown instead of the module's default title in the student/teacher registration to-do list, or the requirement label that will be shown in parentheses next to the title. To add or remove modules, you will need to enable/disable them in the <a href="/manage/{{ program.getUrlBase }}/settings/program">program settings</a>.</p>
 {% inline_program_qsd_block program "manage:modules" %}
 {% end_inline_program_qsd_block %}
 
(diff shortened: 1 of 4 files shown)


imran-techvoyager:
Hi @willgearty just a quick follow-up on this one when you get a chance.

I believe the only blocker right now is the CI workflow approval for forked PRs. Let me know if you’d like me to make any adjustments or add test coverage.

Happy to revise as needed.

willgearty:
We've got quite a backlog of PRs to review, which we probably won't get to until next week. Sorry for the delay, but we're a small core team (at the moment 😉)!

willgearty:
Just took a quick look, and it looks good. We're currently trying to avoid having users need to use the admin pages (they tend to find it scary AND it often results in lots of headaches for us). Could you also integrate this into the Module Management page?

imran-techvoyager:
Thanks, that makes sense.
I’ll integrate this into the Module Management page so users don’t need the admin interface.
Planning to extend the existing edit form rather than introduce a new flow.

imran-techvoyager:
Hi @willgearty just following up on this when you have a chance. Would you mind taking a look and letting me know your thoughts?

Happy to address any feedback.

esp/templates/program/modules/admincore/modules.html
```diff
@@ -55,9 +55,10 @@ <h3>
                 </h3>
                 <ol id="sortable1" class="connectedSortable">
                 {% for mod in learn_modules.required %}
-                    <li id='{{ mod.id }}' class="ui-state-default{% if mod.required_label %} has_label{% endif %}">
+                    <li id='{{ mod.id }}' class="ui-state-default{% if mod.required_label %} has_label{% endif %}{% if mod.link_title %} has_link_title{% endif %}">
                         {{ mod.module.admin_title }}<br/>
-                        <input type="text" name="{{ mod.id }}_label" value="{{ mod.required_label }}" placeholder="(requirement label)"{% if not mod.required_label %} style="display: none;"{% endif %}>
+                        <input type="text" class="label_input" name="{{ mod.id }}_label" value="{{ mod.required_label }}" placeholder="(requirement label)"{% if not mod.required_label %} style="display: none;"{% endif %}>
+                        <input type="text" class="link_title_input" name="{{ mod.id }}_link_title" value="{{ mod.link_title }}" placeholder="(link title override, default: {{ mod.module.link_title }})"{% if not mod.link_title %} style="display: none;"{% endif %}>
```
willgearty: Since the link title comes before the required label in the actual checklist, you should put it first here as well.

willgearty: Also, since the titles are often much longer, this should be a longer input box, probably on it's own line.

willgearty: Finally, I don't think most admins know what the "link title" is, so something like "Registration Step Title" would be better. (same for the text at the top of the page)

imran-techvoyager: Hi @willgearty I’ve pushed commit 0af37db addressing all review comments:

- Moved Registration Step Title before the required label

- Rendered it as a full-width input on its own line

- Renamed “link title” → Registration Step Title throughout (including page text)

- Updated the affected tests to reflect the new behavior

CI is running now. Please let me know if anything else should be adjusted once it’s green.
Thanks again for the detailed feedback!

imran-techvoyager:
> Looks fantastic, thanks so much for your contribution and for addressing my review!

Thanks so much for the review and guidance throughout this, I really appreciate it!
Glad it meets expectations. Happy to help with anything else.

PR #4061, opened 2026-02-20
fix: make merge_users fully atomic with transaction.atomic

Arnav1709:
**Branch:** `fix/transaction-user-merge`
**Closes:** Part of #4054 
## Summary

Moves transaction handling from per-object in `merge()` to the top-level `merge_users()` function, making the entire user merge atomic.

## Context

Previously, each related object transfer had its own `with transaction.atomic()` block inside a loop. This meant a failure partway through would leave a partially-merged user — some relations pointing to the absorber, others still pointing to the absorbee.

## Changes

- Added `@transaction.atomic` decorator to `merge_users()`
- Removed per-object `with transaction.atomic()` from `merge()`
- Kept `IntegrityError` handling per-object (via try/except) for duplicate constraint handling

```diff
 def merge(absorber, absorbee):
     """Transfers everything from absorbee to absorber."""
     for obj, name, m2m in get_related(absorbee):
-        with transaction.atomic():
+        try:
             if m2m:
                 ...
             else:
                 setattr(obj, name, absorber)
                 obj.save()
+        except IntegrityError:
+            pass
```

```diff
+@transaction.atomic
 def merge_users(absorber, absorbee, forward=True, deactivate=False):
```

## Risk Assessment

**Medium** — This changes the transaction scope from per-object to full-merge. The `IntegrityError` catch ensures duplicate constraints are still handled gracefully within the atomic block.


esp/esp/users/controllers/merge.py
```diff
@@ -66,6 +63,8 @@ def merge(absorber, absorbee):
             else:
                 setattr(obj, name, absorber)
                 obj.save()
+        except IntegrityError:
+            pass
```
willgearty: Why do we just pass if there's an integrity error? Shouldn't we log the error? Or even stop the entire process with an error page?

Arnav1709: I've updated the PR to log IntegrityErrors with logger.warning(), including the object class, PK, field name, and both user IDs. I also kept with transaction.atomic() as a per-object savepoint inside the try/except, since without it an IntegrityError would poison the outer @transaction.atomic and break all subsequent operations. IntegrityErrors here are expected when both users share the same related object (e.g., both enrolled in the same class), so stopping the process would prevent legitimate merges from completing.

willgearty: Sounds like it would be better to handle symmetric relations properly like the TODO says

Arnav1709: 1) For reverse M2M: if the related object is the absorber itself (which happens with symmetric fields like "friends"), we now only remove the absorbee without adding a self-referential link
2) For forward M2M: self-referential model fields now exclude the absorber to prevent a user being related to themselves after merge

willgearty:
Looks good, thanks!

PR #4078, opened 2026-02-20
perf: consolidate chained .filter().filter() into single .filter()

### LINKED ISSUE #4077: [Bug]: Consolidate chained `.filter().filter()` to avoid unnecessary JOINs
### Description
Chaining `.filter()` on relations can cause Django to produce unnecessary extra JOINs compared to a single `.filter(field1=val1, field2=val2)`. This results in slower queries, especially on large tables.
### Pattern
```diff
- queryset.filter(field1=val1).filter(field2=val2)
+ queryset.filter(field1=val1, field2=val2)

Arnav1709:
### Summary
Fixes #4077 
Consolidates chained `.filter().filter()` calls into single `.filter(field1=val1, field2=val2)` calls to avoid unnecessary extra JOINs generated by Django's ORM.
### Motivation
When chaining `.filter()` on relations, Django may produce a separate JOIN for each `.filter()` call. Consolidating the conditions into a single `.filter()` allows Django to produce a single JOIN instead, improving query performance.
### Changes
| File | Change |
|------|--------|
| `esp/esp/web/views/bio.py` | Merged `.filter(status__gte=10)...filter(sections__resourceassignment...)` into single `.filter()` |
| `esp/esp/resources/models.py` | Merged `.filter(base_q).filter(name__icontains=label)` into `.filter(base_q, name__icontains=label)` |
| `esp/esp/program/models/__init__.py` | Merged `.filter(section__parent_class__parent_program=self).filter(user__id__in=..., relationship=...)` |
| `esp/esp/program/modules/handlers/jsondatamodule.py` | Merged `.filter(category__category="Lunch").filter(status=ACCEPTED)` |
### Skipped
- `program/models/class_.py` line 1253 — inside a comment/docstring block, not executable code
### Testing
This is a purely mechanical refactor. The consolidated `.filter()` calls produce identical results to the chained version.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@ esp/file_db/*
 esp/esp/database_settings.py.chicago
 esp/esp/database_settings.py.mit
 env
+venv
 useful_logs
 theme_compiled.css
 /.vagrant/*
--- a/docs/admin/program_modules.rst
+++ b/docs/admin/program_modules.rst
@@ -283,7 +283,7 @@ Formstack Application Module
 
 This is the module that embeds a Formstack form on a student-facing page for
 student applications.  For more information, see
-`</docs/admin/student_apps.rst>`_.
+`<student_apps.rst>`_.
 
 Class Change Request Module (ClassChangeRequest)
 ------------------------------------------------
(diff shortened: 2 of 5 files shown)


hwatheod:
Don't merge yet.  I'm reviewing the issue [discussed here](https://docs.djangoproject.com/en/1.11/topics/db/queries/#spanning-multi-valued-relationships) about chained filters

hwatheod:
The difference between chaining two filters versus putting two conditions in one filter is discussed in the Django documentation (linked in my previous comment). It only applies to ManyToMany and reverse ForeignKey. We don't have any those in the queries in this PR, except in the `bio.py` change, the `sections` is many-to-many. But in that case, the other filter is `status` which is just a single number attached to a `ClassSubject` so there is no problem.

PR #4117, opened 2026-02-20
fix(program): allow admin deletion without manual resource cleanup

### LINKED ISSUE #3027: Deleting a program requires deleting resources first
When you go to delete a program (through the admin pages), it first requires you to delete (some of) the resources. This is weird because it then goes ahead and deletes all of the other database objects for you, but it seems to think (some of) the resources for the program are "protected"/"secured"?

Experienced on dev3

aryan7081:
Fixes #3027

Fixes program deletion in admin when program-linked resources/resource requests exist and were triggering protected-delete errors.
Adds cleanup in `Program.delete() `and a regression test to ensure program removal works end-to-end.


hwatheod:
For this issue, I think it might be better to ask why we are using `PROTECT` in the first place. What is it protecting from?  From a quick search of the codebase, I only see [3 instances](https://github.com/search?q=repo%3Alearning-unlimited%2FESP-Website%20models.PROTECT&type=code) where we use `models.PROTECT`. It is unclear to me what is special about those specific instances that they need `PROTECT`.  

If we don't really need `PROTECT`, then we should just change it to something else and avoid adding a bunch of additional code that is just deleting the "protected" resources anyway.

aryan7081:
@hwatheod Done. `Resource/ResourceRequest` now use CASCADE on res_type (migration added) and the custom `Program.delete()` has been removed so deletion is handled by normal cascades.

> It is unclear to me what is special about those specific instances that they need PROTECT

PROTECT was originally there to block deleting a ResourceType that’s still in use, but since we were already deleting those resources in `Program.delete()` when removing a program, we’re dropping `PROTECT` and the manual cleanup and using CASCADE instead. The only remaining PROTECT is on `Transfer.paid_in`, with a short help_text explaining it’s intentional, so we don’t delete a payment transfer that fee transfers still reference.

willgearty:
@aryan7081 please address the merge conflicts. Then hopefully @hwatheod can rereview this.

aryan7081:
@hwatheod This PR is ready for review. Sorry for the delay. Please review when you have time.
Thankyou

kkbrum:
@aryan7081 can you fix the failing test?

aryan7081:
@kkbrum Working on it.

hwatheod (approved):

PR #4118, opened 2026-02-20
fix(web): route unknown complex paths to fallback page

### LINKED ISSUE #3892: Unknown program page doesn't work for complex URLs
The Unknown Program page (e.g., https://chicago.learningu.org) doesn't work for anything that has an even slightly more complex URL path (e.g., https://chicago.learningu.org/test)

aryan7081:
Fixes #3892 

adding a fallback route for unmatched complex URLs (for example `/test`) so unknown program domains render the landing/unknown-program page instead of failing route resolution.

Includes a regression test to verify complex path fallback behavior.

### DIFF
--- a/esp/esp/urls.py
+++ b/esp/esp/urls.py
@@ -162,3 +162,8 @@ urlpatterns += [
     url(r'^manage/templateoverride/(?P<template_id>[0-9]+)',
         esp.utils.views.diff_templateoverride),
 ]
+
+# Keep unknown program domains functional for deeper paths
+urlpatterns += [
+    url(r'^(?P<path>.+)$', main.unknown_program_fallback),
+]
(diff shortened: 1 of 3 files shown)


willgearty:
If I understand correctly, this makes any pages that currently would give a 404 instead redirect to the home page? If so, I think we want to keep the 404 error. Instead, I think this requires a change in our server infrastructure which @milescalabresi should do.

milescalabresi:
I'm not sure I follow. Isn't this change what you requested in #3892?

willgearty:
No, I think the issue I was posting in #3892 was that links like this should direct to the not-a-program page: https://notaprogram.learningu.org/test.

Whereas something like this should just use our built-in 404 handler: https://yale.learningu.org/notapage (as it already is)

aryan7081:
@willgearty @milescalabresi Can you give clarification on what is expected behaviour here?? So that I can update this PR accordingly

willgearty:
The fix for #3892 needs to be done in an apache config file on the server, so a PR in this repo isn't necessary.

PR #4119, opened 2026-02-20
feat(droplets): add separate page and main background colors

### LINKED ISSUE #3939: Main vs. background colors for droplets
Should we add the ability to have separate colors for the background on the page vs. the background of the main content in the droplets theme?

aryan7081:
Fixes #3939 

Adds two new droplet theme settings to control background colours independently: one for the full page and one for the main content area.
Defaults preserve current behaviour, so existing sites keep the same look until customised.

**Quick UI test checklist**

1. Go to /themes/select/ and choose droplets (if not already active).
2. Open /themes/setup/ and set different values for:
-     page_background_color (e.g. #f0f4ff)
-     main_content_background_color (e.g. #ffffff)

3. Save and open homepage + one inner page.
4. Verify:
-     page/outside area uses page color,
-     main content panel uses main content color,
-     no layout break on mobile/desktop widths.


kkbrum (changes requested):
for color customization, we use themes/customize/, not themes/setup/. Could you move the customization to the theme editor form, as opposed to the config?

kkbrum:
@aryan7081 Can you please take a look at this when you get a chance so we can get this PR merged?

aryan7081:
@kkbrum Sorry for the delay. Background colours are now in the theme editor.
Thank you!

kkbrum:
Do you get this error when you go to /themes/customize? I do not get this error on the main branch. Is there something I need to do?

Screenshot 2026-03-15 at 10 03 43 PM

aryan7081:
@kkbrum The branch changed how lessc is invoked switched from `shell=True` to `shell=False`, which can break on some setups. I’ve reverted that change to match main, so `/themes/customize` should work again. Please pull the latest and try again.

kkbrum (changes requested):
It's still not compiling for me.

Image

aryan7081:
@kkbrum Sorry for the delay. I investigated and found out that the root cause was that `droplets/less/variables.less` was using `@bodyBackground`. I have replaced it with the concrete value. I have tested it locally, everything is working as expected. I hope that it will work the same on your system as well.
Thankyou

kkbrum (changes requested):
1) The content background should be in the "scaffolding" section of the editor, not its own section.
2) When I "change" the background color and save, it does not save; it just reverts to the same background color as if I did not select anything.

kkbrum:
Closing due to inactivity.

PR #4127, opened 2026-02-20
fix(mailgate): prevent class-list recursion from BCC/CC aliases

### LINKED ISSUE #3172: bcc'ing class lists causes email recursion

aryan7081:
Fixes #3172

Fixes email recursion when class list aliases appear in recipient headers by sanitizing self-referential list addresses before forwarding and dropping Bcc on forwarded mail.
Adds regression tests covering self-address removal and forwarded-mail Bcc handling.


willgearty:
We probably don't want to touch mailgate.py until #3832 is merged.

willgearty:
Superseded by #3832

PR #4172, opened 2026-02-21
fix(qsd): replace 403 errors with graceful message redirect

### LINKED ISSUE #4168: [Bug]: Quasi-Static Data (QSD) Authorization Failures Result in Hard 403 Pages
### Description

When a user without edit permissions attempts to edit a Quasi-Static Data (QSD) page (e.g., by directly navigating to the `.edit.html` URL), the server raises a raw [Http403](cci:2://file:///c:/Users/Lenovo/devsite/esp/esp/middleware/esperrormiddleware.py:53:0-54:8) exception. This results in the user seeing a harsh "Access Denied" error page instead of a graceful UI error. There's an existing `FIXME: SHOW A REAL ERROR!` comment in the code ([esp/esp/qsd/views.py](cci:7://file:///c:/Users/Lenovo/devsite/esp/esp/qsd/views.py:0:0-0:0)) indicating that this mechanism needs to be updated.

### Steps to Reproduce

1. Log into the ESP website using a regular user account that doesn't have administrative/edit privileges for QSD pages.
2. Navigate to a known QSD page (e.g., `/learn/bar.html`).
3. Manually modify the URL in the address bar to intentionally append `.edit.html` (e.g., `/learn/bar.edit.html`) or submit an edit form manually.
4. Observe the page resulting in an "Access Denied" HTTP 403 page rather than gracefully handling the unauthorized state.

### Expected Behavior

The application should gracefully catch the lack of edit permission and use Django's `messages.error` framework to redirect the user back to the read-only page ([.html](cci:7://file:///c:/Users/Lenovo/devsite/esp/templates/403.html:0:0-0:0)) while displaying a flash message like: *"You don't have permission to edit this page."*

### Actual Behavior

The application throws an [Http403("You don't have permission to edit this page.")](cci:2://file:///c:/Users/Lenovo/devsite/esp/esp/middleware/esperrormiddleware.py:53:0-54:8) exception, completely halting the flow and displaying the generic [403.html](cci:7://file:///c:/Users/Lenovo/devsite/esp/templates/403.html:0:0-0:0) system template.

### Screenshots

_No response_

### Operating System

_No response_

### Browser

_No response_

### Additional Context

The exact locations to fix are in [esp/esp/qsd/views.py](cci:7://file:///c:/Users/Lenovo/devsite/esp/esp/qsd/views.py:0:0-0:0):
- In the [qsd()](cci:1://file:///c:/Users/Lenovo/devsite/esp/esp/qsd/views.py:66:0-214:45) view, roughly around line 155 (under `# Detect POST`): `raise Http403("Sorry, you do not have permission to edit this page.")`
- In the [qsd()](cci:1://file:///c:/Users/Lenovo/devsite/esp/esp/qsd/views.py:66:0-214:45) view, roughly around line 192, where the `# Enforce authorizations (FIXME: SHOW A REAL ERROR!)` comment is located.

Rahul-R79:
## Description

This PR fixes a bug where users without edit permissions attempting to access a QSD `.edit.html` page would be met with a raw [Http403](cci:2://file:///c:/Users/Lenovo/devsite/esp/esp/middleware/esperrormiddleware.py:53:0-54:8) exception and a harsh "Access Denied" error page. The fix replaces the exception with Django's `messages.error` framework to display a graceful UI error ("You don't have permission to edit this page.") and redirect the user smoothly back to the standard read-only [.html](cci:7://file:///c:/Users/Lenovo/devsite/esp/templates/403.html:0:0-0:0) page. 

The redirect uses relative pathing (`../{page}.html`) to guarantee compatibility across sub-directory production environments.

## Related Issue

Closes #4168 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

## AI Disclosure

N/A

## Checklist

- [x] Self-reviewed the code

### DIFF
--- a/.github/scripts/project-tracker/update_project_sheet.py
+++ b/.github/scripts/project-tracker/update_project_sheet.py
@@ -295,12 +295,43 @@ def build_issue_rows(issues: list[dict]) -> list[list[str]]:
     return rows
 
 
+def fetch_merged_pr_counts(logins: list[str]) -> dict[str, int]:
+    """
+    Fetch the number of merged PRs per contributor using batched GraphQL
+    search queries (up to 20 per request to stay within complexity limits).
+    """
+    counts: dict[str, int] = {}
+    batch_size = 20
+
+    for i in range(0, len(logins), batch_size):
+        batch = logins[i : i + batch_size]
+        # Build aliased search queries
+        fragments = []
+        for j, login in enumerate(batch):
+            q = f"repo:{REPO_OWNER}/{REPO_NAME} is:pr is:merged author:{login}"
+            fragments.append(
+                f'u{j}: search(query: "{q}", type: ISSUE) {{ issueCount }}'
+            )
+        query = "query {\n" + "\n".join(fragments) + "\n}"
+        data = graphql(query)
+        for j, login in enumerate(batch):
+            counts[login] = data[f"u{j}"]["issueCount"]
+
+    return counts
+
+
+def link_count(url: str, count: int) -> str:
+    """Google Sheets HYPERLINK formula displaying a count."""
+    return f'=HYPERLINK("{url}", {count})'
+
+
 def build_contributor_rows(
     prs: list[dict], issues: list[dict]
 ) -> list[list[str]]:
     """Build the rows for the Contributor Activity sheet."""
     header = [
-        "Contributor", "Open PRs", "Assigned Open Issues", "Last Activity",
+        "Contributor", "Open PRs", "Merged PRs",
+        "Assigned Open Issues", "Last Activity",
     ]
 
     activity: dict[str, dict] = defaultdict(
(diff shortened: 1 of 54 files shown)


willgearty:

willgearty:
Do we need to alter any templates to make sure the message is shown?

willgearty:
@Rahul-R79 Do we need to alter any templates to make sure the message is shown?

Also, please address the Copilot review before formal review can take place.

Rahul-R79:
Yes, good catch! It turns out there was no global mechanism in the base template to render `messages`, meaning the flashed QSD warning would have just silently sat in the user's session without ever being displayed.

I just pushed a fix for this as well. I've added a global block to render Django's `messages` directly into [templates/main.html](cci:7://file:///c:/Users/Lenovo/devsite/esp/templates/main.html:0:0-0:0) right above the main content area, so they should now properly display as flash alerts across the site.

and fixed the copilot suggested things now its ready for a review @willgearty

willgearty:
If you're making changes to templates/main.html, you'll also need to add those to the theme-specific templates/main.html as well.

Rahul-R79:
Done! I've added the flash message block to the `main.html` templates for all of the specific themes (**barebones, bigpicture, circles, droplets, floaty, and fruitsalad**). I've also verified that the theme tests are fully passing locally @willgearty

esp/esp/qsd/views.py
willgearty: I think you need to update this one as well.

Rahul-R79: I will look into this tomorrow. Thanks for your understanding!

willgearty:
Thanks for adding that test!

Should we redirect the user if they try to edit a nonexistent QSD page? Thoughts @Rahul-R79? And, if so, where do we redirect them? The homepage?

Rahul-R79:
Currently, if an unauthorized user tries to edit a nonexistent page, they get a 403. I think the cleanest approach would be to just return a 404 instead — from the perspective of a user without edit permission, there's no reason to distinguish "this page doesn't exist" from "you can't edit this page." It also avoids leaking information about editability. What do you think?

@willgearty

willgearty:
Sure, that sounds like a good solution to me @Rahul-R79 👍

Rahul-R79:
Done @willgearty

Rahul-R79:
My plessure :)

PR #4180, opened 2026-02-21
Fix: admin dashboard student accepted count

### LINKED ISSUE #1681: Dashboard text still shows accepted student stat when not using applications
The admin dashboard still shows "Students who are accepted to at least one class. - 0" even if the application module isn't being used. This is can be confusing/ambiguous to those who don't know what that is referring to,

Rahul-R79:
## Description

Hide the "Students who are accepted to at least one class" stat from the admin dashboard when the program is not using student applications. Previously this always showed as `- 0`, which was confusing when the application module wasn't in use.

## Related Issue

Closes #1681 

## Type of Change

- [x] Bug fix

## Testing

- [x] Manually verified

## Checklist

- [x] Self-reviewed the code


hwatheod:
It looks like this code is already in the application review module (`AdminReviewApps`), and also, I don't see it shown on the dashboard on my dev server. I think this issue may already be fixed.

To be sure, I created and scheduled a class for this program, and added a student to the class.

Screenshot 2026-02-22 at 23 52 23


@willgearty to confirm.

willgearty:
Good catch @hwatheod. Looks like this was already fixed and these stats should now only be included in the dashboard if the review apps module is enabled. Thanks, though @Rahul-R79 !

willgearty:
Oh, just reread over the issue, and I think the suggestion was to move these stats to the Student Application module, just to make sure that they don't show up even if the Admin Apps module is enabled?

willgearty:
@Rahul-R79 the accepted students stat should be moved to the student application module

willgearty:
@Rahul-R79 any update on this?

willgearty (approved):

PR #4181, opened 2026-02-21
Refactor: apply preferences use set preference

### LINKED ISSUE #1318: Reduce code duplication in extra costs part of accounting controller
The behavior of `apply_preferences()` and `set_preference()` is basically the same. The only difference is that the former sets preferences for all optional costs at once and deletes all optional `Transfer` objects for a student for a program, whereas the latter sets a single preference and only deletes old `Transfer` objects for that preference for that student.

We can reduce code by making `apply_preferences()` call perform the same delete it is doing now and then call `set_preference()` a bunch of times.

Rahul-R79:
## Description

Refactors [apply_preferences()](cci:1://file:///c:/Users/Lenovo/devsite/esp/esp/accounting/controllers.py:326:4-346:21) in [Individual AccountingController](cci:2://file:///c:/Users/Lenovo/devsite/esp/esp/accounting/controllers.py:278:0-646:86) to eliminate duplicated code. It now performs its bulk delete of all optional transfers as before, then delegates transfer creation for each item to [set_preference()](cci:1://file:///c:/Users/Lenovo/devsite/esp/esp/accounting/controllers.py:348:4-374:21) instead of reimplementing the same logic inline.

## Related Issue

Closes #1318 

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

## Checklist

- [x] Self-reviewed the code


Rahul-R79:
@hwatheod please take a look into this if you have some time thankyou!

hwatheod (approved):
Thank you for doing this!

PR #4152, opened 2026-02-20
fix: replace bare except clauses with specific exception types

SuyashJain17:
## Description

This PR replaces all bare except: clauses across 56 files with specific exception types.

Bare except: blocks catch all exceptions — including SystemExit, KeyboardInterrupt, and GeneratorExit — which can mask real errors and make debugging difficult. Each replacement was chosen based on the operations inside the corresponding try block (e.g., ValueError for conversions, DoesNotExist for ORM lookups, KeyError/IndexError for collection access, etc.).

This improves reliability and maintainability without changing intended behavior.

## Related Issue #4108

Closes #4130

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [ ] New tests added (if applicable)
- [x] Manually verified

## Checklist

- [x] Self-reviewed the code
- [ ] Updated documentation (if needed)
- [x] No new warnings or errors introduced

PR #4190, opened 2026-02-21
Replace bare except with specific exceptions in program view

### LINKED ISSUE #4189: [Bug]: Bare except in program view hides errors and can catch system exits
### Description

In `program() `(main program view), the “current program” resolution uses a bare except: that catches all exceptions (including `KeyboardInterrupt` and SystemExit) and turns them into `Http404`. That violates common Python practice, can hide real bugs, and makes debugging harder. Only expected lookup/attribute failures should be caught and converted to “no current program” (404).

### Steps to Reproduce

1. Trigger the “current program” code path (e.g. open a URL that uses the “current” program instance for a given program type).
2. Cause a failure inside the try (e.g. no programs of that type so progs is empty and sorted(progs,...)[0] raises IndexError).
3. Observed: User gets Http404 “No current program of the type …” (intended).
4. Code-quality issue: If an unexpected error (e.g. KeyboardInterrupt, or a bug) occurs in that block, it is still caught and reported as 404 instead of surfacing the real error.

Reproduction is mainly for verifying that after the change, expected cases still return 404 and unexpected exceptions are no longer swallowed.

### Expected Behavior

- Before fix: (1) When there is no current program, user correctly gets Http404 “No current program of the type …”. (2) Any other exception in that block (e.g. bug, KeyboardInterrupt) is incorrectly caught and also returned as 404, hiding the real error.
- After fix: (1) When there is no current program (e.g. empty list → IndexError), user still gets Http404 “No current program of the type …”. (2) Other exceptions (e.g. KeyboardInterrupt, unexpected bugs) are no longer caught; they propagate so the real error is visible and debugging is correct.

### Actual Behavior

N/A

### Screenshots

_No response_

### Operating System

_No response_

### Browser

_No response_

### Additional Context

_No response_

aryan7081:
Fixes #4189 

Replaces the bare except: in the “current program” resolution with except (IndexError, ValueError, AttributeError, KeyError, TypeError). Expected behaviour: “No current program” still returns Http404; all other exceptions propagate so they are not hidden and system exits are not caught.

### DIFF
--- a/esp/esp/web/views/main.py
+++ b/esp/esp/web/views/main.py
@@ -75,7 +75,7 @@ def program(request, tl, one, two, module, extra = None):
             programs = Program.objects.all()
             progs = [(program, program.dates()[-1]) for program in programs if program.program_type == one and len(program.dates()) > 0]
             two = sorted(progs, key=lambda x: x[1], reverse = True)[0][0].program_instance
-        except:
+        except (IndexError, ValueError, AttributeError, KeyError, TypeError):
             raise Http404("No current program of the type '" + one + "'.")
     try:
         prog = Program.by_prog_inst(one, two)


hwatheod:
Duplicate of #4152

PR #4193, opened 2026-02-21
Replace bare except with NameError in database settings

### LINKED ISSUE #4191: [Bug]: Bare except in settings.py when loading database config
### Description

In settings.py, when default DB `USER/PASSWORD/NAME` are empty, the code falls back to `DATABASE_USER / DATABASE_PASSWORD / DATABASE_NAME`. Each fallback uses a bare except:, which catches every exception. The intended case is an `undefined` name `(NameError)`. Catching only NameError makes the intent clear and avoids masking other startup errors.

### Steps to Reproduce

1. Configure the app so `DATABASES['default']['USER'] `(or PASSWORD/NAME) is empty or missing.
2. Ensure the corresponding fallback is not defined (e.g. DATABASE_USER not set in the module settings imports from).
3. Start the Django app (e.g.` manage.py runserver` or any command that loads settings).

### Expected Behavior

- Before fix: (1) When the fallback name is missing, startup fails with the intended message (“You need to supply either … or DATABASE_USER/… in …”). (2) Any other exception in that block (e.g. typo, import error) is caught and reported as the same “supply USER/PASSWORD/NAME” message, hiding the real cause.
- After fix: (1) When the fallback name is missing (NameError), startup still fails with the same clear message. (2) Any other exception in that block is no longer caught; the real error (e.g. typo, import error) is raised and visible for debugging.

### Actual Behavior

N/A

### Screenshots

_No response_

### Operating System

_No response_

### Browser

_No response_

### Additional Context

_No response_

aryan7081:
Fixes #4191 

Replaces the three bare `except`: blocks in database config fallback (USER, PASSWORD, NAME) with `except NameError`:. Expected behaviour: Missing `DATABASE_USER/DATABASE_PASSWORD/DATABASE_NAME` still raises the same clear message; all other startup errors propagate, so the actual cause is visible.

### DIFF
--- a/esp/esp/settings.py
+++ b/esp/esp/settings.py
@@ -61,17 +61,17 @@ TEMPLATES[0]['OPTIONS']['debug'] = DEBUG
 if len(DATABASES['default']['USER']) == 0:
     try:
         DATABASES['default']['USER'] = DATABASE_USER
-    except:
+    except NameError:
         raise Exception("You need to supply either DATABASES['default']['USER'] or DATABASE_USER in database_settings.py")
 if len(DATABASES['default']['PASSWORD']) == 0:
     try:
         DATABASES['default']['PASSWORD'] = DATABASE_PASSWORD
-    except:
+    except NameError:
         raise Exception("You need to supply either DATABASES['default']['PASSWORD'] or DATABASE_PASSWORD in database_settings.py")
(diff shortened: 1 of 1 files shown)


hwatheod:
Duplicate of #4152

Issue #1716, opened 2015-06-23
Remove subqueries in the catalog


benjaminjkraft:
The worst one will be removed in #1715, but I think all of the subqueries around line 216 of esp/esp/program/models/class_.py are likely to be slow, since they are effectively a query per line.  Many of them can now be written using a `select_related`, `prefetch_related`, or other aggregate, and probably once we hit django 1.7 all of them can.  If not, it is probably possible to change them to be extra `SELECT`/`WHERE` parameters, rather than full subqueries, which is probably faser.

mgersh:
I fell down a rabbit hole looking into this. Summary: we can't actually remove all of the subqueries. Also, the change I made in #1717 doesn't work right.

In the original query before I messed with it, there were five subqueries. Here's what's going on with them:
* _num_students is doing an aggregate across class sections. This probably worked fine before I changed it but now it's broken because class sections are also accessed elsewhere in the query, specifically because of the order by sections__meeting_times__start, so grouping isn't happening. I think my preferred solution here is getting rid of the meeting times sort, since it doesn't mean much anyway, but this is not particularly robust against people trying to do more things with this query in the future, and we might just want to also give up and get rid of _num_students.
* teacher_ids is just a prefetch_related. Some surrounding code will also have to change, but nothing weird is going on here that I can see.
* media_count is doing an aggregate on a GenericRelation, which Django can't do.
* _index_qsd is looking for something even less related to the ClassSubject than media_count is, which Django definitely can't do. It's only ever used to check for existence, though, so we can at least change it to a count instead of the weird custom list thing.
* _studentapps_count can be replaced by a normal aggregate if we make sure we're not getting duplicates and are careful not to break it.

PR #4198, opened 2026-02-21
Fix N+1: batch-fetch latest profiles in statistics view

### LINKED ISSUE #4197: [Bug]: N+1 queries in program statistics view when loading user profiles
### Description

The program statistics view builds a list of “latest profile” per user with a list comprehension that calls `user.getLastProfile()` for each user. Each call runs a separate DB query (`RegistrationProfile.objects.filter(user__exact=user).select_related().latest('last_ts')`), so the number of queries grows with the number of users (N+1). For large result sets this makes the statistics page slow and increases database load.

### Steps to Reproduce

1. Log in as a user with access to program statistics (e.g. program admin).
2. Open the statistics page for a program.
3. Run a query that returns a large number of users (e.g. filter that matches hundreds or thousands of users).
4. Submit the form and wait for the statistics result.
5. Observe slow response time and, if possible, many duplicate queries for RegistrationProfile in the DB or in Django Debug Toolbar.

### Expected Behavior

- Before fix: One query per user to load the latest profile (N+1). Page can be very slow for large user sets.
- After fix: A small, fixed number of queries (e.g. one or two) to load all needed profiles. Response time stays reasonable as the number of users grows, and statistics results are unchanged.

### Actual Behavior

N/A

### Screenshots

_No response_

### Operating System

_No response_

### Browser

_No response_

### Additional Context

_No response_

aryan7081:
Fixes #4197 

Replaces the per-user getLastProfile() loop in the program statistics view with a single batched query: fetch latest RegistrationProfile per user by user_id__in=user_ids ordered by user_id, -last_ts, build a user_id → profile map, and build profiles in the same order as user_list. Statistics still receive (user_list, profiles) in matching order. This removes the N+1 and improves performance for large result sets. Can be linked to performance work such as #1716 if desired.


esp/esp/program/views.py
```diff
@@ -1087,12 +1099,20 @@ def get_field_ids(form):
 
             users = ESPUser.objects.filter(users_q).distinct()
             result_dict['num_users'] = users.count()
-            profiles = [user.getLastProfile() for user in users]
+            user_list = list(users)
+            # Batch-fetch latest profile per user to avoid N+1
+            user_ids = [u.id for u in user_list]
+            profile_by_user = {}
+            if user_ids:
+                for p in RegistrationProfile.objects.filter(user_id__in=user_ids).select_related('user').order_by('user_id', '-last_ts'):
```
hwatheod: This will be more performant if you use `.filter(user__in=users)` directly instead of passing in the list of ids from a Python list. I tested it on our server on one of our larger programs.

esp/esp/program/views.py
```diff
@@ -1087,12 +1099,20 @@ def get_field_ids(form):
...
+                    if p.user_id not in profile_by_user:
+                        profile_by_user[p.user_id] = p
+            profiles = [profile_by_user.get(u.id) or RegistrationProfile(user=u) for u in user_list]
 
             #   Accumulate desired information for selected query
             from esp.program import statistics as statistics_functions
             if hasattr(statistics_functions, form.cleaned_data['query']):
-                context['result'] = getattr(statistics_functions, form.cleaned_data['query'])(form, programs, users, profiles, result_dict)
+                context['result'] = getattr(statistics_functions, form.cleaned_data['query'])(form, programs, user_list, profiles, result_dict)
```
hwatheod: You should keep `users` rather than `user_list` here. You shouldn't change the data type being passed in to the statistics functions.

aryan7081:
@hwatheod Done. Switched to `filter(user__in=users)` instead of building a list of IDs, and pass users rather than `user_list` into the statistics functions so the original data type is kept.

hwatheod (approved):

PR #4203, opened 2026-02-21
feat: Refactor send_text_messages.py to use argparse

### LINKED ISSUE #4200: [Feature]: Refactor send_text_messages.py to use argparse module
### Problem Statement

The utility script [esp/useful_scripts/send_text_messages.py](cci:7://file:///c:/Users/Lenovo/devsite/esp/useful_scripts/send_text_messages.py:0:0-0:0) currently relies on manual argument parsing by checking `len(sys.argv)`. This approach is fragile, makes the code harder to read, and prevents the script from offering a standard built-in help (`-h` or `--help`) menu. There is an existing inline comment `TODO(jmoldow): Use Python built-in argparse module` that points out this technical debt.

### Proposed Solution

Refactor [send_text_messages.py](cci:7://file:///c:/Users/Lenovo/devsite/esp/useful_scripts/send_text_messages.py:0:0-0:0) to use Python's built-in `argparse` module. We should define the parser with descriptions and the 5 required arguments to replace the current `sys.argv` manual indexing:

1. `account_sid` (Account SID)
2. `auth_token` (Auth Token)
3. `senders_file` (File with line-separated phone numbers to send from)
4. `recipients_file` (File with line-separated phone numbers recipients)
5. `message_body` (Message to send)

This transition will automatically handle argument validation and provide a formatted help string to the user.

### Alternatives Considered

- Leaving it as is: Not ideal since with 5 expected sequential arguments, manual indexing easily leads to mistakes for whoever executes the script. 
- Using 3rd-party libraries like `click` or `typer`: While the syntax is clean, we should stick with the standard library `argparse` to meet the existing TODO's requirements and avoid adding unnecessary Python dependencies.

### Additional Context

Relevant code section located in `esp/useful_scripts/send_text_messages.py`:

```python
# TODO(jmoldow): Use Python built-in argparse module.
if len(sys.argv) < 6:
  print(("Usage: %s <account SID> <auth token> <file with line-separated phone numbers to send from> <file with line-separated phone numbers recipients> <message to send>" % sys.argv[0]))
  exit(1)

Rahul-R79:
## Description
Refactored [esp/useful_scripts/send_text_messages.py](cci:7://file:///c:/Users/Lenovo/devsite/esp/useful_scripts/send_text_messages.py:0:0-0:0) to use Python's built-in `argparse` module instead of manual `sys.argv` parsing. This resolves an existing TODO in the code, improves argument validation, and provides a standard `--help` menu for users running the script.

## Related Issue
Closes #4200

### DIFF
--- a/esp/useful_scripts/send_text_messages.py
+++ b/esp/useful_scripts/send_text_messages.py
@@ -13,8 +13,10 @@ account_sid = args.account_sid
 auth_token = args.auth_token
 ourNumbers = [x.strip() for x in open(args.senders_file, "r").readlines() if x.strip()]
 if not ourNumbers:
-    parser.error("No sender numbers provided in '{}'".format(args.senders_file))
-recipients = [x.strip() for x in open(args.recipients_file, "r").readlines() if x.strip()]
+with open(args.senders_file, "r") as f:
+    ourNumbers = [x.strip() for x in f.readlines() if x.strip()]
+with open(args.recipients_file, "r") as f:
+    recipients = [x.strip() for x in f.readlines() if x.strip()]
 body = args.message
 
 numberIndex = 0


Bixquix:
Hey @Rahul-R79, I had a look through the changes and spotted a few things :
1.import sys is still there even though it's not needed anymore after the argparse switch, so I think that can go. 
2.The help text for recipients_file is also a tiny bit off grammatically, something like "File with line-separated recipient phone numbers" reads cleaner.
3.There's no check for when senders_file is empty — if it is, ourNumbers becomes an empty list and the script will blow up later at the index lookup which is a pain to debug. 
4.Both files are being opened without a with block which can leak file descriptors, worth wrapping those in with open(...) as f:
Are you still working on this? If not I'm happy to pick it up!

Rahul-R79:
Hey @Bixquix, thanks for the awesome and detailed review! I'm still working on this and would love to collaborate on those fixes.

If you'd like, feel free to drop those specific fixes as Suggested Changes in the 'Files changed' tab and I can commit them directly!

Bixquix:
I was about to suggest the same change here, and Copilot already proposed it in the review. Applying that suggestion should fix this.

PR #4210, opened 2026-02-21
Make admin bar links configurable via theme settings

Pjoshi0805:
This PR makes the admin bar links configurable via theme settings instead of being hardcoded in admin_bar.js.

Changes:
- Added admin_links field to theme config
- Updated admin_bar.js to dynamically render links
- Preserved fallback behavior for backward compatibility.
I am currently keeping it as a draft pull request but I would appreciate it if you could review and test it on your end.  
Once tested, this PR can be finalized and merged.


willgearty:

PR #4211, opened 2026-02-21
Filter available floating resources based on 'returned' status (#3018)

### LINKED ISSUE #3018: Filter available floating resources based on 'returned' status
@kkbrum requested this in https://github.com/learning-unlimited/ESP-Website/pull/3012, but it didn't make it into SR12. The idea is that while floating resources may be "available" according to the database after a class is over, they aren't physically available until they are returned by the teacher. This might be especially important for programs with time blocks with little time between them or for instances where teachers hold on to the floating resources for multiple classes. Really, we should make sure a resource is "available" according to the database (i.e. it isn't assigned at the specified time) and that that same resource has been returned since the last time it was assigned if we are going to allow it to be assigned on the teacher checkin page.

abhinavjha0239:
### Problem

When a floating resource (e.g., a projector) is checked out to a class in timeslot A and that class ends, the resource immediately appears as "available" in the dropdown for timeslot B classes. This is because `getAvailableResources()` only checks whether the timeslot-B copy of the resource has a `ResourceAssignment` — it doesn't check whether the teacher from timeslot A has **physically returned** the equipment.

This leads to a real-world problem during Splash events: an admin checks out Projector #1 to a class at 8 AM, the class ends at 9 AM, but the teacher hasn't brought the projector back yet. The system still shows Projector #1 as available for the 10 AM class, so another admin checks it out — but it's not actually there.

### Root Cause

The `ResourceAssignment` model already has a `returned` boolean field and a checkbox in the teacher checkin UI to toggle it, but `Program.getAvailableResources()` never consulted this field. It only called `Resource.is_available()`, which checks for assignments in the **current** timeslot only.

### Solution

**1. `Program.getAvailableResources()` — batch filter unreturned resources**

Instead of the old one-liner:

```python
return [x for x in self.getFloatingResources(...) if x.is_available()]
```

We now do a single batched query to find all resource names that have an unreturned assignment in any **earlier** timeslot (`event__end__lte=timeslot.start`), then exclude them:

```python
unreturned_names = set(
    ResourceAssignment.objects.filter(
        resource__name__in=floating_names,
        resource__event__end__lte=timeslot.start,
        resource__event__program=self,
        resource__is_unique=True,
        returned=False,
    ).values_list('resource__name', flat=True)
)
return [x for x in floating if x.is_available() and x.name not in unreturned_names]
```

This adds exactly **1 extra DB query** (not N+1). All existing callers (`newassignment`, `getavailableequipment`) iterate the result immediately, so the return type change from QuerySet to list is compatible.

**2. `Resource.has_unreturned_prior_assignment()` — helper method**

Per-resource check used for defense-in-depth and testing.

**3. `Resource.assign_to_section()` — guard clause**

Even if a code path bypasses `getAvailableResources()`, the assignment itself now raises `ESPError` if the resource hasn't been returned from a prior timeslot.

### Edge Cases Handled

- **First timeslot**: No earlier timeslots exist, `unreturned_names` is empty, behavior unchanged
- **Same timeslot**: `event__end__lte=timeslot.start` excludes the current timeslot (a slot's end > its start), so it doesn't double-count with `is_available()`
- **Returned resources**: A resource marked `returned=True` in an earlier slot correctly appears available again

## Related Issue

Closes #3018

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] New tests added (5 test cases in `FloatingResourceAvailabilityTest`)
- [x] Manually verified (see screenshots below)

### Unit Tests

5 new test cases covering:

| Test | What it verifies |
|------|-----------------|
| `test_available_when_no_prior_assignment` | Resource with no assignment in any slot is available |
| `test_available_when_prior_assignment_returned` | Resource with `returned=True` in earlier slot is available |
| `test_unavailable_when_prior_assignment_not_returned` | Resource with `returned=False` in earlier slot is **filtered out** |
| `test_current_timeslot_taken_still_unavailable` | Existing behavior preserved — assigned in current slot = unavailable |
| `test_has_unreturned_prior_assignment` | Helper method returns correct values and responds to status changes |

### Manual Testing Screenshots

**Test setup**: 4 timeslots, 3 floating resources (Projector 1, Projector 2, Speaker 1), 4 classes.
- Projector 1 assigned to Physics 101 (Slot 1), `returned=False`
- Speaker 1 assigned to Chemistry 101 (Slot 2), `returned=True`

**Missing Teachers page overview** — shows resource assignments with returned checkboxes:

02-missingteachers-page

**Before return** — Math 101 (Slot 3) dropdown shows only **Speaker 1** and **Projector 2**. Projector 1 is correctly filtered out because it hasn't been returned from Slot 1:
06-dropdown-unreturned-filtered

**After marking Projector 1 as returned** — Math 101 dropdown now shows all 3 resources (**Speaker 1**, **Projector 2**, **Projector 1**):

07-dropdown-all-available-after-return

## Checklist

- [x] Self-reviewed the code

### DIFF
--- a/esp/esp/program/modules/handlers/batchclassregmodule.py
+++ b/esp/esp/program/modules/handlers/batchclassregmodule.py
@@ -1,0 +1,265 @@
+
+from __future__ import absolute_import
+__author__    = "Individual contributors (see AUTHORS file)"
+__date__      = "$DATE$"
+__rev__       = "$REV$"
+__license__   = "AGPL v.3"
+__copyright__ = """
+This file is part of the ESP Web Site
+Copyright (c) 2024 by the individual contributors
+  (see AUTHORS file)
+
+The ESP Web Site is free software; you can redistribute it and/or
+modify it under the terms of the GNU Affero General Public License
+as published by the Free Software Foundation; either version 3
+of the License, or (at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Affero General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public
+License along with this program; if not, write to the Free Software
+Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+
+Contact information:
+MIT Educational Studies Program
+  84 Massachusetts Ave W20-467, Cambridge, MA 02139
+  Phone: 617-253-4882
+  Email: esp-webmasters@mit.edu
+Learning Unlimited, Inc.
+  527 Franklin St, Cambridge, MA 02139
+  Phone: 617-379-0178
+  Email: web-team@learningu.org
+"""
+from esp.program.modules.base import ProgramModuleObj, needs_admin, main_call, aux_call
+from esp.program.modules.handlers.listgenmodule import ListGenModule
+from esp.utils.web import render_to_response
+from esp.users.models import ESPUser, PersistentQueryFilter
+from esp.users.controllers.usersearch import UserSearchController
+from esp.program.models import ClassSection, ClassSubject
+from esp.program.class_status import ClassStatus
+from esp.middleware import ESPError
+from django.db import transaction
+from django.db.models import Prefetch
+
+import logging
+logger = logging.getLogger(__name__)
+
+
+class BatchClassRegModule(ProgramModuleObj):
+    doc = """Batch register a group of students to a class section."""
+
+    @classmethod
+    def module_properties(cls):
+        return {
+            "admin_title": "Batch Class Registration",
+            "link_title": "Batch Register Students to a Class",
+            "module_type": "manage",
+            "seq": 503,
+            "choosable": 1,
+        }
+
+    @main_call
+    @needs_admin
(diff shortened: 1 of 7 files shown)


abhinavjha0239:
Hi @willgearty , Can you please review this?

willgearty:

abhinavjha0239:
Hi @willgearty, I am working on the copilot suggestions.

abhinavjha0239:
Hi @willgearty , ready for review.

kkbrum:
Can you please resolve the merge conflicts?

abhinavjha0239:
> Can you please resolve the merge conflicts?

@kkbrum resolved the merge conflicts.

abhinavjha0239:
@kkbrum Thank you for the great feedback! Both help text additions are now implemented in a4c1a43:

1. **Help text below the checkout form**: Added "Once the resource is returned, don't forget to hit the checkmark next to the item so that it is available for the next teacher! If you added a resource by accident, hit the X to remove the assignment." underneath the dropdown/buttons row.

2. **Hover text on disabled dropdown**: When "No resources available" is shown, hovering the disabled dropdown now displays "Make sure resources have been recorded as returned on previous time slots!"

Both changes are minimal — one line in `assignment-edit.js` and a new table row in `new_assignment_form.html`.

PR #4234, opened 2026-02-21
Fix npm installation: re-add npm to packages_base.txt

### LINKED ISSUE #3795: Fix issue installing npm
During SR16, installing `npm` threw lots of errors. We are able to proceed without it for this release, but we need to figure out how to upgrade it long-term. This issue is to investigate and resolve. Once that's fixed, add it back to the first line of `esp/packages_base.txt`.

Oval17:
- Add npm back as first line of packages_base.txt (was removed as a workaround but broke lessc installation)
- Remove the now-redundant nodejs apt install from packages_base_manual_install.sh (the npm apt package already depends on nodejs, so both are installed when packages_base.txt is processed)
- Upgrade less from 1.7.5 to 3.13.1 for compatibility with modern Node.js

## Related Issue

Closes #3795

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [ ] Existing tests pass
- [ ] Manually verified

## Checklist

- [ ] Self-reviewed the code


willgearty:
Do we need to update any of our .less files because of the less upgrade? Also, currently failing theme editor tests

Oval17:
Bootstrap 2.3.2's LESS files use IE6/7 property hacks (zoom: 1, display: inline-block, etc.) across 38 places. Less 2.0+ removed support for these hacks, so upgrading to less@3.13.1 caused lessc to fail with a parse error → 500 responses in both testSelector and testEditor.

Have reverted it, it will go with bootstrap upgrade only @willgearty

Oval17:
Have tested and checked all the edge cases , before npm wasnt even getting installed thats why we were getting so many errors that led to npm install less failure anyway now all of that is resolved.

willgearty (approved):
Looks great, thanks! Will merge when the tests pass 👍

PR #4254, opened 2026-02-21
Warn if comm email links to another program

### LINKED ISSUE #2096: Automatically check for URLs in emails for references to past programs
When copying an email from a past program, it is easy to forget to update links with the new program name.  Not sure of the best way to implement this, but somehow the communications panel should warn you if you're about to send a mass email with links to a past program.

aryan7081:
Fixes #2096 

Scans subject and body for `/learn/... `and `/manage/...` program URLs and compares them to the current program. If any link points to a different program, the preview shows a warning and the first "Send" re-shows the preview with a "Send anyway" option so admins can fix links or confirm before sending.


esp/esp/program/modules/handlers/commmodule.py
```diff
@@ -48,6 +49,30 @@
 
 import re
 
+# Match /learn/ProgramName/Instance or /manage/ProgramName/Instance (program url = two path segments)
+_PROGRAM_URL_PATTERN = re.compile(
+    r'(?:/learn|/manage)/([^/\s\'"<>]+/[^/\s\'"<>]+)',
```
hwatheod: The important paths are `learn`, `teach`, `volunteer`.  It is not necessary to include `manage`, which is only used by administrators and unlikely to be mentioned in an email.

aryan7081:
@hwatheod Updated to match learn, teach, and volunteer and removed manage.

Issue #3739, opened 2024-03-22
Enhanced search in toolbar


milescalabresi:
Clark requested the ability to search for more things in the gray toolbar (in addition to users and classes). Suggestions include
- program modules/config pages
- static pages
- other commonly used pages/links

Up for discussion if the search should match the URL, some keywords associated with the page, or something else. Also unclear how fuzzy to make the matching. May relate to https://github.com/learning-unlimited/ESP-Website/issues/3699

ayushk-1801:
Hi @milescalabresi , I would like to work on this issue, Can you assign this to me?

milescalabresi:
Thanks for your interest in contributing to our project! I've assigned you to this issue. Please make sure to check out our [instructions for getting started](https://github.com/learning-unlimited/ESP-Website/wiki#getting-started) and follow our [contribution rules](https://github.com/learning-unlimited/ESP-Website/wiki#rules). This issue is pretty broad at this point. I think before you start coding, it'd be great if you can lay out a plan. It reflects better on your ability if you can do so without the help of AI

PR #4255, opened 2026-02-21
Implement module search feature

### LINKED ISSUE #3602: Search for modules/functionality
Someone at SplashCon suggested that we implement some way to search all of the modules/functionality for a keyword to help them find what they are looking for.

Requested at SplashCon.

SuyashJain17:
## Description

This PR adds a client-side search filter for administrative modules and settings. It introduces a reusable search utility to enable instant filtering on:

- **Program Creation**: Filter module selection questions.
- **Program Dashboard**: Filter the "Additional Modules" list.
- **Module Management**: Filter student and teacher registration modules.

## Related Issue

Closes #3602 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [ ] Existing tests pass
- [x] Manually verified

## Checklist

- [x] Self-reviewed the code

### DIFF
--- a/esp/esp/program/modules/handlers/commmodule.py
+++ b/esp/esp/program/modules/handlers/commmodule.py
@@ -33,6 +33,7 @@ Learning Unlimited, Inc.
   Email: web-team@learningu.org
 """
 from esp.program.modules.base import ProgramModuleObj, needs_admin, main_call, aux_call
+from esp.program.modules.admin_search import AdminSearchEntry
 from esp.program.modules.handlers.listgenmodule import ListGenModule
 from esp.utils.web import render_to_response
 from esp.dbmail.models import MessageRequest, PlainRedirect
@@ -89,6 +90,19 @@ class CommModule(ProgramModuleObj):
             "seq": 10,
             "choosable": 1,
             }
+
+    @classmethod
+    def get_admin_search_entry(cls, program, tl, view_name, pmo):
+        if view_name != "commpanel":
+            return None
+        base = program.getUrlBase()
+        return AdminSearchEntry(
+            id="manage_commpanel",
+            url="/manage/%s/commpanel" % base,
+            title="Email (Communications Panel)",
+            category="Coordinate",
+            keywords=["email", "communications", "commpanel", "messages"],
+        )
 
     @staticmethod
     def get_mailer_warnings(listcount, filterid, sendto_fn_name):


hwatheod:
All of these changes are searching for modules among a list that is already printed on the screen. While this is mildly more convenient for the user, I'm not sure if this is what we want because the user can already use their browser's "Find" function to do this. 

I think that the request was perhaps for some way to find functionality that wasn't already shown on the current page and the user doesn't know where to go to find it.

@willgearty What do you think?

SuyashJain17:
Thanks for the feedback!

I agree that users can use the browser’s Find feature. My intention here was mainly to improve usability by allowing dynamic filtering of modules directly in the UI, especially since some of these lists can get long.

That said, if the original goal of #3602 was more about helping users discover functionality that isn’t already visible on the page, then I may have misunderstood the scope. I’m happy to adjust the implementation based on what was intended.

SuyashJain17:
After thinking more about the feedback, I’d like to adjust the approach to better address discoverability on the Admin Dashboard.

Instead of limiting search to just “Additional Modules”, I’m planning to:

- Move the search bar to the top of the Admin Dashboard.

- Make it filter all dashboard sections (Configure, Coordinate, Logistics, Quick Links, Printables, etc.), not just the additional modules list.

- Hide entire sections if none of their items match the search query.

- Add a “No modules found” message when nothing matches (with a link to Module Management).

This way the search helps users locate functionality across the dashboard, not just within one visible list — which should better address the original concern about discoverability.

I’ll also keep the search behavior consistent across Program Creation and Module Management.

Let me know if this direction aligns better with what was intended before I proceed.

willgearty:
Yes, I believe @hwatheod is right. The scope is a bit broader in that a user may be looking for something that isn't on the dashboard. This would include individual pages which may have specific uses (as a quick example, a very specific printable page). Ultimately, to make this as thorough and useful as possible we would need to document each view with some sort of metadata/keywords, and then we would search through those to provide the user with the page that is closest to their query. It's possible that the metadata could be generated using AI to speed things up. This also probably overlaps a little with https://github.com/learning-unlimited/ESP-Website/issues/3739.

SuyashJain17:
I’ve reworked this PR to better align with the broader discoverability goal.

Instead of filtering only visible module lists, this update introduces a metadata-driven admin search:

- Added an AdminSearchEntry dataclass and serialize_admin_search_entries(program) to define structured metadata (title, category, keywords, URL) for admin views.

- Injected the serialized metadata into the Admin Portal view.

- Added a new “Search admin tools” bar at the top of the dashboard.

- Implemented client-side search over titles and keywords.

- Grouped results by category (Configure, Coordinate, Logistics, Quick Links, Printables).

- Included pages not directly visible in the dashboard (e.g., specific/hidden printables).

This should better support searching across functionality rather than just filtering visible lists.

I’d appreciate another look to confirm this direction aligns with what was intended. Happy to refine further if needed.

esp/templates/program/modules/admincore/directory.html
```diff
@@ -46,6 +57,25 @@
...
+        <input
+            type="text"
+            id="admin_search_input"
+            placeholder="Search for modules, pages, or printables…"
+            style="width: 100%; box-sizing: border-box; padding: 6px; border: 1px solid #ccc; border-radius: 4px;"
+        >
+        <div id="admin_search_results" style="margin-top: 10px; display: none;"></div>
+        <div id="admin_search_empty" style="margin-top: 10px; display: none; font-style: italic;">
+            No modules found for that search. You may be able to find more via
+            <a href="/manage/{{ program.getUrlBase }}/modules">Module Management</a>.
```
kkbrum: The module management page really doesn't show any more modules that are useful relevant to what admins would be searching here; that's really about ordering things for student/teacher reg. Instead, I would say to look under "Additional Modules" at the bottom of the page.

esp/templates/program/modules/admincore/modules.html
```diff
@@ -12,6 +12,15 @@
 {% block xtrajs %}
     {{block.super }}
     <script type="text/javascript" src="/media/scripts/program/modules/module_management.js"></script>
+    <script type="text/javascript" src="/media/scripts/utils/search_filter.js"></script>
```
kkbrum: These student/teacher module searches aren't very useful in my opinion. The lists here really don't get that long.

esp/public/media/scripts/program/modules/modulequestions.js
```diff
@@ -2,6 +2,16 @@ $j(document).ready(function(){
     // when the page loads, check/select the questions induced by the modules
     // that are already selected (if any, from a template program e.g.)
     $j(".hidden-field").parents("tr").hide()
+    
+    // Inject search bar for module questions
+    var searchHtml = '<div style="margin-bottom: 10px;"><input type="text" id="module_search" placeholder="Filter modules..." style="width: 100%; padding: 5px; border: 1px solid #ccc; border-radius: 4px;"></div>';
```
kkbrum: This one also doesn't seem super useful to me.

esp/esp/program/modules/admin_search.py
```diff
@@ -0,0 +1,236 @@
...
+    """
+
+    id: str
+    url: str
+    title: str
+    category: str
+    keywords: List[str]
+
+
+def get_admin_search_entries(program):
```
kkbrum, line 41: This search functionality is super cool and useful! However, I think coding it like this is pretty fragile. For example, if I search "line items" or "extra costs", it does not give any results. When someone codes up a new module, they are unlikely to remember to add it here. I think it would be better to add this metadata within each module itself, and then pull it all together here. That way each module is guaranteed to have /some/ keywords that will work.

kkbrum (changes requested):
I really like the overall search at the top of the management page! The other searches are less useful and I think are just adding clutter. For the overall search, I think it would be better to code up the keywords within each module's file rather than in this search module's file, as I think it is less likely to be forgotten about when people code up new modules.

esp/templates/program/modules/admincore/directory.html
SuyashJain17: Updated - the message now directs users to check Additional Modules at the bottom of the page instead of Module Management.

esp/templates/program/modules/admincore/modules.html
SuyashJain17: Removed these search filters and the `search_filter.js` dependency to avoid adding UI clutter.

esp/public/media/scripts/program/modules/modulequestions.js
SuyashJain17: Removed the injected search bar and related filtering logic from `modulequestions.js`.

esp/esp/program/modules/admin_search.py
SuyashJain17: Refactored the implementation so each module can define its own `get_admin_search_entry()` with keywords.
`admin_search.py` now collects entries dynamically from modules instead of using a centralized list.

SuyashJain17:
@kkbrum  I've addressed the review comments:

- Removed the extra search filters to avoid UI clutter
- Updated the "No modules found" message
- Refactored search metadata so keywords live in each module

Please let me know if anything else should be adjusted.

esp/public/media/scripts/program/modules/modulequestions.js
kkbrum, line 1: Please remove the unnecessary whitespace change.

esp/templates/program/modules/admincore/modules.html
kkbrum, line 1: I think this is leftover from previous changes and should get removed now?

esp/templates/program/newprogram.html
```diff
@@ -11,6 +11,7 @@
 {% block javascript %}
 {{ block.super }}
 <script src="/media/scripts/program/modules/modulequestions.js"></script>
+<script src="/media/scripts/utils/search_filter.js"></script>
```
kkbrum: I think this is leftover from previous changes and should get removed now?

SuyashJain17:
Thanks for the review! I've addressed the requested changes:

- Disambiguated duplicate admin search titles in `admin_search.py` by appending the view name where necessary.
- Restored the intended whitespace formatting in `modulequestions.js`.
- Removed the leftover change in `modules.html`.
- Removed the unused `search_filter.js` script from `newprogram.html`.

Please let me know if anything else should be adjusted.

SuyashJain17:
Thanks for the review! I’ve pushed updates addressing the feedback:

* Fixed the XSS concerns by switching to `json_script` and removing `innerHTML` usage.
* Excluded class-specific views that require IDs so invalid links no longer appear.
* Improved human-readable labels for view names.
* Added tests for the admin search logic.

I also tested the results locally and verified the links point to valid pages. Let me know if anything else should be adjusted.

esp/esp/program/modules/admin_search.py
```diff
@@ -2,6 +2,43 @@
...
+# broken links.
+EXCLUDED_VIEW_NAMES = frozenset([
+    "editclass", "manageclass", "addsection", "deletesection",
+    "deleteclass", "approveclass", "rejectclass", "coteachers",
+    "classavailability", "reviewclass",
+])
+
+# Human-readable labels for view names used in disambiguation. When multiple
+# entries share the same title, we append " (label)" so users can tell them apart.
+VIEW_DISPLAY_NAMES = {
```
kkbrum: again, I think this should go into the module files themselves, as opposed to collected here, since people will undoubtedly forget to update this list in the future.

SuyashJain17:
Thanks for the feedback! I verified the changes locally.

After switching the admin search to an **opt-in model** (modules must explicitly return `AdminSearchEntry`), JSON/AJAX/internal endpoints are no longer listed.

I also noticed that **TeacherClassRegModule** and **ProgramPrintables** were still leaking endpoints into the search results, so I fixed those as well:

* **TeacherClassRegModule** now only exposes the human-readable views `makeaclass` (Register Your Classes) and `copyclasses` (Copy Your Classes), while internal endpoints like `ajaxstudentattendance` are excluded.
* **ProgramPrintables** now returns clean entries for the printable pages (teacher schedules, teacher list, classes by teacher, etc.), replacing the raw links with human-readable titles.

I confirmed that previously exposed endpoints such as `ajaxteachercheckin`, `ajaxclassdetail`, and the JSON data endpoints no longer appear in the search results.

Let me know if there’s anything else you’d like adjusted.

image

esp/esp/program/modules/handlers/commmodule.py
kkbrum, line 1: this file has nothing to do with this PR

SuyashJain17: Yeah, you’re right - I’ve removed those changes from `commmodule.py`, so it’s no longer modified in this PR.

SuyashJain17:
I fixed the broken teacher copy link — the admin search entry was pointing to `copyaclass`, but the actual route/view is `copyclasses`. It now links to `/teach/<base>/copyclasses`, which resolves correctly.

image

esp/esp/program/modules/handlers/commmodule.py
kkbrum: Whoops I'm not sure what I was looking at when I wrote this; the search entry here is helpful. I'll commit it back and then this all looks great.

PR #4266, opened 2026-02-22
fix: remove Python 2/Django compat code and fix performance issues

### LINKED ISSUE #4240: [Bug]: Deprecated Python 2 / Django compat code and performance issues in accounting controllers
## Description

`esp/esp/accounting/controllers.py` has several issues:

### Deprecated Usage (Python 2 / Django < 2.0)
- `python_2_unicode_compatible` decorator and `from __future__` imports are Python 2 shims, unnecessary since Django 2.0+
- `six.moves.range` — `range` is already a builtin in Python 3
- `django.template.defaultfilters.slugify` — should use `django.utils.text.slugify`

### Performance
- `amount_paid()` runs the same query twice (`.exists()` then `.aggregate()`) — the `.exists()` check is redundant
- `amount_donation()` has the same double-query pattern
- `latest_finaid_grant()` uses `.exists()` followed by `.get()` — should use `.first()`
- `get_preferences()` has O(n²) list scanning per transfer
- `amount_due()` triggers redundant cascading calls to `ensure_required_transfers()` and `amount_requested()`

### Security
- `record_payment_from_identifier` doesn't validate identifier format, raising raw `ValueError`/`IndexError` on malformed input
- With `trusted=True`, transfer amounts are overwritten from the identifier without bounds checking

### Steps to Reproduce

1) Start the ESP-Website server with default settings (the middleware is enabled at [django_settings.py:208]

2) Send any HTTP request with an IE-style User-Agent header:

3) Observe a 500 error. The server crashes at [fixiemiddleware.py:34](vscode-file://vscode-app/c:/Users/boopa/AppData/Local/Programs/Microsoft%20VS%20Code/072586267e/resources/app/out/vs/code/electron-browser/workbench/workbench.html)

### Expected Behavior

When a request comes from an IE browser:

Middleware checks the response's MIME type using the modern content_type attribute
If the MIME type is not in the safe list 
1) remove the vary header
2) set Pragma:no-cache
3)Set Cache-Control: no-cache,must-revalidate
 
The response is delivered normally to the user — no crash, no 500 error

### Actual Behavior

Current (Broken) Behavior
When a request comes from an IE browser (User-Agent contains "MSIE"):

Middleware reaches [fixiemiddleware.py:34]: [response.mimetype.lower()
[AttributeError] is raised — [mimetype] was removed in Django 1.7
Server returns a 500 Internal Server Error
The user sees nothing — the intended page is never delivered

### Screenshots

_No response_

### Operating System

_No response_

### Browser

_No response_

### Additional Context

_No response_

moksha-hub:
Fixes deprecated Python 2 / Django compatibility code and performance issues in accounting controllers and middleware.

### Deprecated Usage Fixes
- Remove `from __future__ import absolute_import, unicode_literals` imports
- Remove `@python_2_unicode_compatible` decorator (unnecessary since Django 2.0+)
- Replace `six.moves.range` with builtin `range`
- Replace `django.template.defaultfilters.slugify` with `django.utils.text.slugify`
- Fix `FixIEMiddleware`: replace deprecated `response.mimetype` with `response.content_type` (removed in Django 1.7)

### Performance Optimizations
- `amount_paid()`: Remove redundant `.exists()` check before `.aggregate()` - single query instead of two
- `amount_donation()`: Remove redundant `.exists()` check before `.aggregate()` - single query instead of two
- `latest_finaid_grant()`: Use `.first()` instead of `.exists()` + `.get()` - single query instead of two
- `get_preferences()`: Use dict for O(1) lookup instead of O(n²) list scanning

### Security Improvements
- Add `getattr(response, "content_type", "")` with default to prevent AttributeError
- Add exception handling for `AttributeError` in User-Agent parsing

Closes #4240


esp/esp/accounting/controllers.py
```diff
@@ -78,7 +77,6 @@ def __init__(self, program, *args, **kwargs):
         self.finaid_items = ['Financial aid grant', 'Sibling discount']
         self.admission_items = ["Program admission", "Student payment"]
 
-    @transaction.atomic
```
willgearty, line 80: Why are you removing @transaction.atomic?

moksha-hub: > Why are you removing @transaction.atomic?

I hope its now resolved sir.
Thank you.

esp/esp/accounting/controllers.py
```diff
@@ -283,7 +284,6 @@ def transfers_to_program_exist(self):
                 destination=self.default_program_account(),
                 user=self.user).exists()
 
-    @transaction.atomic
```
willgearty, line 286: Still three instances of `@transaction.atomic` to restore in this file.

esp/esp/accounting/controllers.py
```diff
@@ -381,17 +379,16 @@ def get_transfers(self, line_items=None, **kwargs):
         return Transfer.objects.filter(user=self.user, line_item__in=line_items).order_by('id')
 
     def get_preferences(self, line_items=None):
-        #   Return a list of 4-tuples: (item name, quantity, cost, options)
```
willgearty, line 384: This comment should not be removed.

esp/esp/accounting/controllers.py
```diff
@@ -564,12 +562,12 @@ def amount_siblingdiscount(self):
             return Decimal('0')
 
     def amount_paid(self):
-        #   Compute sum of all transfers from outside (e.g. credit card payments) that are for this user
```
willgearty, line 567: Again, we shouldn't be removing useful comments.

esp/esp/middleware/fixiemiddleware.py
```diff
@@ -15,26 +13,26 @@ class FixIEMiddleware(MiddlewareMixin):
 
     def process_response(self, request, response):
 
-        # a list of mime-types that are decreed "Vary-safe" for IE
```
willgearty, line 18: Please restore these comments

esp/esp/web/views/json_utils.py
willgearty, line 1: Why is this file here?

moksha-hub: accidentally included.

PR #4285, opened 2026-02-22
Fix AJAX helpers to surface 4xx/5xx JSON errors and prevent double submits

### LINKED ISSUE #4279: [Bug]: AjaxErrorMiddleware returns 4xx/5xx for JSON errors, but frontend only handles 2xx responses
### Description

While testing AJAX error handling, I noticed that `AjaxErrorMiddleware` returns JSON error responses with different HTTP status codes depending on the exception type, but the frontend helpers in `ajax_tools.js` only attach a **success** callback and expect to handle errors via the JSON payload.

Because of that:

- Errors that go through `AjaxErrorMiddleware.bad_request` (HTTP 200) are surfaced to the user correctly.
- Errors that go through `AjaxErrorMiddleware.not_found` (HTTP 404) or `server_error` (HTTP 500) never hit the success callback, and since there’s no error callback wired up, those errors effectively disappear from the UI even though the middleware has generated a JSON error body.

This seems unintentional, given that the middleware is explicitly intended to “Return AJAX errors to the browser in a sensible way.”

### Steps to Reproduce

1. Set up a dev instance of ESP-Website and log in as a user who can access a view that uses the generic AJAX helpers from `esp/public/media/scripts/ajax_tools.js`.  
   - For example, student registration AJAX views under `/learn/<program>/ajax_*`, as exercised in `esp/esp/program/modules/tests/ajaxstudentreg.py`.
2. Trigger an AJAX request that hits a **404** or other exception that would be handled by `AjaxErrorMiddleware.not_found` or `server_error`. For example:
   - Call an AJAX endpoint with an invalid ID in the URL or POST data so the view raises `Http404` or triggers a lookup that fails with `ObjectDoesNotExist`.
3. In the browser’s dev tools → Network tab, observe the response:
   - The response body is JSON with an `error` field (coming from `AjaxErrorMiddleware.serialize_error`).
   - The HTTP status code is **404** (or **500** for internal errors).
4. On the page itself, observe that:
   - No alert is shown.
   - No visible feedback is given to the user.
   - `handle_success` in `ajax_tools.js` is never called because jQuery only calls that callback for 2xx responses.

By contrast, if you trigger an error that raises `AjaxErrorMiddleware.AjaxError` (which is sent with HTTP 200), the JSON is returned with status 200, `handle_success` runs, and the user sees the alert as expected.

### Expected Behavior

- All AJAX errors handled by `AjaxErrorMiddleware` should be surfaced consistently to the user when using the generic AJAX helpers.
- If the middleware returns a JSON object with an `error` field, the frontend should either:
  - Always receive it via the success callback (e.g. HTTP 200 everywhere, with `status` encoded in the JSON), or
  - Receive it via an explicit error callback for non‑2xx responses and display it in a similar way.
- From a user’s perspective, if an AJAX action fails, they should see a clear error message instead of the UI silently doing nothing.

### Actual Behavior

- Errors that go through AjaxErrorMiddleware.bad_request :
  
  ```
  def bad_request(self, request, 
  exception):
      return self.serialize_error(200, 
      exception.message)
  ```
  - Response: JSON with status: 200 , HTTP status code 200.
  - jQuery runs the success callback handle_success in ajax_tools.js .
  - handle_success sees 'error' in data and shows an alert.
  - The user sees an error message as intended.
- Errors that go through AjaxErrorMiddleware.not_found or server_error :
  
  ```
  def not_found(self, request, exception)
  :
      return self.serialize_error(404, 
      str(exception))
  
  def server_error(self, request, 
  exception):
      ...
      return self.serialize_error(500, _
      ('Internal error'))
  ```
  - Response: JSON with error and status 404/500; HTTP status code 404 or 500.
  - jQuery does not run the success callback for 404/500.
  - ajax_tools.js does not register an error callback ( .fail(...) ) for these helper calls.
  - Result: the user gets no alert and no visible feedback, even though a structured JSON error was returned.
This mismatch is also reflected in the tests in esp/esp/program/modules/tests/ajaxstudentreg.py , where expect_ajaxerror explicitly assumes status == 200 in the JSON for errors handled via AjaxErrorMiddleware .

### Screenshots

_No response_

### Operating System

macOS

### Browser

Chrome

### Additional Context

Relevant code locations:

- Middleware:
  - esp/esp/middleware/esperrormiddleware.py
    - AjaxErrorMiddleware.process_exception
    - AjaxErrorMiddleware.not_found / bad_request / server_error
- Frontend helpers:
  - esp/public/media/scripts/ajax_tools.js
    - handle_success
    - handle_submit (uses $j.get / $j.post with only a success handler)
- Tests that assume status 200 in JSON for AJAX errors:
  - esp/esp/program/modules/tests/ajaxstudentreg.py ( expect_ajaxerror )
Potential directions (leaving the choice to maintainers):

- Option A: Treat all AjaxErrorMiddleware responses as 200 at the HTTP level
  
  - Have not_found and server_error also return HTTP 200, but keep status inside the JSON as 404/500.
  - This matches the current frontend pattern and the existing tests that expect status == 200 in the JSON when AjaxErrorMiddleware is engaged.
- Option B: Keep 4xx/5xx semantics and update the frontend to handle them
  
  - Update ajax_tools.js to attach .fail(...) handlers that read the JSON error payload for non‑2xx responses.
  - Keep HTTP 404/500 for better semantics and logging, but still show a user‑visible error message based on the JSON.

Nakshatra480:
Closes #4279 

## Summary
This PR fixes the mismatch between `AjaxErrorMiddleware` responses and the generic AJAX helpers in `ajax_tools.js`. Previously, 4xx/5xx JSON error responses were silently dropped because the helpers only listened for success callbacks. It also prevents native form submissions after AJAX posts to avoid duplicate submissions.

## Changes
- Add a shared `handle_error` that reads JSON error payloads for non‑2xx responses.
- Attach `.fail(handle_error)` to AJAX requests in `handle_submit` and `fetch_fragment`.
- Return `false` from `handle_submit`, and propagate that return value through the callback wrappers so native form submission is blocked.

## Why
This makes AJAX error handling consistent across status codes and ensures errors from `AjaxErrorMiddleware` are actually visible to users.

### DIFF
--- a/esp/public/media/scripts/ajax_tools.js
+++ b/esp/public/media/scripts/ajax_tools.js
@@ -157,6 +157,7 @@ var handle_submit = function(mode, attrs, eventObject)
     if (req && req.fail) {
         req.fail(handle_error);
     }
+    return false;
 }
 
 
@@ -175,7 +176,7 @@ function CallbackForm(id, url)
 {
     this.id = id;
     this.url = url;
-    this.callback = function (e) {handle_submit('post', this, e)};
+    this.callback = function (e) {return handle_submit('post', this, e)};
 }
     
 function CallbackLink(id, url, content, post_form)
@@ -186,11 +187,11 @@ function CallbackLink(id, url, content, post_form)
     this.post_form = post_form;
     if (this.post_form)
     {
-        this.callback = function (e) {handle_submit('post', this, e)};
+        this.callback = function (e) {return handle_submit('post', this, e)};
     }
     else
     {
-        this.callback = function (e) {handle_submit('get', this, e)};
+        this.callback = function (e) {return handle_submit('get', this, e)};
     }
 }
 


Nakshatra480:
Hi @willgearty , the PR changes are ready for review. Can you review it and give feedback.
Thanks for your time.

willgearty:

willgearty:
@Nakshatra480 will you be getting back to this any time soon?

PR #4302, opened 2026-02-22
Fix overly restrictive registration redirect permission checks

### LINKED ISSUE #3495: Registration redirect permissions might be too restrictive?
After filling out a profile, we provide a list of programs that a user might want to register for (if there is only one program, we redirect them automatically to the registration of that program). However, it appears that we only populate this list of programs based on a simple set of permissions:
https://github.com/learning-unlimited/ESP-Website/blob/0cc9e6ac19b67d743b8747a8547fbd7a191f4c1f/esp/esp/web/views/main.py#L235-L249

My guess is that these might be too restrictive for some programs, especially now that there are multiple types of student registration now. However, I guess it's possible that these work for almost all cases? Should we expand the permissions that we check to include others (e.g., phase zero)

SuyashJain17:
## Description

This PR expands the permission checks used in the `registration_redirect` logic to ensure users see all programs they are legitimately eligible for.

Previously, program visibility relied on a narrow set of role-based permissions. With the introduction of additional student registration phases and teacher acknowledgment flows, this could result in eligible users not seeing relevant programs.

### Changes Made

**Students**
- Added support for `Student/PhaseZero` and `Student/Applications` in addition to `Student/Classes`.
- Preserved grade-based filtering when determining eligible programs.
- Unified permission checks using a loop for easier future extension.

**Teachers**
- Included `Teacher/Acknowledgement` alongside class creation permissions.
- Refactored permission lookup into a unified loop to simplify maintenance.

The existing behavior for already-registered/taught programs remains unchanged.

---

## Related Issue

Closes #3495

---

## Type of Change

- [x] Bug fix
- [x] Refactor / cleanup

---

## Testing

- [x] Existing tests pass
- [x] Manually verified

---

## AI Disclosure

AI tools were used for assistance in reviewing and refining the permission expansion logic. All implementation decisions and final code changes were manually verified.

---

## Checklist

- [x] Self-reviewed the code
- [ ] Updated documentation (not required)

### DIFF
--- a/esp/esp/web/views/main.py
+++ b/esp/esp/web/views/main.py
@@ -234,7 +234,7 @@ def registration_redirect(request):
         userrole['base'] = 'teach'
         userrole['reg'] = 'teacherreg'
         # check for programs where the teacher has any of the following permissions
-        teacher_perms = ['Teacher/Classes/Create/Class', 'Teacher/Classes/Create/OpenClass', 'Teacher/Acknowledgement']
+        teacher_perms = ['Teacher/Classes/Create/Class', 'Teacher/Classes/Create/OpenClass', 'Teacher/Availability']
         progs_set = set()
         for perm in teacher_perms:
             progs_set |= set(Permission.program_by_perm(user, perm))


esp/esp/web/views/main.py
```diff
@@ -233,8 +233,13 @@ def registration_redirect(request):
         userrole['name'] = 'Teacher'
         userrole['base'] = 'teach'
         userrole['reg'] = 'teacherreg'
-        # first check for which programs they can still create a class
-        progs = list(set(Permission.program_by_perm(user, 'Teacher/Classes/Create/Class')) | set(Permission.program_by_perm(user, 'Teacher/Classes/Create/OpenClass')))
+        # check for programs where the teacher has any of the following permissions
+        teacher_perms = ['Teacher/Classes/Create/Class', 'Teacher/Classes/Create/OpenClass', 'Teacher/Acknowledgement']
```
kkbrum: Availability comes before Acknowledgement, and is required, so I would use that as a permission check as opposed to Acknowledgement

SuyashJain17: Thanks for the suggestion! I’ve updated the permission check to use `Teacher/Availability` instead of `Teacher/Acknowledgement` in `esp/esp/web/views/main.py`, since availability comes first and is required.

Please let me know if any further adjustments are needed.

PR #4334, opened 2026-02-23
Fix statistics view registration type filters

### LINKED ISSUE #4314: statistics view code doesn't make sense
The `student_reg_types` and `teacher_reg_types` lines here don't make sense because the `if` statements have conditions of the form:

 `X and not X`

https://github.com/learning-unlimited/ESP-Website/blob/e2a77c27600c7ac02fc2d0c7a4e256954073250c/esp/esp/program/views.py#L1053-L1058

SuyashJain17:
## Description

Fix the statistics view’s registration type filters so that `student_reg_types` and `teacher_reg_types` are applied when present, instead of using a contradictory `X and not X` condition that prevented any filtering.

## Related Issue

Closes #4314

## Type of Change

- [x] Bug fix

## Testing

- [x] Manually verified

## Checklist

- [x] Self-reviewed the code


hwatheod (approved):
Thank you for doing this!

PR #4346, opened 2026-02-23
Fix coteacher POST trust issue in teacher class registration

### LINKED ISSUE #921: Coteachers minor security flaw
In programs/modules/handlers/teacherclassregmodule.py, around line 400, we have the following snippet:

```
    # set txtTeachers and coteachers....                                                          
    if not request.POST.has_key('coteachers'):
        coteachers = cls.get_teachers()
        coteachers = [ ESPUser(user) for user in coteachers
                       if user.id != request.user.id           ]

        txtTeachers = ",".join([str(user.id) for user in coteachers ])

    else:
        txtTeachers = request.POST['coteachers']
        coteachers = txtTeachers.split(',')
        coteachers = [ x for x in coteachers if x != '' ]
        coteachers = [ ESPUser(User.objects.get(id=userid))
                       for userid in coteachers                ]
        add_list_member("%s_%s-teachers" % (prog.program_type, prog.program_instance), coteachers)
```

This accepts any list of coteachers from a current teacher as long as the deadline is open. Any student or other individual could create a teacher account and add any other teacher accounts to the program list whether they have registered classes or not. This is a not a huge deal, but I don't see any reason why we should accept a user inputted list of coteachers, when we handle the lack of provided list just fine.

Flamki:
## Description
Fixes #921 by removing trust in client-supplied coteacher lists.

## What changed
- `TeacherClassRegModule.coteachers_logic()` now always derives coteachers from the database (`cls.get_teachers()`), never from `request.POST['coteachers']`.
- Removed hidden `<input name="coteachers">` fields from both coteacher templates:
  - `esp/templates/program/modules/teacherclassregmodule/coteachers.html`
  - `esp/templates/program/modules/adminclass/coteachers.html`
- Kept Mailman synchronization, but now sourced from server-side teacher data after POST actions.
- Added regression test in `esp/esp/program/modules/tests/teacherclassregmodule.py`:
  - `test_coteachers_post_data_is_not_trusted`
  - Verifies forged `coteachers` POST data does not trigger false duplicate checks.

## Why this matters
Previously, a user could inject arbitrary IDs in `coteachers` POST data. That list was trusted and used in coteacher handling and program teacher-list side effects. This patch makes server-side DB state authoritative.

## Testing
- `python -m py_compile esp/esp/program/modules/handlers/teacherclassregmodule.py esp/esp/program/modules/tests/teacherclassregmodule.py`

Closes #921


Flamki:
Hi @willgearty, I’ve addressed the Copilot feedback and rebased the branch on latest main.

Updates included:
- convert coteachers to a list before append usage (admin flow fix)
- clarify Mailman comment wording to match behavior
- keep server-side DB as source of truth (no trust in posted coteachers)
- regression test included for forged POST coteachers data

All required checks are passing (Lint + Unit Tests).  
Could you please review when you have time? Thanks!

willgearty (approved):
Thanks for cleaning this up, it works great!

PR #4368, opened 2026-02-24
try_multi: narrow retries to Exception and memcache errors

### LINKED ISSUE #4350: [Bug]: try_multi retries swallow exceptions (bare except) and can hide real failures
### Description

esp/esp/utils/try_multi.py implements a retry decorator that uses a bare except: inside the retry loop. This catches all exceptions (including BaseException subclasses like KeyboardInterrupt / SystemExit) and silently swallows them for the first n-1 attempts with no logging.

### Steps to Reproduce

Open esp/esp/utils/try_multi.py and confirm the retry loop uses a bare except:.

Run this minimal reproduction snippet (can be run in any Python environment where the module is importable):
from esp.utils.try_multi import try_multi

calls = {"n": 0}

@try_multi(3)
def flaky_bug():
    calls["n"] += 1
    # Simulate a real programmer error on the first call
    if calls["n"] == 1:
        raise TypeError("real bug that should not be retried")
    return "ok"

print(flaky_bug())     # Prints "ok" even though a TypeError occurred
print(calls["n"])      # 2 (shows it retried after swallowing the TypeError)

### Expected Behavior

try_multi should not catch BaseException (so interrupts/exits aren’t swallowed).

By default, it should catch only Exception (or, even better, only explicitly provided exception types).

Unexpected exceptions like TypeError should raise immediately (no retry), unless explicitly included in a retryable exceptions tuple.

When retries are exhausted, the last exception should be logged and re-raised.

### Actual Behavior

try_multi catches everything due to bare except:.

Exceptions are silently swallowed for the first n-1 attempts (no logging).

Non-transient exceptions (including programmer errors) can be retried and hidden, causing confusing behavior and delayed diagnosis.

### Screenshots

_No response_

### Operating System

_No response_

### Browser

_No response_

### Additional Context

_No response_

Mashikha1:
## Description
Fixes esp/esp/utils/try_multi.py retry decorator so it no longer uses a bare except: (which can swallow BaseException like KeyboardInterrupt/SystemExit and hide unexpected errors). The decorator now catches retryable exceptions explicitly, logs the final failure, and re-raises after retries are exhausted. Added unit tests to verify retry behavior and that unexpected exceptions are not retried or swallowed.

## Related Issue

Closes #4350

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [ ] Existing tests pass
- [x] Manually verified

## Checklist

- [x] Self-reviewed the code


willgearty:
@Mashikha1 please address the copilot review and fix the merge conflicts.

PR #4371, opened 2026-02-24
Fix: handled undefined greeting in loginbox (#4337)

### LINKED ISSUE #4337: [Bug]: Fix: Navigation greeting displays "undefined" for users with empty profile names.
### Description

The Bug:
* When a logged-in user has not yet completed their profile (specifically leaving First and Last Name fields empty), the navigation greeting incorrectly displays "Hello, undefined undefined!" instead of a clean fallback like the username or "Guest".

Image

Image

### Steps to Reproduce

* Clear User Data: Access the Admin panel at /admin/users/espuser/, delete the "First Name" and "Last Name" for your user account, and click Save.

* Navigate to Profile Page: Return to the main application and navigate to the profile URL: http://localhost:8000/myesp/profile/.

* Verify Empty Fields: Confirm that the "First Name" and "Last Name" input fields on the profile form are currently empty.

* Observe the Bug: Look at the navigation greeting in the top header. It will display the text "Hello, undefined undefined!" instead of a proper fallback.

### Expected Behavior

* The greeting should implement a robust fallback hierarchy to ensure a professional user experience regardless of profile completeness.
* Primary: If the first_name is present, it should display "Hello, [First Name]!".
* Secondary Fallback: If the first_name is missing (even if the last_name exists), it should automatically fall back to the username (e.g., "Hello, vipul7917!") to avoid displaying the "undefined" string.
* Final Fallback: If for any reason the user is not authenticated or the username is unavailable, it should default to a generic "Hello, Guest!".
* The system should never render the literal string "undefined" in the UI.

### Actual Behavior

* The navigation header greeting displays the literal string "Hello, undefined undefined!" when a user is logged in but has not yet provided a First or Last name in their profile. Instead of gracefully falling back to a username or hiding the name, the UI renders "undefined" directly from the JavaScript variables.

### Screenshots

Attached in the description and steps above.

### Operating System

 * Ubuntu 24.04 (via Vagrant/VirtualBox)

### Browser

* Brave Browser

### Additional Context

* I used grep to identify the source in the template files.
* Location: esp/templates/users/loginbox_content.html
* Line: Hello, <span id="user_first_name"></span>

Image

vipulpandey21:
## Description

Fixed the 'undefined' greeting bug in the navbar. The greeting now shows the user's first name if available, falls back to the username for incomplete profiles, and displays "Guest" when neither value is present. The literal string "undefined" will no longer appear in the UI. Verified locally on Vagrant.

## Related Issue

Closes #4337 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [ ] Existing tests pass
- [x] Manually verified

## AI Disclosure
I used Gemini to assist with identifying the Django template logic and debugging the local Vagrant environment setup.

## Checklist

- [x] Self-reviewed the code

### DIFF
--- a/esp/esp/themes/theme_data/droplets/templates/users/loginbox_content.html
+++ b/esp/esp/themes/theme_data/droplets/templates/users/loginbox_content.html
@@ -6,8 +6,7 @@
   <li>
       <li class="dropdown">
         <a class="navbar-header" id="loginbox_user_name" data-toggle="dropdown">
-          Hello, <span id="user_first_name"></span>
-          <span id="user_last_name"></span>!
+          Hello, <span id="user_first_name"></span>!
         </a>
         <a class="dropdown-toggle caret-wrapper" data-toggle="dropdown"><b class="caret"></b></a>
         <ul class="dropdown-menu pull-right">


vipulpandey21:
Hi Mr. Will @willgearty , I have manually verified these changes on a local Vagrant environment. It correctly handles the 'undefined' greeting by falling back to the username when first/last names are missing, and shows 'Guest' for unauthenticated users.

esp/templates/users/loginbox_content.html
```diff
@@ -8,10 +8,18 @@
   <div class="logged_in hidden">
     <ul class="nav">
       <li>
-          <a id="loginbox_user_name">
-              Hello, <span id="user_first_name"></span>
-              <span id="user_last_name"></span>!
```
willgearty, line 13: I think we originally had some javascript populating these fields. If we're not using it anymore, can the javascript just be removed?

vipulpandey21: Hi Mr. Will @willgearty  ! I’ve updated the PR to remove the obsolete logic in user_classes.js as you suggested.
Moving the greeting logic to the Django template (falling back from First Name -> Username -> Guest) makes the header more robust and completely removes the 'undefined' flash on page load. I’ve manually verified on my local Vagrant environment that the greeting works perfectly without the legacy JavaScript. Ready for your review!

willgearty:
Looks like two of our tests are currently failing: https://github.com/learning-unlimited/ESP-Website/actions/runs/22379486492/job/64776858353?pr=4371.

Please update them or change your code accordingly.

vipulpandey21:
Hi Mr. Will @willgearty, Apologies for the delayed response—I have been caught up with my semester exams.

I've looked at the failing tests (testCatalog and testQSD). It seems like moving the greeting logic to the Django template is causing a session/cookie conflict in those specific views. I am currently investigating the code in esp/web/tests.py and will push a fix to address these test failures shortly. Thanks for your patience Mr. Will!

vipulpandey21:
Hi Mr. Will @willgearty,
I reverted the template to match upstream exactly to ensure no session access occurs during rendering (so caching remains intact).
The fix for the “undefined” greeting is implemented purely in user_classes.js, and I confirmed locally that NoVaryOnCookieTest now passes and the greeting fallback works as expected.
Would appreciate a review when you get a chance.

vipulpandey21:
Hi Mr.Will @willgearty, 
Thanks for pointing this out — you're right that unescape(undefined) in user_data.js can produce the literal string "undefined".

For this PR, I focused on ensuring the greeting rendering logic never displays "undefined" in the UI.

Since user_data.js is shared across multiple areas, I avoided modifying it here to keep the scope limited.
Would you prefer that I address the user_data.js guard in this PR as well, or handle that separately?

willgearty (changes requested):
This works great, thanks so much for the contribution! Just one issue that needs to be resolved:

1. We no longer use the last name field in the templates, so there is now a blank space between the name and the exclamation point on most themes. Please either fix your solution here or update the templates accordingly (you'll possibly need to update the template for each theme).

vipulpandey21:
Hi Mr. Will @willgearty, 
Thanks for pointing that out. I've removed the unused last name span from the greeting template to eliminate the extra space before the exclamation point across themes.
Please let me know if there’s anything else you’d like adjusted.

willgearty:
Looks good on the base theme, but there's still the space on other themes. You'll need to update their respective loginbox_content.html templates:

- https://github.com/learning-unlimited/ESP-Website/blob/main/esp/esp/themes/theme_data/bigpicture/templates/users/loginbox_content.html
- https://github.com/learning-unlimited/ESP-Website/blob/main/esp/esp/themes/theme_data/circles/templates/users/loginbox_content.html
- https://github.com/learning-unlimited/ESP-Website/blob/main/esp/esp/themes/theme_data/droplets/templates/users/loginbox_content.html

(I think it's just those three, but you should double check the other themes too)

vipulpandey21:
Hi Mr. Will @willgearty  Thanks for pointing that out!
Sorry for the delay — I wanted to fully verify the whitespace issue across themes before pushing additional changes. I tested the greeting at the template level, DOM level, and runtime to make sure there were no trailing spaces or newline artifacts.
I’ve now updated the affected themed loginbox_content.html templates (bigpicture, circles, droplets) to ensure consistency across themes. Please let me know if you’d like me to check any others as well.

It’s quite late on my end now, so I’ll get some rest — but I’ll check back in a few hours and respond promptly if there’s any further feedback.
Thanks again for your time and guidance!

esp/esp/themes/theme_data/bigpicture/templates/users/loginbox_content.html
```diff
@@ -4,14 +4,16 @@
   <div class="logged_in hidden">
     <form class="navbar-form form-inline">
       <a href="/myesp/profile/" class="username">
-        <i class="icon-user"></i> Hello,
-        <span id="user_first_name"></span> <span id="user_last_name"></span>!
+        <i class="icon-user"></i> Hello, <span id="user_first_name"></span> <span id="user_last_name"></span>!
```
willgearty: This still has the `user_last_name` span.

esp/esp/themes/theme_data/droplets/templates/users/loginbox_content.html
```diff
@@ -3,42 +3,46 @@
...
-          <span id="user_last_name"></span>!
-        </a>
-        <a class="dropdown-toggle caret-wrapper" data-toggle="dropdown"><b class="caret"></b></a>
-        <ul class="dropdown-menu pull-right">
-          <li><a href="/myesp/accountmanage/">Manage Account</a></li>
-        </ul>
-      </li>
+  <li class="dropdown">
+    <a class="navbar-header" id="loginbox_user_name" data-toggle="dropdown">
+      Hello, <span id="user_first_name"></span> <span id="user_last_name"></span>!
```
willgearty: This still has the `user_last_name` span.

esp/esp/themes/theme_data/droplets/templates/users/loginbox_content.html
```diff
@@ -3,42 +3,46 @@
...
-          Hello, <span id="user_first_name"></span>
-          <span id="user_last_name"></span>!
-        </a>
-        <a class="dropdown-toggle caret-wrapper" data-toggle="dropdown"><b class="caret"></b></a>
-        <ul class="dropdown-menu pull-right">
-          <li><a href="/myesp/accountmanage/">Manage Account</a></li>
-        </ul>
-      </li>
+  <li class="dropdown">
+    <a class="navbar-header" id="loginbox_user_name" data-toggle="dropdown">
```
willgearty: Why did you change this?

esp/esp/themes/theme_data/bigpicture/templates/users/loginbox_content.html
vipulpandey21: Thanks for pointing that out. You're right — the `user_last_name` span was still present in the template. I've now removed it so the greeting only uses `user_first_name` in both `bigpicture` and `droplets`. Please let me know if anything else needs adjustment.

esp/esp/themes/theme_data/droplets/templates/users/loginbox_content.html
vipulpandey21: Good catch — that change wasn’t intentional. It happened when I restored the droplets template to match the upstream version while fixing the greeting issue. The only intended change for this PR is removing the `user_last_name` span. If you'd prefer, I can revert the structure so the diff only includes that change.

vipulpandey21:
Hi Mr. Will @willgearty,  Thanks for the feedback earlier. The intermediate commit was to restore the droplets template to the upstream version, and the latest commit now only removes the `user_last_name` span as intended. So overall the final state should match the expected change in both templates. Could you please take another look and let me know if anything else still needs adjustment?

vipulpandey21:
Mr. Will @willgearty, Thank you for the review and guidance throughout the process, **Will — I really appreciate it**. Since my current focus is on regression test expansion, would it make sense to **add a regression test** for the behavior fixed in #4337 to ensure it doesn’t reappear in the future? Happy to open a small follow-up PR if that would be helpful.

willgearty:
No, that's ok, thanks though 👍

PR #4438, opened 2026-02-26
Fix #2299: Recovery of QuasiStaticData depends on the "default" NavBarCategory

### LINKED ISSUE #2287: Deleting NavBarCategory called "default" causes errors everywhere
It's not at all obvious that you shouldn't do this. If we're going to depend on it, we should get_or_create it or something.

### LINKED ISSUE #2299: Restoring deleted QSD pages doesn't work
If, for example, one delete NavBars categories (see #2287), Editable pages are deleted with it. There is a button that is supposed to restore these pages, called "Recover deleted Editables" on `/admin/qsd/quasistaticdata/`, but it does not do anything when clicked. Even if one navigates directly to the URL `/admin/qsd/quasistaticdata/recover/` that this button is meant to link to, the indices of the deleted pages are not the same as the original indices (which may not be a problem itself?), but more to the point, trying to restore the deleted pages from that page does not work, "especially" if it was a page like FAQ that might not have originally had a Title or Content when it was created with `install.py`.

arpit10128:
## Description

The system assumes that a `NavBarCategory` named `"default"` always exists. This assumption is reflected in:

- `default_navbarcategory()` (which explicitly fetches `name='default'`)
- The install logic that creates a default category
- Recovery behavior that falls back to the default category

Allowing deletion of the `"default"` category breaks this invariant and makes recovery of related `QuasiStaticData` objects impossible.

**Implemented Solution**

Rather than modifying the model or recovery logic, I enforced the existing architectural assumption by preventing deletion of the `"default"` `NavBarCategory` via the admin interface.

Specifically:

- Overrode `has_delete_permission()` in `NavBarCategoryAdmin`
- Blocked deletion when `obj.name == "default"`
- Left all other categories fully deletable
- Added tests to verify:

1. `"default"` cannot be deleted
2. Non-default categories remain deletable

This approach preserves the current design, avoids schema changes, and prevents the recovery failure scenario while keeping behavior consistent for other categories.

## Related Issue

Closes #2299 Closes #2287

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

## AI Disclosure

It was used to assist with generating and refining the solution and test cases.
All changes were manually reviewed, tested locally (including Docker), and verified to work correctly before submission.

## Checklist

- [x] Self-reviewed the code

### DIFF
--- a/esp/esp/web/admin.py
+++ b/esp/esp/web/admin.py
@@ -50,7 +50,6 @@ class NavBarCategoryAdmin(admin.ModelAdmin):
                 level = messages.ERROR,
             )
         return super().delete_model(request, obj)
-    
     def has_delete_permission(self, request, obj = None):
         if obj and obj.name == "default":
             return False


esp/esp/web/admin.py
```diff
@@ -33,12 +33,28 @@
...
+class NavBarCategoryAdmin(admin.ModelAdmin):
+    def delete_model(self, request, obj):
+        if obj.name == "default":
+            self.message_user(
+                request,
+                'The "default" nav category cannot be deleted because it is required by the system.',
+                level = messages.ERROR,
+            )
+        return super().delete_model(request, obj)
+
```
willgearty:
```suggestion

```

arpit10128:
> Works great! I especially like that it doesn't even show the delete button when you open the default object. Thanks for your contribution!

Thank you so much for the review and the kind words!
Glad the solution works well. Happy to contribute — looking forward to working on more improvements.

PR #4455, opened 2026-02-26
Refactor sorting logic in sorttable.js for improved performance

### LINKED ISSUE #3398: sorttable doesn't work properly if data in a sorted column is updated after sorting
When you sort by a column with sorttable, it inserts a class "sorttable_sorted" for that `<th>` tag. If you then click the same column, it reverses the sort, and the class changes to "sorttable_sorted_reverse".

However, if you *update* the data in that column after you have sorted, the class "sorttable_sorted" or "sorttable_sorted_reverse" remains.  So sorttable thinks the column is still sorted, even though it (probably) isn't now because you updated the data. If you click on that same column again, sorttable will simply reverse the rows in the dataset, because it thinks it's still sorted. So it isn't sorted properly.

This can be tested on the section attendance page. If you sort by section attendance, then randomly mark/unmark a bunch of students as "attending", then sort again, it will just reverse the rows, whereas we expect all the "attending" students to be together.

If there are other places on the site using sorttable, this bug could be there too.

mrkeshav-05:
## Description
`sorttable.js` previously just reversed the row order when clicking an already-sorted column. If data in that column changed after sorting (e.g., toggling attendance checkboxes), the reversed order was incorrect. Now every click does a full re-sort using current cell values, so the table is always sorted correctly.

## Closes: 

Closes #3398 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

## Checklist

- [x] Self-reviewed the code

### DIFF
--- a/docs/dev/contributing.rst
+++ b/docs/dev/contributing.rst
@@ -38,29 +38,29 @@ From the directory ``/esp``: ::
 Write some code!
 Test your code!
 
-Look at what you’ve changed (``git status`` and/or ``git diff``), and then run ``git commit -a -m``, and type a commit message (see `<http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html>` for good commit message style), like so: ::
+Look at what you've changed (``git status`` and/or ``git diff``), and then run ``git commit -a -m``, and type a commit message (see `<http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html>` for good commit message style), like so: ::
 
   git commit -a -m "Hello world"
 
-Repeat a few times from “Write some code!” if you want to make multiple commits.
+Repeat a few times from "Write some code!" if you want to make multiple commits.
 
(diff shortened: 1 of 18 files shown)


hwatheod:
My testing shows that it's not working correctly.

On the section attendance page, at the URL that looks like this

    onsite/Splash/2026_Spring/section_attendance/120?secid=90

Here is what I have:

Screenshot 2026-02-27 at 23 43 49

If I click the "section attending" once, then the check marks in the last column go to the top, as expected.

However, now if I click "section attending" again, it should revert to the initial state. But instead it looks like:

Screenshot 2026-02-27 at 23 45 43

Now the checked sections are first and last.

willgearty:
@mrkeshav-05 Please address the reviews above.

PR #4473, opened 2026-02-27
Fix csv.writer crash in Python 3 caused by binary file mode

### LINKED ISSUE #4437: [Bug]: csv.writer crashes in Python 3 due to binary mode ('ab+') in studentregsanity.py
### Description

Hey team! While looking through esp/esp/program/controllers/studentregsanity.py, I noticed that CSV log files are being opened in append-binary mode ('ab+') before being passed to csv.writer().

Since the migration to Python 3, csv.writer expects text/strings, not bytes. Passing it a file opened in binary mode causes an immediate TypeError crash when it tries to write a row.

### Steps to Reproduce

1. Trigger the sanitize_walkin or sanitize_lunch methods in StudentRegSanityController.
2. Ensure it runs with logging enabled (csvlog=True and fake=False).
3. The script will open a log file using open(filefullname, 'ab+').
4. See the TypeError: a bytes-like object is required, not 'str' crash when csvwriter.writerow(...) is called.

### Expected Behavior

The script should successfully append rows to the CSV log file without throwing type errors.

### Actual Behavior

The script crashes with a TypeError because Python 3's csv.writer cannot write strings to a file opened in binary mode.

### Screenshots

_No response_

### Operating System

_No response_

### Browser

_No response_

### Additional Context

Proposed fix:

Replace:
open(filefullname, 'ab+')

With:
open(filefullname, 'a', newline='', encoding='utf-8')

Vipeen-Kumar:
## Description

This PR fixes a Python 3 compatibility issue where CSV log files were opened in binary append mode (`ab+`) in `studentregsanity.py`.

Since `csv.writer` in Python 3 expects text streams instead of bytes, this caused a `TypeError` when writing rows.

The fix updates file opening to text append mode with newline handling and UTF-8 encoding:

```
open(filefullname, 'a', newline='', encoding='utf-8')
```

This ensures `csv.writer` writes correctly without runtime crashes while preserving existing behavior.

## Related Issue

Closes #4437

## Type of Change

* [x] Bug fix
* [ ] New feature
* [ ] Breaking change
* [ ] Documentation update
* [ ] Refactor / cleanup
* [ ] CI/CD or build change

## Testing

* [x] Existing tests pass
* [x] Manually verified

## AI Disclosure

AI assistance was used for guidance in drafting the issue description and verifying the Python 3 compatibility approach. All code changes were reviewed and validated manually.

## Checklist

* [x] Self-reviewed the code


esp/esp/program/controllers/studentregsanity.py
```diff
@@ -61,14 +61,15 @@ def __init__(self, program, **kwargs):
...
         category_walkin = ClassCategories.objects.get(category="Walk-in Activity")
         if csvlog and not(fake): #If I'm actually doing things, and I want a log....
             import csv
             if csvwriter is None:
                 closeatend = True
                 if directory is None: directory = self.options['directory']
-                filefullname = directory +'/santitize_walkins_log.csv'
-                csvfile = open(filefullname, 'ab+')
+                filefullname = os.path.join(directory, 'sanitize_walkins_log.csv')
+                csvfile = open(filefullname, 'a', newline='', encoding='utf-8')
```
userAdityaa: I noticed that files are still being opened using `open()` with manual closing via the `closeatend` flag. Would it make sense to switch these to context managers (with open(...) as csvfile:)?

This would ensure files are properly closed even if an exception occurs during processing and would simplify the control flow by removing the need for closeatend.

esp/esp/program/controllers/studentregsanity.py
```diff
@@ -78,27 +79,29 @@ def sanitize_walkin(self, fake=True, csvwriter=None, csvlog=False, directory=Non
...
-            if not fake:
-                if csvlog: csvwriter.writerow([w.title().encode('ascii', 'ignore'), ', '.join(sec.friendly_times()), sr.user.name().encode('ascii', 'ignore'), sr.relationship.__str__().encode('ascii', 'ignore')])
-                sr.expire()
+                for sr in srs:
+                    if not fake:
+                        if csvlog:
+                            csvwriter.writerow([w.title(), ', '.join(sec.friendly_times()), sr.user.name(), str(sr.relationship)])
+                        sr.expire()
         logger.debug(report)
         if closeatend: csvfile.close()
```
userAdityaa: One small defensive suggestion: since csvfile is initialized as None, it might be safer to guard the close call with if closeatend and csvfile:. I know this wasn’t introduced in this PR and already exists in the original code, but adding this check could help prevent potential issues if the control flow changes in the future.

Vipeen-Kumar:
Thanks for the review, @userAdityaa! These were great catches.

I completely agree with using a context manager here. Since csvwriter can optionally be passed in from the caller, wrapping the existing logic directly in a with block would have forced some pretty deep, ugly nesting.

To keep it clean, I extracted the core loop logic into private helper methods (_run_walkin_check, etc.). This allowed me to use with open(...) as csvfile: cleanly, which completely removes the need for the closeatend flag and naturally resolves the None initialization edge case you mentioned below.
+1

I've pushed the updated changes. Let me know if this new structure looks good to you!

samalarussa (approved):

PR #4513, opened 2026-02-28
Fix #234: prevent /tmp permission failures with per-UID tempdir

### LINKED ISSUE #234: Tests that make use of /tmp fail in certain situations
Specifically, if you execute runserver, and then run the tests without first deleting the folder that's created, or changing it's permission in a certain way, you will get: `OSError: [Errno 13] Permission denied: '/tmp/esptmp__testsiteESP/tmpccMHLF'`. This happens for 32 tests for me. It seems to be because the webserver that Django uses changes the permissions on the /tmp/esptmp__<name> folder to be owned by www-data, which the tester user then doesn't have access to (unless you run with sudo I think). Is this just an unavoidable Django flaw, or can we make it so that the permissions on the folder are different? This is not high priority at all, it just bugged me after a while so I thought I'd note it here.

pushkarkumar06:
- settings.py: append os.getuid() to tempdir name so each OS user (e.g. www-data vs developer) gets their own isolated subdirectory under /tmp, eliminating [Errno 13] Permission denied errors.

- web/tests.py: replace string concatenation with os.path.join() for closure temp file paths to fix missing path separator bug.

Fixes #234

## Description

When `runserver` was executed as one OS user (e.g. `www-data`) before running tests as a different user, Django's shared tempdir under `/tmp/esptmp__<CACHE_PREFIX>` was owned by the first user. Any subsequent `tempfile` call by the test runner user hit `[Errno 13] Permission denied`. This affected 32 tests at the time of the original report.

Fix: append `os.getuid()` to the tempdir name so each OS user gets their own private subdirectory (e.g. `/tmp/esptmp__xyz_1000`), eliminating the cross-user ownership collision entirely without changing any other logic.

Also fixed a missing `/` path separator bug in `esp/web/tests.py` where `tempfile.gettempdir() + 'closure.out'` produced a malformed path.

## Related Issue

Closes #234

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

## AI Disclosure

Claude Sonnet (via GitHub Copilot) was used to assist with identifying the root cause and drafting the fix. All suggested changes were manually reviewed and verified before being committed to the codebase.

## Checklist

- [x] Self-reviewed the code


pushkarkumar06:
Hi @willgearty,

I've completed the investigation and fix for Issue #234 — "Tests that make use of /tmp fail in certain situations."

---

**Findings:**
After auditing the full codebase, the root cause was confirmed to still exist in `esp/esp/settings.py`. The settings module creates a shared tempdir at `/tmp/esptmp__<CACHE_PREFIX>` owned by whichever OS user runs first (e.g. `www-data` via `runserver`). Any test runner process belonging to a different user would then hit `[Errno 13] Permission denied` on all `tempfile` operations.

No tests in `esp/program/` used `/tmp` directly, but `esp/web/tests.py` had an additional minor path separator bug.

---

**Changes Made:**

1. `esp/esp/settings.py` — appended `os.getuid()` to the tempdir name, giving each OS user their own isolated subdirectory (e.g. `/tmp/esptmp__xyz_1000`). This eliminates the cross-user ownership collision without changing any other logic.

2. `esp/esp/web/tests.py` — replaced `tempfile.gettempdir() + 'closure.out'` with `os.path.join(tempfile.gettempdir(), 'closure.out')` to fix a missing path separator.

---

**PR:** #4513

Please let me know if any changes are needed before review.

Thanks!
Pushkar

willgearty:

PR #4531, opened 2026-02-28
Fix/attendance sort race condition

### LINKED ISSUE #3394: Students are being automatically marked as attending a class
In the last 2 Rainstorms (May 2021, Aug 2021), as a moderator, I observed students showing up on the class attendance page as attended (the "attending section" is checked) even though I did not check that box. We should figure out how this is happening.

SuyashJain17:
## Description

Fixes a race condition on the class attendance page where students could appear as “Attending” even when no valid `Attended` registration existed.

The issue was likely caused by sorting the attendance table while an attendance update was being processed via AJAX. If the table was reordered before the AJAX callback completed, the checkbox state could be applied to the wrong row, causing UI inconsistency.

This PR:

- Ensures the attendance table has a proper `<thead>` and `<tbody>` structure for stable sorting behavior.

- Prevents sorting while an attendance AJAX request is in progress by introducing a temporary `sorttable-busy` guard.

- Restores sorting functionality after the AJAX request completes (both success and error paths).

 These changes stabilize DOM behavior and ensure checkbox state always corresponds to the correct student.

## Related Issue

Closes #3394

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

## AI Disclosure

AI was used to help structure the PR description and refine explanation wording. All debugging, root cause analysis, and code changes were performed manually.

## Checklist

- [x] Self-reviewed the code

### DIFF
--- a/esp/public/media/scripts/program/modules/attendance.js
+++ b/esp/public/media/scripts/program/modules/attendance.js
@@ -50,6 +50,8 @@ $j(function(){
         var $checkedin = $me.closest("tr").find("[name=checkedin]");
         var $table = $me.closest("table.sortable");
         $me.prop('disabled', true);
+        var inFlight = ($table.data('in-flight') || 0) + 1;
+        $table.data('in-flight', inFlight);
         $table.addClass('sorttable-busy');
         markAttendance(username, secid, !checked, function(response) {
             if (response.error) {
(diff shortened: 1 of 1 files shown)


kkbrum:
What do you mean by "race"?

SuyashJain17:
By “race condition” I mean a timing issue between the AJAX attendance update and table sorting.

If the table is sorted while the AJAX request is still processing, the rows in the DOM can move. When the callback then updates the checkbox state, it may end up applying the update to the wrong row.

That could explain why a student sometimes appears as “Attending” even though no `Attended` registration exists.

This PR prevents sorting while the AJAX request is in progress so the update is applied to the correct row.

Issue #4530, opened 2026-02-28
[Bug]: Admin Toolbar Overlaps Page Content on Theme Page


prem-2006:
### Description

When logged in as an admin and navigating to /themes/, the admin toolbar dropdown overlaps and hides underlying page content instead of properly pushing it downward or staying layered correctly.

The dropdown menu (Admin Toolbar → Other Important Links) visually covers the content below it, making parts of the page unreadable and inaccessible.

### Steps to Reproduce

Run the site locally.

Log in as an admin user.

Navigate to /themes/.

Click Admin Toolbar.

Observe that the dropdown overlaps and hides the page content.

### Expected Behavior

<img width="696" height="471" alt="Image" src="https://github.com/user-attachments/assets/1030dffa-07e8-4916-b63b-141b7170d806" />

The admin toolbar dropdown should:

Either push page content downward

Or remain properly layered without obstructing readable content

Maintain clean layout alignment

### Actual Behavior

<img width="615" height="487" alt="Image" src="https://github.com/user-attachments/assets/720cc113-7a83-476b-958e-1f41da604fdd" />

The dropdown uses positioning that causes it to overlap page elements, making content partially hidden.

Environment

OS: Windows 11

Browser: Chrome

Project running via Docker

Current theme: barebones

prem-2006:
Hi @willgearty I looked into this and was able to reproduce the toolbar overlap issue locally also I have worked on it.

PR #4532, opened 2026-02-28
Fix admin toolbar overlapping page content (#4530)

### LINKED ISSUE #3919: Admin toolbar covers content in Barebones
The admin toolbar has a tendency to cover content when the barebones theme is installed (may be an issue with other themes as well)

prem-2006:
### Summary

This PR fixes a layout issue where the admin toolbar dropdown was overlapping and hiding page content on the `/themes/` page.

### Changes Made

- Adjusted CSS positioning to prevent the toolbar from overlapping content.
- Ensured the layout remains stable and unaffected in other areas.

### Testing

- Reproduced the issue locally.
- Verified that the toolbar no longer overlaps content.
- Confirmed no visible layout regressions.

Fixes #3919

### DIFF
--- a/esp/public/media/default_styles/admin_bar.css
+++ b/esp/public/media/default_styles/admin_bar.css
@@ -3,7 +3,7 @@
 	border: 1px solid #AAAAAA;
 	left: 20px;
 	position: absolute;
-	top: 10px;
+	top: 65px;
 	width: 200px;
 	z-index: 1000000;
 	text-align: center;


willgearty:
Please remove the commits that are unrelated to this PR.

willgearty:
Please share before and after screenshots to demonstrate the issue is fixed.

prem-2006:
> Please share before and after screenshots to demonstrate the issue is fixed.

Fixed image:
image

Actual image:
Screenshot 2026-03-19 221309

esp/public/media/default_styles/admin_bar.css
```diff
@@ -1,68 +1,63 @@
+:root {
+	/* Offset to prevent the admin bar from overlapping with themes' headers */
+	--adminbar-offset-top: 55px;
+}
+
 #adminbar {
 	background: none repeat scroll 0 0 #EEEEEE;
 	border: 1px solid #AAAAAA;
-	left: 20px;
+	right: 20px;
```
willgearty: This completely changes which side of the screen the admin toolbar is on, which will interfere with the muscle memory of existing users. Furthermore, it now overlaps with an element on the bigpicture theme.

This isn't obvious, but the entire admin toolbar can be collapsed by clicking on the "Admin Toolbar" header. So I don't think a dramatic shift is really necessary here, especially if we're just fixing this because of the barebones theme, which no one currently uses.

prem-2006:
image

HI @willgearty, I think now its look good if further change is needed please let me know.

PR #4553, opened 2026-03-01
Fix AJAX helpers to surface 4xx/5xx JSON errors (Fixes #4279)

### LINKED ISSUE #4279: [Bug]: AjaxErrorMiddleware returns 4xx/5xx for JSON errors, but frontend only handles 2xx responses

Asmi768:
## Description

This PR fixes an issue where AJAX error responses returned with HTTP 4xx/5xx
status codes were not surfaced to users.

Previously, ajax_tools.js only handled 2xx responses via the success callback.
When AjaxErrorMiddleware returned structured JSON errors with proper HTTP
status codes (e.g., 404 or 500), jQuery did not trigger the success handler,
causing the UI to silently ignore those errors.

This change:

- Attaches `.done()` and `.fail()` handlers to AJAX requests
- Parses structured JSON error payloads from non-2xx responses
- Displays user-visible error messages consistently
- Preserves correct HTTP semantics (no backend changes required)

Users now receive clear feedback for all AJAX failures.

---

## Related Issue

Closes #4279

---

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

---

## Testing

- [x] Existing tests pass
- [x] Manually verified

Manual verification included:
- Triggering AJAX request with valid data (2xx response works as before)
- Triggering 404 response (error displayed correctly)
- Triggering 500 response (error displayed correctly)
- Confirming no silent failures occur

---

## AI Disclosure

- Tool used: ChatGPT (OpenAI GPT-5)
- Scope of assistance: Refactoring guidance and structured error-handling logic suggestions
- Human verification: All code was reviewed, modified, and validated manually. Final implementation decisions and architectural choices were made independently.

---

## Checklist

- [x] Self-reviewed the code

### DIFF
--- a/esp/public/media/scripts/barcodescanner.js
+++ b/esp/public/media/scripts/barcodescanner.js
@@ -52,7 +52,7 @@ $j(function() {
                 },
             }, function(err) {
                   if (err) {
-                      console.log(err);
+                      log(err);
                       return
                   }
                   Quagga.start();
--- a/esp/public/media/scripts/commpanel.js
+++ b/esp/public/media/scripts/commpanel.js
@@ -259,27 +259,25 @@ function initialize()
 
     for (var i = 0; i < list_names.length; i++)
     {
-        with ({list_name: list_names[i]})
-        {
-            //  Make the ANDs turn off the ORs and vice versa
-            $j("input[name=checkbox_and_" + list_name + "]").on('change', function () {
-                if ($j("input[name=checkbox_and_" + list_name + "]").prop("checked")
-                    && $j("input[name=checkbox_or_" + list_name + "]").prop("checked"))
-                    $j("input[name=checkbox_or_" + list_name + "]").click();
-            });
-            $j("input[name=checkbox_or_" + list_name + "]").on('change', function () {
-                if ($j("input[name=checkbox_and_" + list_name + "]").prop("checked")
-                    && $j("input[name=checkbox_or_" + list_name + "]").prop("checked"))
-                    $j("input[name=checkbox_and_" + list_name + "]").click();
-            });
-            //  NOT can't be selected by itself
-            $j("input[name=checkbox_not_" + list_name + "]").on('change', function () {
-                if ($j("input[name=checkbox_not_" + list_name + "]").prop("checked")
-                    && !$j("input[name=checkbox_and_" + list_name + "]").prop("checked")
-                    && !$j("input[name=checkbox_or_" + list_name + "]").prop("checked"))
-                    $j("input[name=checkbox_and_" + list_name + "]").click();
-            });
-        }
+        let list_name = list_names[i];
+        //  Make the ANDs turn off the ORs and vice versa
+        $j("input[name=checkbox_and_" + list_name + "]").on('change', function () {
+            if ($j("input[name=checkbox_and_" + list_name + "]").prop("checked")
+                && $j("input[name=checkbox_or_" + list_name + "]").prop("checked"))
+                $j("input[name=checkbox_or_" + list_name + "]").click();
+        });
+        $j("input[name=checkbox_or_" + list_name + "]").on('change', function () {
+            if ($j("input[name=checkbox_and_" + list_name + "]").prop("checked")
+                && $j("input[name=checkbox_or_" + list_name + "]").prop("checked"))
+                $j("input[name=checkbox_and_" + list_name + "]").click();
+        });
+        //  NOT can't be selected by itself
+        $j("input[name=checkbox_not_" + list_name + "]").on('change', function () {
+            if ($j("input[name=checkbox_not_" + list_name + "]").prop("checked")
+                && !$j("input[name=checkbox_and_" + list_name + "]").prop("checked")
+                && !$j("input[name=checkbox_or_" + list_name + "]").prop("checked"))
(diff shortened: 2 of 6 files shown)


willgearty:
Please do not submit pull requests before being assigned to an issue.

PR #4569, opened 2026-03-01
Fix mailgate relay: polymorphic handler dispatch + bounce notificatio…

### LINKED ISSUE #3943: SendGrid Mailgate relay improvements
Two suggested improvements:
1. Handle typos and bad email addresses. It is nice to send the user feedback (some kind of a bounce message). The problem is that we would end up spending money sending a lot of mail back to bots. Possible options include limiting to one bounce message per user per day, but that comes with its own problems. Among others, humans tend to need more feedback, and bots tend to be burners. Relevant areas are `dbmail.MessageRequests` and  line 111ish of `mailgate.py`.
2. In the long term, it would be better to implement polymorphism so that class lists and individual user aliases both have `recipients`. Relevant area is 
line 88ish of `mailgate.py`.

abhinavjha0239:
## Summary

Refactors the mailgate inbound email system to use polymorphic handler dispatch and adds bounce notifications for known ESP users.

**Two changes requested in #3943:**
1. Make `UserEmail` use `self.recipients` like every other handler (eliminate `direct_send` hack)
2. Bounce undeliverable emails back to known users; silently drop emails from unknown senders

**Drive-by fixes found during code review:**
- `classlist.py` caught `ESPUser.DoesNotExist` instead of `ClassSubject.DoesNotExist` (copy-paste bug, never actually caught the right exception)
- `sectionlist.py` used bare `except:` which swallows everything including `KeyboardInterrupt`

### Polymorphism: `preserve_headers` flag

The old mailgate had a procedural `if hasattr(instance, "direct_send")` check that special-cased `UserEmail` — it would skip the header-stripping logic and send the message as-is. Every other handler (`ClassList`, `SectionList`, `PlainList`) went through a different code path that strips `To`, `Cc`, `Message-ID`, prefixes `Subject`, and overrides `From`.

The problem with just removing `direct_send` and making `UserEmail` set `self.recipients`: the unified code path would destroy `Cc` headers (breaking Reply All) and regenerate `Message-ID` (breaking email threading in Gmail/Outlook). These headers matter for individual user aliases but should be stripped for group broadcasts.

Solution: `BaseHandler` now has a `preserve_headers = False` class attribute. `UserEmail` sets `self.preserve_headers = True`. The mailgate checks this flag instead of checking handler identity:

- `preserve_headers=False` (default) — strips To/Cc/Message-ID, prefixes Subject, overrides From. Used by ClassList, SectionList, PlainList.
- `preserve_headers=True` — skips the entire stripping block. Only the `To` header changes (alias to real email). Used by UserEmail.

### Bounce logic

After the handler loop, if no regex matched at all (`matched_any_handler=False`), we parse the `From` header and check if the sender exists in `ESPUser`. If yes, send a bounce via `django.core.mail.send_mail`. If no, silently drop.

### Edge cases handled beyond the original scope

1. **`matched_any_handler` flag:** Distinguishes "no regex matched" (true invalid address, send bounce) from "regex matched but handler chose not to send" (e.g. student emailing a teacher-only alias — don't send a misleading "address doesn't exist" bounce).
2. **Anti-loop defense:** Bounces sent back to the `SUPPORT` address would re-enter the mailgate, fail to match, and bounce again infinitely. We now drop bounces when `sender_email == SUPPORT`.
3. **Python 3 stdin encoding (Crash fix):** The old code used `email.message_from_file(sys.stdin)`. In Python 3, `sys.stdin` is strictly a UTF-8 text stream. Malformed spam (e.g. ISO-8859 declared as UTF-8) causes `UnicodeDecodeError` backend crashes. Changed to `email.message_from_bytes(sys.stdin.buffer.read())`.
4. **Duplicate `To` header in `cc_all` path:** Python's `email.message` module *appends* headers on assignment (`message['To'] = x` adds a new `To` header, doesn't overwrite). For `preserve_headers=True` handlers where the original `To` still exists, the `cc_all` block would create duplicate headers. Added [del(message['To'])](cci:1://file:///Users/abhinavjha/Documents/esp/ESP-Website/esp/esp/dbmail/tests.py:71:4-78:51) before assignment.
5. **`sha` module removal:** The legacy Python 2 `sha` module was removed and replaced safely with `hashlib.sha1`.
6. **MTA Exit Code Safety:** Always `sys.exit(0)` after the bounce logic block to ensure Exim considers delivery handled, preventing MTA-native NDR bounces.

## Files changed

| File | Change |
|------|--------|
| `esp/esp/dbmail/base.py` | Added `preserve_headers = False` to `BaseHandler` |
| `esp/esp/dbmail/receivers/useremail.py` | Replaced `direct_send` with `self.recipients` + `self.preserve_headers = True` |
| `esp/esp/dbmail/receivers/classlist.py` | Fixed `ESPUser.DoesNotExist` to `ClassSubject.DoesNotExist` (both methods) |
| `esp/esp/dbmail/receivers/sectionlist.py` | Fixed bare `except:` to `except (ClassSubject.DoesNotExist, IndexError, ValueError):` (both methods) |
| `esp/mailgates/mailgate.py` | `preserve_headers` guard, bounce logic, `sha`->`hashlib`, `stdin.buffer`, `del(To)` fix |
| `esp/esp/dbmail/tests.py` | 8 new tests across 4 test classes |

## Test plan

- [x] Full test suite passesed
- [x] `grep -r direct_send` confirms no references remain (except the test that asserts its absence)

Closes #3943


esp/mailgates/mailgate.py
```diff
@@ -74,40 +81,34 @@ def send_mail(message):
         if not instance.send:
             continue
 
-        if hasattr(instance, "direct_send") and instance.direct_send:
-            if message['Bcc']:
-                bcc_recipients = [x.strip() for x in message['Bcc'].split(',')]
-                del(message['Bcc'])
-                message['Bcc'] = ", ".join(bcc_recipients)
-
```
hwatheod, line 82: There used to be this logic with Bcc that doesn't seem to be in your new code. Doesn't Bcc need some special handling?

abhinavjha0239: old Bcc block was only doing whitespace normalization (split by comma, strip each address, rejoin). It wasn't adding or removing any Bcc recipients.

In the new code with preserve_headers=True, the header-stripping block is skipped entirely, so Bcc stays in the message as-is. When we pipe to sendmail with -t, it still reads and delivers to Bcc addresses the same way.

For the non-preserve path (ClassList, SectionList, PlainList), the old code also never touched Bcc — only To, Cc, and Message-ID were stripped. So no change there either.

Basically Bcc delivery works identically in both old and new code — the only thing lost is the cosmetic whitespace cleanup, which sendmail handles fine regardless. Happy to add the normalization back if you'd prefer keeping it as a defensive measure though!

Issue #3501, opened 2022-03-24
Requiring the credit card module breaks it


willgearty:
The credit card module checks that all required modules have been completed first:

https://github.com/learning-unlimited/ESP-Website/blob/ea049a4e3460189e521a506e48c9668af9aebb42/esp/esp/program/modules/handlers/creditcardmodule_stripe.py#L147-L153

However, if you require the credit card module, then it is included in this list, and this check will always be false, preventing the user from accessing the credit card module.

I'd suggest that we filter out the credit card module from the list of modules. Furthermore, if the credit card module is required, it should probably always be the last required module (seq-wise), regardless of the seq value it actually has?

imran-techvoyager:
Hi @willgearty! I’d like to work on this issue.

I reviewed the logic in creditcardmodule_stripe.py, and the problem appears to be that when the credit card module is marked as required, it is included in the prog.getModules(...) loop that checks whether all required modules are completed. Since the credit card module itself won’t be “completed” until after payment, the check will always fail, effectively blocking access to the payment page.

To fix this, I propose:

Exclude the credit card module from the required-module completion check, so it doesn't block itself.

Keep the existing ordering behavior (since it already defaults to a high seq value), unless you’d prefer enforcing payment modules to always appear last regardless of admin-configured seq.

I’m inclined to implement the minimal fix first (skipping self in the completion check), and then we can decide if ordering guarantees should be handled separately.

Does that approach sound good?

willgearty:
Yep, that sounds like a solid plan to approach this, thanks 👍

PR #4576, opened 2026-03-02
Conditionally require Credit Card module when balance >= $0.50 (#161,…

### LINKED ISSUE #161: Conditionally Make the Credit Card Module Required
Create a Tag setting such that, when True, the credit card module becomes required if you select certain Extra Cost items. This would be useful if a director decides that students can only reserve meal tickets in advance if they pay in advance.

This is a specific instance of Issue #156. If that issue is solved, then this is no longer an issue.

abhinavjha0239:
# Conditionally require Credit Card module when balance is due (#161, #3501)

Add `creditcard_required_if_amount_due` tag that makes the Credit Card module required when a student has an outstanding balance >= $0.50. Also fix the self-blocking circular dependency in payonline() and harden isCompleted() to check full payment instead of any payment.

## Description

This PR adds conditional logic to both `CreditCardModule_Stripe` and `CreditCardModule_Cybersource` so that the Credit Card module becomes **required** when a student owes money, controlled by a per-program boolean tag.

**Key changes:**
- **`isRequired()` override**: When `creditcard_required_if_amount_due` tag is enabled for a program and the student has an outstanding balance >= $0.50, the CC module becomes required. The $0.50 threshold matches Stripe's minimum charge amount — sub-$0.50 micro-balances from fractional financial aid are safely ignored to prevent a deadlock where Stripe rejects the charge but the module blocks registration.
- **`isCompleted()` fix**: Changed from `has_paid()` to `has_paid(in_full=True)` so that students who pay admission then add extra-cost items (e.g. t-shirts) are correctly shown as incomplete until they pay the remaining balance.
- **`payonline()` self-blocking fix (#3501)**: The payment page checks that all required modules are completed before allowing payment. Without this fix, making the CC module required creates a circular dependency — the module blocks access to its own payment page. Fixed by skipping `module.module_id == self.module_id` in the completion loop. Note: `module != self` doesn't work because `getModules()` returns different cached object instances.
- **New tag definition**: `creditcard_required_if_amount_due` added to `tagdict/__init__.py` with help text and proper categorization.

## Related Issue

Closes #161

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [ ] Manually verified

**22 new tests** across 3 test classes:

| Class | Tests | Covers |
|-------|-------|--------|
| `CreditCardRequiredTest` | 16 | `isRequired()` default/tag/finaid/threshold/unauthenticated/free program/scoped, `isCompleted()` before/after/partial payment, extra cost bypass, `getBooleanTag` |
| `CreditCardCybersourceRequiredTest` | 3 | Cybersource tag enabled/default/paid |
| `CreditCardSelfBlockingTest` | 3 | `payonline()` skips self, still checks other modules |

**Full test suite**: Passed.

## AI Disclosure

## Checklist

- [x] Self-reviewed the code

### DIFF
--- a/.github/workflows/enforce-pr-assignment.yml
+++ b/.github/workflows/enforce-pr-assignment.yml
@@ -1,0 +1,130 @@
+name: Enforce PR Assignment
+
+on:
+  pull_request_target:
+    types: [opened, reopened, edited, synchronize]
+
+permissions:
+  pull-requests: write
+  issues: write
+
+jobs:
+  check-assignment:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check issue assignment
+        uses: actions/github-script@v7
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const pr = context.payload.pull_request;
+            const prAuthor = pr.user.login;
+            const prNumber = pr.number;
+            const body = pr.body || '';
+
+            console.log(`Checking PR #${prNumber} by @${prAuthor}`);
+
+            // Check if the PR author is an admin or maintainer (exempt)
+            let permission = 'none';
+            try {
+              const { data: permissionLevel } = await github.rest.repos.getCollaboratorPermissionLevel({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                username: prAuthor,
+              });
+              permission = permissionLevel.permission;
+              console.log(`@${prAuthor} has '${permission}' permission`);
+            } catch (error) {
+              if (error && error.status === 404) {
+                console.log(`@${prAuthor} is not a collaborator; treating as non-admin (permission='none').`);
+              } else {
+                throw error;
+              }
+            }
+
+            if (permission === 'admin' || permission === 'maintain') {
+              console.log('PR author is admin/maintainer — skipping check.');
+              return;
+            }
+
+            // Extract issue numbers from the PR body
+            // Matches: Closes #123, Fixes #456, Resolves #789 (case-insensitive)
+            const issuePattern = /(?:close[sd]?|fix(?:e[sd])?|resolve[sd]?)\s+#(\d+)/gi;
+            const issueNumbers = [];
+            let match;
+            while ((match = issuePattern.exec(body)) !== null) {
+              issueNumbers.push(parseInt(match[1], 10));
+            }
+
+            if (issueNumbers.length === 0) {
+              console.log('No linked issues found in PR body — allowing through.');
+              return;
+            }
+
+            console.log(`Found linked issues: ${issueNumbers.join(', ')}`);
+
+            // Check each linked issue for assignees
+            for (const issueNumber of issueNumbers) {
+              let issue;
+              try {
+                const { data } = await github.rest.issues.get({
+                  owner: context.repo.owner,
+                  repo: context.repo.repo,
+                  issue_number: issueNumber,
+                });
+                issue = data;
+              } catch (error) {
+                console.log(`Could not fetch issue #${issueNumber} — skipping.`);
+                continue;
+              }
+
+              const assignees = issue.assignees.map(a => a.login);
+
+              if (assignees.length === 0) {
+                // Issue is unassigned — close the PR
+                await github.rest.issues.createComment({
+                  owner: context.repo.owner,
+                  repo: context.repo.repo,
+                  issue_number: prNumber,
+                  body: `👋 Hi @${prAuthor}, thanks for your interest in contributing!\n\nHowever, issue #${issueNumber} is not yet assigned to anyone. To avoid duplicate work, this PR has been automatically closed.\n\nPlease:\n1. Comment on the issue asking to be assigned\n2. Wait for a maintainer to assign you\n3. Reopen this PR\n\nWe appreciate your enthusiasm and look forward to your contributions! 🙌`,
+                });
+
+                await github.rest.pulls.update({
+                  owner: context.repo.owner,
+                  repo: context.repo.repo,
+                  pull_number: prNumber,
+                  state: 'closed',
+                });
+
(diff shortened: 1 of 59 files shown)


abhinavjha0239:
Hi @willgearty ready for review.

willgearty:

hwatheod:
This doesn't quite meet the requirements. The issue being discussed In #161 is this:

> Create a Tag setting such that, when True, the credit card module becomes required if you select certain Extra Cost items. This would be useful if a director decides that students can only reserve meal tickets in advance if they pay in advance.

The Extra Cost items are those costs other than program admission (e.g. meals). We don't want our programs to require students to pay by credit card for admission. The Tag should be able to specify which Extra Cost items this credit card requirement will apply to (so I think it shouldn't just be a boolean).

abhinavjha0239:
> This doesn't quite meet the requirements. The issue being discussed In #161 is this:
> 
> > Create a Tag setting such that, when True, the credit card module becomes required if you select certain Extra Cost items. This would be useful if a director decides that students can only reserve meal tickets in advance if they pay in advance.
> 
> The Extra Cost items are those costs other than program admission (e.g. meals). We don't want our programs to require students to pay by credit card for admission. The Tag should be able to specify which Extra Cost items this credit card requirement will apply to (so I think it shouldn't just be a boolean).

Hey @hwatheod , Right now it checks `amount_due()` which includes admission costs too, but the issue specifically says:

> We don't want our programs to require students to pay by credit card for admission. The Tag should be able to specify which Extra Cost items this credit card requirement will apply to.

So I'm planning to rework this. Instead of a boolean tag that fires on any balance, I'm thinking of a string tag (`creditcard_required_for_extracosts`) where:

- Not set → feature off (same as now)
- `*` → CC required if the student selected any extra cost item
- `Meal Ticket,T-Shirt` → CC required only if they picked one of those specific items

That way admission-only students aren't forced through CC, and directors can control exactly which items need upfront payment.

Does that match what you had in mind?

hwatheod:
Yes, that sounds good.

abhinavjha0239:
Hey @hwatheod , Reworked the entire implementation based on your feedback. The old boolean tag (`creditcard_required_if_amount_due`) that checked `amount_due()` is gone — it was wrong because `amount_due()` includes program admission, and we don't want to force CC payment just for admission.

### New approach

Replaced with a **string tag** `creditcard_required_for_extracosts` that checks actual extra cost item **Transfer records** instead of the total balance:

| Tag value | Behavior |
|-----------|----------|
| _(not set)_ | Feature off, same as before |
| `*` | CC required if student selected **any** extra cost item |
| `Meal Ticket,T-Shirt` | CC required only if student selected one of those specific items |

The logic explicitly excludes admission items (`pac.admission_items` = `["Program admission", "Student payment"]`), so a student who only owes admission is never forced through credit card payment.

There's also a `$0.50` minimum balance check — if financial aid or partial payment brings the balance below Stripe's minimum charge amount, the module won't block registration for a charge that Stripe would reject anyway.

### Also addressed all Copilot review comments

- `module.id == self.id` for the payonline() self-skip (clearer than `module_id`)
- Using `ProgramModuleObj.getFromProgModule()` in test helper instead of reimplementing it
- Self-blocking test now explicitly asserts `completedAll == True` when only CC is incomplete
- Other-modules test explicitly creates a required `StudentRegCore` module and asserts it blocks

### #3501 self-blocking fix still included

The `payonline()` loop now skips `module.id == self.id` so the CC module doesn't block access to its own payment page. Without this, making CC required creates a deadlock — student can't pay because they haven't paid.

### E2E verification

**Tag configuration** — `creditcard_required_for_extracosts = *` scoped to the program:


01-admin-tag-configuration

**Line item types** — Meal Ticket ($8.00) and T-Shirt ($15.00) are extra cost items, Program admission is separate:


02-line-item-types-extracosts



**Student extra costs — nothing selected** — Both items unchecked, $0.00 due, CC module is NOT required:



06-student-extracosts-items-unchecked


**Student extra costs — Meal Ticket selected** — Checked Meal Ticket, $8.00 due, CC module IS now required:



07-student-extracosts-meal-ticket-selected

**All 8 E2E scenarios pass** — Covers tag=`*`, specific item match, non-match, comma-separated list, full payment, tag removal:



03-e2e-isRequired-tests-all-pass


### Test results

23 unit tests across 3 classes, all passing. 

| Class | Tests | What it covers |
|-------|-------|----------------|
| `CreditCardRequiredTest` | 18 | tag `*` / specific items / comma-separated, admission-only never triggers, finaid covers all, micro-balance ignored, anonymous user, free program, tag scoped to program, isCompleted before/after/partial payment, extra cost added after admission payment |
| `CreditCardCybersourceRequiredTest` | 3 | Same tag works on Cybersource module, default not required, admission-only doesn't trigger |
| `CreditCardSelfBlockingTest` | 2 | payonline() skips self (#3501), other required modules still block |


@hwatheod does this look aligned with what you had in mind, or do you want any changes to the tag behavior / naming?

abhinavjha0239:
@hwatheod, does this looks fine or need any changes? Since you have reviewed it please let me know?

imran-techvoyager:
@willgearty , my PR #4390 fixes the self-blocking behavior described in #3501 (where the credit card module could block access to its own payment page). Since making the CC module conditionally required would otherwise create a circular dependency, that fix seemed like a necessary prerequisite.

Once that PR is merged, I’d be happy to work on implementing the conditional CC requirement described here. Would it make sense for me to take this issue as the follow-up?

willgearty:
Ah, looks like the issue was that the OP said this closes #3501. I've fixed the OP and this should be fine to keep under @abhinavjha0239.

PR #4588, opened 2026-03-02
Fix: getLastForProgram no longer auto-saves profiles

### LINKED ISSUE #1450: RegistrationProfiles created for users without permission to register for program

Dhirenderchoudhary:
Date : 2 March 2026

## Description

`RegistrationProfile.getLastForProgram()` no longer auto-saves profiles. Previously, visiting a registration page would create database records even without form submission. Now profiles are only persisted when explicitly saved by the caller.

## Related Issue

Closes #1450

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

## Checklist

- [x] Self-reviewed the code


esp/esp/program/models/__init__.py
```diff
@@ -1620,25 +1596,9 @@ def getLastForProgram(user, program, tl = None):
...
-                # assume the info is up-to-date and save it now
-                else:
-                    # but, if the profile was for a previous program, we should keep the old profile
-                    # and make a new one for this program by wiping the id, then saving
-                    if prog is not None:
-                        regProf.id = None
-                    # otherwise, it was a profile without a program,
-                    # and we can just associate it with this program now, so just save
-                    regProf.save()
+                regProf.id = None
```
Oval17, line 1676: If @cache_function returns the same object by reference (not a deep copy), any subsequent call to getLastProfile(user) that hits the cache will return a profile with id=None and the wrong program.
From what I can see iss that old code had the same mutation of .program, but now .id is also always wiped. 
The fix should be to copy before mutating n Django this is typically copy.copy(regProf) then clear pk/id.
Can you explore this once if possible??

Oval17: import copy
regProf = copy.copy(RegistrationProfile.getLastProfile(user))
regProf.program = program
if regProf.id is not None:
    regProf.id = None
something like this should help you maybe 
need a bit of exploration if you can do ?

esp/esp/program/models/__init__.py
Oval17, line 1683: umm this phrase "our attempts to be smart" refers to the old auto-save logic that this PR removes. The cache invalidation rule is still correct (it depends on any profile for the user, not just the (user, program) pair), but the comment should be updated to reflect that the reason is now "because we fall back to profiles from other programs when no program-specific one exists", not because of auto-saving.
though inline but yeah can be useful for someone reading code ....

esp/esp/program/modules/tests/regprofilemodule.py
```diff
@@ -64,18 +64,16 @@ def runTest(self):
...
         self.assertTrue( self.students[0].registrationprofile_set.count() <= 0, "Profile was saved when it shouldn't have been." )
-        # Test migration of initial non-program profile to a program
         prof = self.students[0].getLastProfile()
         prof.program = None
         prof.save()
         self.assertTrue( self.students[0].registrationprofile_set.count() >= 1, "Profile failed to save." )
         self.assertTrue( self.students[0].registrationprofile_set.count() <= 1, "Too many profiles." )
-        self.assertTrue( self.moduleobj.isCompleted(self.students[0]), "Profile id wiped." )
-        self.assertTrue( self.students[0].registrationprofile_set.all()[0].program == self.program, "Profile failed to migrate to program." )
+        self.assertFalse( self.moduleobj.isCompleted(), "Profile should not be complete until explicitly saved with program." )
```
Oval17: Every other isCompleted() call in this test file passes the user explicitly you can see just above line 65 uses isCompleted(student), and then below again  line 91 uses isCompleted(self.students[1]). This call relies silently on get_current_request().user being self.students[0] (set at line 68), which is somewhat fragile. Pass the user explicitly for consistency: self.moduleobj.isCompleted(self.students[0]).

esp/esp/program/tests.py
```diff
@@ -806,6 +805,53 @@ def create_past_program(self):
             end_time = start_time + timedelta(minutes=past_settings['timeslot_length'])
             event, created = Event.objects.get_or_create(program=self.new_prog, event_type=event_type, start=start_time, end=end_time, short_description='Slot %i' % i, description=start_time.strftime("%H:%M %m/%d/%Y"))
 
+class RegistrationProfileTest(ProgramFrameworkTest):
+
+    def setUp(self):
+        super(RegistrationProfileTest, self).setUp()
```
Oval17: This setUp does nothing except call super(), which Python does automatically when setUp is not overridden. should we remove it, or add meaningful per-test setup here?

Oval17 (commented):
Please look into mutating cache comment it will need some exploration from your side.

Tests seem fine to me 
Thanks

Oval17:
Test Results : 
The ones that are written in and the one's that I simulated

## Test Results

| Category                     | Test Name                                                       | Status |
|-----------------------------|------------------------------------------------------------------|--------|
| Existing Tests              | ProfileTest::testAcctCreate                                     | PASSED |
| Existing Tests              | RegProfileModuleTest::runTest                                   | PASSED |
| Existing Tests              | RegProfileModuleTest::runTest                                   | PASSED |
| PR's New Tests (Simulated)  | test_getLastForProgram_does_not_auto_save                       | PASSED |
| PR's New Tests (Simulated)  | test_getLastForProgram_does_not_mutate_cached_last_profile      | PASSED |
| PR's New Tests (Simulated)  | test_getLastForProgram_with_existing_profile                    | PASSED |

Oval17:
LGTM.
Thanks

Oval17 (approved):

PR #4281, opened 2026-02-22
Feature/migrate vagrant to docker

Arnav1709:
## Description

This PR addresses issue #4215 by completely replacing the legacy Vagrant/VirtualBox/Fabric local development environment with a modern, fast, and cross-platform Docker Compose setup. 

As requested by the maintainers, Docker is now the single source of truth for local development. All Vagrant and Fabric artifacts have been entirely removed, and the documentation has been fully rewritten to reflect Docker as the new standard.

Key technical changes:
- Created [Dockerfile](cci:7://file:///home/arnav/Desktop/ESP-Website/Dockerfile:0:0-0:0) (Python 3.7) and [docker-compose.yml](cci:7://file:///home/arnav/Desktop/ESP-Website/docker-compose.yml:0:0-0:0) (`web`, `db` Postgres 14, and `memcached`).
- Added robust [docker-entrypoint.sh](cci:7://file:///home/arnav/Desktop/ESP-Website/docker-entrypoint.sh:0:0-0:0) to handle DB wait, media symlinks, migrations, and static collection.
- Cleaned up obsolete VirtualBox flags in [update_deps.sh](cci:7://file:///home/arnav/Desktop/ESP-Website/esp/update_deps.sh:0:0-0:0).
- Fixed a bug in [esp/esp/settings.py](cci:7://file:///home/arnav/Desktop/ESP-Website/esp/esp/settings.py:0:0-0:0) where `CACHES` was unconditionally overridden, which broke [local_settings.py](cci:7://file:///home/arnav/Desktop/ESP-Website/esp/esp/local_settings.py:0:0-0:0) overrides.

## Related Issue

Closes #4215

## Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [X] Documentation update
- [X] Refactor / cleanup
- [X] CI/CD or build change

## Testing

- [X] Existing tests pass
- [ ] New tests added (if applicable)
- [X] Manually verified

*Tested `docker-compose up --build` locally and verified that containers start successfully, migrations execute smoothly, and the Django development server successfully binds to `0.0.0.0:8000`. Also verified that `createsuperuser` runs correctly via `docker-compose exec web python esp/manage.py createsuperuser` and ran a codebase-wide regex search to guarantee complete removal of all `vagrant`, `virtualbox`, and `fabric` references.*

## AI Disclosure

AI tools were used to assist in writing and updating the documentation.

## Checklist

- [X] Self-reviewed the code
- [X] Updated documentation (if needed)
[...]

PR #4589, opened 2026-03-02
Prevent .pyc file creation in dev environments

### LINKED ISSUE #1563: Avoid creating *.pyc files
Stale *.pyc files can mess things up. We can certainly make changes (such as #1562) to clean these files more aggressively. However, we might want to see if we can just avoid creating these files.

There is an environment variable (PYTHONDONTWRITEBYTECODE) that can be set to stop Python from creating *.pyc files. I wonder if we can set that in manage.py and have that take effect before the rest of the esp modules get imported, or if it's too late by then. We should try this and see what happens.

Of course, it's possible that the *.pyc files are necessary for performance. So before merging such a change, we should find out if it has a noticeable affect on performance. If we think or know that it might, then maybe we can somehow only disable *.pyc files on dev servers. Between that and #1562, that would be enough to avoid any *.pyc troubles.

Dhirenderchoudhary:
Date : 2 March 2026

## Description

Prevent Python from creating `.pyc` files in development environments by:
- Setting `sys.dont_write_bytecode = True` early in `manage.py`
- Exporting `PYTHONDONTWRITEBYTECODE=1` in `docker-entrypoint.sh`

This avoids stale bytecode issues when switching branches or deleting Python files.

## Related Issue

Closes #1563

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

## Checklist

- [x] Self-reviewed the code

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -1,8 +1,8 @@
 #!/bin/bash
 set -e
 
-# Prevent Python from creating .pyc files (avoids stale bytecode issues)
-export PYTHONDONTWRITEBYTECODE=1
+# Prevent Python from creating .pyc files (avoids stale bytecode issues);
+# PYTHONDONTWRITEBYTECODE=1 is set in the Dockerfile.
 
 # Copy Docker-specific settings if local_settings.py doesn't exist
 if [ ! -f /app/esp/esp/local_settings.py ]; then


willgearty:
Closing because this was already implemented in #4281

PR #4602, opened 2026-03-02
fix: customforms/tests: stale loop var and dict alias in response…

### LINKED ISSUE #4582: [Bug]: Custom forms test uses stale variable in response verification loop, most fields never checked
### Description

In `tests.py`, the test that verifies admin-viewed response data has a bug where the loop variable `field_spec` is never used, instead, the assertion uses **key**, which is a stale variable from a previous for loop:

```js
for field_spec in indiv_response:
    self.assertEqual(self.map_form_value(responses_corrected[key]), indiv_response[key])
```

This means the loop iterates over all fields in `indiv_response` but checks the exact same field key every iteration. Most response fields are never verified. Data corruption bugs in response retrieval would not be caught.

Additionally, on `tests.py`, `responses_corrected = responses_initial` creates an alias (not a copy), so mutations to one affect the other, further weakening the test.

### Steps to Reproduce

1. Introduce an intentional bug in getResponseData() that corrupts one field's value.
2. Run python manage.py test esp.customforms.
3. Observe that the test still passes.

### Expected Behavior

The test should iterate over each field in the response and verify it against the expected value. A data corruption bug should cause a test failure.

### Actual Behavior

The test only checks one field repeatedly, so bugs in other fields go undetected.

userAdityaa:
… test

Use dict.copy() so mutations to responses_corrected do not affect responses_initial. Replace stale loop variable key with field_spec and skip metadata-only fields so every question field is verified.

## Description
Fixes a bug in the admin-viewed response test where only one field was
checked repeatedly due to a stale loop variable. Now, every field in
the response is verified. Also replaces dict aliasing with a copy to
prevent unintended mutations and ensure reliable test comparisons.

## Related Issue

Closes #4582 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

## Checklist

- [x] Self-reviewed the code

cc: @willgearty


willgearty:

PR #4604, opened 2026-03-02
Fix/expire records on deactivation

### LINKED ISSUE #1478: Records for upcoming programs should not include inactive users
Deactivating a user should expire/remove all associated records, registrations, and maybe permissions for upcoming programs. I think it's important that this fix doesn't affect old records or info from previous programs because it would mess with correct stats. (This helped cause #1464.)

Dhirenderchoudhary:
Date: 2 March 2026

## Description

When a user account is deactivated, this change automatically expires their registration records for upcoming programs while preserving records for programs that have already ended.

## Related Issue

Closes #1478 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

## Checklist

- [x] Self-reviewed the code


esp/esp/users/models/__init__.py
```diff
@@ -1450,6 +1446,103 @@ def update_email(**kwargs):
             for l in lists:
                 mailman.remove_list_member(l, old_email)
 
+
+@dispatch.receiver(signals.pre_save, sender=ESPUser,
```
Oval17, line 1547: This signal uses pre_save, which fires before the UPDATE users SET is_active=False SQL executes. If user.save() subsequently fails (DB error, constraint violation, etc.), the records will already have been expired while the user is still active shoule we change to post_save and add if kwargs.get('created'): return as an early guard, since newly-created users can't be "deactivated"??

esp/esp/users/models/__init__.py
```diff
@@ -1450,6 +1446,103 @@ def update_email(**kwargs):
...
+                   dispatch_uid='expire_upcoming_records_on_deactivation')
+def expire_upcoming_records_on_deactivation(**kwargs):
+    """Expire registrations/permissions for upcoming programs when user is deactivated."""
+    new_user = kwargs['instance']
+
+    if new_user.id is None:
+        return
+
+    old_user = ESPUser.objects.get(id=new_user.id)
```
Oval17: This fires on every ESPUser.save(), even when the user is being saved while already inactive (e.g. admin editing an inactive account) or while already active.
can you explore this once if possible cause we can save double redundant query lookup here 
cause lets say Add if new_user.is_active: return before this line to skip the extra SELECT query in the common case where deactivation is not happening.

esp/esp/users/models/__init__.py
```diff
@@ -1450,6 +1446,103 @@ def update_email(**kwargs):
...
+        user=user
+    )
+    perm_count = permission_qs.update(end_date=now)
+
+    vol_qs = VolunteerOffer.objects.filter(
+        request__program_id__in=upcoming_program_ids,
+        user=user
+    )
+    vol_count = vol_qs.count()
+    vol_qs.delete()
```
Oval17: QuerySet.delete() already returns a (total_count, {model: count}) tuple, so the .count() call is an unnecessary extra DB query. Replace with:

vol_count, _ = vol_qs.delete()

esp/esp/users/models/__init__.py
```diff
@@ -1450,6 +1446,103 @@ def update_email(**kwargs):
...
+def expire_user_records_for_upcoming_programs_for_queryset(user_qs):
+    """Expire registrations and permissions for all users in the given queryset
+    for programs that haven't ended yet.
+
+    This is intended for use after bulk deactivation operations performed via
+    QuerySet.update(...), which do not trigger model save signals.
+    """
+    for user in user_qs:
+        expire_user_records_for_upcoming_programs(user)
+
```
Oval17, line 1688: For a queryset of N users, this runs the "find upcoming program IDs" DB queries like N times, even though those queries return the same result for every user I believe 
can you verify this once if possible 
. For bulk deactivations this can be very slow. The upcoming_program_ids computation should be extracted and done once, then a single bulk update applied with user__in=user_qs. For example:

def expire_user_records_for_upcoming_programs_for_queryset(user_qs):
    upcoming_program_ids = _get_upcoming_program_ids()  # called once
    now = datetime.now()
    StudentRegistration.valid_objects(now).filter(
        section__parent_class__parent_program_id__in=upcoming_program_ids,
        user__in=user_qs
    ).update(end_date=now)
    # same pattern for SSI, Permission, VolunteerOffer

Can you explore this  as the above patch is somewhat pseudo code written 
This reduces the DB load from O(N × 6 queries) to O(6 queries total).

Oval17 (changes requested):
Hi 
There is a lot of db Optimization scope here 
I would suggest to go through them once we can save  a lot lookup query 
duplicate round trips

I have also given a performance optimization 
if you wish you can read about maps and lookup as well this will give you a solid idea on how o(1) conversion happens

Dhirenderchoudhary:
fixed

willgearty:
Thanks for putting this together @Dhirenderchoudhary. I've had a little chat with the other maintainers, and we agree that this is valuable, but we also worry that performing this many additional DB changes in association with deactivating an account is concerning, especially given that the user (either a user unsubscribing or an admin deactivating another user's account) is never told these various records/etc will be expired.

I realize this'll broaden the scope of this PR, but I think the reasonable fix for this is to be more explicit about what will happen if a user is deactivated. This should be done in two places:

1. The unsubscribe page needs a broader disclaimer of what will happen when the user's account is deactivated (https://github.com/learning-unlimited/ESP-Website/blob/main/esp/templates/users/unsubscribe.html)
2. The userview page needs a confirmation upon clicking the "Deactivate user" button, with a similar disclaimer (https://github.com/learning-unlimited/ESP-Website/blob/main/esp/templates/users/userview.html)

Let me know if you have any questions. @Oval17 also mentioned some things above that it doesn't look like you've addressed yet. Please address them or reply to his comments with why you aren't implementing the suggestions.

Dhirenderchoudhary:
@willgearty i'll look into this and try to fix and let you know asap.

PR #4609, opened 2026-03-02
chore: customforms cache: add thread-safety and cache invalidation

### LINKED ISSUE #4591: [Enhancement]: Make CustomFormsCache thread-safe and add cache invalidation
### Problem Statement

`CustomFormsCache` in `linkfields.py`  uses the Borg design pattern (shared mutable `__dict__`) with no thread synchronization. The `_populate()` method checks `self.loaded` without a lock. In a multi-threaded WSGI server (for example, Gunicorn running with `--threads`), multiple threads can race through `_populate()`, which can lead to a partially populated or duplicated cache state.

Additionally, the cache never invalidates. Once `self.loaded = True`, the cache serves the same data for the lifetime of the process, even if the underlying models or field configurations change.

### Proposed Solution

1. Add a threading.Lock() around `_populate()` to prevent concurrent initialization:
```py
import threading
_lock = threading.Lock()

def _populate(self):
    if self.loaded:
        return
    with self._lock:
        if self.loaded:  # double-check
            return
        # ... existing population logic ...
        self.loaded = True
```
2. Add an invalidate() method that resets `loaded` to False and clears the cached dicts.
3. Alternatively, refactor to use @functools.lru_cache on a module-level function.

### Additional Context

This enhancement is focused, low-risk, and high-value. It will make the custom forms system robust in real-world deployments and easier to maintain. Only one file is strictly required to change.

userAdityaa:
## Description
Added a threading lock and double-check locking to CustomFormsCache to prevent race conditions during cache population in multi-threaded WSGI servers. Introduced an invalidate() method to allow safe cache refresh when models or field configurations change.

## Related Issue

Closes #4591 

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

## Checklist

- [x] Self-reviewed the code

cc: @willgearty

### DIFF
--- a/esp/esp/customforms/linkfields.py
+++ b/esp/esp/customforms/linkfields.py
@@ -1,3 +1,5 @@
+import threading
+
 from django import forms
 from django.forms.models import fields_for_model
 from django.apps import apps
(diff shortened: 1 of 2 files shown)


userAdityaa:
All copilot comments has been resolved, this PR is ready for a review!

willgearty:
Thanks for this work, but upon review I don't believe this is necessary and just adds extra mechanics for a problem that doesn't exist.

PR #4614, opened 2026-03-02
Fix #3929: Move Clear Change Log button inside loading overlay

### LINKED ISSUE #3929: Clear change log on scheduler should supercede the "loading" overlay
I ran into a situation where the scheduler wouldn't load, and the fix was to clear the change log, but the button for that is unavailable while the scheduler is loading. The clear change log button should show/be available on top of the loading overlay.

abhinavjha0239:
## Description

When the AJAX scheduler's changelog fails to load (e.g., corrupted data, server error), `ChangelogFetcher.applyChangeLog()` never executes, so `$j("#loadingOverlay").remove()` never runs. The loading overlay (`z-index: 9999`, `position: fixed`, full viewport) stays permanently on screen.

The "Clear Change Log" button — the intended recovery mechanism for this exact situation — was inside `#side-panel-wrapper` behind the overlay, making it completely unreachable when it's needed most.

**Fix**: Move the button **inside** `#loadingOverlay` so it:
1. Is always clickable when the overlay is visible (the failure case)
2. Auto-disappears when the overlay is removed on successful load

## Related Issue

Closes #3929

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

## Checklist

- [x] Self-reviewed the code


samalarussa (changes requested):
The clear change log button now disappears with the overlay, making it unavailable during normal use of the scheduler. This button is commonly used during normal scheduler use, to synchronize the interface between different users/browser windows if the change log loses synchronization. Ideally it should be visible on top of the overlay but also after the overlay disappears.

The current state of it being present after the overlay is removed is essential for proper use.

abhinavjha0239:
Thanks for the review @samalarussa! Great catch — I didn't realize the button is used during normal scheduler operation for re-syncing between users/windows.

**Fixed:** The button is now placed **outside** the overlay as a sibling element with `position: fixed; z-index: 10000` (above the overlay's `z-index: 9999`). This means it:

1. **Floats above the overlay** when loading fails (the original bug fix)
2. **Remains visible after the overlay is removed** (normal scheduler use)

samalarussa:
image

While this fixes the issue of the button disappearing after page load, it now covers content. The above image is your most recent PR, while the below shows the typical button placement on the page:

image

Could you resolve this issue so that there is no text covered by the button?

samalarussa (dismissed):
See my comment above; currently this new button placement covers text

willgearty (approved):
Looks good now 👍

PR #4619, opened 2026-03-02
Fix custom form crash when re-linking to program without modules (#3868)

### LINKED ISSUE #3868: Custom form issue with changing program link
If you change a custom form that was previously linked to a different program and had responses such that it is now linked to a new program, you get a whoops error when a user submits it.

abhinavjha0239:
## Description

Fix custom form crash when an admin re-links a form to a program that doesn't have custom form modules enabled.

When the target program lacks custom form modules, the module dropdown in the form builder is empty. jQuery's `$j().val()` returns `null` for an empty `<select>`, which gets sent as `link_module: null` in the AJAX payload. This triggered a chain of failures on the server that left the database in an inconsistent state and caused students to see a "Whoops!" error on the registration page.

## Related Issue

Closes #3868

## Type of Change

- [x] Bug fix

## Root Cause

Six interrelated bugs in `views.py` (`onModify`/`onSubmit`) and `custom_form.js`:

1. **JS sends `link_module: null`** for empty module dropdown — server receives the key with a `None` value
2. **`if 'link_module' in metadata:`** evaluates to `True` even when value is `None` — enters the validation branch when it shouldn't
3. **`return ESPError(...)`** returns an exception object instead of raising it — Django tries `.status_code` on it → `AttributeError` → 500
4. **Tags deleted before validation** — `Tag.objects.filter(...).delete()` runs at line 204, but validation at line 207+ can fail, leaving tags destroyed with no new ones created
5. **`@transaction.atomic` commits partial state** — the `except` block catches errors and returns a JSON response (normal return), so Django commits the destructive tag deletion
6. **`form.link_id` (int) != `metadata['link_id']` (string)** in Python 3 — causes unnecessary `change_only_fkey` calls

## Bug Reproduction 

**Setup:** Form 2 linked to Program 6 (has StudentCustomFormModule). Program 5 has no modules.

1. Student submits form successfully via `/customforms/view/2/` — works fine
2. Admin re-links form to Program 5 via form builder (sends `link_module: null`)
3. Server returns **500** (`'ESPError_Log' object has no attribute 'status_code'`)
4. DB state is now **inconsistent**: tags deleted, but `form.link_id` still points to Program 6
5. Student visits `/learn/TestProg/2026_Summer/extraform` → **"Whoops! An error has occurred!"**

**Why this was hard to reproduce:** The bug only triggers when the target program does NOT have the custom form module enabled. If the target program has the module, `link_module` is a valid string and everything works fine.

### `custom_form.js`
- Only include `link_module` in the payload when `$j("#links_id_module").val()` is truthy — prevents `null` from reaching the server

### `views.py` — `onModify`
- **Validate before deleting:** Move all `link_module` validation (program exists, module enabled, module recognized) before `Tag.objects.filter(...).delete()` so tags are preserved if validation fails
- Use `metadata.get('link_module')` with truthiness check instead of `'link_module' in metadata`
- Change `return ESPError(...)` to `raise ESPError(...)` for correct exception handling
- Add `transaction.set_rollback(True)` in the `except` block so any error rolls back instead of committing partial state
- Convert `metadata['link_id']` to `int()` for proper comparison with `form.link_id`

### `views.py` — `onSubmit`
- Same `metadata.get('link_module')` / `raise` / `transaction.set_rollback(True)` fixes (identical bug pattern)

## Testing

- [x] Existing tests passed
- [x] Manually verified :
  - **Before fix:** 500 error, tags destroyed, "Whoops!" on registration page
  - **After fix:** 200 OK, form re-linked correctly, DB state consistent

## AI Disclosure

## Checklist

- [x] Self-reviewed the code


samalarussa (changes requested):
When I go to reproduce the error, the resulting custom form is still no longer editable in the custom form editor after step #2. This indicates that the form is no longer functional, even with your fix.

abhinavjha0239:
@samalarussa good catch — ran the repro and confirmed the form was still not editable after re-link. Root cause was in the form-builder JS, not the server. `rebuildData()` returns `link_tl=''` / `link_module=''` when the form is linked to a program with no custom form modules, and the builder blindly did:

```js
$j('#links_id_tl').val(metadata['link_tl']).trigger("change");
```

`#links_id_tl` only has `learn`/`teach` options. `.val('')` leaves it deselected, then `onChangeLinksTL` runs `modules[''].length` → TypeError → the whole `rebuild()` stops mid-way and the editor never fully loads.

Fix in b8b7c527d: only apply the saved `link_tl` / `link_module` when they're truthy. If they're empty, keep the defaults the earlier `#links_id_pick` change already set up, so the editor lands in a clean state that the admin can save or re-edit.

Also adding `test_form_editor_reloads_after_relink_to_unmoduled_program` which hits `/customforms/metadata/` after a re-link and locks in the contract the JS depends on.

@Copilot both of your comments asked for the same scenario (relink to a program without modules) from different angles; tests are in place:
- `test_submit_omitting_link_module_succeeds` / `test_submit_with_null_link_module_succeeds` — cover omitting/nulling `link_module` on `/customforms/submit/`.
- `test_relink_to_program_without_module_succeeds_when_unlinking` — cover the `/customforms/modify/` re-link path; asserts 200, `form.link_id` updated, tags cleared.
- `test_relink_with_invalid_link_module_rolls_back_completely` / `test_relink_with_unknown_module_rolls_back` / `test_relink_with_missing_program_rolls_back` — cover the rollback/destructive-before-validate paths.

Full `esp` test suite: 983 passed, 2 skipped (Firefox).

samalarussa (changes requested):
While your latest commit does address the form editing glitch I observed, I am still able to assign the form to a program that does not have forms enabled. Furthermore, doing this reassignment erases all of the previous responses for some reason, and these are not restored by reassigning the custom form to the original program. Could you address this issue, possibly by preventing the form from being reassigned to a program that does not have forms enabled in the first place?

willgearty:

PR #4620, opened 2026-03-02
Implement class replacement with conflict confirmation #561

### LINKED ISSUE #561: Class changes alternatives interface idea
Another idea from a Splash parent:

> For the add/drop period, offer a page where I can view a list of available alternatives to that class BEFORE I drop the class. Maybe have an option to filter by classes that I marked as interesting in the lottery.  This might even decrease your server load as you won't have to send out the same huge list so many times (I often reloaded those pages trying to figure out how to make the best schedule).

This might be useful, or it might not help.  Thoughts?

Supriyo760:
## Description
This PR implements a feature that allows students to replace existing classes with new ones that have schedule conflicts. Instead of seeing a generic "conflict" error, students are now presented with a clear confirmation dialog that lists the conflicting classes and offers an automatic "drop and replace" option.

## Key Changes

**Backend Logic:**
- Added `get_conflicts(user)` method to `ClassSection` to dynamically identify schedule overlaps for a specific student
- Modified `ajax_addclass` in `StudentClassRegModule` to detect conflicts and return a JavaScript payload for the confirmation flow
- Implemented `force_replace` parameter to automatically unregister students from conflicting sections before enrolling them in the new one

**Frontend Interaction:**
- Integrated a `confirm()` browser dialog triggered by the AJAX response
- Automatically re-submits the registration request with the `force_replace` flag if the student confirms the action

## Related Issue
Closes #561

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup

## Testing
- [x] Existing tests pass
- [x] New automated test added (`issue_561_test.py`) to verify conflict detection and replacement logic
- [x] Manually verified the confirmation flow and automatic drop/add behavior in a local Docker environment

## AI Disclosure
AI was just used for suggestion.

## Checklist
- [x] Self-reviewed the code
- [x] Updated documentation (walkthrough provided)


Supriyo760:
all changes done.

esp/esp/program/modules/handlers/studentclassregmodule.py
```diff
@@ -448,6 +469,39 @@ def ajax_addclass(self, request, tl, one, two, module, extra, prog):
...
+                                '    form.append(%s);'
+                                '    form.submit();'
+                                '    form.find(\'input[name="force_replace"]\').remove();'
+                                '  }'
+                                '}'
+                            ) % (
+                                json.dumps(confirm_msg),
+                                json.dumps(form_selector),
+                                json.dumps(force_replace_input),
+                            )
```
willgearty: This seems really fragile. Instead, you should return structured json to the template that you then process with client-side javascript.

esp/esp/program/modules/tests/test_class_replacement.py
willgearty, line 1: File names and class names should not be based on issue numbers. Please update this to reflect the actual functionality.

willgearty (changes requested):
I tried to test this out, but I'm not able to select a class that conflicts with a class that's already on my schedule:

Image

It looks like you haven't changed any of the template logic to actual allow students to try to enroll in conflicting classes.

I've included a couple of other comments just based on the sourcecode changes.

When you update this, please make sure to test both normal student registration (e.g., /fillslot) and the student onsite webapp (e.g., /onsitecatalog).

willgearty:
I'm getting this error message when I try to add a conflicting class in either location:

image

esp/esp/utils/apps.py
```diff
@@ -19,3 +19,7 @@ class InstallConfig(AppConfig):
 
     def ready(self):
         signals.post_migrate.connect(run_install, sender=self)
+
+class UtilsConfig(AppConfig):
+    name = 'esp.utils'
+
```
willgearty, line 25: This seems unrelated to this PR.

esp/esp/django_settings.py
```diff
@@ -243,7 +243,7 @@
     'esp.survey',
     'esp.accounting.apps.AccountingConfig',
     'esp.customforms.apps.CustomformsConfig',
-    'esp.utils',    # Not a real app, but, has test cases that the test-case runner needs to find
+    'esp.utils.apps.UtilsConfig',    # Not a real app, but, has test cases that the test-case runner needs to find
```
willgearty, line 246: This seems unrelated to this PR.

esp/templates/program/modules/studentonsite/conflict_confirm.html
```diff
@@ -0,0 +1,38 @@
...
+            <tr class="class_row">
+                <td colspan="2" style="padding: 20px; text-align: center;">
+                    <p style="font-size: 1.2em;"><b>{{ confirm_msg }}</b></p>
+                    <form method="post" action="{{ prereg_url }}" style="margin-top: 20px; display: inline-block;">
+                        {% csrf_token %}
+                        <input type="hidden" name="class_id" value="{{ class_id }}" />
+                        <input type="hidden" name="section_id" value="{{ section_id }}" />
+                        <input type="hidden" name="force_replace" value="true" />
+                        <button type="submit" class="btn" style="background-color: #d9534f; color: white; margin-right: 10px;">Yes, replace class</button>
+                        <a href="{{ program.get_learn_url }}studentonsite" class="btn" style="background-color: #5bc0de; color: white; text-decoration: none; padding: 10px 15px; border-radius: 5px;">Cancel</a>
```
willgearty: These two buttons should have similar styling:

Image

esp/esp/program/modules/handlers/studentonsite.py
```diff
@@ -161,8 +161,56 @@ def onsitesurvey(self, request, tl, one, two, module, extra, prog):
...
+            if "conflicts with your schedule" in error_message:
+                try:
+                    sectionid = request.POST.get('section_id')
+                    classid = request.POST.get('class_id')
+                    if sectionid:
+                        section = ClassSection.objects.filter(id=sectionid, parent_class__parent_program=prog).first()
+                        if section:
+                            conflicts = section.get_conflicts(request.user)
+                            conflict_titles = ", ".join([c.emailcode() for c in conflicts])
+                            confirm_msg = "This class conflicts with your schedule! If you add this class, you will be removed from %s. Do you want to proceed?" % conflict_titles
```
willgearty: For normal student registration you used title, but here you used emailcode. I would use title in both, students aren't going to know which email code corresponds to which class.

esp/esp/program/modules/handlers/studentonsite.py
```diff
@@ -64,7 +64,7 @@ def module_properties(cls):
 
     @main_call
     @needs_student_in_grade
-    @meets_deadline('/Webapp')
+    # @meets_deadline('/Webapp')
```
willgearty: I assume this was for testing, but you should restore this decorator

esp/esp/program/modules/handlers/studentonsite.py
```diff
@@ -83,7 +83,7 @@ def studentonsite(self, request, tl, one, two, module, extra, prog):
 
     @aux_call
     @needs_student_in_grade
-    @meets_deadline('/Webapp')
+    # @meets_deadline('/Webapp')
```
willgearty: I assume this was for testing, but you should restore this decorator

esp/public/media/scripts/ajax_tools.js
willgearty, line 1: Note that this script isn't loaded in the [/fillslot template](https://github.com/learning-unlimited/ESP-Website/blob/main/esp/templates/program/modules/studentclassregmodule/fillslot.html), only in the student catalog (there are two places to add a class for normal student registration: `/learn/[program]/catalog` and `/learn/[program]/fillslot/[slot id]`)

willgearty (changes requested):
The student onsite replacement pathway works really well now, thanks!

But it's very clear the normal student registration pathway was never tested (as I've noted below, you need to check /catalog and /fillslot). Both pages still give a full new error page when trying to add a replacement class:

Image

Right now, both sets of buttons submit a POST directly to the `addclass` endpoint, which raises the error. You've only modified the `ajax_addclass` endpoint, which is never touched.

Two options:

1. Actually handle the button pressing/form submission with javascript (it seems this is what ajax_tools.js is trying to do). Your changes to `ajax_addclass` and `ajax_tools.js` should then work and raise a a confirmation message to the user.
2. Repeat the changes you made for the webapp to just show a new confirmation page if a student tries to add a replacement class in normal student registration or from the catalog.

esp/templates/admin/login.html
```diff
@@ -1,6 +1,6 @@
 {% extends "admin/base_site.html" %}
 
-{% load i18n admin_static %}
```
willgearty: This change seems unrelated?

esp/public/media/scripts/ajax_tools.js
willgearty, line 1: I don't think this is ever used in the new confirmation pathways.

esp/esp/program/modules/handlers/studentclassregmodule.py
```diff
@@ -448,6 +516,28 @@ def ajax_addclass(self, request, tl, one, two, module, extra, prog):
...
+                            form_selector = '#prereg_%s' % sectionid
+                            resp = HttpResponse(content_type='application/json')
+                            resp.content = json.dumps({
+                                'conflict': True,
+                                'confirm_msg': confirm_msg,
+                                'form_selector': form_selector
+                            })
+                            return resp
+                    except (ClassSection.DoesNotExist, ValueError):
+                        pass
```
willgearty: This is never used for any of the confirmation pathways.

esp/templates/program/modules/studentonsite/conflict_confirm.html
```diff
@@ -0,0 +1,38 @@
+{% extends "program/modules/studentonsite/webapp.html" %}
+
+{% block head %}
+<script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/jquery/{{ settings.JQUERY_VERSION }}/jquery.min.js" integrity="{{ settings.JQUERY_HASH }}" crossorigin="anonymous"></script>
+<script src="https://code.jquery.com/jquery-migrate-3.3.2.{% if not settings.DEBUG %}min.{% endif %}js"></script>
+<script type="text/javascript" src="/media/scripts/jquery.cookie.js"></script>
```
willgearty: I don't think we need jquery on this page, right?

willgearty (changes requested):
This works great now, thanks! And I really appreciate the addition of some more tests, too!

All that's left is some cleanup of used code and unrelated changes (see my comments below and my comments above from my previous review).

esp/templates/admin/login.html
Supriyo760: This template was loading admin_static, which was deprecated in Django 2.1 and completely removed in Django 3.0. Since the codebase is running on Django 3.2, attempting to load this page with {% load admin_static %} throws a TemplateSyntaxError and crashes the admin login page. This change is necessary to keep the admin page functioning, even though it's technically unrelated to the class replacement feature itself.

esp/esp/program/modules/handlers/studentclassregmodule.py
Supriyo760: These changes are active and critical for standard desktop users. By default, the desktop registration catalog uses AJAX (ajax_addclass) to add/swap classes.
When a schedule conflict occurs, the backend catches the error and returns a JSON response containing {conflict: true, confirm_msg: ..., form_selector: ...}. ajax_tools.js intercepts this via apply_fragment_changes to show the browser's native confirm() popup. If the user confirms, it appends force_replace=true and submits.
The standalone studentclassregmodule/conflict_confirm.html template only serves as a fallback when JavaScript is disabled. If we remove the AJAX JSON pathway, desktop users with JS enabled will just get a generic schedule conflict error and will not be prompted to replace conflicting classes.

esp/templates/program/modules/studentonsite/conflict_confirm.html
Supriyo760: Yah this is a standalone HTML page that submits a standard POST form, it doesn't need to load jQuery, jquery-migrate, cookies, or csrf_init.js. I will clean this up

esp/esp/program/modules/handlers/studentclassregmodule.py
willgearty: I just tested the catalog, and I confirmed this pathway is never triggered (I put a print statement right at the beginning of the `ajax_addclass` method). Instead, it always uses the `addclass` method and shows the separate confirmation page.

`ajax_tools.js` is indeed loaded for the catalog, but the setup is never triggered because there is no element in the catalog with id "student_schedule":
https://github.com/learning-unlimited/ESP-Website/blob/1f149bc4dff6d9aac17e24860aec6200b344acf8/esp/templates/inclusion/program/class_catalog.html#L88-L90

esp/templates/admin/login.html
willgearty: OK, thanks!

esp/esp/django_settings.py
willgearty: @Supriyo760 please revert if not needed.

esp/esp/utils/apps.py
willgearty: @Supriyo760 please revert if not needed.

willgearty (approved):
Thanks for cleaning up those unneeded bits, looks good now! This will save lots of students from the headache of removing existing classes before replacing them!

PR #4634, opened 2026-03-03
Fix missingteachers 'when' to default to 'date' when viewing 

### LINKED ISSUE #2581: "when" for missingteachers doesn't match "date"
When looking at the teacher check-in page for an entire day, I realized that it bases the check-in status of a teacher on whether they are currently checked-in (for the current day), not the day that you are looking at.
This would be fine if you were looking at the check-in for the current day (probably the most common use case), but for future days this doesn't make much sense. I'll propose two solutions that I think would work well:

1. Change the urls on the teachercheckin page to have a default `when` GET parameter, which would be the same as the `date` GET parameter (as a side note, clicking the "Back" link on any `missingteachers` page seems to preserve the `when` GET parameter on the `teachercheckin` page; is this intended behavior?)
2. Change the default for `when` in the `missingteachers` view such that it defaults to `date` if `date` is provided

Dhirenderchoudhary:
Date : 3 March 2026

## Description
Fix the missingteachers view to default the 'when' parameter to the 'date' being viewed when that date is not today. Previously, when viewing the teacher check-in page for a future or past day, the check-in status was incorrectly based on the current date (today) rather than the date being viewed.

## Related Issue
Closes #2581

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

## Checklist

- [x] Self-reviewed the code


esp/esp/program/modules/handlers/teachercheckinmodule.py
```diff
@@ -513,6 +489,12 @@ def missingteachers(self, request, tl, one, two, module, extra, prog):
                 context['url_when'] = request.GET['when']
         else:
             when = None
+
+        # Default 'when' to start of date if viewing a non-today date
+        if when is None and date is not None and date != date_type.today():
+            when = datetime.combine(date, time.min)
+            context['when'] = when
```
Oval17, line 551: The template (missingteachers.html) renders url_when in 5 places whenever when is  truthy 
e.g. in lines 344 all produce links like href="teachercheckin?when={{ url_when }}". The existing code (lines 488–489) always sets context['url_when'] alongside context['when'].
 Are we omitting context['url_when'] will this lead to something like you know lets say links like ("Back", "Previous day", "Next day") will silently render with a blank when= query param, breaking the preserved-when-filter behavior. 

Maybe this when give better result 
context['url_when'] = when.strftime("%m/%d/%Y %H:%M")
Look into this once..

esp/esp/program/modules/tests/teachercheckinmodule.py
```diff
@@ -142,3 +142,22 @@ def test_phone_numbers_on_checkin_page(self):
...
+
+    def test_missingteachers_when_defaults_to_date(self):
+        """Viewing missingteachers for a non-today date should default 'when' to that date."""
+        self.assertTrue(self.client.login(username=self.admin.username, password='password'))
+
+        program_dates = self.program.dates()
+        today = datetime.date.today()
+        test_date = next((d for d in program_dates if d != today), None)
+
+        if test_date is not None:
```
Oval17: Umm here I guess seem like a silently passing test 
if program_dates contains only today's date (or is empty), test_date is None and the entire if block will be skipped. The test appears to pass in CI with zero assertions actually checked.

Oval17 (changes requested):
Units tests are failing here, can you look here @Dhirenderchoudhary, rest most of the changes LGTM

willgearty (approved):
Just a few minor fixes for edge cases. Should be good to merge now, thanks for your contribution!

PR #4650, opened 2026-03-03
Refactor hardcoded "Lunch" category checks to use ClassCategories.is_lunch

### LINKED ISSUE #770: Lunch category should be treated similar to open class category
@luac and I were just talking this over. program.class_categories.all() returns all categories including lunch, but excluding walk-ins, for example. It seems to make sense to do a similar thing for Lunch -- have both it and open class category be non-program-specific categories that are linked to via program attributes.

Boopana-M:
Description
This PR resolves issue #770 by removing hardcoded "Lunch" category checks and centralizing the logic in a single model property.
A new is_lunch property has been added to the ClassCategories model, and all existing string comparisons against "Lunch" have been replaced with this property. This removes fragile string-based logic and improves maintainability while keeping behavior unchanged.

Related Issue
Closes #770

Type of Change
-> Bug fix
-> New feature
-> Breaking change
-> Documentation update
-> Refactor / cleanup
-> CI/CD or build change

Changes Made
->Added ClassCategories.is_lunch property (case-insensitive and safe for None)
->Replaced hardcoded "Lunch" comparisons in:
->studentlunchselection.py
->studentregtwophase.py
->Removed obsolete FIXME comment referencing issue #770

Testing
->Existing tests pass
->New tests added (not required for this refactor)
->Manually verified
->Behavior remains unchanged; this is strictly a refactor to centralize Lunch category detection logic.

AI Disclosure
AI tools were used to help locate occurrences of hardcoded Lunch logic within the codebase. All changes were reviewed, validated, and finalized by me.

Checklist
->Self-reviewed the code
->Documentation update not required
->No unrelated changes included
->Branch created cleanly from main


Boopana-M:
Rebased on main and resolved conflicts. Ready for review

Boopana-M:
Hii @willgearty 
Just checking in on PR #4650. All checks are passing and the changes are scoped only to issue #770 (centralizing Lunch detection via model property and replacing hardcoded checks). If there are any concerns or requested adjustments, I’d be happy to address them quickly 
thank you

willgearty:
While this does indeed refactor the hardcoded "Lunch" category checks to a single place (instead of 3), this doesn't actually address the request of #770.

Issue #770 says "Lunch category should be treated similar to open class category." The `open class` pattern has a **database boolean field** `is_open_class` on `ClassCategories`.

This PR adds a **property** `is_lunch` that just checks `self.category.lower() == 'lunch'`. It replaces `category.category == 'Lunch'` scattered across the codebase with `category.is_lunch` — but `is_lunch` still just checks if the string is "Lunch". The hardcoded string comparison is now in one place instead of three, but it's still a hardcoded string comparison.

The real fix (matching the `is_open_class` pattern) would be adding a `BooleanField` `is_lunch` to `ClassCategories` with a data migration that sets it `True` for existing "Lunch" categories. That way chapters could name their lunch category whatever they want and it would still work (e.g., if it's actually a "Dinner").

Oval17 (changes requested):
Hi @Boopana-M 
Unit tests are failing here , can you look here.
Thanks

willgearty (approved):
I made some additional changes to make sure there is only ever one lunch category. Everything works smoothly locally, so looks like this is good to go. Thanks!

PR #4660, opened 2026-03-03
security: migrate SHA-1 to SHA-256 across codebase

### LINKED ISSUE #4659: [Feature]: Migrate SHA-1 to SHA-256 across codebase
### Problem Statement

SHA-1 is used in 5 files for hashing. SHA-1 has been cryptographically broken since the SHAttered collision attack (2017) and is deprecated by NIST. Additionally, `program/tests.py` uses `hashlib.sha1(str(random.random()))` for salt generation, combining a deprecated hash with an insecure PRNG — a bad pattern that could be copied into production code.

OWASP Reference: A04:2025 Cryptographic Failures — Deprecated Hash Algorithms.

### Proposed Solution

- `users/models/__init__.py`: Replace `hashlib.sha1()` → `hashlib.sha256()` in `PersistentQueryFilter` (3 locations)
- `qsd/models.py`: Replace in `qsd_edit_id()`
- `themes/controllers.py`: Replace in file integrity checks (2 locations)
- `survey/templatetags/survey.py`: Replace for cache key generation
- `program/tests.py`: Replace `hashlib.sha1(str(random.random()))` with `secrets.token_hex(3)`

The `PersistentQueryFilter.sha1_hash` DB column is **not** renamed to avoid a schema migration. A follow-up can rename the column separately.

Arnav1709:
## Description

Migrate all `hashlib.sha1()` calls to `hashlib.sha256()` across 5 files. Replace insecure `random.random()` + SHA-1 salt generation in tests with `secrets.token_hex()`. The DB column `sha1_hash` is intentionally not renamed to avoid a schema migration.

## Related Issue

Closes #4659 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

## Checklist

- [x] Self-reviewed the code


userAdityaa:
Hey @Arnav1709, CodeQL is flagging this because full_filename is derived from user-controlled input and used directly in file operations (open, getsize), which can lead to a path traversal / arbitrary file read vulnerability.

**Two possible fixes:**
1. Constrain the path to a fixed base directory and validate using os.path.abspath + prefix check.
2. Strip directory components and only allow safe filenames (e.g., Path(user_input).name) or enforce a strict whitelist.

This isn’t directly caused by the changes in this PR, but since it’s a legitimate security issue, it would be good to address it here if possible.

esp/esp/themes/controllers.py
```diff
@@ -402,7 +402,7 @@
                 result += self.get_file_summaries(full_filename)
             else:
                 file_data = open(full_filename, 'rb').read()
-                result.append((full_filename, os.path.getsize(full_filename), hashlib.sha1(file_data).hexdigest()))
+                result.append((full_filename, os.path.getsize(full_filename), hashlib.sha256(file_data).hexdigest()))
```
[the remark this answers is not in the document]
willgearty: This is a preexisting issue, it can be ignored for now.

esp/esp/users/models/__init__.py
```diff
@@ -2114,7 +2062,7 @@ class Meta:
         db_table = 'users_k12school'
 
     @classmethod
```
willgearty, line 2209: You can't start saving new objects with a different hashing algorithm, because then the column won't be consistent with existing objects. You either need to revert this or add a migration for all of the existing objects to convert them to sha256.

willgearty (changes requested):

esp/esp/users/models/__init__.py
Arnav1709: I'll add a data migration to rehash all existing rows to SHA-256 so the column stays consistent. Also noticed create_from_Q still has a stale sha1() call on the lookup side, will fix that too. Will rebase onto main as well.

Arnav1709:
### Migration Plan

Adding a Django data migration (`users/migrations/0044_rehash_pqf_sha256.py`):

```python
from django.db import migrations
import hashlib

def rehash_sha1_to_sha256(apps, schema_editor):
    PersistentQueryFilter = apps.get_model('users', 'PersistentQueryFilter')
    for pqf in PersistentQueryFilter.objects.all():
        pqf.sha1_hash = hashlib.sha256(bytes(pqf.q_filter)).hexdigest()
        pqf.save(update_fields=['sha1_hash'])

def rehash_sha256_to_sha1(apps, schema_editor):
    PersistentQueryFilter = apps.get_model('users', 'PersistentQueryFilter')
    for pqf in PersistentQueryFilter.objects.all():
        pqf.sha1_hash = hashlib.sha1(bytes(pqf.q_filter)).hexdigest()
        pqf.save(update_fields=['sha1_hash'])

class Migration(migrations.Migration):
    dependencies = [
        ('users', '<previous_migration>'),
    ]
    operations = [
        migrations.RunPython(rehash_sha1_to_sha256, rehash_sha256_to_sha1),
    ]
```

This rehashes every existing `PersistentQueryFilter.sha1_hash` from SHA-1 → SHA-256 so the column stays consistent with the new code. Reverse migration included for rollback safety.

willgearty (approved):
After consulting with Claude, I was wrong to request a migration (which would have caused all of the rows from before 2024 to have the same hash value). Instead, we'll just switch cold turkey to sha256. This might cause the creation of new rows that match old rows semantically, but at least all of the hash values will be different.

Should be good to go now, thanks for your contribution!

PR #4666, opened 2026-03-03
Fix QSD GUI breaks with HTML comment edge cases

### LINKED ISSUE #2746: QSD GUI breaks in weird edge cases
I'm not really sure what exactly causes the problem, but there seem to be some weird edge cases where the new QSD GUI breaks. The QSD then appears to have the old editor, but all of the buttons and such disappear:

image

When you try to save the HTML, you get an error:

image

However, when you go to .edit.html (for a static page), it still has the GUI, but even after saving, the GUI is still gone on the page itself.

I've found the easiest way around this is to always return to the rich-text mode before saving (instead of saving while in source mode). This ensures that the code is in a format that the package likes.

If the problem occurs, the easiest solution is to copy the source code for later, edit the QSD record in the admin pages to something simple (e.g. "Hello World"), and then go back to the page and edit it to the text you saved earlier. You'll want to add the text back in incrementally to figure out which part broke it (and if so, you may need to remove parts from the QSD record in the admin pages again).

Dhirenderchoudhary:
Date : 4 March 2026

## Description

Fix QSD (Quasi-Static Data) editor GUI breaking when HTML content contains edge case comments like nested markers ( -->) or comments with blank lines.

## Related Issue

Closes #2746 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

## Checklist

- [x] Self-reviewed the code


esp/esp/qsd/views.py
```diff
@@ -62,10 +62,20 @@
...
 
 logger = logging.getLogger(__name__)
 
+
+def _sanitize_html_comments(value):
+    """Remove nested comment markers that break HTML parsing."""
+    def fix_comment(match):
+        content = match.group(1).replace('<!--', '').replace('-->', '')
+        return '<!--' + content + '-->'
+    return re.sub(r'<!--(.*?)-->', fix_comment, value, flags=re.DOTALL)
```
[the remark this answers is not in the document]
Dhirenderchoudhary: fixed it please resolve the comment

esp/esp/utils/templatetags/markup.py
```diff
@@ -10,6 +10,8 @@
 pin to a given version of markdown, and don't use markdown extensions.
 """
 
+import re
```
Oval17: unused imprt

esp/esp/qsd/views.py
```diff
@@ -63,10 +62,32 @@
 
 import logging
 import os
+import re
```
Oval17: unused

esp/esp/utils/templatetags/markup.py
```diff
@@ -20,7 +22,34 @@
 register = template.Library()
 
 
+def _sanitize_html_comments(value):
```
Oval17: this and above method in views py are exact replica 
Will it be possible to move it to some shared location and use it?

Also see a bit about DRY in programming ..

Oval17 (commented):
Just a observation does `QsdModel.html()` bypasses sanitization in anyway 
can you see this once and confirm it?
cause I think cached html() method renders QSD content to HTML without going through _sanitize_html_comments.
if this is not case then it might be fine...

Dhirenderchoudhary:
Fixed

willgearty (changes requested):
These changes look good, but it doesn't actually address the issue in #2746, which was about blank spaces within comments (even though your PR description says this handles those).

Please add a fix for blank lines and address the Copilot comments.

And for future PRs, please make sure to also note what other changes were made. There are a bunch of JS changes here totally unrelated to #2746. They're useful, so we should keep them, but you didn't note these changes in your PR description at all.

willgearty:
@Dhirenderchoudhary just reminding you about this PR whenever you are free. You'll also need to address the merge conflicts.

Dhirenderchoudhary:
I'll fix the needed changes asap as in 2-3 days my exams are over so i'll be free for next 2 months and i'll contribute with my full dedication. Sorry for delay.

willgearty (approved):
Just some minor cleanup, otherwise works well, thanks!

PR #4668, opened 2026-03-03
Fix ChangelogFetcher bugs causing scheduled classes to disappear (#3591)

### LINKED ISSUE #3591: Caching issue with AJAX scheduler
In certain cases (needs reproduction), after classes are scheduled, they will disappear from the AJAX scheduler (though they are still saved in the database correctly). This issue appears to be related to server caching, as clearing the browser cache does not solve it but waiting a day does. Sometimes classes appear and disappear in groups consistently.

abhinavjha0239:
# PR: Fix caching issue with AJAX scheduler

**Branch**: `fix/ajax-scheduler-caching-3591`
**Issue**: Closes #3591

---

## Summary

Fix two JavaScript bugs in `ChangelogFetcher.js` that caused scheduled classes to "disappear" from the AJAX scheduler view even though they were saved correctly in the database.

## Root Cause

Two bugs in the client-side changelog polling mechanism:

### Bug 1: `this` scoping error in constructor

```js
// BEFORE (broken)
$j.getJSON('ajax_schedule_last_changed', function(data, status) {
    if (status == "success") {
        this.last_applied_index = data['latest_index'];
        //  ^ 'this' is jQuery's AJAX settings object, NOT ChangelogFetcher
    }
});
```

Inside a `$.getJSON` callback, `this` refers to the jQuery AJAX settings object, not the `ChangelogFetcher` instance. So `last_applied_index` was **always stuck at 0**, regardless of the server's response. This meant the client never correctly tracked its position in the changelog.

### Bug 2: Server `reload` command silently ignored

When the changelog is pruned (entries older than 12 hours are deleted), the server sends a reload command:
```json
{ "other": [{ "command": "reload", ... }] }
```

Note: this response has **no `changelog` key**.

The old client code did:
```js
$j.each(data.changelog, function(id, change) { ... });
```

`$j.each(undefined, fn)` silently does nothing in jQuery. The `reload` command in `data.other` was **never checked**. The client continued polling with a stale index, getting the same ignored reload response every 5 seconds. The scheduler displayed stale JS-cached data indefinitely.

### Why existing workarounds worked
- **"Waiting a day"**: User eventually reloads the page, getting fresh data from `schedule_assignments`
- **"Clearing the AJAX change log"**: The admin view redirects the page, forcing a full reload

## Fix

**1 file changed**: `esp/public/media/scripts/ajaxschedulingmodule/ESP/ChangelogFetcher.js`

| Change | What | Why |
|--------|------|-----|
| `var self = this` capture | Use `self` instead of `this` in `$.getJSON` callback | Fix scoping so `last_applied_index` is correctly set on the ChangelogFetcher instance |
| `data.other` reload handling | Check for `reload` command before processing changelog | Detect when server says data is out of sync |
| `if (data.changelog)` guard | Wrap `$j.each` in a null check | Prevent silent no-op when `data.changelog` is undefined |
| `showReloadNotification()` | New method: shows a red banner with "Reload now" link | Non-destructive UX: user decides when to reload (preserves unsaved state) |
| `pollInterval` tracking | Store interval ID + clear before re-setting | Prevent interval leaks and allow stopping polls when data is stale |

### Why a notification banner (not auto-reload)

A hard `window.location.reload()` would:
- Destroy any unsaved state (mid-drag, pending server requests)
- Trigger silently from background polling with no warning
- Give the user zero control

The banner approach preserves all local state, communicates clearly, and lets the user reload when ready. Polling is stopped to avoid repeated server hits.

## Testing

- All 6 `ajaxschedulingmodule` tests pass
- Full test suite passed

### Notification banner shown
New code — reload command detected, polling stopped, user-friendly banner displayed.

reload-banner-visible

### Normal scheduler state (no regression)
Scheduler works normally when changelog responses contain valid data.

scheduler-normal-state


willgearty:
@abhinavjha0239 Sorry for the delay, and thanks for identifying these bugs! Will you be able to address the two Copilot comments? Moving the banner to `scheduling.css` would make things a lot cleaner (ignore its suggested fix here which doesn't actually do what it recommends). Adding tests to cover this new behavior would be nice, but not entirely necessary.

If you won't be able to work on this at the moment, let me know and I can take it from here.

abhinavjha0239:
> @abhinavjha0239 Sorry for the delay, and thanks for identifying these bugs! Will you be able to address the two Copilot comments? Moving the banner to `scheduling.css` would make things a lot cleaner (ignore its suggested fix here which doesn't actually do what it recommends). Adding tests to cover this new behavior would be nice, but not entirely necessary.
> 
> If you won't be able to work on this at the moment, let me know and I can take it from here.

I will do it by tonight.

willgearty (approved):
Thanks for cleaning this up. It works well for me locally.
This will be a huge relief to our users. We've been trying to figure out this bug for years! Thanks!

PR #4672, opened 2026-03-03
fix target is null JS error in fruitsalad (#1291)

### LINKED ISSUE #1291: Javascript errors
I think we are getting some Javascript errors that may need to be tracked down (my "server log" e-mail folder has 368,000 messages, although this is not the only cause).

When I navigate to one of our sites or a dev server in Firefox 31 (Windows 8), there is a Javascript error:
"Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://c328740.ssl.cf1.rackcdn.com/mathjax/latest/config/TeX-AMS-MML_HTMLorMML.js. This can be fixed by moving the resource to the same domain or enabling CORS."
I'm not sure why this is happening as the MathJax script is included in the page with a script tag.  Also, the references to the LU CDN (Cloudfront addresses) don't seem to be causing trouble, unless things are being executed out of order.

I have also seen some different errors in Chrome.  Are these impacting other things?  I wouldn't want to see more CSRF issues...

Supriyo760:
## Description

This pull request resolves the remaining Javascript error mentioned in Issue #1291 (`target is null`) that was spamming the server logs.

A null check was added to [esp/esp/themes/theme_data/fruitsalad/templates/bubblesfront.html](cci:7://file:///c:/Users/Dange/Downloads/ESP-Website-main/ESP-Website-fresh/esp/esp/themes/theme_data/fruitsalad/templates/bubblesfront.html:0:0-0:0) to ensure the Javascript wrapper behavior does not throw a `"target is null"` exception when a user is already authenticated. (The login inputs are wrapped in a Django conditional `{% if not request.user.is_authenticated %}`, meaning the DOM nodes do not exist when an authenticated user visits the index page. This check gracefully exits if those elements aren't present).

*(Note: The remaining error mentioned in the issue regarding MathJax CORS appears to have already been fixed when the CDN link was updated in `esp/templates/elements/html` to a proper Cloudflare URL.)*

## Related Issue

Closes #1291

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

## AI Disclosure

AI was just used for suggestion.

## Checklist

- [x] Self-reviewed the code

### DIFF
--- a/esp/esp/themes/theme_data/fruitsalad/templates/bubblesfront.html
+++ b/esp/esp/themes/theme_data/fruitsalad/templates/bubblesfront.html
@@ -36,7 +36,7 @@
       target.onblur();
     };
 
-    // set behavior on Username, Password, and Search fields
+    // set behavior on the login Username and Password fields
     var loginUser = document.getElementById('login_user_index');
     if (loginUser) {
       wrap(loginUser, 'Username');
@@ -47,7 +47,9 @@
 
   function set_teachlearn(tl) {
     if (document.getElementById('remember').checked) {
-      document.cookie = 'teachlearn=' + tl + '; expires=' + new Date("January 1, 2025").toGMTString() + '; path=/';
+      var expires = new Date();
+      expires.setFullYear(expires.getFullYear() + 1);
+      document.cookie = 'teachlearn=' + tl + '; expires=' + expires.toUTCString() + '; path=/';
     }
   }
 </script>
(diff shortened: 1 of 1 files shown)


willgearty (approved):
Looks good, thanks!

PR #4675, opened 2026-03-04
Separate moderator vs teacher unavailability in scheduler (#3654)

### LINKED ISSUE #3654: Separating notions of moderator vs teacher unavailability in scheduler
Lingering thoughts from #3644:

When a moderator is unavailable but all teachers are available, I feel like we should denote that differently than when a teacher is unavailable, since it would let the admin know that if they remove the moderator then it would work. One idea I have is that the blocks where it is ONLY a moderator that is unavailable would be grey with blue slashes through them or something like that.

Note this may be a lot of work to implement according to Will, as we combine all of the teacher and moderator availabilities before we then change the colors of the sections.

abhinavjha0239:
## Description

When a section is selected in the AJAX scheduling module, teacher and moderator availabilities are combined using `_.intersection()` before highlighting cells. This means if a moderator is unavailable but all teachers ARE available, the timeslot looks exactly the same as when a teacher is unavailable — just dim/dark. Admins have no way to tell that simply removing the moderator would make the timeslot schedulable.

This PR separates the two cases visually. Timeslots where all teachers are free but a moderator is not are now shown with **grey-with-blue diagonal stripes** instead of plain dark grey. This gives admins an actionable signal: "remove the moderator and you can schedule here."

### What changed

- `getAvailableTimeslots()` in `Sections.js` now computes teacher-only and combined intersections separately, returning a third element: the list of timeslots where only the moderator blocks scheduling
- `highlightTimeslots()` and `unhighlightTimeslots()` in `Matrix.js` apply/remove the new `moderator-only-unavailable-cell` CSS class for those timeslots
- New CSS class in `scheduling.css` with grey + blue stripe pattern
- New legend entry (only shown when moderator module is enabled) explaining the stripe pattern

### Before — both look the same

**S13s1** — has a moderator (Mod One). The moderator is unavailable at Slot 2, but the teacher (Alice Smith) IS available at both slots:

before-S13s1-matrix before-S13s1-info

**A14s1** — no moderator. The teacher (Bob Jones) is only available at Slot 1:

before-A14s1-matrix before-A14s1-info

Slot 2 looks identical in both — plain grey. But the situations are completely different: for S13s1, removing the moderator would fix it. For A14s1, nothing can be done.

### After — now you can tell the difference

**S13s1** — Slot 2 now shows blue stripes instead of grey:

after-S13s1-matrix after-S13s1-info

**A14s1** — still plain grey (teacher is unavailable, not a moderator issue):

after-A14s1-matrix after-A14s1-info

The legend also gets a new entry explaining what the blue stripes mean:

after-legend

## Related Issue

Closes #3654

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

All 6 `ajaxschedulingmodule` Django tests pass. Full test suite passed.

Manually verified:
- Blue stripes appear for moderator-only-unavailable timeslots
- Sections without moderators show no blue stripes (same as before)
- Yellow (busy) takes priority over blue stripes
- Dim cells correctly show when a teacher is unavailable
- Blue-striped cells are not clickable/schedulable
- Stripes clean up properly when switching between sections
- Override mode shows all green (no blue stripes)
- Legend entry appears only with moderator module enabled

## AI Disclosure

## Checklist

- [x] Self-reviewed the code


samalarussa (changes requested):
The formatting is successful, but in my testing it misses the case of a moderator being unavailable due to being assigned to teach a section during a conflicting block. Currently for that case, the block is still highlighted in yellow vs gray with blue stripes. Could you address the case of a moderator being unavailable due to being scheduled to teach during that block, and also any merge conflicts?

samalarussa:
Closing due to inactivity and outstanding requested changes. Please re-assign yourself to this issue and reopen this PR if you wish to continue with this PR.

PR #4701, opened 2026-03-04
fix(onsite): resolve UnboundLocalError in printschedule_status user lookup

### LINKED ISSUE #4700: [Bug]: printschedule_status crashes with UnboundLocalError when user lookup fails
### Description

In the on-site endpoint, a try/except block attempts to look up a student by ID. When the lookup fails (_invalid ID, non-integer input, or non-existent user_), the except block sets an error message but does not return. Execution continues into the next lines, which reference `user_obj`, a variable that was never assigned, causing an **UnboundLocalError and an HTTP 500** instead of the intended graceful error response.

The except clause is also a bare except, meaning it silently swallows all exceptions, including database failures.

### Steps to Reproduce

1. Log in as a user with onsite permissions
2. Navigate to the print schedule endpoint directly with a bad user ID:
`GET /onsite/<program_url>/printschedule_status?user=999999`
3. or with garbage input:
`GET /onsite/<program_url>/printschedule_status?user=<notanumber or #!@>`
4. Expected: JSON response with {"message": "Could not find user 999999."}
5. Actual: HTTP 500 — UnboundLocalError: local variable 'user_obj' referenced before assignment

### Expected Behavior

A JSON response with a human-readable error message:

`{"message": "Could not find user 999999."}` or something related

### Actual Behavior

HTTP 500 — UnboundLocalError: local variable 'user_obj' referenced before assignment

### Screenshots

Image

### Operating System

Any (reproduced on Ubuntu 22.04)

### Browser

Any (this is a JSON API endpoint, reproducible via curl or browser)

### Additional Context

The bare except: also masks real errors like database connection failures, giving a misleading "Could not find user" message when the actual problem is infrastructure-level.

BalaSubramaniam12007:
## Description

The printschedule_status view in onsiteclasslist.py had a try/except block that caught user lookup errors, but did not return after handling them. This caused execution to fall through to lines that used user_obj, which was never assigned, resulting in an UnboundLocalError and an HTTP 500 on every failed lookup.

## Fix :
Split the bare except into two specific handlers:
- TypeError/ValueError for non-integer ?user= params
- ESPUser.DoesNotExist for valid integers with no matching user

Previously, the single except block caught the error, set a message, but did not return, causing user_obj to be referenced before assignment and raising UnboundLocalError on every failed lookup (HTTP 500).

Also replaces bare with specific exception types to avoid silently swallowing unrelated errors like DB failures.

unbounded error fix

## Related Issue

Closes #4700 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor/cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

## Checklist

- [x] Self-reviewed the code


esp/esp/program/modules/handlers/onsiteclasslist.py
```diff
@@ -328,21 +322,28 @@ def update_schedule_json(self, request, tl, one, two, module, extra, prog):
...
             user_obj = ESPUser.objects.get(id=user)
-        except (ValueError, TypeError, KeyError, ESPUser.DoesNotExist):
-            resp.status_code = 400
-            result['message'] = "Could not find user %s." % request.GET.get('user', None)
+        except ESPUser.DoesNotExist:
+            result['message'] = "Could not find user %s." % user
             json.dump(result, resp)
             return resp
+        except (ValueError, TypeError, KeyError, ESPUser.DoesNotExist):
+            result['message'] = "Could not find user %s." % request.GET.get('user', None)
```
willgearty, line 341: I believe this can now be removed.

willgearty (changes requested):

willgearty:

BalaSubramaniam12007:
> Closing due to unresponsiveness.

Hi, I saw the PR was closed due to inactivity. I am sorry, but I am unable to work on this at the moment. I will revisit and submit a new PR later once I am available.

Happy to work with you all :)

PR #4704, opened 2026-03-04
Fix texescape filter crashing pdflatex when \LaTeX is used in math mo…

### LINKED ISSUE #2928: Using $\LaTeX$ in a course description causes catalog compilation to fail with error
When compiling, it gives the error "You can't use '\spacefactor' in math mode". Other LaTeX commands seem to work fine, but $\LaTeX$ does not work.

abhinavjha0239:
## Description

When teachers write `$$\LaTeX$$` in course descriptions, the `texescape` template filter converts it to `$\LaTeX$` (math mode). But `\LaTeX` uses `\spacefactor` internally — a text-mode-only primitive — so pdflatex crashes with *"You can't use `\spacefactor' in math mode"* and the entire catalog PDF fails to generate.

This fix detects logo commands (`\LaTeX`, `\TeX`, `\XeLaTeX`, `\LuaLaTeX`, `\AmSTeX`) inside math segments and wraps them in `\mbox{}`, which switches back to text mode. `\mbox` is a base LaTeX primitive — zero package dependency, works unconditionally.

### Before — 
before_catalog_error

### After — catalog compiles cleanly
after_catalog_pdf

## Related Issue

Closes #2928

## Type of Change

- [x] Bug fix

## What Changed

| File | Change |
|------|--------|
| `esp/esp/utils/templatetags/latex.py` | Regex to wrap `\spacefactor`-using logo commands in `\mbox{}` inside math blocks |
| `esp/esp/utils/tests.py` | 12 new `TexEscapeTest` cases (first-ever tests for `texescape`) |

## Testing

- [x] Existing tests passed
- [x] 12 new unit tests added covering all 5 logo commands, edge cases, and non-regression
- [x] Verified with pdflatex compilation  (with and without `amsmath`)
- [x] Verified end-to-end via catalog PDF generation on running ESP

## Checklist

- [x] Self-reviewed the code


kkbrum (approved):
Thanks!

PR #4710, opened 2026-03-05
Fix incorrect response index in get_class_application_2 and get_class_application_3

### LINKED ISSUE #4709: [Bug]: Incorrect response text returned in get_class_application_2() and get_class_application_3()
### Description

While reviewing listgenmodule.py, I noticed that the methods get_class_application_2() and get_class_application_3() return the wrong response text.

Both methods correctly reference the appropriate response when retrieving the question subject (responses[1] and responses[2]). However, the response text is still taken from responses[0], which means the first response is always displayed regardless of which application field is selected.

Because of this, the output combines the correct question subject with the wrong response text.

Location in Code

File:

esp/esp/program/modules/handlers/listgenmodule.py

Relevant section (around lines ~256–273):

def get_class_application_2(self):
    responses = self.user.listAppResponses(self.program)
    if len(responses) > 1:
        return str(responses[1].question.subject) + ':  ' + str(responses[0])
    else:
        return None

def get_class_application_3(self):
    responses = self.user.listAppResponses(self.program)
    if len(responses) > 2:
        return str(responses[2].question.subject) + ':  ' + str(responses[0])
    else:
        return None

Suggested Fix

Update the response index used in the return statement.

return str(responses[1].question.subject) + ':  ' + str(responses[1])
return str(responses[2].question.subject) + ':  ' + str(responses[2])

### Steps to Reproduce

1. Submit a student application with at least 3 responses.

2. Go to the List Generator module as an admin.

3. Select the fields:
        Class Application 2
        Class Application 3

4. Generate the list.

### Expected Behavior

<Question 2 subject>: <Response 2 text>
<Question 3 subject>: <Response 3 text>

### Actual Behavior

<Question 2 subject>: <Response 1 text>
<Question 3 subject>: <Response 1 text>

### Screenshots

_No response_

### Operating System

_No response_

### Browser

_No response_

### Additional Context

_No response_

Vipeen-Kumar:
## Description

While reviewing `listgenmodule.py`, I noticed that `get_class_application_2()` and `get_class_application_3()` were returning the wrong response text.

Both methods correctly use `responses[1]` and `responses[2]` to fetch the question subject, but the response text itself was still being taken from `responses[0]`. As a result, the first response's text was always displayed even when the second or third application field was selected.

This PR updates the index used for the response text so that each method returns the correct response corresponding to its question.

## Related Issue

Closes #4709

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Manually verified

Verified by reviewing the logic to ensure each method now references the correct response index.

## AI Disclosure

No AI tools were used to generate the final code changes.

## Checklist

- [x] Self-reviewed the code
- [ ] Updated documentation (not required for this change)

### DIFF
--- a/esp/esp/program/modules/handlers/listgenmodule.py
+++ b/esp/esp/program/modules/handlers/listgenmodule.py
@@ -263,13 +263,13 @@ class UserAttributeGetter(object):
     def get_class_application_2(self):
         responses = self.user.listAppResponses(self.program)
         if len(responses) > 1:
-            return str(responses[1].question.subject) + ':  ' + str(responses[0])
+            return str(responses[1].question.subject) + ':  ' + str(responses[1])
         else:
             return None
     def get_class_application_3(self):
         responses = self.user.listAppResponses(self.program)
         if len(responses) > 2:
-            return str(responses[2].question.subject) + ':  ' + str(responses[0])
+            return str(responses[2].question.subject) + ':  ' + str(responses[2])
         else:
             return None
 


willgearty (approved):
Looks good, thanks!

PR #4747, opened 2026-03-06
Fix: replace plaintext activation token with HMAC-signed token

### LINKED ISSUE #4744: [Bug]: Account activation token stored as plaintext suffix in password hash field
### Description

The account activation system appends a plaintext token directly to the user's hashed password in the `auth_user.password` database column:

This is architecturally fragile and has three independent failure modes, each of which can silently break activation or expose the token.

### Steps to Reproduce

**Token exposure (DB read access):**
1. Gain read access to the `auth_user` table (DB backup, SQL error, compromised read replica)
2. Read `password` column of any inactive user
3. Extract suffix after last `_` — this is the raw activation token
4. Navigate to `/myesp/activate/?username=<user>&key=<token>` to activate the account

**Broken activation via `rfind("_")` (reproducible):**
1. Register a new account
2. Before activating, have an admin set the user's password via the Django admin panel
3. The activation token suffix is destroyed — the emailed link now returns "Incorrect key" with no recovery path

**Wrong token extracted (edge case):**
1. A PBKDF2 base64 hash component happens to contain `_` (e.g., ends in `...a3b_c4d`)
2. `user.password[user.password.rfind("_")+1:]` extracts `c4d` instead of the appended token
3. The activation email is sent with the wrong key — link can never work

### Expected Behavior

Activation tokens should be:
- Stored independently of the password hash (not appended to it)
- HMAC-signed and time-limited so they expire automatically
- Invalidated immediately upon use
- Non-extractable even with full read access to the database

Django already ships `PasswordResetTokenGenerator` which satisfies all of the above and is already used correctly in the password reset flow (`ESPPasswordResetConfirmView`).

### Actual Behavior

Current code in `registration.py`:

```python
# Activation check :
if not u.password.endswith("_%s" % request.GET['key']):
    raise ESPError("Incorrect key...")

# Token extraction for resend :
userkey = user.password[user.password.rfind("_")+1:]

### PR DESCRIPTION
**Pull request by SHREESHANTH99**
## Description

Account activation previously stored a random integer token appended directly to the user's password hash field . This is a security vulnerability: the token is plaintext in the database, can be extracted by anyone with DB read access, and is destroyed if the user changes their password before activating.

This PR replaces it with a stateless HMAC-signed token using Django's
`PasswordResetTokenGenerator`, encoded with a `uid`/`token` URL pattern.
The token is never stored in the database and automatically expires via
`PASSWORD_RESET_TIMEOUT` (default 3 days).

## Related Issue

Closes #4744 

## Testing

- [x] Existing tests pass (`Ran 49 tests in 23.036s OK`)
- [x] Manually verified

### CODE REVIEW

**Comment by SHREESHANTH99** [comment]
i have  address all capilot reviews ,Could you please take another look when you have a chance? Thank you.

File: esp/esp/users/tests.py [hunk 1]
```diff
@@ -475,17 +475,68 @@ def phase_2(self):
...
+        self.assertIsNotNone(match, "Activation email does not contain the expected /activate/<uid>/<token>/ URL")
+        # Verify the token is valid for this user
+        uid = force_str(urlsafe_base64_decode(match.group("uid")))
+        self.assertEqual(str(u.pk), uid)
+        self.assertTrue(account_activation_token.check_token(u, match.group("token")))
+
+    def test_activate_account_with_token(self):
+        """Test that new-style HMAC token activation actually activates the user."""
+        if not Tag.getBooleanTag('require_email_validation'):
+            return
```
willgearty: This guard makes the whole test a no-op, so the new activation endpoint is currently untested.

`require_email_validation` defaults to `False` (`esp/esp/tagdict/__init__.py:420-426`), and this test never sets it. Tests within a class run in alphabetical order, so `test_activate_account_with_token` runs before `test_phase_2` (which is the only place that flips the tag), and `TestCase` rolls back each test's changes anyway. The result is that this returns at line 501 every run.

That matters because this is the only test exercising the new `activate_account` view — the security-critical path this PR introduces.

The guard also isn't needed: the test builds its own inactive user and hits the endpoint directly, so it doesn't depend on the tag at all. Please drop lines 500-501, and consider moving the test into the `ActivateAccountTest` class that now exists on `main` (`esp/esp/users/tests.py:1315`) alongside the legacy-flow tests.

---
_Generated by [Claude Code](https://claude.ai/code)_

esp/esp/users/tokens.py
```diff
@@ -0,0 +1,20 @@
...
+    """
+    HMAC-signed, time-limited token for account activation.
+    Automatically invalidated when is_active changes (i.e. after first use).
+    Token is never stored in the database.
+
+    In Django 2.2, tokens inherit Django's PASSWORD_RESET_TIMEOUT_DAYS setting
+    (default: 3 days), ensuring activation links automatically expire.
+    Override PASSWORD_RESET_TIMEOUT_DAYS in local_settings.py to adjust.
+    For newer Django versions, the equivalent setting is PASSWORD_RESET_TIMEOUT
+    (in seconds).
```
willgearty: This is inverted relative to the current tree. `main` pins `Django==5.2.16` (`esp/requirements.txt:2`), not 2.2, and `PASSWORD_RESET_TIMEOUT_DAYS` was **removed** in Django 4.0 — setting it in `local_settings.py` would do nothing here.

The applicable setting is `PASSWORD_RESET_TIMEOUT` (seconds). It isn't set anywhere in the repo, so activation links inherit Django's default of 259200 seconds (3 days), which matches the intent described in the PR body.

Suggested rewrite of lines 10-14:

```
    Tokens expire according to Django's PASSWORD_RESET_TIMEOUT setting
    (seconds; default 259200 = 3 days).  Override it in local_settings.py
    to adjust how long activation links stay valid.
```

---
_Generated by [Claude Code](https://claude.ai/code)_

esp/esp/users/tokens.py
```diff
@@ -0,0 +1,20 @@
...
+    """
+    def _make_hash_value(self, user, timestamp):
+        return f"{user.pk}{timestamp}{user.is_active}"
```
willgearty: Two things on the generator itself:

**Set an explicit `key_salt`.** The subclass inherits `key_salt = "django.contrib.auth.tokens.PasswordResetTokenGenerator"` from the parent, so activation tokens and password-reset tokens are HMAC'd under the same salt. They can't collide today because the hash inputs differ in shape (`is_active` always renders as `True`/`False`, never empty), but domain separation shouldn't depend on that coincidence:

```python
class AccountActivationTokenGenerator(PasswordResetTokenGenerator):
    key_salt = "esp.users.tokens.AccountActivationTokenGenerator"
```

**`is_active` alone makes the token reusable after a deactivation.** Because the hash covers only `pk`, `timestamp` and `is_active`, a link issued shortly before an admin deactivates the account becomes valid again — the holder can flip `is_active` back to `True`. `PASSWORD_RESET_TIMEOUT` caps this at 3 days from issue, so the exposure is narrow, but hashing a dedicated "activation pending" flag instead of `is_active` (see the main review comment) would close it and make the single-use property exact.

---
_Generated by [Claude Code](https://claude.ai/code)_

esp/esp/users/urls.py
```diff
@@ -11,12 +11,14 @@
         name='esp.users.views.user_registration_phase1'),
     re_path(r'^register/information/?$', views.user_registration_phase2,
         name='esp.users.views.user_registration_phase2'),
-    re_path(r'^activate/?$', views.registration.activate_account),
+    re_path(r'^activate/?$', views.registration.activate_account_legacy),
+    re_path(r'^activate/(?P<uidb64>[0-9A-Za-z_\-]+)/(?P<token>[0-9A-Za-z\-]+)/$',
+        views.registration.activate_account, name='activate_account'),
```
willgearty: Moving the `activate_account` URL name onto the new two-argument route will break 8 tests once this is rebased.

`main` now declares the legacy route as `re_path(r'^activate/?$', views.registration.activate_account, name='activate_account')` and added an `ActivateAccountTest` class (`esp/esp/users/tests.py:1315`) whose `setUp` does `self.url = reverse('activate_account')`. With this PR, that name resolves to a pattern requiring `uidb64` and `token`, so `reverse()` raises `NoReverseMatch` in `setUp` and every test in the class errors out.

Least-disruptive fix — keep the existing name on the legacy route and give the new one its own:

```python
re_path(r'^activate/?$', views.registration.activate_account_legacy,
    name='activate_account'),
re_path(r'^activate/(?P<uidb64>[0-9A-Za-z_\-]+)/(?P<token>[0-9A-Za-z\-]+)/?$',
    views.registration.activate_account, name='activate_account_token'),
```

Note the `/?$` on the new route, to match the trailing-slash-optional convention used by every other pattern in this file.

---
_Generated by [Claude Code](https://claude.ai/code)_

esp/esp/users/forms/user_reg.py
```diff
@@ -133,13 +133,22 @@ def __init__(self, *args, **kwargs):
...
+        # Legacy: password ends with _<token>
+        # New: is_active=False and never logged in
+        awaiting_activation = Q(
+            is_active=False,
+            password__regex='\$(.*)_'
+        ) | Q(
+            is_active=False,
+            last_login__isnull=True
+        )
+        if not ESPUser.objects.filter(username__iexact=data).exclude(password='emailuser').filter(awaiting_activation).exists():
```
willgearty: Two problems here — one about what this predicate lets in, and one about a sibling call site that wasn't updated.

**`last_login__isnull=True` matches accounts that were deactivated on purpose.** This form gates the *public* resend-activation view, so any inactive account that never logged in can now be sent a fresh activation link by anyone who knows the username — and clicking it sets `is_active=True`. That includes accounts disabled deliberately:

- merge absorbees — `esp/esp/users/controllers/merge.py:111` sets `absorbee.is_active = False`
- mass deactivation — `esp/esp/program/modules/handlers/deactivationmodule.py:93` does `users.update(is_active=False)`
- admin deactivation via the "Deactivate user" control in `esp/templates/users/userview.html:62`

So this reintroduces the concern from the earlier review round rather than resolving it: `last_login IS NULL` is a proxy for "never activated", but the two aren't the same set.

**`UserRegForm.clean_username` (line 103, not in this diff) still uses the legacy regex, and that breaks re-registration.** That check deliberately *excludes* awaiting-activation accounts so a user who never activated can re-claim their username — `user_registration_validate` supports exactly that, fetching `username=..., is_active=False` to reuse the row. Under the new scheme a pending account's password no longer carries the `_<key>` suffix, so it stops matching `\$(.*)_`, stops being excluded, and the user gets "Username already in use."

I checked the regex against both hashers this repo uses, and neither produces a match without the legacy suffix:

| password value | matches `\$(.*)_` |
| --- | --- |
| `pbkdf2_sha256$1000000$AbC123xyzSalt$Zm9vYmFy...` (new scheme) | no |
| `pbkdf2_sha256$1000000$AbC123xyzSalt$Zm9vYmFy..._123456` (legacy) | yes |
| `md5$$acbd18db...` (`test_settings.py` hasher, new scheme) | no |
| `md5$$acbd18db..._123456` (legacy) | yes |

Salts use `[a-zA-Z0-9]` and the hash is base64/hex, so no underscore can appear after the first `$`. This will fire for every account created after the change.

Two more call sites on `main` need the same treatment (they postdate this branch): `_username_live_validation` (`esp/esp/users/views/registration.py:69`) and `_email_live_validation` (`registration.py:134`) both still use the legacy regex, so the AJAX feedback will contradict the server-side check.

**Suggestion:** rather than patching four predicates independently, add an explicit `is_activation_pending` boolean to `ESPUser` with a migration backfilling it from the legacy `_` suffix. That makes "awaiting activation" a fact rather than an inference, fixes all four sites consistently, keeps deliberately-disabled accounts out, and gives `AccountActivationTokenGenerator._make_hash_value` a precise flag to hash. It's more work than swapping the predicate, but the inference-based approach is what produced this whole class of bug.

---
_Generated by [Claude Code](https://claude.ai/code)_

esp/esp/users/views/registration.py
```diff
@@ -161,31 +169,56 @@ def user_registration_phase2(request):
...
+        raise ESPError("Invalid activation link. Please use the <a href='/myesp/resend/'>resend form</a> to get a new one.", log=False)
+
+    if u.is_active:
+        raise ESPError('This account is already activated. <a href="/accounts/login/?next=/myesp/profile/">Log in</a> or <a href="/myesp/passwdrecover/">recover your password</a>.', log=False)
+
+    if not account_activation_token.check_token(u, token):
+        raise ESPError("This activation link is invalid or has expired. Please use the <a href='/myesp/resend/'>resend form</a> to get a new link.", log=False)
+
+    # Remove legacy activation token suffix from password field if present
+    u.password = re.sub(r'_\d+$', '', u.password)
```
willgearty: This line is correct and necessary, but it reads like dead code — worth a comment explaining *why* a new-style activation would ever encounter a legacy suffix.

The path is: a user who registered under the old scheme (password carries `_<key>`) uses the resend form, which now always issues an HMAC link, so they activate through *this* view rather than the legacy one and their suffix has to be stripped here.

It's also safe against corrupting a real hash, which is worth stating since the regex is being applied to every activating user's password: the `$` anchor means only a trailing `_<digits>` matches, and no hasher in use here can produce that — pbkdf2/md5 salts are `[a-zA-Z0-9]` and the digest is base64 or hex, so no underscore appears after the first `$`.

Something like:

```python
# A user who registered under the legacy scheme still has "_<key>" appended
# to their password hash; if they request a new link via the resend form they
# activate through this view, so strip the suffix to restore a valid hash.
# Only a trailing _<digits> can match, which no hasher we use can produce.
u.password = re.sub(r'_\d+$', '', u.password)
```

Also: `main` has since switched these redirects to `reverse('myesp_profile')` instead of a hardcoded `/myesp/profile/` — please match that when rebasing (line 194 here, and the same in `activate_account_legacy`).

---
_Generated by [Claude Code](https://claude.ai/code)_

esp/esp/users/views/registration.py
```diff
@@ -197,15 +230,28 @@ def resend_activation_view(request):
...
             return render_to_response('registration/resend.html', request,
                                       {'form':form, 'site': Site.objects.get_current()})
-        user=ESPUser.objects.get(username=form.cleaned_data['username'])
-        userkey=user.password[user.password.rfind("_")+1:]
-        send_activation_email(user, userkey)
+        username = form.cleaned_data['username']
+        try:
+            user = ESPUser.objects.get(username=username)
+        except ESPUser.DoesNotExist:
+            form.add_error('username', 'No account with this username is awaiting activation.')
```
willgearty: Case-sensitivity mismatch between the form and this lookup. `AwaitingActivationEmailForm.clean_username` validates with `username__iexact`, but line 235 does an exact `get(username=...)`, so a user who types `TestUser` instead of `testuser` passes validation and then gets told "No account with this username is awaiting activation."

Handling `DoesNotExist` at all is a real improvement over `main` (which would raise and 500 here), so this is just about closing the remaining gap:

```python
user = ESPUser.objects.get(username__iexact=username)
```

Also worth noting: this view discloses whether a given username exists and is awaiting activation. That's pre-existing behaviour, not something this PR introduces — flagging only so it isn't mistaken for new.

---
_Generated by [Claude Code](https://claude.ai/code)_

esp/esp/users/tests.py
```diff
@@ -475,17 +475,68 @@ def phase_2(self):
...
+
+        # Create inactive user
+        u = ESPUser.objects.create_user(username='testuser',
+                                       first_name='Test',
+                                       last_name='User',
+                                       email='testuser@example.com',
+                                       password='testpass123')
+        u.is_active = False
+        u.save()
+
```
willgearty: This blank line (and several others) contains trailing whitespace, which fails the repo's lint job. `esp/setup.cfg` selects `W29`, so `W293` is an error, and `deploy/lint` runs `flake8` over the whole tree in CI.

Running the repo's own flake8 config against this branch:

```
esp/esp/users/tests.py:517:1: W293 blank line contains whitespace
esp/esp/users/tests.py:521:1: W293 blank line contains whitespace
esp/esp/users/tests.py:525:1: W293 blank line contains whitespace
esp/esp/users/tests.py:529:1: W293 blank line contains whitespace
esp/esp/users/tests.py:532:1: W293 blank line contains whitespace
esp/esp/users/tests.py:536:1: W293 blank line contains whitespace
esp/esp/users/views/registration.py:240:1: W293 blank line contains whitespace
esp/esp/users/views/registration.py:245:1: W293 blank line contains whitespace
```

Stripping the whitespace from those 8 lines clears it. `deploy/lint --staged` will check just your changed files before you push.

---
_Generated by [Claude Code](https://claude.ai/code)_

willgearty (commented):
Thanks for picking this up — getting the activation token out of the `password` column is the right fix, and a `PasswordResetTokenGenerator` subclass keyed on `is_active` is a sound way to do it. The shape of the change is good. There are a few things to sort out before it can land, and the largest is that the branch has drifted well behind `main`.

## Blocking

**1. The branch no longer merges, and the rebase isn't mechanical.** GitHub reports `mergeable_state: dirty`. `main` has moved a lot since this was cut, in ways that touch exactly the code you're changing:

- **Django is now 5.2.16** (`esp/requirements.txt:2`), not 2.2. `PASSWORD_RESET_TIMEOUT_DAYS` was removed in Django 4.0, so `tokens.py`'s guidance is backwards for this tree — details inline.
- **`main` added an `ActivateAccountTest` class** (`esp/esp/users/tests.py:1315`) with 8 tests that reach the legacy flow via `reverse('activate_account')`. This PR moves that URL name to the new two-argument route, so `reverse()` raises `NoReverseMatch` in `setUp` and all 8 error out. Inline suggestion on `urls.py`.
- **`main` added AJAX live-validation endpoints** — `_username_live_validation` (`views/registration.py:69`) and `_email_live_validation` (`registration.py:134`) — which both still use the legacy `password__regex=r'\$(.*)_'` heuristic. They need the same update as the sites you did change, or the inline registration feedback will contradict the server.
- `main` now uses `reverse('myesp_profile')` rather than hardcoded `/myesp/profile/`.

**2. The new test never executes.** `test_activate_account_with_token` returns immediately because `require_email_validation` defaults to `False` and the test never sets it. Since it's the only coverage for the new endpoint, the security-critical path this PR adds is currently untested — which also means the "Existing tests pass / new tests added" checklist is passing on a test that does nothing. Inline detail.

**3. Lint fails.** 8 × `W293 blank line contains whitespace`; `esp/setup.cfg` selects `W29` and `deploy/lint` runs in CI. Verified locally against the repo's own config; full list inline.

**4. `UserRegForm.clean_username` was missed, which breaks re-registration.** That check deliberately excludes awaiting-activation accounts so a user who never activated can re-claim their username — `user_registration_validate` relies on it. New-scheme pending accounts no longer match the legacy regex, so they stop being excluded and the user hits "Username already in use." I verified the regex against both hashers in play (pbkdf2 in prod, MD5 in `test_settings.py`); truth table inline on `forms/user_reg.py`.

## Worth addressing

**5. `last_login__isnull=True` is too loose a stand-in for "awaiting activation."** It also matches accounts that were disabled on purpose — merge absorbees (`controllers/merge.py:111`), mass deactivation (`handlers/deactivationmodule.py:93`), and admin deactivation (`templates/users/userview.html:62`). Because `AwaitingActivationEmailForm` gates the *public* resend form, anyone who knows such a username can have an activation mail sent and re-enable the account. This is the same concern raised in the earlier review round; the `last_login` proxy narrows it but doesn't remove it.

My recommendation, and it would simplify 4 and 5 together: add an explicit `is_activation_pending` boolean to `ESPUser` with a migration backfilling from the legacy `_` suffix. Then "awaiting activation" is a stored fact instead of an inference over password formatting, all four call sites agree, deliberately-disabled accounts are excluded by construction, and `_make_hash_value` gets a precise flag to hash instead of `is_active`. It's more work than swapping predicates, but the inference-based approach is what generated this whole cluster of issues.

**6. Give the generator its own `key_salt`** — it currently shares the parent's, so activation and password-reset tokens are HMAC'd under the same salt. Inline.

**7. A deactivation re-opens an unexpired link,** since only `is_active` distinguishes used from unused. Bounded to 3 days by `PASSWORD_RESET_TIMEOUT`, and a dedicated pending flag would close it. Inline.

## Nits

- Case-sensitivity mismatch in `resend_activation_view` (inline).
- The `resetpassword/done` URL reorder is unrelated to this change and only adds conflict surface — worth dropping.
- The new `activate/<uid>/<token>/` route requires a trailing slash, unlike every other pattern in `urls.py` which uses `/?$`.

## Testing gaps beyond #2

Once the guard is removed, the flows I'd want covered are: a legacy pending user going through the resend form and activating via the new view, with the `_<key>` suffix correctly stripped; an expired token rejected; and a token rejected after the account is already active. The first is the migration path most likely to break silently in production, since it's the one case where old and new schemes meet.

Happy to look again once it's rebased — the core approach is right, and most of the above is a consequence of the branch age rather than the design.

---
_Generated by [Claude Code](https://claude.ai/code)_

willgearty:
One correction to my review, and a draft you can pull from.

**Correction:** I suggested adding an `is_activation_pending` boolean to `ESPUser`. That isn't possible as written — `ESPUser` is a proxy model (`esp/esp/users/models/__init__.py:1348`, `class Meta: proxy = True`), so it can't declare new fields. The idea still holds, it just needs somewhere else to live: a one-row-per-pending-account side table.

I've drafted the whole change on `claude/pr-4747-review-bpj32h`, rebased onto current `main`, if it's useful to compare against or cherry-pick from. What it does differently from this PR:

- **`PendingActivation`**, a `OneToOneField` side table, plus migration `0048` backfilling it from the legacy `_<key>` suffix. Presence of a row means "awaiting activation", and it's deleted on activation. No secret is stored — tokens stay stateless HMACs.
- **`ESPUser.awaiting_activation_Q()`** as the single predicate, used by both `user_reg` forms, the phase-1 duplicate check, and the two AJAX validation endpoints — so all five sites agree, including `UserRegForm.clean_username`, which is what keeps username re-registration working.
- **The token hashes the pending row, not `is_active`.** `is_active` returns to `False` on every admin deactivation, so hashing it lets a pre-deactivation link switch the account back on; the row only ever goes set → cleared, so a spent token stays spent. It also hashes the password, like Django's own reset tokens, so re-registering over a pending account supersedes earlier links.
- **`key_salt`** set on the generator.
- The `_<key>` strip is kept in the new view, for the case where a legacy pending account uses the resend form and activates through an HMAC link.
- Pending rows are also cleared on the other two paths that activate an account (password reset, admin activation from userview), so a stale row can't make an activated account look pending again.
- The legacy URL keeps working under the name `activate_account_legacy`; the existing `ActivateAccountTest` was repointed at it.

**Testing:** 26 activation/awaiting-activation tests, plus the updated `phase_1`/`phase_2`. Full suite: **2211 passed, 15 failed**, and those same 15 fail identically on unmodified `main` in the same environment (LaTeX, npm/bootswatch, and a dependency-version check that's environmental here), so they're not from the change. `makemigrations --check --dry-run` is clean and `deploy/lint` passes.

The security fix in this PR is the right one — this is just the surrounding call sites and the "awaiting activation" question, which turned out to be the harder half.

---
_Generated by [Claude Code](https://claude.ai/code)_

willgearty (approved):
I cleaned this up with Claude to make a separate column for the pending activation status. I also removed all of the legacy machinery since we won't need to support legacy links after the switchover. (Claude also added a bunch of tests)

PR #4755, opened 2026-03-06
Refactor fillslot grade filtering to remove redundancy and fix user_g…

### LINKED ISSUE #4736: [Bug]: fillslot() applies grade-range filter twice in studentclassregmodule.py
### Description

While reviewing the fillslot() method in studentclassregmodule.py, I noticed that the class list is filtered by grade range twice in succession.
The first filter already restricts classes based on the student's grade, Immediately afterward, a second filter applies the exact same condition again when user_grade != 0.
Because the list has already been filtered once, the second filtering step does not change the result and effectively becomes redundant.
Additionally, the surrounding logic suggests that grade filtering was intended to be skipped when user_grade == 0 (for users whose grade is not recorded). However, since the first filter is always applied, users with user_grade == 0 may end up seeing no classes.

### Steps to Reproduce

1. Open the file `esp/program/modules/handlers/studentclassregmodule.py`.

2. Locate the `fillslot()` method.

3. Find the section where classes are filtered by grade (around lines 488–491).

4. Observe that the class list is filtered by the same grade-range condition twice:
   - First unconditionally.
   - Then again inside `if user_grade != 0`.

5. Notice that when `user_grade == 0`, the first filter still runs, which may remove most or all classes even though the conditional suggests grade filtering should be skipped.

### Expected Behavior

If a user has a valid grade (user_grade != 0), the class list should be filtered once based on the grade range.

If the user's grade is unknown (user_grade == 0), the grade filter should be skipped so that the user can still see available classes.

### Actual Behavior

The grade filter is applied unconditionally before the conditional check, which means:

The second filter becomes redundant when user_grade != 0.

When user_grade == 0, the first filter still runs and may remove most or all classes because 0 usually does not fall within normal class grade ranges.

This can result in users with an unset grade seeing no classes.

### Screenshots

_No response_

### Operating System

_No response_

### Browser

_No response_

### Additional Context

If I’ve misunderstood the intended behavior here, please feel free to correct me. I’d be happy to adjust the approach or explore an alternative fix if there’s a better way to handle this.

Anas-Blue:
## Description

I refactored it to fetch the catalog once up front, then apply the grade and closed-registration filters inside a single conditional that correctly skips them for onsite users. The grade filter now only runs when user_grade != 0, so users with no recorded grade are no longer shown an empty class list.

This also eliminates the duplicated catalog() call that existed in both branches of the original if/else.

## Related Issue

Closes #4736

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

## AI Disclosure

Used AI assistance to validate the refactored logic.

## Checklist

- [x] Self-reviewed the code

### DIFF
--- a/esp/esp/program/modules/handlers/studentclassregmodule.py
+++ b/esp/esp/program/modules/handlers/studentclassregmodule.py
@@ -499,12 +499,11 @@ class StudentClassRegModule(ProgramModuleObj):
 
         #   Override both grade limits and size limits during onsite registration
         #   Classes are sorted like the catalog
-        if is_onsite and not 'filter' in request.GET:
-            classes = list(ClassSubject.objects.catalog(self.program, ts))
-        else:
-            classes = [c for c in list(ClassSubject.objects.catalog(self.program, ts)) if c.grade_min <= user_grade and c.grade_max >= user_grade]
+        classes = list(ClassSubject.objects.catalog(self.program, ts))
+        should_filter_invalid = (not is_onsite) or ('filter' in request.GET)
+        if should_filter_invalid:
             if user_grade != 0:
-                classes = [c for c in classes if c.grade_min <=user_grade and c.grade_max >= user_grade]
+                classes = [c for c in classes if c.grade_min <= user_grade and c.grade_max >= user_grade]
             classes = [c for c in classes if not c.isRegClosed()]
 
         categories_sort = self.sort_categories(classes, self.program)


esp/esp/program/modules/handlers/studentclassregmodule.py
```diff
@@ -499,12 +499,11 @@ def fillslot(self, request, tl, one, two, module, extra, prog):
...
-            classes = list(ClassSubject.objects.catalog(self.program, ts))
-        else:
-            classes = [c for c in list(ClassSubject.objects.catalog(self.program, ts)) if c.grade_min <= user_grade and c.grade_max >= user_grade]
+        classes = list(ClassSubject.objects.catalog(self.program, ts))
+        should_filter_invalid = (not is_onsite) or ('filter' in request.GET)
+        if should_filter_invalid:
             if user_grade != 0:
-                classes = [c for c in classes if c.grade_min <=user_grade and c.grade_max >= user_grade]
+                classes = [c for c in classes if c.grade_min <= user_grade and c.grade_max >= user_grade]
             classes = [c for c in classes if not c.isRegClosed()]
```
willgearty, line 507: Thanks for removing the duplicate `catalog()` call and the redundant grade filter.

One thing though: the original code also had this for non-onsite users:

```python
classes = [c for c in classes if not c.isRegClosed()]
```

That got dropped in the refactor, which means non-onsite students will now see classes with closed registration. Your description mentions keeping the closed-registration filter, so I'm guessing this was accidental?

Should just need to add it back after the grade filter:

```python
if should_filter_invalid:
    if user_grade != 0:
        classes = [c for c in classes if c.grade_min <= user_grade and c.grade_max >= user_grade]
    classes = [c for c in classes if not c.isRegClosed()]
```

willgearty (changes requested):

esp/esp/program/modules/handlers/studentclassregmodule.py
Anas-Blue: Good catch, sorry about dropping that! I've updated the PR to restore the isRegClosed() filter inside the should_filter_invalid block so the original logic for non-onsite users is intact.

willgearty (approved):
Looks good, thanks!

PR #4768, opened 2026-03-06
fix(lottery): cast priority index to int in get_computed_schedule

### LINKED ISSUE #4759: [Bug]: TypeError in get_computed_schedule — String used as list index (priority_N mode)
### Description

The method get_computed_schedule(student_id, mode) supports a mode argument in the form "priority_N" (e.g., "priority_2") to fetch a student's schedule based on a specific priority level. It extracts N via a regex, but then uses the result as a string to index into self.priority (which is a list). In Python 3, list indices must be integers — passing a string causes a TypeError at runtime.

### Steps to Reproduce

Initialize a LotteryAssignmentController for a program with multiple priority levels.
Call get_computed_schedule(student_id, mode='priority_2').
Observe the crash.

### Expected Behavior

The method should parse the priority level from the mode string, convert it to an integer, and use it to index into self.priority successfully — returning the student's computed schedule assignments for that priority level without error.
`# lottery.py, inside get_computed_schedule()
p = re.search('(?<=priority_)\d*', mode).group(0)   # p is a STRING e.g. "2"
if p:
    assignments = numpy.nonzero(self.priority[p][...])  # ❌ TypeError: list indices must be integers, not str`

### Actual Behavior

When get_computed_schedule(student_id, mode='priority_2') is called, the regex correctly extracts "2" from the mode string, but it is never cast to an integer. Indexing self.priority["2"] raises a TypeError immediately, crashing the method before any schedule data is returned:
`
TypeError: list indices must be integers or slices, not str
`

prem-2006:
Fixes #4759

## Description

Fixes a TypeError in get_computed_schedule() where the priority level extracted from the mode string ("priority_N") was used as a string when indexing self.priority, which is a list. Python requires list indices to be integers, causing a runtime crash.

##Problem

The method supports a mode parameter in the form:

`priority_N`

For example:

```
priority_1
priority_2
priority_3
```

The priority value N is extracted using a regex:

`p = re.search('(?<=priority_)\d*', mode).group(0)`

However, p is returned as a string (e.g., "2").
Later it is used to index into self.priority, which is a list:

`self.priority[p]`

Since list indices must be integers, this raises:

`TypeError: list indices must be integers or slices, not str`

This causes get_computed_schedule() to fail whenever a "priority_N" mode is used.

##Solution

Convert the extracted priority value to an integer before indexing the list.

Before

`assignments = numpy.nonzero(self.priority[p][self.student_indices[student_id],:])[0]`

After

`assignments = numpy.nonzero(self.priority[int(p)][self.student_indices[student_id],:])[0]`

This ensures the correct integer index is used.

### DIFF
--- a/esp/esp/program/controllers/lottery.py
+++ b/esp/esp/program/controllers/lottery.py
@@ -690,7 +690,7 @@ class LotteryAssignmentController(object):
             assignments = numpy.nonzero(self.priority[1][self.student_indices[student_id],:])[0]
         else:
             import re
-            p = re.search('(?<=priority_)\d*', mode).group(0)
+            p = re.search(r'(?<=priority_)\d*', mode).group(0)
             if p:
                 assignments = numpy.nonzero(self.priority[int(p)][self.student_indices[student_id],:])[0]
         result = []


esp/esp/program/controllers/lottery.py
```diff
@@ -692,7 +692,7 @@ def get_computed_schedule(self, student_id, mode='assigned'):
             import re
             p = re.search('(?<=priority_)\d*', mode).group(0)
```
hwatheod: Can you also fix this line by making it a raw string?  This will remove this warning from Python:

`SyntaxWarning: "\d" is an invalid escape sequence. Such sequences will not work in the future. Did you mean "\\d"? A raw string is also an option.`

prem-2006: thanks! I'll update the regex to use a raw string

hwatheod (approved):
Thank you for doing this!

PR #4770, opened 2026-03-06
Add decorator-order auth test and fix module view violations

### LINKED ISSUE #2003: Test for ordering of program module decorators
Similar to the test from #1954, it may be useful to have a test that asserts that the decorators on module views are in the right order.  I think the ordering should be apprxoimately the following (from outside to inside):
* `@aux_call`/`@main_call`
* `@needs_<group>`
* `@meets_grade`
* `@meets_deadline[s]`
* `@meets_cap` (after #1830 is merged)
This will ensure that users always see the right error message.  (All of these are optional unless otherwise specified, although we could also assert, say, that anything with `@needs_student` also has `@meets_grade` (except the profile).)

GxAditya:
## Summary

This PR adds a decorator-order regression test for program module views and fixes current violations in existing handlers.

## What changed

- Extended `esp/esp/program/modules/tests/auth.py` with `testViewDecoratorOrder`.
- The test parses module handler files via AST and enforces this order:
  - `@aux_call/@main_call`
  - `@needs_<group>`
  - `@meets_grade`
  - `@meets_deadline` / `@meets_any_deadline`
  - `@meets_cap`
- Reordered decorators in:
  - `esp/esp/program/modules/handlers/studentclassregmodule.py`
    - `catalog_json`
    - `catalog`
    - `catalog_pdf`
  - `esp/esp/program/modules/handlers/studentlunchselection.py`
    - `select_lunch`

## Related Issue

Closes #2003 

# Screenshots

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

## AI Disclosure

- Used GPT 5.3-Codex xHigh for Codebase Exploration and Test Verification .

## Checklist

- [x] Self-reviewed the code

### DIFF
--- a/esp/esp/program/modules/handlers/jsondatamodule.py
+++ b/esp/esp/program/modules/handlers/jsondatamodule.py
@@ -617,9 +617,9 @@ class JSONDataModule(ProgramModuleObj, CoreModule):
         return {'sections': sections}
 
     @aux_call
+    @no_auth
     @cache_control(public=True, max_age=300)
     @json_response()
-    @no_auth
     @cached_module_view
     def class_info(self, request, tl, one, two, module, extra, prog):
         return_key = None
@@ -679,9 +679,9 @@ class JSONDataModule(ProgramModuleObj, CoreModule):
     class_info.cached_function.depend_on_model(ClassSection)
 
     @aux_call
+    @no_auth
     @cache_control(public=True, max_age=300)
     @json_response()
-    @no_auth
     def class_size_info(self, request, tl, one, two, module, extra, prog):
         return_key = None
         if 'return_key' in request.GET:
--- a/esp/esp/program/modules/tests/auth.py
+++ b/esp/esp/program/modules/tests/auth.py
@@ -42,9 +42,11 @@ class ProgramModuleAuthTest(ProgramFrameworkTest):
     """Validate that all program modules have some property."""
 
     CALL_DECORATORS = {'main_call', 'aux_call'}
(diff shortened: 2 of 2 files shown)


pushkarkumar06 (changes requested):
hii @GxAditya  , Thanks for working on this — the idea of enforcing decorator order with a regression test is really helpful.

While reviewing the changes I noticed a couple of things that might need adjustment:

* The AST test doesn’t currently enforce the full decorator order described in the PR (for example `@no_auth` and `@cache_control` aren’t explicitly checked), so some ordering issues could still slip through.
* Since caching and auth ordering can affect security behavior, it might be safer for the test to explicitly verify the relative order of `@no_auth` and `@cache_control`.

Also, it may help to add a short comment in the test explaining the intended decorator order and any limitations of the AST approach.

Once those points are addressed this looks good overall.

GxAditya:
@pushkarkumar06  I've made the changes you asked . 

- In auth.py, the AST test now treats @no_auth as part of the auth layer and explicitly checks that @no_auth stays outside @cache_control when both are present. 

- I also added a short note in the test docstring explaining the intended order and the AST limitation: it only sees decorators declared on the method itself, not wrappers applied later at runtime.

- To keep the tightened rule passing, I reordered two existing JSON module views in jsondatamodule.py  so @no_auth comes before @cache_control.

PR #4773, opened 2026-03-06
Optimize TemplateOverride.next_version() by replacing two queries with aggregate(Max)

### LINKED ISSUE #4757: [Feature]:  Optimize TemplateOverride.next_version to use database aggregation
### Problem Statement

While looking through esp/utils/models.py, I noticed that next_version() on the TemplateOverride model currently makes two separate database queries every time a new template override is saved:

```
def next_version(self):
    qs = TemplateOverride.objects.filter(name=self.name)
    if qs.exists():                                                          # Query 1 — COUNT(*)
        return qs.order_by('-version').values_list('version', flat=True)[0] + 1  # Query 2 — SELECT + ORDER BY
    else:
        return 1
```
The first query checks whether any rows exist, and the second sorts them in descending order to grab the highest version number. Every time something edits a template override, this runs and as the number of overrides for any given template grows, the `ORDER BY` becomes increasingly wasteful when all we actually need is a single integer: the max version.

### Proposed Solution

Replace the two-query approach with a single `aggregate(Max(...))` call, which lets the database compute the maximum version directly.

This is cleaner for a few reasons:
-One query instead of two.
-No `ORDER BY` — `MAX()` is a simple aggregation that databases handle very efficiently.
-The empty case (`None or 0`) is handled naturally without a separate .exists() check.

### Alternatives Considered

_No response_

### Additional Context

_No response_

Anas-Blue:
## Description

This PR optimizes the `TemplateOverride.next_version()` method in `esp/utils/models.py`. The previous implementation performed two database queries:
1) `.exists()` to check if any rows exist
2) `.order_by('-version')` to retrieve the highest version

This change replaces that logic with a single `aggregate(Max('version'))` query, allowing the database to compute the maximum version directly.

Benefits:

-Reduces database queries from two to one
-Removes unnecessary `ORDER BY`
-Keeps the original behavior when no template overrides exist

## Related Issue

Closes #4757 

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

## AI Disclosure
--

## Checklist

- [x] Self-reviewed the code


Anas-Blue:
Hi maintainer,
Should I need to countinously update the branch for PR, or leave it as it is?

kkbrum:
You can leave it as-is until a maintainer reviews it and asks for merge conflict resolutions.

Anas-Blue:
ok

willgearty (approved):
Sorry for the delay. This looks good, thanks!

PR #4781, opened 2026-03-06
Perf: Move RecordType query outside loop in checkout module

### LINKED ISSUE #4778: [Feature]:  Move RecordType query outside loop in checkout module for better performance
### Problem Statement

In onsitecheckoutmodule.py (line 66), the "Check Out All Students" feature runs 
RecordType.objects.get(name="checked_out") inside a for-loop. This means if you're 
checking out 200 students at end of day, it hits the database 200 times to fetch the 
exact same thing. That's 199 wasted queries.

### Proposed Solution

Just move that one line above the for-loop so it runs once and reuse the result. 
Nothing else changes.

### Alternatives Considered

Could also use bulk_create() to insert all records in one go, but that's a bigger 
change and might break any post-save hooks. Moving the query out of the loop is 
the simplest and safest fix.

### Additional Context

_No response_

Rakesh-18112006:
## Description

In the "Check Out All Students" feature, `RecordType.objects.get(name="checked_out")` was being called inside the for-loop. Since the RecordType doesn't change between students, this results in the same query hitting the database once per student. For a Splash event with 200 students, thats 199 wasted queries.
Moved the query above the loop so it runs once and the result is reused.

## Related Issue

Closes #4778 

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [X] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [X] Existing tests pass
- [ ] Manually verified

## Checklist

- [X] Self-reviewed the code

### DIFF
--- a/esp/esp/program/modules/handlers/onsitecheckoutmodule.py
+++ b/esp/esp/program/modules/handlers/onsitecheckoutmodule.py
@@ -62,8 +62,8 @@ class OnSiteCheckoutModule(ProgramModuleObj):
 
         if "checkoutall" in request.POST and "confirm" in request.POST:
             students = prog.currentlyCheckedInStudents()
+            rt = RecordType.objects.get(name="checked_out")
             for student in students:
-                rt = RecordType.objects.get(name="checked_out")
                 Record.objects.create(user=student, event=rt, program=prog)
             context['checkout_all_message'] = "Successfully checked out %s students" % (students.count())
 


Rakesh-18112006:
Hi @willgearty, just checking if any changes are needed from my side. Happy to update.

PR #4783, opened 2026-03-06
Security: replace pickle.loads() RCE with JSON serialization

### LINKED ISSUE #4748: Security: Replace pickle.loads() with JSON serialization to eliminate RCE vector
### Description

Two subsystems in the codebase deserialize data from the database using Python's
`pickle.loads()`. This is a known Remote Code Execution (RCE) vector — if an
attacker can write arbitrary bytes to those database fields (via SQL injection,
a compromised admin account, or a malicious database restore), they achieve full
arbitrary code execution on the server.

Python's own documentation warns: *"The pickle module is not secure. Only unpickle
data you trust."* Database rows are not trusted data.

---

### Steps to Reproduce

1. Gain write access to the database (via SQLi, compromised admin account,
   or access to a DB backup/dump)
2. Craft a malicious pickle payload in Python:
      import pickle, os
      class Exploit(object):
          def __reduce__(self):
              return (os.system, ('id > /tmp/pwned',))
      payload = pickle.dumps(Exploit())
3. Write the payload bytes into the `pickled_provider` column of the
   `dbmail_messagevars` table, OR into the `q_filter` column of the
   `users_persistentqueryfilter` table
4. Trigger email sending (any outgoing ESP email) OR load any user list
   that uses a saved query filter
5. Observe the command executed on the server

### Expected Behavior

Stored database data should be deserialized using a safe, structured format
(such as JSON) that cannot execute arbitrary code. Reading a tampered database
field should raise a validation error, not execute OS commands.

### Actual Behavior

pickle.loads() is called directly on bytes read from the database with no
integrity check, no class whitelist, and no sandboxing. Any crafted pickle
payload stored in the `pickled_provider` or `q_filter` columns executes
arbitrary Python code on the server at the privilege level of the web
application process.

SHREESHANTH99:
## Description

`MessageVars.pickled_provider` and `PersistentQueryFilter.q_filter` stored arbitrary Python objects as pickle blobs in the database. `pickle.loads()` executes arbitrary code on deserialization — any attacker able to write to these fields could achieve Remote Code Execution. This PR replaces both with safe JSON serialization, adds an explicit provider allowlist, and backfills existing rows via data migrations.

- Add `providers.py` with safe lazy-import provider registry
- `MessageVars`: add `provider_info` JSONField, deprecate `pickled_provider`
- `PersistentQueryFilter`: add `q_filter_json` JSONField, deprecate `q_filter`
- Pickle fallback now hard-rejects classes not in the allowlist
- `connector` in Q objects validated against `{'AND', 'OR', 'XOR'}` allowlist
- On-the-fly migration: pickle rows auto-upgraded to JSON on first read
- Schema migrations: `users.0042`, `dbmail.0012`
- Data backfill migrations: `users.0043`, `dbmail.0013`

## Related Issue
Closes #4748 

## Type of Change

- [x] Bug fix
- [x] Breaking change
- [x] Refactor / cleanup

## Testing

 Existing tests pass (76/76)
 New tests added (if applicable)
 Manually verified (migrations applied and confirmed via showmigrations)
 
## AI Disclosure
Used Chatgpt as an assistant during development.
All code was personally reviewed, tested, and verified before submission.


SHREESHANTH99:
i have address capilot review, Could you please take another look when you have a chance? Thank you.

Oval17 (approved):
LGTM.

willgearty:
Closing this for now because I don't think it's necessary. We can revisit #4748 at some other point if it becomes clear it's a real problem.

PR #4791, opened 2026-03-06
Respect Student/Catalog deadline for catalog pages

### LINKED ISSUE #2402: Despite deadline, class catalog cannot be closed to students (?)
I'm not entirely sure if I'm missing something since it feels like we've done this recently, but we haven't been able to close the catalog to students. Closing `/Student/Catalog` does nothing. Looking at the code, there is a `catalog_student` with a permission decorator attached, but it is not an `aux_call`. It appears that it was meant to be called by the `catalog` method if the current user is a student, but this commit removes that logic:
https://github.com/learning-unlimited/ESP-Website/commit/9bfd5c2e3980e3b205f9d0e53cc487d026c16de2#diff-cd310f15494269c1d6790b375cf2001bR681

Dhirenderchoudhary:
Date : 6 March 2026

## Description

The catalog endpoint (/Student/Catalog) was ignoring deadline settings because it uses [@no_auth](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) for caching. This checks the non-user-specific Permission so we can still cache while respecting when admins close the catalog.

## Related Issue

Closes #2402

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

## Checklist

- [x] Self-reviewed the code


pushkarkumar06 (changes requested):
hii @Dhirenderchoudhary  , Thanks for working on this — the overall direction looks good and it’s nice to see the catalog endpoint respecting the Student/Catalog deadline.

While testing the change, I noticed a couple of things that might need attention before merging:

* The permission check currently performs two DB lookups (`exists()` + `null_user_has_perm()`). Since this appears in multiple places, it might be worth extracting into a helper to avoid duplicate queries.
* Because this endpoint is publicly cached, there is a small risk of stale responses around deadline transitions (especially for `catalog_json`). It would be good to confirm that cache behavior won’t serve outdated open/closed responses.
* I also couldn’t find regression tests covering the JSON path when the catalog is closed (expected 403 response). Adding a small test for both open and closed states would help prevent regressions.

Once those are addressed, the fix should be much safer to merge.

Dhirenderchoudhary:
@pushkarkumar06 Nice catch of regression tests covering the JSON added.
thanks

esp/esp/program/modules/handlers/studentclassregmodule.py
```diff
@@ -609,15 +566,22 @@ def timeslots_json(self, request, tl, one, two, module, extra, prog, timeslot=No
 
         return resp"""
 
-    @cache_control(public=True, max_age=3600)
     @no_auth
     @aux_call
     def catalog_json(self, request, tl, one, two, module, extra, prog, timeslot=None):
         """ Return the program class catalog """
+        # Check Student/Catalog deadline if one exists.
+        # are not served to clients.
```
Oval17: Comment is truncated here, either remove this or complete this.

esp/esp/program/modules/handlers/studentclassregmodule.py
```diff
@@ -651,13 +615,19 @@ def catalog_registered_classes_json(self, request, tl, one, two, module, extra,
...
     def catalog(self, request, tl, one, two, module, extra, prog, timeslot=None):
+        if self._catalog_deadline_closed(prog):
+            return render_deadline_for_tl('learn', request,
+                    {'extension': 'the deadline Student/Catalog was', 'moduleObj': self})
         return self.catalog_render(request, tl, one, two, module, extra, prog, timeslot)
 
     @disable_csrf_cookie_update
     @aux_call
     @no_auth
     @cache_control(public=True, max_age=120)
```
Oval17: for this line and line 616 you can see `render_deadline_for_tl` returns a standard HTTP 200 response right?
 Both catalog and catalog_pdf still have @cache_control(public=True, max_age=120) wrapping the entire view, so the "Student Registration Closed" deadline page will be publicly cached for 2 minutes. After an admin re-opens the catalog, users would still see the closed page from cache. catalog_json already handles this correctly by removing @cache_control and setting Cache-Control manually per response path. 

I think the same treatment should be applied here: remove the @cache_control decorator from catalog and catalog_pdf, and set resp['Cache-Control'] = 'public, max-age=120' only on the successful (open) return path

or you can think of roundabout for this if possible?

Oval17 (commented):
one thing I observed in test 
No test covers catalog_pdf when closed. Three endpoints were changed (catalog, catalog_json, catalog_pdf) but only two are tested.

Dhirenderchoudhary:
Added coverage for the closed Student/Catalog deadline path on catalog_pdf, and also tightened cache-header assertions on closed responses. See studentreg tests around test_catalog_pdf_closed and test_catalog_deadline.

esp/esp/program/modules/handlers/studentclassregmodule.py
Dhirenderchoudhary: Applied the same treatment as discussed: removed decorator-based cache behavior from catalog and catalog_pdf, and now set Cache-Control manually only on successful open responses (public, max-age=120). Closed deadline responses for both endpoints now return Cache-Control: no-store so stale closed pages are not cached.

Oval17 (commented):
LGTM
1- address the failing test case

willgearty (approved):
The Student/Catalog was removed in #2976, so that needed to be restored to make this actually work. Should be good to go now.

PR #4820, opened 2026-03-07
Feature: admin activity & audit log

### LINKED ISSUE #4786: [Feature]: Admin Activity / Audit Log System
### Problem Statement

There is currently no audit trail for sensitive admin actions in the system. 
Administrators can finalize financial aid grants, approve/reject grade change 
requests, and bulk-expire or renew user permissions — but none of these actions 
are recorded with who did them, when, or what changed. If a dispute arises or 
an account is compromised, there is no way to trace what was modified.

### Proposed Solution

Add a lightweight audit log system that records sensitive admin actions:

- A new `AuditLogEntry` model with fields: `actor` (user), `action`, 
  `model_name`, `object_pk`, `changes` (JSON before/after), `timestamp`
- An `AuditedModelAdmin` mixin that hooks into `save_model`, `delete_model`, 
  and bulk actions to automatically create log entries
- Apply the mixin to: `FinancialAidGrantAdmin`, `GradeChangeRequestAdmin`, 
  and `PermissionAdmin`
- A read-only admin view to browse logs, filterable by user, model, date range

### Alternatives Considered

- Using a third-party package like `django-simple-history` — avoided to keep 
  dependencies minimal and stay consistent with the existing codebase patterns
- Logging to files only — rejected because it's not queryable or visible to 
  admins without server access

### Additional Context

_No response_

SHREESHANTH99:
## Description

Adds an immutable admin activity audit log. Every create, update, delete, and bulk action through the Django admin is recorded with the actor, their IP, the affected object, and a per-field JSON diff.

- `AuditLogEntry` model in `esp.utils` (ContentType FK, object_id, JSONField diffs, actor_ip)
- `AuditedModelAdmin` mixin with `IGNORED_FIELDS` and pre/post save diffing
- Mixin applied to `FinancialAidGrantAdmin`, `PermissionAdmin`, `GradeChangeRequestAdmin`
- Read-only `AuditLogEntryAdmin` (no add/change/delete)
- Migrations: `utils.0002_add_auditlogentry`, `utils.0003_auditlogentry_add_actor_ip`

## Related Issue

Closes #4786 

## Testing

- [x] Existing tests pass (131 tests — esp.users, esp.accounting, esp.utils)
- [x] Manually verified

## AI Disclosure
Used Chatgpt as an assistant during development.
All code was personally reviewed, tested, and verified before submission.


SHREESHANTH99:
i have address capilot review, Could you please take another look when you have a chance? Thank you.

kkbrum:
Great! No need to comment though; we can see the commits addressing the review. We will review when we are able, which may be quite a long time due to the long queue.

esp/esp/utils/migrations/0003_auditlogentry_add_actor_ip.py
kkbrum, line 1: I'm not that familiar with the best practice for migrations, but couldn't this be included in the 0002 migration you've added? Having a ton of migrations is annoying, and I don't think you need two separate ones here?

esp/esp/users/models/__init__.py
```diff
@@ -2916,6 +2933,7 @@ def get_admin_url(self):
     def __str__(self):
         return  "%s requests a grade change to %s" % (self.requesting_student, self.claimed_grade) + (" (Approved)" if self.approved else "")
 
+
```
kkbrum, line 2913: In the future, please don't push whitespace changes like this.

esp/esp/utils/models.py
```diff
@@ -79,3 +81,63 @@ class PrintRequest(models.Model):
     user = AjaxForeignKey(ESPUser, on_delete=models.CASCADE)
     time_requested = models.DateTimeField(auto_now_add=True)
     time_executed = models.DateTimeField(blank=True, null=True)
+
+
+class AuditLogEntry(models.Model):
+    """
+    Immutable record of a sensitive admin action.
+    Written by AuditedModelAdmin whenever an object is saved or deleted
```
kkbrum: it seems like this isn't just for any object, but just for financial aid, permissions, and grade changes. Is that correct? This documentation makes it sound like it is for any object.

esp/esp/users/admin.py
```diff
@@ -168,9 +169,29 @@ def save_model(self, request, obj, form, change):
...
 admin_site.register(GradeChangeRequest, GradeChangeRequestAdmin)
 
 #   Include admin pages for Django group
 admin_site.register(Group, GroupAdmin)
 
+
+class AuditLogEntryAdmin(admin.ModelAdmin):
+    """Read-only browse view for the admin audit log."""
+    list_display  = ['timestamp', 'actor', 'actor_ip', 'action', 'content_type', 'object_id', 'object_repr', 'extra']
+    list_filter   = ['action', 'content_type', 'timestamp']
```
kkbrum: The filter shows all possible content_types, but it seems like this feature is only for ~3 content types. Can you not show any content_types here that would never be recorded? It is confusing to see in the filter, for example, "resource", when it does not actually log changes to resources.

esp/esp/accounting/admin.py
```diff
@@ -73,7 +74,7 @@ class AccountAdmin(admin.ModelAdmin):
 def finalize_finaid_grants(modeladmin, request, queryset):
     for grant in queryset:
         grant.finalize()
-class FinancialAidGrantAdmin(admin.ModelAdmin):
+class FinancialAidGrantAdmin(AuditedModelAdmin, admin.ModelAdmin):
```
kkbrum, line 77: It seems like this only records it when I finalize grants through the admin pages. However, that is not the primary way of approving financial aid. Rather, admins are supposed to go to http://localhost:8000/manage/Splash/2026_Summer/finaidapprove/. I added financial aid there and it was not recorded. If we want financial aid actions recorded, it should be recorded through that page as well.

esp/esp/users/admin.py
```diff
@@ -85,7 +86,7 @@ def queryset(self, request, queryset):
         elif self.value() == 'expired':
             return queryset.filter(end_date__lte=datetime.datetime.now())
 
-class PermissionAdmin(admin.ModelAdmin):
+class PermissionAdmin(AuditedModelAdmin, admin.ModelAdmin):
```
kkbrum, line 95: permissions and deadlines are set by admins at http://localhost:8000/manage/Splash/2026_Summer/deadlines/. An action on this page was not recorded. If we want to record permission changes, then this page needs to be tracked as well, not just permissions set through the admin pages.

esp/esp/users/admin.py
```diff
@@ -158,7 +164,7 @@ def contact_name(self, obj):
 
 admin_site.register(K12School, K12SchoolAdmin)
 
-class GradeChangeRequestAdmin(admin.ModelAdmin):
+class GradeChangeRequestAdmin(AuditedModelAdmin, admin.ModelAdmin):
```
kkbrum, line 167: Grade changes are approved by admins directly on the http://localhost:8000/manage/userview?username=Student page. If we want to record grade changes, grade changes accepted or rejected directly through userview will also need to be tracked, not just grade changes through the admin pages.

kkbrum (changes requested):

kkbrum:
Please fix the failing tests and merge conflicts.

kkbrum (changes requested):
Thanks for these changes. 

For permissions, it now logs when a permission is created via the deadlines management page, but it does not log when that permission is ended or deleted via that page. Those should also get logged. 

Also, in addition to logging permissions via that page, I think it would be good if deadline changes were also logged. Deadlines are very pesky and important, so that's something where additional accountability could be useful.

For financial aid, when I deleted a grant as a bulk action via the admin pages, the deletion got logged twice (maybe because you have to confirm the bulk action?). Same thing when I deleted a permission as a bulk action via the admin pages.

kkbrum:
Closing due to inactivity.

SHREESHANTH99:
@kkbrum i was busy with my university exams so i was unable to work on this pr .Now that my exams are over i would like to get this pr merged by working on requested changes

PR #4827, opened 2026-03-07
Fix login empty fields 4818

### LINKED ISSUE #4818: [Bug]: Login form submits even when username or password is empty
### Description

While testing the ESP website login page, I noticed that the login form can be submitted even when the username or password field is empty.

This can cause unnecessary requests to the server and may confuse users because the page simply reloads without clearly indicating what went wrong.

Adding simple client-side validation before submitting the form would improve the user experience.

### Steps to Reproduce

1. Open the login page of an ESP site.
2. Leave the username or password field empty.
3. Click the "Login" button.
4. The form submits even though required fields are missing.

### Expected Behavior

The login form should prevent submission if username or password is empty and display a clear message such as:

"Username and password are required."

### Actual Behavior

The login form submits even when fields are empty, and the page reloads without clear feedback to the user.

### Screenshots

_No response_

### Operating System

Windows

### Browser

Google Chrome 

### Additional Context

This issue affects the user experience on the login page because users may not immediately understand why the login attempt failed when fields are empty.

Adding simple client-side validation (for example using JavaScript or HTML required attributes) could prevent unnecessary form submissions and provide immediate feedback to users.

A possible improvement would be displaying an inline error message such as:

"Please enter both username and password before logging in."

This change would improve usability and reduce unnecessary requests to the backend.

Monu26052006:
Fix #4818: prevent login submission when username or password fields are empty

This PR fixes the login form submission issue where users could submit the form with empty username or password fields.

The fix adds the `required` attribute to the username and password input fields in the login form.

Closes #4818


willgearty:
@Monu26052006 Please address the Copilot review. Especially make sure to remove the new file that is unrelated.

PR #4833, opened 2026-03-07
Fix case-insensitive username lookup in resend_activation_view

### LINKED ISSUE #4798: [Bug]: resend_activation_view  crashes with DoesNotExist when username case differs from registration
### Description

The Resend Activation Email flow crashes with an unhandled `ESPUser.DoesNotExist` exception when a user enters their username with different casing than the one used during registration.

The form validation checks usernames case-insensitively, but the subsequent lookup in the view performs a case-sensitive query. Because the form returns the username exactly as typed by the user (e.g., `JOHNDOE`) while the database may store it as `johndoe`, the lookup fails and raises `ESPUser.DoesNotExist`, resulting in a 500 server error.

### Steps to Reproduce

1) Register an account with a lowercase username, e.g. johndoe, with email validation required (require_email_validation tag enabled).
2) Do not activate the account.
3) Navigate to /myesp/resend_activation/.
4) Submit the form with the username typed in a different case, e.g. JohnDoe or JOHNDOE.
5) Observe an unhandled server error (500).

### Expected Behavior

The username should be treated case-insensitively, and the activation email should be resent successfully.

### Actual Behavior

The view performs a case-sensitive lookup, causing ESPUser.DoesNotExist and returning a 500 Internal Server Error.

### Screenshots

_No response_

### Operating System

_No response_

### Browser

_No response_

### Additional Context

A safer alternative approach is to avoid using `get()` and instead perform a filtered lookup that handles missing results gracefully.

### Alternative Fix

Use `filter()` with a case-insensitive lookup and retrieve the first matching user:

```python
user = ESPUser.objects.filter(
    username__iexact=form.cleaned_data['username'],
    awaiting_activation=True
).first()

if user is None:
    raise ESPError("User not found.", log=False)
```

### Rationale

Unlike `get()`, which raises exceptions (`DoesNotExist` or `MultipleObjectsReturned`), using `filter().first()` prevents the view from crashing and allows the code to handle unexpected cases safely. This makes the resend activation flow more robust even if validation assumptions fail.

rupayanjas:
## Description

Fixes a bug where `resend_activation_view` performed a case-sensitive username lookup.

Previously the code used:
`ESPUser.objects.get(username=form.cleaned_data['username'])`

If the user entered the username with different casing (for example `JohnDoe` instead of `johndoe`), the lookup could fail and raise `ESPUser.DoesNotExist`, which could lead to an error.

This change replaces the lookup with a case-insensitive query:

`ESPUser.objects.filter(username__iexact=form.cleaned_data['username']).first()`

It also safely handles the case where no matching user is found.

## Related Issue

Closes #4798

## Type of Change

* [x] Bug fix
* [ ] New feature
* [ ] Breaking change
* [ ] Documentation update
* [ ] Refactor / cleanup
* [ ] CI/CD or build change

## Testing

* [x] Existing tests pass
* [x] Manually verified

Testing steps:

* Ran the project locally using Docker.
* Created a test user with username `johndoe`.
* Submitted a resend activation request using `JohnDoe`.
* Verified that the request works correctly and no error occurs.

## AI Disclosure

AI tools were used for guidance in understanding the repository structure and workflow. The code changes were reviewed and implemented by me manually.

## Checklist

* [x] Self-reviewed the code

### DIFF
--- a/esp/esp/users/views/registration.py
+++ b/esp/esp/users/views/registration.py
@@ -197,7 +197,10 @@ def resend_activation_view(request):
         if not form.is_valid():
             return render_to_response('registration/resend.html', request,
                                       {'form':form, 'site': Site.objects.get_current()})
-        user=ESPUser.objects.get(username=form.cleaned_data['username'])
+        user = ESPUser.objects.filter( username__iexact=form.cleaned_data['username']).first()
+        if not user:
+            return render_to_response('registration/resend.html', request,
+                                       {'form': form, 'site': Site.objects.get_current()})
         userkey=user.password[user.password.rfind("_")+1:]
         send_activation_email(user, userkey)
         return render_to_response('registration/resend_done.html', request,


willgearty:

PR #4848, opened 2026-03-07
Fix flaky testUserIDSearchOneResult

### LINKED ISSUE #1835: Flaky testUserIDSearchOneResult?
```
ERROR: testUserIDSearchOneResult (esp.program.tests.ViewUserInfoTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/learning-unlimited/ESP-Website/esp/esp/program/tests.py", line 141, in testUserIDSearchOneResult
    self.assertStringContains(response['location'], "/manage/userview?username=adminuser124353")
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/django/http/response.py", line 208, in __getitem__
    return self._headers[header.lower()][1]
KeyError: 'location'
```

(https://travis-ci.org/learning-unlimited/ESP-Website/builds/87389625)

May have been introduced in 1.8?

vipulpandey21:
## Description

Fixes a flaky test in `ViewUserInfoTest`.

The test `testUserIDSearchOneResult` previously searched using the admin user's ID. In some cases this could match multiple users depending on the database state, causing the `/manage/usersearch` view to return a search results page (HTTP 200) instead of redirecting (HTTP 302). This made the test flaky.

The test now searches using the username instead, which guarantees a unique match and ensures deterministic redirect behavior. The test also explicitly verifies that the response status code is `302` before checking the redirect location.

## Related Issue

Closes #1835

## Type of Change

* [x] Bug fix
* [ ] New feature
* [ ] Breaking change
* [ ] Documentation update
* [ ] Refactor / cleanup
* [ ] CI/CD or build change

## Testing

* [x] Existing tests pass
* [x] Manually verified

## AI Disclosure

AI assistance was used to help reason about the flaky test and propose the deterministic test change.

## Checklist

* [x] Self-reviewed the code
* [ ] Updated documentation (not required)


vipulpandey21:
Hi Mr. Will @willgearty, quick context: this change makes `testUserIDSearchOneResult` deterministic by searching via username instead of ID and explicitly asserting the redirect status. The full test suite was run multiple times locally and all checks are passing. Let me know if this approach looks fine to you?

hwatheod:
This is not the right fix, because it defeats the purpose of the test. The name of the test is `testUserIDSearchOneResult` which means it should be testing searching by user id, not username.

vipulpandey21:
Hi @hwatheod I tried to reproduce the intermittent `KeyError: 'location'` that was reported for this test.

From reading the test and the traceback, the failure would occur if `/manage/usersearch` returns a **non-redirect response**. The test currently accesses `response['location']` directly, which assumes the view returns a `302` redirect. If the view instead returns a normal results page (`200`)—for example if the search does not resolve to a single direct match—the `location` header would not exist and Django raises `KeyError`.

To check this, I temporarily modified the test to guard against the missing header (by first checking the response status / redirect) and ran the test repeatedly to see whether the intermittent failure disappeared. After that experiment, I **reverted the test back to the original implementation** and reran the test multiple times using:

`for i in {1..30}; do python manage.py test esp.program.tests.ViewUserInfoTest.testUserIDSearchOneResult --keepdb; done`

On my local setup the test passed consistently across all runs with the original code, and I was not able to reproduce a case where the response lacked the `location` header.

At the moment I cannot reproduce the failure locally, so the condition causing the non-redirect response does not seem to occur in my environment. If there is a CI failure log or a known scenario where `/manage/usersearch` returns a non-redirect response for this query, that would help narrow down the exact cause of the flakiness.

willgearty:
@vipulpandey21 Did you test the edge case that was mentioned in the issue discussion?
> this test will fail whenever the admin's id is contained in the user's first name

vipulpandey21:
Hi Mr. Will @willgearty, I checked this edge case and instead of just fixing the test, I made a small change in find_user() in views.py.
If the search string is purely numeric, I’m skipping the first_name__icontains, last_name__icontains, and similar name filters. Numeric input is more likely to be ID, username, or phone, so applying name-based filters didn’t really make sense and was causing inconsistent matches.
Now for numeric input, it only checks ID and username, so the redirect should be more predictable.

I ran the test multiple times locally and it’s been consistent so far.
(sorry for being late Mr. Will)

willgearty:
Thanks for making those changes. That sounds good to me.

Could you now please address the comment from @hwatheod above. The test should use a user ID.

> This is not the right fix, because it defeats the purpose of the test. The name of the test is `testUserIDSearchOneResult` which means it should be testing searching by user id, not username.

vipulpandey21:
Hi Mr. Will @willgearty,

I’ve updated the test to search by self.admin.id, which aligns with what the test name implies, and also addressed the underlying non-determinism in find_user.

The flakiness was coming from name-based filters (first_name__icontains, last_name__icontains) being applied even for purely numeric input. In cases where another user’s name happened to contain the admin’s ID digits as a substring, the query would return multiple results instead of a redirect, which led to a KeyError: 'location'.

The fix skips only the name-based filters when the input is numeric. Identifier-based filters still apply — username__iexact, id=userstr, and phone matching — so numeric usernames like "12345" are still handled correctly. A numeric input maps more naturally to identifiers than names, so applying name-contains filters here was both unnecessary and a source of DB-dependent behavior.

I also added an assertEqual(response.status_code, 302) before accessing response['location'] so failures are clearer, and replaced the hardcoded "adminuser124353" with self.admin.username.

willgearty (approved):
Looks good, thanks!

PR #4864, opened 2026-03-07
Fix PlainRedirect exposing all recipient addresses in To: header (#2393)

### LINKED ISSUE #2393: Plain redirects should protect recipient info

AnushkaPandit-21:
## Problem

When an email is sent to a forwarding address (e.g., `directors@site.com`), the `PlainList` handler exposes **all destination addresses** in the To: header to every recipient. This is a privacy concern — recipients can see each other's personal email addresses, even if the original alias was BCC'd.

As @milescalabresi noted in #2393, this also leaks redirect destinations in Mailer Daemon bounce messages when case mismatches occur.

## Root Cause

In `plainlist.py`, `self.recipients` was set to a comma-separated destination string, so all addresses ended up in a single To: header — exposing everyone's address.

### 1. Split comma-separated destinations (`plainlist.py`)
Each `PlainRedirect.destination` field is now properly split into individual addresses:

```python
self.recipients = []
for redirect in redirects:
    for addr in redirect.destination.split(','):
        addr = addr.strip()
        if addr:
            self.recipients.append(addr)
```

### 2. SMTP envelope routing for full privacy (`mailgate.py`)
Per @milescalabresi's feedback, recipients should not see each other's addresses at all — not even their own real address in the To: header. The fix uses **SMTP envelope routing**:

- The `To:` header shows the **alias address** (e.g., `directors@site.com`), preserving the appearance of a normal mailing list
- The actual recipient is passed via the **SMTP envelope** using `sendmail -i <recipient>` (no `-t` flag), so the MTA delivers to the real address without it appearing in headers
- Uses `subprocess.Popen` instead of `os.popen` to pass arguments as a list (avoids shell injection)

```python
# When hide_recipients is True:
message['To'] = 'alias@site.com'         # What the recipient sees
send_mail(str(message), envelope_recipient=real@example.com)  # Where it actually goes
```

This is the same approach used by Mailman and other mailing list software.

## Testing

Added `PlainListRecipientPrivacyTest` with tests covering:
- `hide_recipients` flag is set on PlainList handler
- Original alias address is stored for To: header substitution
- Case-insensitive matching preserves original address casing
- Single and multi-address destination splitting
- Multiple PlainRedirect objects for the same alias
- Whitespace trimming
- No-match scenario

All tests pass. @SuyashJain17 confirmed the fix works locally.

**Test output:**

image
image

## AI Disclosure

AI assistant used for code suggestions and test scaffolding. Root cause analysis, architectural decisions, and all testing performed by contributor.

Closes #2393

### DIFF
--- a/esp/esp/dbmail/receivers/plainlist.py
+++ b/esp/esp/dbmail/receivers/plainlist.py
@@ -12,7 +12,12 @@ class PlainList(BaseHandler):
         if len(redirects.values('id')[:1]) == 0:
             return
 
-        self.recipients = [redirect.destination for redirect in redirects]
+        self.recipients = []
+        for redirect in redirects:
+            for addr in redirect.destination.split(','):
+                addr = addr.strip()
+                if addr:
+                    self.recipients.append(addr)
 
         self.send = True
 
--- a/esp/esp/dbmail/receivers/plainlist.py
+++ b/esp/esp/dbmail/receivers/plainlist.py
@@ -5,12 +5,18 @@ from esp.dbmail.models import PlainRedirect
 
 class PlainList(BaseHandler):
 
+    # Hide real recipient addresses from the sender by using SMTP envelope
+    # routing instead of exposing them in the To: header.
+    hide_recipients = True
+
     def process(self, user, user_match):
 
         redirects = PlainRedirect.objects.filter(original__iexact = user)
 
         if len(redirects.values('id')[:1]) == 0:
             return
+
+        self.original_address = user
 
         self.recipients = []
         for redirect in redirects:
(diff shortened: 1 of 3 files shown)


SuyashJain17:
Nice work on this! I tested it locally and the change seems to address the issue.

AnushkaPandit-21:
Thanks for testing @SuyashJain17! Glad to hear it's working as expected.

@milescalabresi — I've addressed your feedback by implementing SMTP envelope routing so the actual recipient addresses are completely hidden from the sender. The `To:` header now shows only the alias address, while the real recipient is passed via the SMTP envelope. Would love to get your thoughts on whether this approach looks good for merge.

willgearty:
Superseded by #3832

PR #4869, opened 2026-03-07
Fix: Prevent infinite loop in onsiteattendance.py

### LINKED ISSUE #4857: [Bug]: Infinite loop risk in times_attending_class() when class end time precedes start time
### Description

In `esp/esp/program/modules/handlers/onsiteattendance.py`, the `times_attending_class()` method has a `while(True)` loop that iterates hourly from `start_time` to `end_time`. However, if `end_time` is earlier than `start_time` (due to timezone issues, data corruption, or a multi-week program date adjustment error), the loop will never terminate because `time` starts at `start_time` and increments past an `end_time` that is already in the past.

The relevant code:
```python
start_time = sr.start_date.replace(minute=0, second=0, microsecond=0)
end_time = sr.section.end_time().end.replace(
    year=sr.start_date.year, month=sr.start_date.month, day=sr.start_date.day,
    minute=0, second=0, microsecond=0)
# ...
while(True):
    # ...
    if time > end_time:
        break
```

`esp/esp/program/modules/handlers/onsiteattendance.py`

This is a @cache_function_for(105) cached function, so a single corrupt record could lock a server thread for the full 105-second cache duration before being called again.

### Steps to Reproduce

1. Have a ClassSection whose end_time() returns a time earlier than sr.start_date (possible with timezone issues or corrupt data)
2. 2. Navigate to the attendance page as an admin
3. 3. The server thread hangs indefinitely in the while(True) loop

### Expected Behavior

The function should skip StudentRegistration records where end_time < start_time, or cap the loop iterations.

### Actual Behavior

The while(True) loop never terminates, causing the server thread to hang.

Proposed Fix: Add a guard before the while loop: `if end_time < start_time: continue`

I discovered this while auditing the onsite modules for test coverage as part of issue #3780. I would be happy to fix this and submit a PR with a regression test. Could I be assigned to this issue?

naman005:
Fixes #4857 

Modified `esp/esp/program/modules/handlers/onsiteattendance.py` to add a guard clause checking `if end_time < start_time`. This prevents a while(True) loop from hanging the server thread when encountering corrupt or timezone-misaligned data, as reported in the issue.

## AI Usage Disclosure
Tool & Scope: Gemini 3 Flash; used for professional phrasing of the commit message and PR description to ensure standard technical English.

### DIFF
--- a/esp/esp/program/modules/handlers/onsiteattendance.py
+++ b/esp/esp/program/modules/handlers/onsiteattendance.py
@@ -176,7 +176,7 @@ class OnSiteAttendance(ProgramModuleObj):
                 # treat this as a cross-midnight class and shift the end time forward by one day.
                 if section_end_dt.time() < start_time.time():
                     end_time = end_time + datetime.timedelta(days=1)
-                
+
                 # If after adjustment the end time is still invalid, skip this record.
                 if end_time < start_time:
                     continue


esp/esp/program/modules/handlers/financialaidappmodule.py
willgearty, line 1: This change is unrelated, please remove it.

esp/esp/program/modules/handlers/onsiteattendance.py
```diff
@@ -170,6 +170,16 @@ def times_attending_class(self, prog):
             user = sr.user
             time = start_time
             # loop through hours until we get to the end time of the section
+            if end_time < start_time:
+                # If the original section end time-of-day is earlier than the start time-of-day,
+                # treat this as a cross-midnight class and shift the end time forward by one day.
+                if section_end_dt.time() < start_time.time():
```
willgearty, line 177: `section_end_dt` is not a defined variable

naman005: sorry, I was fixing that copilot's suggestion and missed changes in line +167, +168, fixed that now

naman005:
Hi @willgearty, I've addressed all the changes.

I accidentally pushed the financial aid refactor here because I missed a branch switch earlier sorry about that!
I also cleaned up the attendance logic which I was implementing the suggested fix by copilot but missed the variable definition, which caused that linting error. Thanks for pointing it out promptly.

Everything is fixed now and all 7/7 tests (including linting and unit tests) have passed. Kindly review when you have a chance!

PR #4256, opened 2026-02-21
Improve rapid check-in: first name search, filtering, and barcode vis…

SuyashJain17:
## Description

This PR improves the student rapid check-in process by enhancing search flexibility and adding administrative filters:

- **First Name Search**: Enabled autocomplete matching by first name (in addition to last name and username).
- **New Filters**: Added Grade (7-12) and Last Name Range (A-G, H-N, O-Z) dropdowns to narrow down student lists.
- **UI Enhancement**: Promoted the Barcode Check-in link to a high-visibility button at the top of the interface.
- **Backend Refactor**: `Updated ESPUser.ajax_autocomplete` and `db/views.py` to support these new dynamic filtering parameters.

## Related Issue

Closes #3600

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [ ] Existing tests pass
- [x] New tests added (if applicable)
- [x] Manually verified

## AI Disclosure

I used Antigravity (AI assistant) to help research the onsite check-in architecture and implement the search/filtering logic.

## Checklist

- [x] Self-reviewed the code
- [ ] Updated documentation (if needed)
- [x] No new warnings or errors introduced

PR #4876, opened 2026-03-07
Fix: catch both KeyError and ValueError in ajax_autocomplete

### LINKED ISSUE #4870: [Bug]: `except KeyError as ValueError` in ajax_autocomplete doesn't catch ValueError
### Description

In `esp/esp/db/views.py:40`, the exception handler uses `except KeyError as ValueError`, which catches `KeyError` and binds it to a variable named `ValueError` — it does **not** catch `ValueError`. This means if a user sends a non-numeric `limit` parameter (e.g., `?limit=abc`), the `int()` call on line 34 raises a `ValueError` that goes uncaught, resulting in a 500 Internal Server Error instead of the intended 400 Bad Request.

Additionally, this shadows Python's builtin `ValueError` within the except block scope.

```python
# Line 34-40
limit = int(request.GET.get('limit', 10))  # ValueError if limit=abc
...
except KeyError as ValueError:  # Only catches KeyError, binds it to name "ValueError"
    response = HttpResponse('Malformed Input')
    response.status_code = 400
    return response
```

### Steps to Reproduce

1. Log in to the site as any authenticated user
2. Send a GET request to the ajax_autocomplete endpoint with a non-numeric `limit` parameter, e.g. `?limit=abc&model_module=...&model_name=...&ajax_data=...&prog=...`
3. The `int('abc')` call raises `ValueError`
4. Observe a 500 error instead of the expected 400 "Malformed Input" response

### Expected Behavior

The endpoint should return a 400 Bad Request with "Malformed Input" for any invalid input, including non-numeric `limit` values.

### Actual Behavior

A `ValueError` is raised and not caught, resulting in a 500 Internal Server Error. Only `KeyError` (missing GET parameters) is caught.

### Screenshots

_No response_

### Operating System

_No response_

### Browser

_No response_

### Additional Context

**File:** `esp/esp/db/views.py:40`

The fix is straightforward — change `except KeyError as ValueError:` to `except (KeyError, ValueError):` so both exception types are caught.

Utkarsh09102004:
## Description

`except KeyError as ValueError` on line 40 of `esp/db/views.py` catches `KeyError` and binds it to a variable named `ValueError` — it does **not** catch actual `ValueError`. This means a non-numeric `limit` parameter (e.g., `?limit=abc`) causes `int()` to raise an uncaught `ValueError`, resulting in a 500 instead of the intended 400.

The fix changes the handler to `except (KeyError, ValueError):` so both exception types are caught and return a proper 400 response.

## Related Issue

Closes #4870

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

## AI Disclosure

Used AI for researching code. Bug identification done while research. Fix implemented with help of AI.

## Checklist

- [x] Self-reviewed the code


Utkarsh09102004:
@willgearty This one's ready for review whenever you get a chance — single line fix. Thanks!

willgearty:
Already fixed in https://github.com/learning-unlimited/ESP-Website/pull/4256. Thanks, though!

PR #4890, opened 2026-03-08
commit with lint failures

hwatheod:
## Description

This is just me testing that if I uninstall the pre-commit tool then the lint failure is still detected on the PR.


hwatheod:
Lint failure is detected. Good.

PR #4897, opened 2026-03-08
fix: Prevent 500 errors by replacing unsafe .objects.get() calls (#4895)

### LINKED ISSUE #4895: [Feature]: Prevent 500 server errors by replacing unsafe .objects.get() calls in student class registration
### Problem Statement

While reading through esp/esp/program/modules/handlers/studentclassregmodule.py, I noticed that the addclass_logic method uses .objects.get() to look up the ClassSection and ClassSubject based on IDs passed from the frontend form.
If these form IDs are ever missing, corrupted, or tampered with by a user, the .objects.get() call will crash the page with an ugly 500 Internal Server Error instead of showing a proper "Not Found" message.

### Proposed Solution

We can easily fix this by swapping out .objects.get() for Django's built-in get_object_or_404().
This tiny change means that if a bad ID is ever passed from the frontend, the server will gracefully show a standard 404 page instead of throwing a 500 error and creating noise in the server logs.

### Alternatives Considered

We could wrap the code in a try/except block and raise an ESPError, but get_object_or_404() does the exact same thing in just one line of code, which keeps things clean.

### Additional Context

I found this while studying the codebase to prepare for GSoC 2026. I already have the fix ready locally and would love to submit a PR for it as soon as this is approved!
Create the issue with this text. Once you have the issue number, let me know, and I'll push the branch right away!

Rakesh-18112006:
## Description

Replaces unsafe .objects.get() calls in studentclassregmodule.py ( addclass_logic )  with Django's built-in get_object_or_404(). If a user submits an invalid or non-existent section_id or class_id via POST request, this prevents the application from throwing an unhandled 500 Internal Server Error. Instead, it gracefully returns a standard 404 response. 

## Related Issue

Closes #4895 

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [X] Existing tests pass
- [X] Manually verified

## Checklist

- [X] Self-reviewed the code

### DIFF
--- a/esp/esp/program/modules/handlers/studentclassregmodule.py
+++ b/esp/esp/program/modules/handlers/studentclassregmodule.py
@@ -41,6 +41,7 @@ from decimal import Decimal
 
 from django.contrib.auth.models import User
 from django.db.models.query import Q, QuerySet
+from django.shortcuts import get_object_or_404
 from django.http import HttpResponse, Http404
 from django.views.decorators.cache import cache_control
 from django.views.decorators.vary import vary_on_cookie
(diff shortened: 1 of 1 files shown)


willgearty (approved):
Looks good, thanks!

PR #4914, opened 2026-03-08
fix: add undo button for accidental student attendance marking

### LINKED ISSUE #4906: [Feature]: Undo automatic checkin that occurs if accidentally marking student as attending a class
### Problem Statement

If you mark a student as attending a class, and they were not already checked in to the program, it automatically checks them in to the program. This makes sense. However, suppose you marked them as attending the class by accident, and then uncheck the box to undo this. It does NOT undo the automatic check in to the program, and in fact, that box is greyed out, so you cannot manually undo the automatic check in to the program (from that page). 

### Proposed Solution

I think if you uncheck the attending class check box, and if the time stamp of that attendance record is the same as the time stamp of the check in to the program, then the check in to the program should be deleted as well. 

### Alternatives Considered

The time stamp solution isn't foolproof, but the foolproof alternatives would involve complicated extra record types / fields / etc to indicate that the check in occurred via attendance of class. I'm not sure it's worth that extra complication, but perhaps.

### Additional Context

Thoughts, @willgearty?

AnushkaPandit-21:
Closes #4906

## Summary
When marking a student as attending a class, an automatic program check-in Record is created if the student wasn't already checked in. Previously, unchecking the attendance checkbox did not undo this automatic check-in, and the check-in checkbox was greyed out — leaving no way to reverse the accidental check-in from the attendance page.

This adds an "Undo" button (following the pattern from teacher check-in) that appears after marking a student as attended. The button tracks whether an auto-check-in occurred via a `data-checkedin` attribute, and when clicked, reverses both the attendance marking and the associated program check-in.

## Changes
- **Backend** (`teacherclassregmodule.py`): Added `undo_checkin` POST parameter to `ajaxstudentattendance()`. When set, deletes the program-level "attended" Record and returns `uncheckedin: true` in the response.
- **Frontend** (`attendance.js`): After marking attendance, an "Undo" button appears with `data-checkedin` attribute. Clicking it calls the endpoint with `undo=true` and `undo_checkin=true` (if auto-check-in occurred), then updates both checkboxes in the UI.
- **CSS** (`attendance.css`): Minor styling for the undo button.

## Test Plan
- [x] Mark a student as attending who is NOT checked in → check-in box auto-checks, undo button appears
- [x] Click undo → both attendance and check-in are reversed
- [x] Mark a student as attending who IS already checked in → undo button appears with `data-checkedin="false"`
- [x] Click undo → only attendance is reversed, check-in remains
- [x] Error handling: if undo AJAX fails, button re-enables for retry
- [x] Webapp icon toggle (`check_box`/`check_box_outline_blank`) works correctly on undo

## AI Disclosure
Claude Code was used for code exploration and development assistance. All code was reviewed, tested, and validated by the author. Architectural decisions (following the teacher check-in undo pattern per @willgearty's suggestion in #4906) were made by the human contributor.


willgearty (approved):
I've addressed the Copilot comments (except for adding a test which is out-of-scope), so this is good to go. It works great, thanks!

PR #4916, opened 2026-03-08
Fix/theme recompile settings

### LINKED ISSUE #1429: Theme editor errors on the first recompile
I don't 100% know when this issue happens, but when people set up a dev server, load a database dump, and then first recompile the theme from /themes/, it gives them an `IOError`.  The theme does get recompiled, though.  If you use the command line to recompile it, it errors the first time, and then works the second.  (I think; @mgersh was the one that tried it that way.)  The issue is that the Tag that saves the theme information has a file for saved settings that doesn't exist.

Irfan762:
Fix error when recompiling theme for the first time.

When a developer sets up the project with a database dump and recompiles the theme from the /themes/ directory, an IOError appears the first time.

This happens because the theme settings file does not exist yet, but the system tries to read it during recompilation. Because of this, the first run shows an error, even though the theme still recompiles.

If the command is run again, it works correctly because the settings file has already been created.

Proposed Fix
- Check if the settings file exists before reading it
- If it doesn't exist, create it with default settings
- Otherwise continue the normal recompilation process

Closes #1429

### DIFF
--- a/esp/esp/themes/controllers.py
+++ b/esp/esp/themes/controllers.py
@@ -321,7 +321,7 @@ class ThemeController(object):
 
         # Save current parameters and palette before they are cleared
         current_vars = self.get_current_params()
-        current_palette = self.get_palette()['custom']
+        current_palette = json.loads(Tag.getTag('current_theme_palette', default='[]'))
 
         backup_info = self.clear_theme(keep_files=keep_files)
         self.load_theme(theme_name, backup_info=backup_info)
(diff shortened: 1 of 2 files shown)


Irfan762:
Hi @benjaminjkraft ,
This PR fixes issue https://github.com/learning-unlimited/ESP-Website/issues/1429. Could someone please approve the workflows so the CI checks can run?
Thank you!

willgearty:
You have changes related to multiple PRs in this one. Please remove them such that only changes related to fixing the theme recompile are in this PR.

Irfan762:
Thanks for the feedback. I’ll remove the unrelated changes and update this PR so that it only contains the fix for the theme recompile issue.

willgearty:
@Irfan762 please address the Copilot comments.

Irfan762:
@willgearty Sure, I’ll review and address the Copilot comments shortly.

Irfan762:
@willgearty ,
Could you please review this PR when you have time?

willgearty:
I agree with Copilot here, you should restore the original behavior and not merge the loaded settings into the current settings.

Irfan762:
Thank you so much @willgearty  Really appreciate your time and for merging my PR

PR #4934, opened 2026-03-09
Fix mutable default arguments in Formstack API methods

### LINKED ISSUE #4907: [Bug]: Mutable default arguments used in Formstack API methods
### Description

Several methods in `esp/formsstack/api.py` use mutable default arguments (`args = {}`).

Examples:

def delete(self, id, args = {}):
def create_field(self, form, args = {}):
def __request(self, method, args = {})

Using mutable objects like `{}` as default parameters in Python is unsafe because the same object instance is reused across function calls. If the dictionary is modified during one call, the modified state persists in subsequent calls, which can lead to unintended side effects and hard-to-trace bugs.

The recommended pattern is to use `None` as the default value and initialize the dictionary inside the function.

### Steps to Reproduce

1. Open the file `esp/formsstack/api.py`.
2. Locate the following methods:
   - delete(self, id, args = {})
   - create_field(self, form, args = {})
   - __request(self, method, args = {})
3. Observe that a mutable default argument (`{}`) is used.
4. If the dictionary is modified during a call, the same mutated object may be reused in later calls.

### Expected Behavior

Functions should avoid mutable default arguments.

The safer pattern is:

def delete(self, id, args=None):
    if args is None:
        args = {}

This ensures a new dictionary is created for every function call.

### Actual Behavior

The methods currently define `args` with a default value of `{}`, which causes Python to reuse the same dictionary instance between function calls.

This may lead to unexpected behavior if the dictionary is modified during execution.

### Screenshots

_No response_

### Operating System

_No response_

### Browser

_No response_

### Additional Context

_No response_

rupayanjas:
## Description
This PR fixes the use of mutable default arguments (`args = {}`) in several methods inside `esp/formsstack/api.py`.

Using mutable objects like `{}` as default parameters in Python is unsafe because the same object instance is reused across function calls. If the dictionary is modified during one call, the modified state persists in subsequent calls, which can lead to unintended side effects and difficult-to-debug behavior.

This change replaces the mutable default with `None` and initializes the dictionary inside the function.

Updated methods:
- delete()
- create_field()
- __request()

Example fix pattern used:

def func(..., args=None):
    if args is None:
        args = {}

## Related Issue

Closes #4907

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

## AI Disclosure

AI assistance was used to help understand the repository structure.
All code changes were reviewed and verified before submission.

## Checklist

- [x] Self-reviewed the code

### DIFF
--- a/esp/esp/formstack/api.py
+++ b/esp/esp/formstack/api.py
@@ -80,19 +80,25 @@ class Formstack(object):
         args['id'] = id
         return self.__request('edit', args)
 
-    def delete(self, id, args = {}):
+    def delete(self, id, args=None):
+        if args is None:
+            args = {}
         """Deletes an existing submission."""
 
         args['id'] = id
         return self.__request('delete', args)
 
-    def create_field(self, form, args = {}):
+    def create_field(self, form, args=None):
+        if args is None:
+            args = {}
         """ Creates a field."""
 
         args['form'] = form
         return self.__request('createField', args)
 
-    def __request(self, method, args = {}):
+    def __request(self, method, args=None):
+        if args is None:
+            args = {}
         """ Makes a Formstack API call and returns the response as an array."""
 
         args['api_key'] = self.__api_key


hwatheod:
Closing all formstack PRs as very few schools use it.

PR #4944, opened 2026-03-09
test: add unit tests for ClassCreationController methods

### LINKED ISSUE #4943: [Feature]: Add direct unit tests for ClassCreationController methods (set_class_data, update_class_sections, force_availability)
### Problem Statement

ClassCreationController in esp/program/controllers/classreg.py is a critical class responsible for handling teacher class creation and editing logic. However, its internal methods currently lack isolated unit tests.

The only direct test ClassCreationControllerTest in esp/program/tests.py effectively does not validate any behavior because it swallows all exceptions

```python
def test_init_stores_program(self):
    try:
        controller = ClassCreationController(self.program)
        self.assertEqual(controller.program, self.program)
    except Exception:
        pass
```
Most current coverage comes from HTTP integration tests introduced in PR #4292 that test the makeaclass and editclass views. While these tests are valuable, they do not validate the controller's internal logic directly.

Because of this:
- A regression in set_class_data() field assignment could go unnoticed.
- The identity comparison bug (k[:8] is not 'section_') has no regression guard.
- update_class_sections() logic for adding/removing sections is not tested directly.
- force_availability() conditional logic (grant-all vs no-op) is not tested at the unit level.

### Proposed Solution

Add a new ClassCreationControllerUnitTest class to test_class_creation.py that calls each controller method directly on model objects — no HTTP client, no view layer involved.

The class reuses the existing ClassCreationTestMixin + ProgramFrameworkTest infrastructure already used by the integration tests in the same file.

7 new tests

set_class_data() — 2 tests

- test_set_class_data_updates_title_in_db — verifies that title and category from cleaned_data are persisted to the DB via cls.save(). Also acts as a regression guard for the is not bug on line 121.
- test_set_class_data_does_not_overwrite_with_section_prefixed_keys — injects a section_grade_min: 999 key into cleaned_data and asserts it is not written onto ClassSubject.grade_min, directly testing the k[:8] != 'section_' guard.

update_class_sections() — 3 tests

- test_update_class_sections_adds_sections — verifies new ClassSection rows are created in the DB when num_sections increases.
- test_update_class_sections_removes_sections — verifies surplus ClassSection rows are deleted when num_sections decreases.
- test_update_class_sections_sets_duration_on_all_sections — verifies all sections (including pre-existing ones with a stale duration) are synced to cls.duration.

force_availability() — 2 tests

- test_force_availability_grants_all_timeslots_when_teacher_has_none — verifies all program timeslots are granted when the teacher starts with zero availability.
- test_force_availability_does_nothing_when_teacher_already_has_times — verifies availability is unchanged and send_availability_email is never called when the teacher already has ≥ 1 timeslot. (send_availability_email is patched with unittest.mock.patch.object so no email infrastructure is needed.)

Files to change
- test_class_creation.py — add ClassCreationControllerUnitTest
- __init__.py — register the new class for test discovery

All 7 tests have been verified passing locally

### Alternatives Considered

_No response_

### Additional Context

- Closes the unit-test gap left by PR #4292 ("Add tests for makeaclass/editclass views"), which addressed issue #794 but only at the HTTP integration level.
- The test test_set_class_data_does_not_overwrite_with_section_prefixed_keys surfaces a pre-existing bug: line 121 of classreg.py uses is not (identity comparison) instead of != (equality comparison) for a string check. This generates a SyntaxWarning on Python 3.8+ and is unreliable across CPython versions. The fix is a 1-character change: is not → !=. That fix can be included in the same PR as the tests.

rajeshuchil:
## Description

Adds direct unit tests for ClassCreationController methods in test_class_creation.py.  
These tests call controller methods directly on model objects without going through the HTTP view layer, covering gaps left by the existing integration tests (MakeAClassViewTest, EditClassBaseTest).

7 new test cases were added in ClassCreationControllerUnitTest

- test_set_class_data_updates_title_in_db  
  Verifies set_class_data() persists the title to the database.  
  Also acts as a regression guard for the is not string comparison bug in classreg.py.

- test_set_class_data_does_not_overwrite_with_section_prefixed_keys
  Ensures `section_-prefixed keys in cleaned_data are ignored by the controller.

- test_update_class_sections_adds_sections
  Confirms that `update_class_sections()` creates new sections when the requested number increases.

- test_update_class_sections_removes_sections
  Confirms that excess sections are removed when the requested number decreases.

- test_update_class_sections_sets_duration_on_all_sections 
  Ensures that all sections receive the updated duration value.

- test_force_availability_grants_all_timeslots_when_teacher_has_none
  Ensures that a teacher with no availability is granted all program timeslots.

- test_force_availability_does_nothing_when_teacher_already_has_times
  Ensures that teachers with existing availability are not modified.

These tests improve coverage of controller logic that was previously only exercised through integration tests.

## Related Issue

Closes #4943 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [x] CI/CD or build change

## Testing

- [x] Existing tests pass
- [ ] Manually verified

## Checklist

- [x] Self-reviewed the code


rajeshuchil:
Hi @kkbrum  I have addressed all the suggest changes, could you please review when you have time ?

kkbrum:
Great! No need to comment unless you have a question; we will review when we are able. It is likely to be a long time due to the hundreds of other PRs.

willgearty (approved):
Thanks for adding these tests for the ClassCreationController! I did just a little cleanup, but it should be good to merge now!

PR #4948, opened 2026-03-10
Add difficulty description to catalog

### LINKED ISSUE #3708: Class difficulties are saved as strings
The options for a class's difficulty is set by the global "teacherreg_difficulty_choices" tag. This tag is basically an ordered dictionary consisting of rankings/abbreviations (keys) and their descriptions (values). During class creation/editing, only the descriptions are shown, but the chosen key is saved as a string. This key is then shown on the course catalog and during student registration. However, this presents several problems:

1. The difficulty description is never shown outside of teacher registration (as far as I can tell). This means that students can only infer what the difficulty is based on the abbreviation. The default for this is some number of stars, but it's unclear to students how many stars is the max, etc.
2. When the choices tag is changed, this can possibly invalidate all course registration forms because the old chosen strings are no longer valid options. This similarly causes the difficulty filter on the course catalog to break (because the new options that are populated into the filter dropdown don't match the strings in the catalog).

I'm not too sure what the fix for this should be. Here are some possible options:

- We could have "difficulty option" model objects like we have flag categories which can be added and removed to individual programs? They would need to have an order such that they are listed in order in the course reg form and catalog filter dropdowns. But how do we backport this behavior to existing difficulty settings for existing classes? We could find all unique difficulty strings and add those as objects in a migration?
- We could force the tag to just be a list, but then we save the list index for the difficulty field. When we render the course in the catalog, we use the value at that list index. Unfortunately, this doesn't completely solve the tag change issues because the tag length could be shortened, causing the same problems. This would also have the problem related to backporting.
- Maybe we are fine with the current behavior regarding invalidating the reg form, and all we really need to fix is the catalog filter? Instead of using the choices tag, we could pull all unique difficulties from the classes in the catalog (although the entire page is cached which could be an issue), then just sort them alphanumerically? Although this still presents the issue where descriptions are never shown to students.

SuyashJain17:
## Description

This PR improves how class difficulty is displayed to students in the catalog and registration pages.

Previously, only the difficulty abbreviation (e.g. `**`) was shown, which could be unclear to students. This change retrieves the corresponding description from the `teacherreg_difficulty_choices` tag and displays both the abbreviation and its description (e.g. `* – Introductory`).

The storage format for difficulty remains unchanged (the key is still stored as a string), so this update focuses purely on improving visibility and clarity in the UI.

## Related Issue

Closes #3708 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

## Future Work

This PR focuses only on improving the UX without modifying how difficulty values are stored.  
If desired, a follow-up PR could explore a more structural solution (e.g., modeling difficulty as a first-class entity) to address potential data consistency issues when the difficulty tag changes.

## AI Disclosure

AI assistance was used only for wording improvements in documentation.

## Checklist

- [x] Self-reviewed the code


SuyashJain17:
Screenshot showing the updated catalog UI where the difficulty abbreviation is now displayed together with its description (e.g. ** – Intermediate).

image

samalarussa (approved):
Approving, appears to work properly once global difficulty tags are set

PR #4956, opened 2026-03-10
adding tests for esp/random (fixes #4837)

### LINKED ISSUE #4837: [Feature]:  Add unit tests for esp/random/utils.py
### Problem Statement

Some parts of the module esp/random/utils.py appear to have low or missing test coverage. 
This may allow regressions or unexpected behavior to go unnoticed during development.

### Proposed Solution

Add unit tests for the key functions in esp/random/utils.py.

The tests will:
- verify correct output for normal inputs
- check edge cases
- ensure that functions behave correctly under invalid inputs

This will improve test coverage and reliability of the module.
Related to #3780 (Create more unit tests).

### Alternatives Considered

_No response_

### Additional Context

_No response_

Dewolf1:
Closes #4837

### Problem
The issue #4837 pointed out that we dont have any unit tests for the functions in `esp/random/views.py`. This leaves `main`, `ajax`, and `good_random_class` open to regressons when people change things later.

### What I did
1. Added a new `tests.py` in `esp/random`. Im using the `ProgramFrameworkTest` class so it mocks everything correctly with the db stuff. The coverage is now at 100% for that module.
2. While testing the edgecases I noticed a bug in `ClassSubject.random_class()`. If a program blacklists every single class, the query returns 0 items and then it crashes with `ValueError: empty range for randrange()`, so I added a quick `count == 0` check to just return None gracefully instead.
3. Found another small bug where `good_random_class` would try to do `json.loads(Tag.getTag('random_constraints'))`. If the tag gets deleted or is empty, it returns None and crashes the server. modified it to default to an empty dict.

all the tests pass for me locally running `./manage.py test esp.random`


willgearty (approved):
Cleaned it up a bit given the overlap with some other tests that were added. I also addressed the concerns from GitHub. Good to merge now.

PR #4958, opened 2026-03-10
fixed potential shell injection vulnerability in mailgate

### LINKED ISSUE #4949: [Bug]: shell injection in mailgate.py
### Description

using os.popen with string formatting is a shell injection vulnerability. In mailgate.py, the code executes `p = os.popen("%s -i -t" % MAIL_PATH, 'w')`

if `MAIL_PATH` or any part of that string can be influenced by env variables, config files or user input in future, it allows arbitrary command execution on the server. 

### Steps to Reproduce

1. navigate to esp/mailgates/mailgate.py
2.  in line 55 in the send_mail function, find the use of os.popen with the string formatting `p = os.popen("%s -i -t" % MAIL_PATH, 'w')`

### Expected Behavior

when trying to send mail, the application should execute the send_mail process securely, without opening the system shell, thereby preventing shell injections. 
this can be fixed by replacing `os.popen` with the `subprocess` module and passing the arguments as a list.

### Actual Behavior

the system currently is vulnerable to shell injection if any part of the string is manipulated through external input

### Screenshots

_No response_

### Operating System

_No response_

### Browser

_No response_

### Additional Context

_No response_

sruthysatheesh:
## Description

os.popen sends the concatenated string to the shell. if any part of the string is manipulated by external input, any arbitrary command can be executed on the server.
in mailgates/mailgate.py, changed os.popen to subprocess.popen to prevent potential shell injection .

## Related Issue

Closes #4949 

## Type of Change

- [x] Bug fix

## Testing

- [x] Manually verified

## AI Disclosure

the fix for this issue was identified using ai assistance, and it has been manually verified that no new errors have been introduced by sending a dummy mail,  before opening the PR.

## Checklist

- [x] Self-reviewed the code

### DIFF
--- a/.github/scripts/project-tracker/requirements.txt
+++ b/.github/scripts/project-tracker/requirements.txt
@@ -1,0 +1,2 @@
+gspread>=6.0.0
+google-auth>=2.0.0
--- a/.github/scripts/project-tracker/update_project_sheet.py
+++ b/.github/scripts/project-tracker/update_project_sheet.py
@@ -1,0 +1,409 @@
+"""
+update_project_sheet.py
+
+Fetches open pull requests and issues from a GitHub repository via the
+GraphQL API and writes three summary sheets to a Google Spreadsheet:
+  1. Open Pull Requests
+  2. Open Issues
+  3. Contributor Activity
+
+Required environment variables:
+  GITHUB_TOKEN   – GitHub token with repo read access (provided by Actions)
+  GOOGLE_SA_KEY  – JSON string of a Google service-account key
+  SHEET_ID       – ID of the target Google Spreadsheet
+  REPO_OWNER     – GitHub repository owner (org or user)
+  REPO_NAME      – GitHub repository name
+"""
+
+import json
+import os
+import sys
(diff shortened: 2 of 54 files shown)


AnushkaPandit-21:
Good catch — I made a similar change in PR #4864 when adding SMTP envelope routing for the PlainRedirect privacy fix. A couple of notes from working with this code:

1. The `subprocess.Popen` approach is definitely the right call. Using `os.popen` with string formatting opens up shell injection if `MAIL_PATH` were ever to contain special characters.

2. You might want to consider using `stdin=subprocess.PIPE` with explicit encoding instead of `universal_newlines=True`, since email messages can contain mixed encodings. In my implementation I used:
```python
p.communicate(message.encode('utf-8') if isinstance(message, str) else message)
```
This handles both `str` and `bytes` inputs gracefully, which matters for the email pipeline where messages may come in either form.

3. The `import subprocess` should probably be moved to the top of the file with the other imports rather than inside the function.

willgearty:
Closed due to unresponsiveness.

PR #4959, opened 2026-03-10
Fix oversized memcached values by chunking cache entries

### LINKED ISSUE #603: Dashboard causes memcached to be sad for sufficiently large programs
The Splash 2012 dashboard tends to fail on dev servers because the memcached item size is too big. This can be fixed by configuring memcached with -I 2M (or similar), but stackoverflow says doing this is discouraged for performance reasons.

There's probably a way to fix the caching so that it doesn't need to cache a single item quite that big; maybe worth looking into?

deepaksoni47:
## Description

Fixes memcached failures for large dashboard-related cached payloads by splitting oversized values across multiple cache keys.

This updates the custom memcached backend to:
- Detect oversized cache values.
- Store them as chunked entries plus a metadata pointer key.
- Transparently reassemble on get/get_many.
- Clean up chunk keys when deleting the parent key.
- Use per-write unique chunk prefixes and a two-phase pointer update so partial chunk-write failures do not corrupt existing values.
- Clean up all old chunk keys after successful overwrites, including overwrite-to-smaller payloads.
- Preserve batched cache behavior for get_many (single base-key batch + single chunk-key batch).

Also adds tests to verify:
- Large-value round-trip behavior.
- Mixed get_many behavior (small + large + missing keys).
- Chunk cleanup on delete.
- Overwrite cleanup when replacing a larger chunked value with a smaller one.
- Partial-write rollback safety (existing cached value remains valid if a chunk write fails).

## Related Issue

Closes #603

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

## AI Disclosure

GitHub Copilot was consulted to gain a better understanding of the issue and to review potential suggestions. All implemented changes were carefully examined and applied only after manual verification to ensure correctness and reliability.

## Checklist

- [x] Self-reviewed the code


Oval17 (commented):
Changes are really good , 
1- is there any possibility In-band sentinel can collide 
Tests are solid

deepaksoni47:
Thanks! Yes, it's theoretically possible, but the chance is extremely low since the sentinel is intentionally unique and the metadata format is validated before reconstruction. Happy to switch to an out-of-band metadata scheme (or another encoding approach) instead of using an in-band sentinel if you would like.

Oval17 (approved):
Changes lgtm.

willgearty (approved):
Thanks for putting this together! I had Claude Code run a full test on this, and it looks like the caching backend behavior had changed between Django 2 and Django 5, so your implementation actually stopped working. I've used Claude Code to update everything accordingly, rewrite the tests, and then make a dummy set of 800 classes to test that the chunking is functioning properly. Good to go now.

PR #4968, opened 2026-03-11
Fix #1101: prevent custom form schema operations from hanging due to database locks

### LINKED ISSUE #1101: Custom forms can hang the entire site
Since custom form creation involves editing table schemas, it can block on a lot of different operations, and a lot of different operations can block on it.  If someone leaves a shell_plus open that has a lock on certain tables, while someone else tries to create a custom form, this can hang the entire site.  It would be nice if this error were somehow caught, even if it's just with a timeout.

Prm01:
## Description
Custom form schema operations (such as creating tables or modifying fields) may hang indefinitely if database tables are locked by another process.

This PR prevents the site from hanging by setting a PostgreSQL `lock_timeout` before performing schema operations. If the lock cannot be acquired within the timeout period, the operation fails gracefully instead of blocking indefinitely.

## Changes Made
- Added `_set_lock_timeout()` helper method.
- Configured PostgreSQL `lock_timeout` before schema operations.
- Wrapped schema operations in `try/except OperationalError` to handle lock timeout failures.
- Applied the fix to:
  - `createTable`
  - `deleteTable`
  - `addField`
  - `updateField`
  - `removeField`

## Related Issue
Closes #1101

## Testing
- Ran the project locally using Docker.
- Verified that the server starts correctly.
- Confirmed schema operations now use a lock timeout instead of hanging indefinitely.

## AI Usage Disclosure
- Tool: ChatGPT (GPT-5)
- Scope of assistance: Helped understand the `DynamicModelHandler` code and reason about a lock-timeout-based safeguard for schema operations.
- Human verification: I manually reviewed the code, implemented the changes myself, and tested locally to verify that the server starts successfully after the modifications.


willgearty (approved):
Claude Code recommended rewriting this with module-level transaction helpers. That's now applied and appears to work well. Thanks for the initial contribution; should be good to merge now.

PR #4970, opened 2026-03-11
Fix: correct upcoming timeslot pre-check logic in unenrollmodule

### LINKED ISSUE #4971: [Bug]: upcoming timeslot pre-check logic incorrect in unenrollmodule
### Description

The `upcoming` flag in `unenrollmodule.py` line 107 is incorrectly 
defined, causing past timeslots to be pre-checked in the right column 
of the unenroll page.

### Steps to Reproduce

1. Start a program with multiple timeslots
2. Let some timeslots pass
3. Navigate to the unenroll page as admin
4. Observe the right column checkboxes

### Expected Behavior

Only timeslots that started within the last hour or haven't started 
yet should be pre-checked in the right column.

### Actual Behavior

All past timeslots are pre-checked in the right column because 
`timeslot.start < now + hour` evaluates to True for every past slot.

### Screenshots

N/A - this is a code logic issue

### Operating System

Platform independent - this is a backend logic bug

### Browser

N/A

### Additional Context

N/A

Sameer0297:
Closes #4971

## Issue
The 'upcoming' flag in 'unenrollmodule.py' was incorrectly defined, 
causing past timeslots to be pre-checked in the right column of the 
unenroll page.

## Root Cause
The condition 'timeslot.start < now + hour' evaluates to 'True' for 
ALL past timeslots, not just upcoming ones. For example, if it is 
2:00 PM, slots from 10:00 AM, 11:00 AM, and 12:00 PM would all be 
incorrectly pre-checked.

## Fix
Changed the condition to 'timeslot.start > now - hour' so that only 
slots that started within the last hour or haven't started yet are 
pre-checked in the right column.

## Before
'''python
'upcoming': timeslot.start < now + hour,
'''

## After
'''python
'upcoming': timeslot.start > now - hour,
'''

### DIFF
--- a/esp/esp/program/modules/handlers/unenrollmodule.py
+++ b/esp/esp/program/modules/handlers/unenrollmodule.py
@@ -104,7 +104,7 @@ class UnenrollModule(ProgramModuleObj):
             'slot': timeslot,
             'seq': seq,
             'passed': timeslot.start < now - hour,
-            'upcoming': timeslot.start < now + hour,
+            'upcoming': timeslot.start > now - hour,
         } for seq, timeslot in enumerate(timeslots)]
         context = {}
         context['selections'] = selections


Sameer0297:
I have been assigned to issue #4971. Please reopen this PR.

Issue #1836, opened 2015-10-26
Two phase lottery needs better error handling


mgersh:
Right now, if you try to register for a class, it'll check if you're in the right grade and silently drop your registration if you aren't. Ideally it would error more loudly.

luac:
It's only supposed to display classes that are in the right grade, so this would only happen if there is a bug somewhere. But in general, yes, better error messaging would be good.

mgersh:
Yeah, this came up because there was in fact a bug somewhere which was harder to debug because of this.

pushkarkumar06:
Hi @willgearty ,

I’d like to work on this issue #1836  if it’s still open. Please let me know if I can take it.

Thanks!

PR #4973, opened 2026-03-11
Fix #1837: standardize grade 0 handling across the system

### LINKED ISSUE #1837: Handling of grade 0 is inconsistent
This is the underlying issue which caused #1836 to be relevant (or, the most directly related of multiple underlying issues). In some places, such as `meets_grade`, it's treated as something like "any grade". In other places, such as selecting lottery choices in two-phase, it's treated as literal grade 0.

pushkarkumar06:
Grade 0 is the sentinel value returned by `getGrade()` when a student has no graduation year on their profile. It means "unknown / unset" and should behave as "any grade", but several codepaths were treating it as a literal grade value,
silently blocking students.

## Description

`ESPUser.getGrade()` returns `0` when a student has no graduation year set on their profile. The intended meaning of grade 0 is "unknown / any grade", and the access-control decorators (`meets_grade`, `needs_student_in_grade`) already
handle it correctly by letting grade-0 students through. However, several downstream codepaths were missing this guard and treated grade 0 as a real grade, causing silent failures:

- `ClassSubject.cannotAdd` returned "not in grade range" for grade-0 students.
- `save_interests` in two-phase registration silently discarded all interests for grade-0 students because `grade_min__lte=0` matched no real classes.
- `save_priorities` silently skipped every section for grade-0 students.
- The lottery numpy filter excluded grade-0 students from every section with `grade_min >= 1`, blocking them from the lottery entirely.
- `onsitecatalog` showed an empty class list to grade-0 students.
- The class-change request module and `fillslot` view filtered out all section  for grade-0 students.

**Fix:** Added a single `ESPUser.grade_in_range(grade, grade_min, grade_max)` static helper that returns `True` unconditionally when `grade == 0`, and `grade_min <= grade <= grade_max` otherwise. All affected codepaths now use
this helper, establishing one consistent rule across the system.

## Related Issue

Closes #1837
Related to #1836

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

Manual verification steps performed:
1. Confirmed that a student with no graduation year (grade 0) can reach the two-phase registration page without hitting a "wrong grade" error.
2. Confirmed that `save_interests` accepts class IDs for a grade-0 student  instead of silently discarding them.
3. Confirmed that `save_priorities` registers sections correctly for a grade-0  student instead of logging "not in grade range" and continuing.
4. Confirmed that `cannotAdd` returns `None` (no error) for grade-0 students on classes with `grade_min >= 1`.
5. Verified the numpy lottery filter logic in isolation — grade-0 students are now included in every section's eligible pool while normal grade enforcement  is unchanged.
6. Confirmed `onsitecatalog` shows the full class list for grade-0 students.

## AI Disclosure

GitHub Copilot was used to assist with locating affected code areas and suggesting the initial structure of the `grade_in_range` helper. However, every changed file was manually reviewed in detail before and after the edit. The
fix logic, the identification of all 8 affected call sites (including the double-filter bug in `studentclassregmodule.py`), and the final correctness were verified manually multiple times. No AI-generated code was accepted without thorough human review.

## Checklist

- [x] Self-reviewed the code

### DIFF
--- a/esp/esp/program/controllers/lottery.py
+++ b/esp/esp/program/controllers/lottery.py
@@ -386,9 +386,12 @@ class LotteryAssignmentController(object):
         possible_students = numpy.copy(signup[:, si])
 
         #   Filter students by the section's grade limits
+        #   Grade 0 means "unknown/unset" – treat as any grade (allow through).
         if self.options['check_grade'] and not (priority == self.effective_priority_limit and self.grade_range_exceptions):
-            possible_students *= (self.student_grades >= self.section_grade_min[si])
-            possible_students *= (self.student_grades <= self.section_grade_max[si])
+            grade_ok = ((self.student_grades == 0) |
+                        ((self.student_grades >= self.section_grade_min[si]) &
+                         (self.student_grades <= self.section_grade_max[si])))
+            possible_students *= grade_ok
 
         if self.options['use_student_apps']:
             possible_students *= (self.ranks[:, si] == rank)
--- a/esp/esp/program/models/class_.py
+++ b/esp/esp/program/models/class_.py
@@ -1821,8 +1821,8 @@ class ClassSubject(models.Model, CustomFormsLinkModel):
             scrmi = self.parent_program.studentclassregmoduleinfo
             return scrmi.temporarily_full_text
 
-        if user.getGrade(self.parent_program) < self.grade_min or \
-               user.getGrade(self.parent_program) > self.grade_max:
+        user_grade = user.getGrade(self.parent_program)
+        if not ESPUser.grade_in_range(user_grade, self.grade_min, self.grade_max):
             if not Permission.user_has_perm(user, "GradeOverride", self.parent_program):
                 return 'You are not in the requested grade range for this class.'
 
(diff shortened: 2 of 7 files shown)


samalarussa:
Not having a grade should behave as an error, since this is not a state that a user's account should be in if they intend to register for classes. What we do not want is for it to behave as a "cheat code" to get into any class (as this PR intends to to do), since we want to respect the grade ranges set by teachers for their classes. Per this comment, we want it to behave as "no grade", not "any grade": https://github.com/learning-unlimited/ESP-Website/issues/1837#issuecomment-151310068 . Generally fixing this behavior requires more verbose error handling, not changing the underlying behavior around grade 0

pushkarkumar06:
Hi @samalarussa  ,
@willgearty 

Thanks for the detailed feedback and clarification regarding the expected behavior around missing grades. I understand now that grade `0` should behave as “no grade” instead of “any grade”, and that the fix requires proper error handling rather than changing the underlying logic.

I’d like to work on this issue again with the corrected approach. If possible, could you please reopen this PR, or let me know if you would prefer that I create a new PR instead?

Thanks again for the guidance!

willgearty:
Sorry for the delay @pushkarkumar06. If you'd like to take another stab at this, you're welcome to reassign yourself and open a new PR. Thanks!

PR #4975, opened 2026-03-11
Implement Engine for automatically adding class flags (#1368)

### LINKED ISSUE #1368: Engine for automatically adding class flags
It should be possible to specify that, when a class is added or edited, it can be assigned certain class flags if the class satisfies certain conditions. For example, we can automatically tag a class if it asks for a budget or materials.

This can be preferable to just adding new search filters, because the tags can be commented on and resolved.

Supriyo760:
## Description

This PR implements an engine for automatically adding class flags to [ClassSubject](cci:2://file:///c:/Users/Dange/Downloads/ESP-Website-main/ESP-Website-fresh/esp/esp/program/models/class_.py:1432:0-2150:29) objects based on predefined rules. Rules are created directly from the Class Search interface using the existing Query Builder logic. 

Key features:
* **Automatic Application**: A `post_save` signal evaluates all rules for a program whenever a class is created or updated.
* **Integrated UI**: Admins can turn any search query into an auto-flag rule with a single click in the search results.
* **Robust Signaling**: Handles request-less saves (e.g., background tasks) by using a fallback system user and integrates with the existing teacher notification system.
* **Extensible**: Rules are stored as JSON-serialized Query Builder states, maintaining full compatibility with all existing search filters.

## Related Issue

Closes #1368

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

## AI Disclosure

AI was just used for suggestion

## Checklist

- [x] Self-reviewed the code


Supriyo760:
All changes are done.

willgearty (changes requested):
Just had to clean up the migrations, but this looks and works great, thanks for putting it together! It would be nice to a have nice frontend for deleting rules, but the admin pages is probably fine for now.

I think the only thing that is needed is, as your own comment suggests, applying the flag to existing classes. I think this should be an option in the modal (a checkbox), and if it's checked, the flag should also get applied to all of the existing classes that match the query.

esp/esp/program/modules/handlers/classsearchmodule.py
```diff
@@ -234,6 +237,62 @@ def query_builder(self):
...
+            flag_type = self.program.flag_types.get(id=flag_type_id)
+            # Create the rule
+            rule = AutoClassFlagRule.objects.create(
+                program=self.program,
+                flag_type=flag_type,
+                rule_data=query_data,
+                comment=comment
+            )
+
+            # If checkbox is checked, apply flag to all existing matching classes
```
willgearty, line 263: Most of the following code is duplicated in the test. It would make the most sense to refactor this to be a method of the `AutoClassFlagRule` class that you can then call in the two locations.

esp/esp/program/models/class_.py
```diff
@@ -2194,3 +2194,62 @@ def install():
...
+        return
+
+    # Use ClassSearchModule to get a QueryBuilder instance for this program
+    module = ClassSearchModule(program=instance.parent_program)
+    qb = module.query_builder()
+
+    current_request = get_current_request()
+
+    for rule in rules:
+        try:
```
willgearty, line 2214: It might make sense to refactor the following code as a method of the `AutoClassFlagRule` class.

willgearty (changes requested):
Works great, thanks! Just some last refactoring required.

willgearty (approved):
Looks great, thanks for cleaning up that duplicated code!

PR #4996, opened 2026-03-12
Fix: Replace vulnerable eval() with secure DOM injection in ajax_tools.js #4128

### LINKED ISSUE #4128: [Bug]: Security: Replace eval() in ajax_tools.js with a Safe Alternative
### Description
[apply_fragment_changes()](cci:1://file:///Users/sourya/Desktop/learning_/ESP-Website/esp/public/media/scripts/ajax_tools.js:70:0-98:1) in [esp/public/media/scripts/ajax_tools.js](cci:7://file:///Users/sourya/Desktop/learning_/ESP-Website/esp/public/media/scripts/ajax_tools.js:0:0-0:0) uses `eval()` to execute JavaScript returned by the server:

eval(data[key]); // executes server-returned script

### Expected Behavior

1. Refactor server responses to avoid returning executable scripts.
2. Replace eval() with explicit DOM operations or a whitelist-based update mechanism.
3. If eval() must be retained temporarily, enforce strict Content Security Policy (CSP) headers.

This is a critical cross-site scripting (XSS) vulnerability. If the server response is compromised or tampered with, arbitrary code can be executed in the user's browser.

### Actual Behavior

....

### Screenshots

_No response_

### Operating System

_No response_

### Browser

_No response_

### Additional Context

_No response_

Sameer-00001:
## Description
This PR addresses a critical cross-site scripting (XSS) vulnerability in `esp/public/media/scripts/ajax_tools.js`. 
I removed the dangerous `eval(data[key])` call within the `apply_fragment_changes()` function. It has been replaced with a secure, explicit DOM operation that dynamically creates a `<script>` node, injects the text content, appends it to the document head, and immediately removes it. This ensures any returned scripts are safely processed through the browser's Content Security Policy (CSP) validators.

## Related Issue
Fixes #4128 (Partial fix: Frontend DOM Injection)

## Type of Change
[x] Bug fix
[ ] New feature
[ ] Breaking change
[ ] Documentation update
[x] Refactor/cleanup
[ ] CI/CD or build change

## Testing
[ ] Existing tests pass
[ ] New tests added (if applicable)
[x] Manually verified via local browser console simulation


willgearty:
@Sameer-00001 please address the Copilot comments.

Sameer-00001:
Hey @willgearty great catch, Since the server is intentionally returning executable JavaScript in the data['script'] payload, swapping eval() for a <script> node doesn't actually neutralize the XSS risk if the payload is compromised.

Before I attempt another fix, how would the maintainers prefer to handle this? Should we be sanitizing this payload with a specific library, or is there a broader plan to stop sending executable scripts via JSON entirely? Let me know what architectural direction you'd prefer! I'll be focusing on my university exams for the next few days, but I will dive straight into refactoring this the moment I'm back.

deepaksoni47 (commented):
Hey, I pulled the PR locally and tested it.

You’re right that switching from `eval()` to script injection doesn’t fully eliminate the risk if the payload itself is compromised. But it’s still a meaningful improvement since it avoids direct arbitrary execution and aligns better with CSP.

From what I see, the current backend is already designed to send executable JS, so a full fix would require changing that contract (moving to structured JSON and handling logic on the client). That’s a larger architectural change.

For this PR, it’s probably better to:

* Keep the DOM-based approach as an incremental security improvement
* Fix the style issues (indentation, braces)
* Avoid over-engineering sanitization here unless maintainers explicitly want it

If maintainers later decide to move away from script payloads entirely, that can be handled in a separate refactor.

Just my take after testing locally 🙂

moksha-hub (commented):
I took a closer look at the full request/response path here, and Sameer's concern is justified: this change swaps one execution mechanism for another, but it does not remove the executable-JS contract that makes the sink dangerous in the first place. The server is still serializing JavaScript into `result['script']` in at least `esp/esp/program/modules/handlers/studentclassregmodule.py` (`ajax_schedule()`) and `esp/esp/program/views.py` (statistics AJAX responses), and `apply_fragment_changes()` still runs whatever string arrives under that key. Appending a `<script>` node here therefore preserves the same trust boundary problem as `eval(...)`; if that payload is compromised, it still executes. Because of that, the current patch does not look like a real fix for #4128, and the comment about this being a "secure alternative" is misleading. The architectural fix seems to be moving away from shipping executable JS inside JSON responses and instead returning structured data / HTML fragments that preloaded client code can handle without evaluating new code from the payload.

Sameer-00001:
Hey @deepaksoni47, thanks so much for pulling this down and taking the time to test it locally!

I completely agree with your logic here. Trying to re-architect the server's payload contract to strip out executable JS entirely is definitely out of scope for this specific PR. Moving away from eval() to a direct DOM injection is a solid incremental step forward and aligns the code closer to modern CSP standards.

I'll go ahead and push up the formatting fixes (indentations and braces) to clear out the remaining Copilot suggestions. Thanks again for the review and the pragmatic approach!

Sameer-00001:
Hey @moksha-hub, thank you for doing that deep dive into studentclassregmodule.py and views.py! You are 100% correct about the backend contract—the server is still serializing the executable JS, which is exactly why I mentioned earlier that this doesn't neutralize the XSS risk if the payload itself is compromised.

My comment agreeing with Deepak was just regarding the scoping of this specific PR. Rewriting those Python views to strip out the JS payload entirely is definitely the correct long-term architectural move, but since this PR originated just to tackle the frontend eval() call from issue #4128, I didn't want to unilaterally rewrite the backend data contracts without the maintainers' green light.

Let's see what Will or Miles prefer when they get through the GSoC queue. If they want the full Python refactor done as part of this PR, I'm absolutely ready to jump into it!

deepaksoni47 (commented):
This looks like a good incremental improvement over `eval()`, but it doesn’t fully resolve the underlying XSS risk since executable JS is still being returned and executed.

Because of that, this probably shouldn’t close #4128 yet. It would be better to treat this as a partial fix and handle the full solution (removing executable JS from responses) in a follow-up PR.

It may also be better to update the PR description from “Closes #4128” to something like “Fixes #4128 (partial)” to reflect the current scope more accurately.

If maintainers prefer to address the full fix in this PR itself, I’m happy to help review those changes as well.

Sameer-00001:
Hey @deepaksoni47, great call! I actually noticed my description had a typo (1128) anyway, so I've gone ahead and updated it to Addresses #4128 (Partial fix) to ensure we keep the main issue open for the backend refactor.

I completely agree with your assessment. If Will and the maintainers want the full Python payload refactor done inside this PR, I'm absolutely ready to push my limits and jump right into those backend changes. Thanks for the review!

willgearty (approved):
Tested on /manage/statistics and /learn///catalog and everything works properly.

PR #5002, opened 2026-03-12
Fix: fall back to current program in userview when user has no profile

### LINKED ISSUE #2646: get_last_program_with_profile doesn't work for volunteers
#2417 added functionality to specify a program to retrieve user information for `/userview`, and if no program is specified, we use `get_last_program_with_profile()`, which will work fine on students and teachers because they fill out a program profile when registering. However, users that are strictly volunteers (not also teachers/students) never fill out a profile for a program, so this returns nothing, which means there is no `program` attribute in the `context`, which means #2587 doesn't work unless you actually specify a program using the dropdown.

I'm not quite sure how widely used `get_last_program_with_profile()` is, but perhaps we should be creating a profile for the program when a volunteer registers for a program? Or perhaps have a different default when `get_last_program_with_profile()` returns nothing, such as the next upcoming program?

Sameer0297:
Closes #2646

## Summary
When a volunteer (who never fills out a registration profile) visits 
`/userview`, `get_last_program_with_profile()` returns `None`, leaving 
the view with no program context. This fix adds a fallback to 
`Program.current_programs()` so volunteers see relevant program 
information instead of a blank view.

### program/views.py
- After `get_last_program_with_profile()` returns `None`, fall back 
  to the first result of `Program.current_programs()`
- If no current program exists either, behavior remains unchanged 
  (profile falls back to `user.getLastProfile()` as before)

## Root Cause
Volunteers never fill out a registration profile for a program, so 
`RegistrationProfile` has no matching row, causing 
`get_last_program_with_profile()` to return `None` as noted in #2646.

## Testing
- Manually traced the fallback logic through `current_programs()`
- Existing behavior preserved when a profile exists
- Volunteers now get the most current program as context


willgearty (approved):
Just added an additional fix to cover other uses of `get_last_program_with_profile()`. Should be good now. Thanks!

PR #5010, opened 2026-03-12
Fix statistics query including non-student users in zip code statistics

### LINKED ISSUE #4315: [Bug]: zip code statistics don't restrict to just students
### Description

The zipcode statistics question is "What are the most common zip codes for students to come from?"  But the [statistics query](https://github.com/learning-unlimited/ESP-Website/blob/e2a77c27600c7ac02fc2d0c7a4e256954073250c/esp/esp/program/statistics.py#L61) doesn't restrict to just students. It includes teachers, volunteers, etc.

### Steps to Reproduce

On my dev server right now, there are 46 `ESPUser` of which 34 are students. But on the `/manage/statistics/` page, if I select `What are the most common zip codes for students to come from?` and click "Go", it says "46 users were matched by the query." 

### Expected Behavior

34 users were matched by the query.

### Actual Behavior

46 users were matched by the query.

### Screenshots

_No response_

### Operating System

_No response_

### Browser

_No response_

### Additional Context

_No response_

SuyashJain17:
## Description

Fixes an issue where zip code statistics queries could include non-student users (such as teachers or volunteers).
This occurred because the statistics view built `users_q` using registration-based queries that could match users with program-related records (e.g., attended) even if they were not actually students.

The fix ensures that registration-based filters are explicitly restricted to the appropriate user type:

- Student queries are restricted to users of type Student

- Teacher queries are restricted to users of type Teacher

This prevents non-student users from appearing in student-only statistics queries.

## Related Issue

Closes #4315 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

## AI Disclosure

AI assistance was used to help review the logic. All code changes were reviewed and tested manually.

## Checklist

- [x] Self-reviewed the code

### DIFF
--- a/esp/esp/users/tests.py
+++ b/esp/esp/users/tests.py
@@ -468,8 +468,7 @@ class AccountCreationTest(TestCase):
                                   first_name="first",
                                   last_name="last",
                                   email="tsutton125@gmail.com")
-        except ESPUser.DoesNotExist as xxx_todo_changeme:
-            ESPUser.MultipleObjectsReturned = xxx_todo_changeme
+        except (ESPUser.DoesNotExist, ESPUser.MultipleObjectsReturned):
             self.fail("User not created correctly or created multiple times")
 
         if not Tag.getBooleanTag('require_email_validation'):
--- a/esp/useful_scripts/pqf_from_emails.py
+++ b/esp/useful_scripts/pqf_from_emails.py
@@ -1,11 +1,13 @@
 #!/usr/bin/env python
 
+import os
+
 from esp.users.models import User, ESPUser, PersistentQueryFilter
 from django.db.models.query import Q
 from io import open
 
 url = 'https://splashchicago.learningu.org/manage/Splash/2010_Fall/maincomm'
-filename = '/home/pricem/past_students.txt'
+filename = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'past_students.txt')
 
 file = open(filename)
 data = file.readlines()


SuyashJain17:
Fixed the query so that it now restricts results to student users. In my test environment there are 10 users in total and 5 students, and the statistics query now correctly matches 5 users. Attaching screenshots for reference.


image
image

hwatheod:
I have 1255 users with 1035 students in my development db, and performance on the statistics queries is extremely slow after this change. In fact, it seems like the server is just hanging. Can you improve the performance?

To generate a lot of users, you can use the script [here](https://github.com/learning-unlimited/ESP-Website/blob/main/esp/useful_scripts/generate_bulk_fake_data.py)

SuyashJain17:
Yeah, I think the slowdown was because the groups join was getting repeated for each OR branch.
I moved the Student/Teacher filter to run once after combining the Q objects, so we don’t hit that join multiple times anymore. Should help with performance while keeping the fix the same.

hwatheod:
It's still too slow.  I already limited it to just one program. Attached is the query that is running.  It is for the zip code statistics query.

[big_statistics_query_2.sql](https://github.com/user-attachments/files/26163423/big_statistics_query_2.sql)

I got this query by doing `docker compose down` while the query is running.  Then it shows the active query before it takes down the container.  There might be better way to do it.

SuyashJain17:
I found that the query was still including the teacher side for zip code stats. I’ve adjusted it to only build what’s needed and also removed the duplicate evaluation.

Could you pull and try it again (or run esp.program.test_statistics in Docker) and see if it’s better now?

esp/esp/program/views.py
```diff
@@ -1252,19 +1252,68 @@ def get_field_ids(form):
...
-                            users_q = users_q | students_objects[reg_type]
-                elif 'teacher_reg_types' in form.cleaned_data and form.cleaned_data['teacher_reg_types']:
-                    teachers_objects = program.teachers(QObjects=True)
-                    for reg_type in form.cleaned_data['teacher_reg_types']:
-                        if reg_type in list(teachers_objects.keys()):
-                            users_q = users_q | teachers_objects[reg_type]
+                student_q = None
+                if not teacher_only:
+                    if form.cleaned_data.get('student_reg_type_all'):
+                        student_q = program.students_union(QObject=True)
```
hwatheod: This `students_union()` isn't really a good function to use here. It includes categories like "enrolled_past" which is "All students who have enrolled in a program that happened before this one."  That doesn't make sense for this query. In `student_reg_type_all`, I would only include the categories (registration types) of students that are actually listed on the statistics form to be selected.

Also queries like `enrolled_past` are expensive since they have to look back into all programs.

esp/esp/program/views.py
```diff
@@ -1252,19 +1252,68 @@ def get_field_ids(form):
...
+                    if student_q:
+                        if student_users_q is None:
+                            student_users_q = student_q
+                        else:
+                            student_users_q |= student_q
+
+                teacher_q = None
+                if teacher_only:
+                    if form.cleaned_data.get('teacher_reg_type_all'):
+                        teacher_q = program.teachers_union(QObject=True)
```
hwatheod: Same for `teachers_union()` here. For `teacher_reg_type_all`, use only the list of teacher reg types on the form.

hwatheod (approved):
Thank you for doing this!

PR #5012, opened 2026-03-12
fix: save financial aid requests through the ModelForm instance

### LINKED ISSUE #5003: [Bug]: financialaidappmodule.py uses unsafe __dict__.update() bypassing Django model field descriptors
### Description

In `FinancialAidAppModule.finaid()`, student form data is persisted via:

    app.__dict__.update(form.cleaned_data)

This bypasses Django field descriptors and property setters entirely.
On fields with custom validation or caching decorators, this causes silent
data corruption — the wrong value may be stored in the database, or caches
may not be invalidated.

### Reproduction
1. Enable the Financial Aid module for a test program.
2. POST valid form data to `/learn/<prog>/finaid`.
3. Inspect the saved `FinancialAidRequest` — fields like `household_income`
   written via `__dict__` skip the ORM descriptor chain.

### Suggested Fix
Replace with `setattr(app, field, value)` for each field, or pass
`instance=app` directly to the ModelForm and call `form.save()`.

### Impact
HIGH — student financial data could be persisted incorrectly, silently.
This is a security and data integrity issue on a highly sensitive pathway.

moksha-hub:
- bind the financial aid form to the existing FinancialAidRequest instance on both GET and POST
- persist changes via orm.save(commit=False) instead of writing directly into __dict__
- add a regression test that confirms Django field save hooks are used during submission

Fixes #5003.

## Testing
- python -m py_compile esp\\esp\\program\\modules\\handlers\\financialaidappmodule.py esp\\esp\\program\\modules\\tests\\studentreg.py
- full pytest not run locally because this machine does not currently have the ESP Django dependencies installed


pushkarkumar06 (changes requested):
@moksha-hub  Thanks for working on this fix — switching to `ModelForm(instance=app)` looks like the correct approach and aligns with Django best practices. I tested the flow locally and the changes appear to persist correctly through `app.save()`.

One small thing I noticed: since the code now uses `form.save(commit=False)`, it might be safer to call `form.save_m2m()` after `app.save()` as well. The current model doesn't have ManyToMany fields, so this isn't an immediate issue, but it could become a problem later if one is added.

Apart from that, the change looks good and the regression test also seems to cover the intended behavior.

moksha-hub:
@Oval17  have a review.

esp/esp/program/modules/tests/studentreg.py
```diff
@@ -430,6 +432,33 @@ def test_finaid(self):
         self.assertIn('/learn/%s/studentreg' % self.program.url, response['Location'])
         self.assertEqual(iac.amount_due(), 0)
 
+    def test_finaid_uses_model_form_field_assignment(self):
+        student = random.choice(self.students)
```
Oval17: random.choice makes this non-deterministic if it ever flakes it'll be hard to reproduce. self.students[0] is simpler and reliable, especially with xdist parallel runs.
But its fine both ways, can be ignored.

Oval17 (commented):
LGTM.
Added a minor note.

Oval17:
LGTM.

moksha-hub:
@Oval17 what does it mean??

> LGTM.

Oval17:
> @Oval17 what does it mean??
> 
> > LGTM.

Definition : The primary full form of LGTM is "Looks Good To Me", a widely used acronym in software development for approving code changes, pull requests, or design documents

😅

PR #5013, opened 2026-03-12
fix: skip invalid time windows in OnSiteAttendance

### LINKED ISSUE #5004: [Bug]: Infinite loop risk in onsiteattendance.times_attending_class() when end_time <= start_time
### Description

`times_attending_class()` uses an unbounded `while True` loop:

    time = start_time
    while True:
        ...
        if time > end_time:
            break

If a section has `end_time <= start_time` (DST transition, data corruption,
or multi-day event edge case), the loop never terminates, hanging the server.

### Reproduction
Create a ClassSection whose end_time is earlier than its start_time
(e.g. across a DST boundary). Call `times_attending_class()` — it will
never exit.

### Suggested Fix
Add a guard before entering the loop:

    if end_time <= start_time:
        continue  # log a warning, skip this section

Or add a hard iteration cap (e.g. 48 iterations max).

### Impact
HIGH — runs during live Event Day attendance tracking. A hang freezes
admin operations for the entire program.

moksha-hub:
- replace the open-ended attendance loop with a bounded while time <= end_time loop
- skip impossible time windows where a registration starts after the section ends instead of recording a bogus hour
- add a regression test covering invalid attendance windows

Fixes #5004.

## Testing
- python -m py_compile esp\\esp\\program\\modules\\handlers\\onsiteattendance.py esp\\esp\\program\\modules\\tests\\onsiteattendance.py esp\\esp\\program\\modules\\tests\\__init__.py
- full pytest not run locally because this machine does not currently have the ESP Django dependencies installed


AnushkaPandit-21:
Clean fix — replacing the `while True` + `break` with `while time <= end_time` is more Pythonic and directly prevents the infinite loop when `start_time > end_time`.

The test coverage is a nice addition too. One edge case to consider: what happens if a registration's `start_date` is exactly equal to the section's end time? With the current `while time <= end_time` guard, it would count that boundary hour once. Depending on whether that's the intended behavior for the onsite attendance tracking (should a student who registered at the exact end of a section count as "attending" that hour?), it might be worth adding an assertion for that edge case in the test.

moksha-hub:
The remaining failing check appears unrelated to the OnSiteAttendance changes in this PR.

The failure is in `esp/program/tests.py::ViewUserInfoTest::testUserIDSearchOneResult` (`/manage/usersearch`), not in the onsite attendance handler or its tests. I opened a separate fix of pr  for that here: #5212.

Once that unrelated failure is resolved on the base branch, this PR should be re-runnable/reviewable on its own scope.

deepaksoni47 (commented):
Thanks for the update, this looks like a solid improvement. The bounded loop and skipping invalid windows make the flow much clearer.

Had a few thoughts while going through it:

- For overnight handling, it seems the logic still depends on time-of-day comparison. I might be missing something, but would it be safer to compute a same-day `candidate_end` and only shift it to the next day if it is actually before the attendance start timestamp?

- Related to that, maybe the invalid-window check should happen after this normalization so it is evaluated against the adjusted window.

- It could be helpful to add a regression test for a valid overnight case where the section spans midnight and attendance starts after midnight but before section end, just to ensure there is no spill into the next day.

- Also wondering about the boundary case where `start_date == section end`. Should that count for the last hour or be skipped? Might be worth adding a small assertion to make the expected behavior explicit.

Overall this is looking good to me, just a few edge cases that might be worth tightening.

PR #5031, opened 2026-03-12
Fix: Issue Data Integrity/Race Conditions in Student Enrollment

### LINKED ISSUE #5025: [Bug]: [Security]: Remove unnecessary @csrf_exempt from Formstack webhook and program registration views
### Description

**Labels:** `security`, `bug`, `enhancement`
 
---
 
## 1. Why This Issue Matters
 
Four views across the codebase use `@csrf_exempt`, disabling Django's built-in Cross-Site Request Forgery protection. While one usage (`unsubscribe_oneclick` in `esp/users/views/__init__.py`) is justified by RFC 8058 List-Unsubscribe-Post semantics, the remaining three — two in `esp/formstack/views.py` and one in `esp/program/views.py` (line 814) — are applied to views that accept POST requests and mutate server-side state. This means any malicious website can craft a hidden form that submits data to these endpoints on behalf of an authenticated admin or teacher, potentially creating or modifying program data without the user's knowledge.
 
Program administrators and teachers are the primary users at risk. During registration periods — when these views are most actively used — an admin browsing a compromised site could unknowingly trigger state changes in the ESP system.
 
---
 
## 2. Impact of Solving This Issue
 
- Eliminates a concrete CSRF attack vector on 3 state-mutating endpoints
- Brings the codebase closer to Django security best practices (CSRF protection is enabled by default for a reason)
- Reduces the attack surface during high-traffic registration windows when program data is most sensitive
- Sets a precedent for future contributors to avoid `@csrf_exempt` unless rigorously justified
 
---
 
## 3. Detailed Problem Description
 
**Affected files and lines:**
 
```python
# esp/esp/formstack/views.py — lines 3-6 and 24-30
from django.views.decorators.csrf import csrf_exempt
 
@csrf_exempt
def formstack_webhook(request):
    # Processes incoming Formstack form submissions
    # Mutates database: creates/updates student application records
    ...
 
@csrf_exempt
def formstack_post(request):
    # Handles Formstack POST data
    ...
 
# esp/esp/program/views.py — line 814
@csrf_exempt
@transaction.non_atomic_requests
def newprogram(request):
    # Creates entirely new programs with timeslots, resources, permissions
    # This is admin-only but CSRF protection is still critical
    ...
```
 
The `formstack_webhook` view is a genuine external webhook endpoint and may legitimately need CSRF exemption, but it should validate the request using Formstack's webhook signature (HMAC) instead of being completely unprotected. The `formstack_post` and `newprogram` views are internal browser-initiated POST requests and have no legitimate reason to bypass CSRF.
 
The `unsubscribe_oneclick` view in `esp/users/views/__init__.py` (line 240) is correctly exempt per RFC 8058 and should be left as-is.
 
---
 
### Step 1 — `esp/formstack/views.py` → `formstack_post`
 
Remove `@csrf_exempt`. Ensure the form template that POSTs to this view includes `{% csrf_token %}`. Test by submitting a Formstack-linked form through the browser.
 
### Step 2 — `esp/formstack/views.py` → `formstack_webhook`
 
Keep `@csrf_exempt` but add HMAC signature verification using the Formstack handshake key (stored in Django settings):
 
```python
import hashlib
import hmac
from django.http import HttpResponseForbidden
 
@csrf_exempt
def formstack_webhook(request):
    secret = getattr(settings, "FORMSTACK_WEBHOOK_SECRET", None)
    if secret:
        signature = request.META.get("HTTP_X_FS_SIGNATURE", "")
        expected = hmac.new(
            secret.encode(), request.body, hashlib.sha256
        ).hexdigest()
        if not hmac.compare_digest(signature, expected):
            return HttpResponseForbidden("Invalid signature")
    ...
```
 
Add `FORMSTACK_WEBHOOK_SECRET` to Django settings and document it in `docs/dev/settings.rst`.
 
### Step 3 — `esp/program/views.py` → `newprogram`
 
Remove `@csrf_exempt`. The `newprogram.html` template already renders a form — verify it includes `{% csrf_token %}`. This view is already restricted to admins via `@admin_required`, but CSRF protection provides defense-in-depth.
 
### Step 4 — Regression Tests
 
Add tests in `esp/program/tests.py` confirming that POST requests to these endpoints without a CSRF token return `403`:
 
```python
def test_newprogram_requires_csrf(self):
    self.client.enforce_csrf_checks = True
    response = self.client.post(reverse("newprogram"), data={...})
    self.assertEqual(response.status_code, 403)
```
 
### Step 5 — CI Guard
 
Add a simple grep-based CI check that flags new uses of `@csrf_exempt` for manual review:
 
```yaml
# .github/workflows/lint.yml
- name: Flag csrf_exempt usage
  run: |
    COUNT=$(grep -r "csrf_exempt" esp/ --include="*.py" | grep -v "unsubscribe_oneclick" | wc -l)
    if [ "$COUNT" -gt 1 ]; then
      echo "Unexpected csrf_exempt usage detected. Review required."
      exit 1
    fi
```
 
---
 

### Steps to Reproduce

Here are the concrete Steps to Reproduce the vulnerability for Issue 1, which you can add as a new sub-section (e.g., #### 5. Steps to Reproduce) to the issue.

5. Steps to Reproduce
This demonstrates the CSRF vulnerability on the newprogram admin endpoint (the same concept applies to formstack_post):

Start the ESP-Website development server (docker compose up).
Log into the website as an Administrator in your browser.
Create a local HTML file on your computer named csrf_exploit.html with the following content. Replace http://localhost:8000 with the correct base URL if different, and adjust the action URL to match the routing for newprogram:
<!DOCTYPE html>
<html>
  <body>
    <h1>You won a free iPad! Click here to claim!</h1>
    
    <form action="http://localhost:8000/manage/newprogram/" method="POST" id="exploit-form">
      <input type="hidden" name="name" value="Malicious CSRF Program" />
      <input type="hidden" name="url" value="csrf-hacked-2026" />
      
      <input type="submit" value="Claim Prize" />
    </form>
    
    
    <script>
      // document.getElementById('exploit-form').submit();
    </script>
  </body>
</html>
Open csrf_exploit.html in a new tab within the same browser session.
Click the "Claim Prize" button (or uncomment the JavaScript for auto-submission).
Expected Secure Behavior: Django should intercept the POST request and return a 403 Forbidden: CSRF verification failed. Request aborted.
Actual Vulnerable Behavior: Because the view has the @csrf_exempt decorator, the server accepts the request and creates the "Malicious CSRF Program" using the authenticated admin's session cookies.

### Expected Behavior

For Internal Views (formstack_post, newprogram): Any POST request to these endpoints must require a valid Django CSRF token (passed via the csrfmiddlewaretoken form field or the X-CSRFToken HTTP header). If the token is missing, invalid, or does not match the session cookie, the server must automatically reject the request and return an HTTP 403 Forbidden status (CSRF verification failed. Request aborted.). Legitimate form submissions from within the ESP-Website UI will naturally contain this token and succeed.
For the Webhook View (formstack_webhook): Because this is an external endpoint receiving data from Formstack's servers, it cannot use a standard Django CSRF cookie. However, the expected behavior is that the view strictly cryptographically verifies the incoming request. It must compute an HMAC-SHA256 signature using the request payload and the secret Formstack handshake key, and compare it to the X-FS-Signature header. If the signature is missing or invalid, the server must return HTTP 403 Forbidden. Only verified payloads should trigger state and database changes.

### Actual Behavior

Due to the presence of the @csrf_exempt decorator on these state-mutating views (formstack_post, formstack_webhook, and newprogram), Django entirely bypasses its built-in Cross-Site Request Forgery checks.

When a malicious POST request is sent to these endpoints without a CSRF token (or with an invalid one), the server processes the request as if it were legitimate. If the victim (e.g., a logged-in Administrator) unknowingly triggers the request via a compromised third-party site, their active session cookies are automatically sent with the payload. The views accept these credentials and execute the database modifications (such as creating a new educational program or modifying student application data) silently in the background, without any warning or validation failure.

### Screenshots

_No response_

### Operating System

_No response_

### Browser

_No response_

### Additional Context

_No response_

### LINKED ISSUE #5030: [Bug]: [Data Integrity]: Student enrollment via StudentRegistration.objects.get_or_create is susceptible to race conditions under concurrent lottery/registration load
## 1. Why This Issue Matters
 
During student registration (especially lottery-based enrollment), multiple concurrent requests can attempt to enroll students into the same class section simultaneously. The enrollment logic in `esp/program/controllers/lottery.py` (line 744) and `esp/program/controllers/classchange.py` (line 692) uses `RegistrationType.objects.get_or_create(name='Enrolled')` followed by separate enrollment record creation, but not all enrollment paths are wrapped in `transaction.atomic()`. This creates a TOCTOU (time-of-check-to-time-of-use) race condition where two requests can both pass the capacity check and both successfully enroll a student, resulting in over-enrolled sections.
 
This directly impacts students (who may be bumped after the fact), teachers (who must accommodate extra students), and admins (who must manually fix enrollment counts). During MIT Splash, with 3,000+ students registering simultaneously, this is not a theoretical concern.
 
---
 
## 2. Impact of Solving This Issue
 
- Prevents over-enrollment of class sections during high-concurrency registration
- Eliminates manual admin intervention to fix enrollment count mismatches
- Ensures fairness: students who register within capacity get their spots, period
- Improves data integrity of `StudentRegistration` records for downstream analytics
 
---
 
## 3. Detailed Problem Description
 
The core issue is in the enrollment flow:
 
```python
# esp/program/controllers/lottery.py — lines 735-744
with transaction.atomic():
    # ... lottery assignment logic ...
    # But the capacity check and enrollment creation are not
    # protected by SELECT ... FOR UPDATE
 
# esp/program/controllers/classchange.py — line 692
relationship, created = RegistrationType.objects.get_or_create(name='Enrolled')
# No transaction.atomic() wrapping the subsequent enrollment creation
```
 
**The race condition:**
 
1. Request A checks `section.enrolled_students < section.max_students` → True (49 < 50)
2. Request B checks `section.enrolled_students < section.max_students` → True (49 < 50)
3. Request A creates `StudentRegistration` → count is now 50
4. Request B creates `StudentRegistration` → count is now 51 (**over capacity!**)
 
The `ClassSection.enrolled_students` field is a cached count that is updated asynchronously, making the race window even wider.
 
---
 
### Step 1 — Wrap enrollment in `transaction.atomic()` with `select_for_update()`
 
```python
from django.db import transaction
 
with transaction.atomic():
    section = ClassSection.objects.select_for_update().get(pk=section_id)
    if section.num_students() < section.class_size_max:
        StudentRegistration.objects.create(
            user=student,
            section=section,
            relationship=enrolled_type
        )
    else:
        raise CapacityExceeded(...)
```
 
### Step 2 — Apply to all enrollment paths
 
Search for all locations that create `StudentRegistration` objects and wrap each with the atomic + lock pattern:
 
- `esp/program/controllers/lottery.py` — lottery assignment
- `esp/program/controllers/classchange.py` — class swap logic
- `esp/program/modules/handlers/studentregtwophase.py` — two-phase registration
- `esp/program/models/class_.py` — `ClassSection` enrollment methods
 
### Step 3 — Add a database-level unique constraint
 
Add a constraint to prevent duplicate `(user, section, relationship)` tuples as a last line of defense:
 
```python
class StudentRegistration(models.Model):
    class Meta:
        constraints = [
            models.UniqueConstraint(
                fields=['user', 'section', 'relationship'],
                condition=Q(expired=False),  # respects soft-delete if used
                name='unique_active_enrollment'
            )
        ]
```
 
Generate and apply the corresponding migration after adding the constraint.
 
### Step 4 — Write a concurrency regression test
 
Use Django's `TransactionTestCase` (not `TestCase`) to simulate concurrent enrollment:
 
```python
import threading
from django.test import TransactionTestCase
 
class EnrollmentRaceConditionTest(TransactionTestCase):
    def test_no_over_enrollment(self):
        section = ClassSectionFactory(max_students=1)
        students = [StudentFactory() for _ in range(10)]
        results = []
 
        def attempt_enroll(student):
            try:
                enroll_student(student, section)
                results.append('success')
            except CapacityExceeded:
                results.append('rejected')
 
        threads = [threading.Thread(target=attempt_enroll, args=(s,)) for s in students]
        for t in threads:
            t.start()
        for t in threads:
            t.join()
 
        self.assertEqual(results.count('success'), 1)
        self.assertEqual(section.num_students(), 1)
```

### Steps to Reproduce

# Steps to Reproduce
 
Here is how you can reliably reproduce this race condition in a local or staging environment.
 
---
 
## Prerequisites
 
- A test program where student registration is currently open.
- A test class section (`section_id=123`) that has at least 1 seat available.
- A test student account that is eligible to enroll in that class.
 
---
 
## Step 1 — Prepare the Concurrent Requests
 
Log in as the test student and intercept the form submission for class registration (or write a quick Python/Bash script using the student's session cookie) so you can fire identical HTTP POST requests.
 
Prepare two identical POST requests to the `addclass` or registration endpoint:
 
```
Request A: POST /learn/Splash/addclass/  with section_id=123
Request B: POST /learn/Splash/addclass/  with section_id=123
```
 
---
 
## Step 2 — Execute Requests Simultaneously
 
Fire both requests at the server at the exact same time using `curl` with bash background execution, or using `threading` in Python:
 
```bash
# Bash script to trigger the race condition
curl -X POST -b "sessionid=test_cookie" -d "section_id=123" "http://localhost:8000/learn/Splash/addclass/" &
curl -X POST -b "sessionid=test_cookie" -d "section_id=123" "http://localhost:8000/learn/Splash/addclass/" &
wait
```
 
---
 
## Step 3 — Observe the Data Integrity Failure
 
Check the database for the test student's registrations in this section. Two active `StudentRegistration` records will have been created for the same student and the same section:
 
```python
# In Django shell
from esp.program.models import StudentRegistration
 
StudentRegistration.valid_objects().filter(user=test_student, section_id=123).count()
# Expected: 1
# Actual (race condition): 2
```
 
---
 
## Step 4 — Alternative: Overfilling a Class
 
If the class section has exactly **1 seat remaining**, firing two concurrent requests with **two different students** will result in both being enrolled, pushing the section over its maximum capacity:
 
```
Request A evaluates self.isFull() → False
Request B evaluates self.isFull() → False
Both requests proceed to create StudentRegistration records.
Section is now over capacity.
```
 

### Expected Behavior

# Expected Behavior
 
Under concurrent load (e.g., during high-traffic lottery runs, registration opening rushes, or if a user accidentally double-clicks the submit button), the student enrollment system should strictly preserve data integrity and enforce capacity limits without race conditions.
 
Specifically, the system should guarantee the following:
 
---
 
## 1. No Duplicate Registrations
 
A student should only ever have a single active, valid `StudentRegistration` record for a specific class section and relationship type (e.g., `"Enrolled"`) at any given time.
 
If two concurrent requests attempt to enroll the same student in the same section, the first request should succeed, and the second request should gracefully recognize that the student is already registered — or fail safely — without creating a duplicate database row.
 
---
 
## 2. Strict Enforcement of Class Capacity
 
When multiple distinct students attempt to enroll in a class section simultaneously and the number of requests exceeds the remaining seating capacity, the system must properly serialize the capacity checks.
 
If only 1 seat remains, the first request to process should claim it, and the second concurrent request must be rejected because the class is now full. The system must never overfill a class beyond its defined maximum size due to a race condition.
 
---
 
## 3. Atomic Operations
 
The evaluation of `self.isFull()` and the subsequent creation of the `StudentRegistration` record must be treated as a **single, indivisible atomic operation** at the database level.
 
Relying solely on application-level checks like `if not qs.exists(): save()` or standard `get_or_create()` is insufficient for high-concurrency environments. The expected behavior requires the database to act as the ultimate source of truth, using:
 
- **Row-level locking** via `select_for_update()` when checking section capacity and student registration existence — ensuring no two transactions can read and act on stale state simultaneously.
- **Database-level unique constraints** on `(user, section, relationship)`, taking soft-deletion/validity into account, so that even if application-level guards are bypassed, the database rejects the duplicate at the persistence layer.
 

### Actual Behavior

# Actual Behavior
 
Under concurrent load (such as during the high-traffic phase when program registration initially opens, or during a bulk lottery processing script), the current registration logic fails to serialize capacity and existence checks, resulting in data integrity corruption.
 
Specifically, the system exhibits the following behaviors:
 
---
 
## 1. Vulnerable Check-Then-Act Workflow
 
In methods like `ClassSection.preregister_student()`, the system performs application-level checks without acquiring any database locks:
 
```python
# esp/program/models/class_.py
if overridefull or fast_force_create or not self.isFull(webapp=webapp):
    # [snip]
    qs = self.registrations.filter(...)
    if fast_force_create or not qs.exists():
        sr = StudentRegistration(user=user, section=self, relationship=rt)
        sr.save()
```
 
Because Django's `qs.exists()` and `.save()` are completely independent transactions, two concurrent requests for the same student can evaluate `qs.exists()` simultaneously. Both receive `False` from the database and both proceed to create a new `StudentRegistration` record — producing duplicate rows for the same student-class relationship.
 
---
 
## 2. Failing Class Capacity Limits (Overbooking)
 
The same vulnerability applies to the capacity check in `self.isFull()`. If a class has exactly 1 seat remaining and two different students attempt to register simultaneously, both requests evaluate `self.isFull()` before either has committed an enrollment to the database. Both receive `False` and both proceed to create enrollment records — effectively bypassing `class_size_max` and overfilling the class.
 
---
 
## 3. Inadequate Use of `get_or_create()`
 
Throughout modules like `studentclassregmodule.py`, `teacherclassregmodule.py`, `lottery.py`, and `adminreviewapps.py`, the codebase relies on `StudentRegistration.objects.get_or_create()`.
 
While `get_or_create()` catches `IntegrityError` to retry, it only works safely when a strict database-level unique constraint exists on the fields being retrieved or created. Because `StudentRegistration` uses soft deletion (maintaining history via validity constraints rather than hard deletes), a simple unique constraint cannot be applied directly to the table. Without either a partial unique constraint at the schema level or explicit row locks via `select_for_update()`, concurrent Python processes can race through `get_or_create()` and successfully insert duplicate `StudentRegistration` rows.
 

### Screenshots

_No response_

### Operating System

_No response_

### Browser

_No response_

### Additional Context

_No response_

khushthecoder:
## Description

This PR addresses the race condition vulnerability (Issue #5030) that occurs during high concurrency registration and lottery assignment periods. Previously, checking `ClassSection.isFull()` and creating the `StudentRegistration` record were distinct database transactions, meaning concurrent requests could simultaneously check capacity and both successfully enroll, bypassing max capacity limits.

To fix this data integrity issue, this PR introduces:
1. **Row-Level Locking:** Uses `select_for_update()` within a `transaction.atomic()` block in `ClassSection.preregister_student()` to strictly serialize access during capacity checks, preventing TOCTOU races.
2. **Database Constraint:** Adds a `UniqueConstraint` on `StudentRegistration` for `(user, section, relationship)` where `end_date IS NULL`. This ensures the database firmly prevents multiple active enrollments for the same student and section, even if application logic were to fail.
3. **Transaction Safety:** Wraps `ClassChangeController.save_assignments()` in entirely atomic transactions to ensure safely isolated concurrency operations.

## Related Issue

Closes #5030

## Type of Change

- x[x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] New tests added (`ClassConcurrencyTest` uses threads to execute 10 simultaneous registration attempts on a class with exactly `1` seat remaining, asserting that only 1 student officially enrolls and database integrity holds).
- [ ] Manually verified

## AI Disclosure

This PR was generated with the assistance of GitHub Copilot and Google Deepmind's Agentic AI to diagnose race conditions and write the concurrency tests.

## Checklist

- [x] Self-reviewed the code


Oval17 (commented):
This and #5033 are duplicates
cc @willgearty

PR #5033, opened 2026-03-12
Fix: Race conditions in student enrollment (Issue #5030)

### LINKED ISSUE #5030: [Bug]: [Data Integrity]: Student enrollment via StudentRegistration.objects.get_or_create is susceptible to race conditions under concurrent lottery/registration load

khushthecoder:
## Description

This PR fixes a critical **TOCTOU (Time-Of-Check-To-Time-Of-Use) race condition** in the student enrollment pipeline that can cause class sections to exceed their maximum capacity during high-concurrency registration periods (e.g., MIT Splash with 3,000+ concurrent students).

### Problem

`ClassSection.preregister_student()` evaluates `self.isFull()` and subsequently creates a `StudentRegistration` record as two independent database operations within the same transaction. Under concurrent load, multiple requests can simultaneously pass the capacity check before any of them commit an enrollment, resulting in:

- **Over-enrolled sections** — more students than `class_size_max` allows
- **Duplicate active registrations** — the same student enrolled twice for the same section
- **Manual admin cleanup** required after every high-traffic registration event

### Fix

| Change | File | Purpose |
|--------|------|---------|
| `select_for_update()` row lock | `esp/program/models/class_.py` | Locks the `ClassSection` row before evaluating `isFull()`, serializing all concurrent capacity checks |
| `UniqueConstraint` | `esp/program/models/__init__.py` | Database-level guard: `(user, section, relationship)` where `end_date IS NULL` prevents duplicate active enrollments even if application logic fails |
| `@transaction.atomic` | `esp/program/controllers/classchange.py` | Wraps `ClassChangeController.save_assignments()` in an atomic transaction for class-change operations |
| Migration `0033` | `esp/program/migrations/` | Applies the `unique_active_enrollment` constraint at the database level |
| `ClassConcurrencyTest` | `esp/program/tests.py` | 10-thread regression test: spawns concurrent enrollment requests for a capacity-1 section, asserts exactly 1 succeeds |

### How the lock works

```python
# Before (vulnerable to TOCTOU):
if not self.isFull():           # Thread A: False  |  Thread B: False
    StudentRegistration.save()  # Thread A: saves  |  Thread B: also saves → OVER-ENROLLED

# After (serialized):
locked = ClassSection.objects.select_for_update().get(id=self.id)
if not locked.isFull():         # Thread A acquires lock, checks → False
    StudentRegistration.save()  # Thread A saves, releases lock
                                # Thread B NOW acquires lock, checks → True (full) → rejected
```

## Related Issue

Closes #5030

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] New tests added — `ClassConcurrencyTest` uses `TransactionTestCase` with 10 threads racing to enroll in a capacity-1 section; asserts exactly 1 `StudentRegistration` row is created
- [ ] Manually verified

## AI Disclosure

AI tools were used to assist with boilerplate generation (migration file scaffolding and test setup/teardown structure). All architectural decisions, code logic, and review were done manually.

## Checklist

- [x] Self-reviewed the code


AnushkaPandit-21:
Solid approach using `select_for_update()` to serialize the capacity check. The TOCTOU race condition is a real concern for programs like MIT Splash with thousands of concurrent registrations.

A few observations:

1. The `UniqueConstraint` with `condition=Q(end_date__isnull=True)` is a good database-level safety net. Worth noting this uses a partial unique index which requires PostgreSQL — the codebase currently targets PostgreSQL so that should be fine, but it's worth mentioning in the PR description for maintainers' awareness.

2. For the migration: since this adds a constraint that could fail on existing data (if there are already duplicate active enrollments), you might want to add a data migration step that cleans up any existing duplicates before applying the constraint. Otherwise the migration could fail on production databases.

3. The concurrent test using threads is clever. One thing to watch out for: Django's `TransactionTestCase` uses real DB transactions (not the wrapped ones in `TestCase`), so this test will be slower — that's the right tradeoff here since you need real concurrency.

khushthecoder:
> Solid approach using `select_for_update()` to serialize the capacity check. The TOCTOU race condition is a real concern for programs like MIT Splash with thousands of concurrent registrations.
> 
> A few observations:
> 
> 1. The `UniqueConstraint` with `condition=Q(end_date__isnull=True)` is a good database-level safety net. Worth noting this uses a partial unique index which requires PostgreSQL — the codebase currently targets PostgreSQL so that should be fine, but it's worth mentioning in the PR description for maintainers' awareness.
> 2. For the migration: since this adds a constraint that could fail on existing data (if there are already duplicate active enrollments), you might want to add a data migration step that cleans up any existing duplicates before applying the constraint. Otherwise the migration could fail on production databases.
> 3. The concurrent test using threads is clever. One thing to watch out for: Django's `TransactionTestCase` uses real DB transactions (not the wrapped ones in `TestCase`), so this test will be slower — that's the right tradeoff here since you need real concurrency.

Thanks for the thoughtful review, @AnushkaPandit-21! Great observations — here's how I'll address each:

1. PostgreSQL requirement for the partial unique index: Good catch — I'll add a note in the PR description explicitly calling out that the UniqueConstraint with condition=Q(end_date__isnull=True) creates a partial unique index, which is a PostgreSQL-specific feature. Since ESP-Website targets PostgreSQL, this is compatible, but it's important for maintainers to be aware.

2. Data migration to clean existing duplicates: Excellent point. I'll split the migration into two steps:

Step 1 (data migration): A RunPython operation that finds and soft-deletes (sets end_date) any existing duplicate active enrollments, keeping only the most recent one per (user, section, relationship) group.
Step 2 (schema migration): The existing AddConstraint operation, which will now safely apply since duplicates are cleaned up first.
This ensures the migration won't fail on production databases with pre-existing duplicate data.

3. TransactionTestCase tradeoff: Agreed — TransactionTestCase is the right choice here since we need real concurrency with actual DB transactions. The slower test execution is an acceptable tradeoff for properly validating the race condition fix.

I'll push the updated migration shortly. Thanks again for the feedback!

willgearty:
Duplicate of #5031

PR #5047, opened 2026-03-13
Fix json response mutable default 

### LINKED ISSUE #5045: [Bug]: Mutable default argument used in json_response decorator
### Problem
In `decorators.py`, the function `json_response` defines a mutable default argument:

```python
def json_response(field_map={}):
```

Using mutable objects such as dictionaries or lists as default arguments in Python can lead to **unexpected behavior** because the same object instance is reused across multiple function calls.

Even if the current implementation does not modify `field_map`, this pattern introduces a **potential hidden bug** and does not follow Python best practices.

---

### File Location

```
devsite/esp/esp/utils/decorators.py
```

---

### Current Implementation

```python
def json_response(field_map={}):
    """ Converts a serializable data structure into the appropriate HTTP response.
        Allows changing the field names using field_map, which might be complicated
        if related lookups were used.
    """
```

---

### Proposed Solution

Replace the mutable default argument with `None` and initialize the dictionary inside the function to ensure that a new dictionary is created for each function call.

---

### Improved Implementation

```python
def json_response(field_map=None):
    """ Converts a serializable data structure into the appropriate HTTP response.
        Allows changing the field names using field_map, which might be complicated
        if related lookups were used.
    """

    if field_map is None:
        field_map = {}

    # Here instead of at the top because of circular imports
    from esp.utils.web import render_to_response
```

---

### Benefits

- Prevents bugs caused by **shared mutable state**
- Follows **Python best practices**
- Improves code reliability and maintainability
- Makes the decorator safer for future changes

### Steps to Reproduce

1. Open the file `devsite/esp/esp/utils/decorators.py`.
2. Locate the `json_response` decorator function.
3. Observe that the function defines a mutable default argument:

```python
def json_response(field_map={}):
```

4. In Python, mutable default arguments are shared between function calls, meaning the same dictionary instance will be reused if the function is called multiple times.

---

### Actual Behavior

The function defines `field_map` with a mutable default value (`{}`).  
If the dictionary is modified inside the function during execution, the changes would persist across future calls of the function, potentially causing **unexpected side effects and hard-to-trace bugs**.

---

### Expected Behavior

The function should avoid using mutable default arguments.  
Instead, it should use `None` as the default value and initialize the dictionary inside the function, ensuring that **a new dictionary is created for every call**.

Example:

```python
def json_response(field_map=None):
    if field_map is None:
        field_map = {}
```

### Expected Behavior

1. Open the file `devsite/esp/esp/utils/decorators.py`.
2. Locate the `json_response` decorator function.
3. Observe that the function defines a mutable default argument:

```python
def json_response(field_map={}):
```

4. In Python, mutable default arguments are shared between function calls, meaning the same dictionary instance will be reused if the function is called multiple times.

### Actual Behavior

The function defines `field_map` with a mutable default value (`{}`).  
If the dictionary is modified inside the function during execution, the changes would persist across future calls of the function, potentially causing **unexpected side effects and hard-to-trace bugs**.

---

### Expected Behavior

The function should avoid using mutable default arguments.  
Instead, it should use `None` as the default value and initialize the dictionary inside the function, ensuring that **a new dictionary is created for every call**.

Example:

```python
def json_response(field_map=None):
    if field_map is None:
        field_map = {}
```

### Screenshots

_No response_

### Operating System

Ubuntu 22.04

### Browser

_No response_

### Additional Context

Using mutable objects as default arguments is a well-known Python pitfall.  
When a mutable object such as a dictionary or list is used as a default parameter, the same object instance is reused across all function calls.

Although the current implementation of `json_response` does not explicitly modify `field_map`, future changes or external modifications could unintentionally mutate this dictionary, leading to unexpected behavior.

Adopting the recommended pattern of using `None` as the default value and initializing the dictionary inside the function ensures safer and more predictable behavior. This also aligns the codebase with Python best practices and improves long-term maintainability.

akashverma712:
## Related Issue

Closes #5045

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [ ] Existing tests pass
- [x] Manually verified

## Checklist

- [x] Self-reviewed the code


akashverma712:
@willgearty Sir, I am sorry to bother you but can you have a check on this and let me know if there are any changes I have fixed the  json response mutable default in the decorators.py can you have a look over it once you are free.

Once again thanks for the constant guidance throughout the contribution process.

willgearty:

willgearty (approved):
I cleaned up some whitespace changes and unrelated documentation changes. Otherwise good to go, thanks!

PR #5049, opened 2026-03-13
Fix ClassSection/ClassSubject __cmp__ O(n log n) DB queries

### LINKED ISSUE #4793: [Feature]: Fix ClassSection/ClassSubject __cmp__ methods that cause O(n log n) DB queries
### Problem Statement

Both `ClassSection.__cmp__` and `ClassSubject.__cmp__` in `esp/esp/program/models/class_.py` (lines 1151 and 2030) hit the database ~4 times per comparison. When sorting a list of n sections, this results in O(n log n) database queries. The code itself documents this issue:

```python
# Warning: this hits the DB around four times per comparison, i.e.,
# O(n log n) times for a list.  Consider using prefetch_related and
# then sorting with the key self.start_time_prefetched(), which will
# hit the DB only once at the start, and compute the start time of each
# class only once.
```

### Proposed Solution

Follow the suggestion in the existing code comments — prefetch meeting times on the queryset, then sort using a key function:

```python
sections = sections.prefetch_related('meeting_times')
sections = sorted(sections, key=lambda s: s.start_time_prefetchable())
```

Callers that sort `ClassSection` or `ClassSubject` objects should be updated to use this pattern instead of relying on the default comparison.

### Additional Context

Affected file: `esp/esp/program/models/class_.py`, lines 1151–1172 and 2030–2050. This affects any view that sorts class lists.

Arnav1709:
## Description

Fixes an O(n log n) database query issue caused by the `ClassSection.__cmp__` and `ClassSubject.__cmp__` methods. Previously, both methods invoked `firstBlockEvent()` which hit the database ~4 times per comparison during sorting.

This PR resolves the issue by:

1. Adding a `_sort_key()` helper method to both `ClassSection` and `ClassSubject` that works efficiently with prefetched data.
2. Utilizing the `start_time_prefetchable()` method (and adding its counterpart to `ClassSubject`).
3. Updating the caller sites across `programprintables.py`, `teacherpreviewmodule.py`, `teacheronsite.py`, and `adminreviewapps.py` to use `prefetch_related('meeting_times')` and explicitly sort using `key=lambda s: s._sort_key()`.

The pre-existing `__cmp__` definitions are safely retained behind-the-scenes as fallback for any un-updated code.

## Related Issue

Closes #4793

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

## Checklist

- [x] Self-reviewed the code


willgearty (approved):
Thanks for improving the performance of these queries! I've done just a couple of extra performance improvements, but it should be good to merge now!

PR #5051, opened 2026-03-13
Add save_as = True to ModelAdmin classes to enable object duplication (Fixes #596)

### LINKED ISSUE #596: Copy object in admin panel
Often when creating objects in the admin panel, you want to create n relatively similar objects, which vary in one or two fields.  It would be nice to have a "copy" button, which would make a new object, with visible fields identical to those of the existing object (it would likely start out with empty non-visible fields), for you to edit.  This would make it a lot easier to create a number of similar UserBIts/StudentRegistrations or other objects that we often create several of via the admin panel.

abdullahlko:
Fixes #596

## Problem
The admin panel had no way to duplicate/copy existing objects. 
Users had to manually recreate similar objects from scratch every time.

## Solution
Added `save_as = True` to all ModelAdmin classes across the codebase.
This enables Django's built-in "Save as new" button on all admin object pages,
allowing users to duplicate any object and modify only the fields they need.

## Files Changed
- esp/esp/accounting/admin.py
- esp/esp/application/admin.py
- esp/esp/cal/admin.py
- esp/esp/customforms/admin.py
- esp/esp/dbmail/admin.py
- esp/esp/program/admin.py
- esp/esp/program/modules/admin.py
- esp/esp/qsdmedia/admin.py
- esp/esp/resources/admin.py
- esp/esp/survey/admin.py
- esp/esp/tagdict/admin.py
- esp/esp/users/admin.py
- esp/esp/utils/admin.py
- esp/esp/web/admin.py

## Testing
Verified locally that "Save as new" button appears on admin object pages.
screenshot below:
image

### DIFF
--- a/esp/esp/accounting/admin.py
+++ b/esp/esp/accounting/admin.py
@@ -41,6 +41,7 @@ class LIOInline(admin.TabularInline):
     model = LineItemOptions
 
 class LITAdmin(admin.ModelAdmin):
+    save_as = True
     list_display = ['text', 'amount_dec', 'program', 'required', 'for_finaid', 'num_options', 'max_quantity']
     search_fields = ['text', 'amount_dec', 'program__url', 'program__name']
     list_filter = ['program']
@@ -48,6 +49,7 @@ class LITAdmin(admin.ModelAdmin):
 admin_site.register(LineItemType, LITAdmin)
 
 class TransferAdmin(admin.ModelAdmin):
+    save_as = True
     def option_description(self, obj):
         if obj.option:
             return obj.option.description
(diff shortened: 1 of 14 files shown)


abdullahlko:
Resolved merge conflict in `esp/esp/web/admin.py` by rebasing onto latest main.

- `save_as = True` addition is preserved
- Alongside the new `list_display`, `search_fields` and `get_readonly_fields` changes

PR is now up to date with main.

SuyashJain17 (commented):
This is a nice quality-of-life improvement for the admin, especially for models where we create similar entries often.

That said, adding `save_as = True` across all admin classes feels a bit broad. Some of these models have uniqueness constraints or represent things like financial records, configs, or audit data, where duplication might either fail or lead to unintended results.

Might be better to enable this only for models where duplication clearly makes sense, instead of applying it everywhere.

PR #5060, opened 2026-03-13
Fix waitlist_subscribe Record creation bug (#5043)

### LINKED ISSUE #5043: [Bug]: studentregcore.py waitlist_subscribe() uses invalid Record.objects.create() kwargs
### Summary
In `waitlist_subscribe()`, the waitlist record is created with:

    Record.objects.create(event__name="waitlist", user=request.user, program=prog)

`event__name` uses Django's double-underscore traversal syntax, which is
valid in `.filter()` queries but NOT in `.create()` or `.get_or_create()`.
Django will attempt to set a literal field named `event__name` on the
Record model, which does not exist, raising an unexpected `TypeError` or
simply silently failing to set the correct event.

### Expected Behavior
The Record should be created with the correct RecordType event:

    rt = RecordType.objects.get(name="waitlist")
    Record.objects.create(event=rt, user=request.user, program=prog)

### Impact
HIGH — student waitlist records are never actually created. Students
think they are on the waitlist but are not enrolled.

Yashin-53:
Fixes #5043

## Summary
The waitlist_subscribe() function used `event__name="waitlist"` in
Record.objects.create(). Django ORM does not support double-underscore
lookups in create() operations.

## Fix
Fetch the correct RecordType object and pass it to the event field.

### Before
Record.objects.create(event__name="waitlist", user=request.user, program=prog)

### After
rt = RecordType.objects.get(name="waitlist")
Record.objects.create(event=rt, user=request.user, program=prog)

## Impact
Ensures waitlist records are correctly created for students.


AnushkaPandit-21:
Clean fix. One thing to consider: `RecordType.objects.get(name="waitlist")` will raise `DoesNotExist` if that RecordType hasn't been created yet (e.g., on a fresh install or if someone deleted it). The other onsite modules handle this by just letting it propagate, so this is consistent — but it's worth noting in case anyone encounters it during setup.

kkbrum:
@Yashin-53 can you address Copilot’s feedback? Note that you don’t necessarily have to do what it says, but you should explain why it’s not a valid critique if so.

willgearty (approved):
Looks good, thanks!

PR #5073, opened 2026-03-14
Issue #3307: Add moderation movement dependency loop diagnostic

### LINKED ISSUE #3307: Moderation Loops Scheduling Check
Purpose is to catch moderation loops and movement dependencies.

Example of this can be seen here: 
image

Logic for this wouldn't be too difficult.

1. Iterating block - first check to see if there is a future time block which is scheduled immediately after the current block.
2. Next iterate through the list of moderators for the block.
3. Check and see the moderator is moderating/in a different room the next block
4. If yes check and see if the person replacing them in also moderating in the current block
5. If yes start building dependency list (ex for Teach A: Dep on B....)
6. Check teacher B for the same conditions and keep iterating on list
7. Check ends when - list discovers a novel person or returns the teacher being checked
8. Returning to the teacher being checked is critical and should return BAD THING cause that's a loop and no one in the chain will be able to move until admin intervention
9. Lists can be categorized by problem level, a 1 dependency list is probably not a problem but 10 could be a huge issue w/o admin intervention 

Some optimizations - when you iterate through the moderators for a block, you can skip everyone who is already in a previously recorded list because the check has already accounted for them.

Also in max length dependency chain of A -> B  -> C -> D ... etc. if teacher C is evaluated first the teacher C list could later be rejected because it will be a shorter duplicate of the teacher A list and therefore not really relevant.

The above could be implemented in the following way -
When doing the chain that starts with A check at each stage if there is another loop that contains the teacher. For example the loop would go A -> B -> C and at C would find the already existing C -> D ... chain. The A chain could then be prepended onto the C chain and the loop stopped.

Outside of virtual programs this kind of check could be used for classroom optimization. While very similar to the running teachers check - this would also give you a chain of running teachers which could help with the optimization of scheduling.

If you have a teacher chain like the one above, this check would make it simple to realize that an optimal schedule would only involve moving A -> room C, B -> room A, and C -> room B in the first block to optimize the schedule.

Boopana-M:
Summary
Implements a new Scheduling Diagnostics check that detects moderator movement dependencies across consecutive time blocks and identifies dependency chains that form loops requiring administrator intervention.

Description
This PR introduces a Scheduling Diagnostics check to detect moderator movement dependency chains across adjacent scheduling blocks and flag hard loops where the dependency chain returns to the starting moderator.

The diagnostic analyzes how moderators are reassigned between consecutive blocks, constructs a dependency graph for those transitions, and traverses the graph to identify dependency chains and circular dependencies.

Key Features
This PR adds:

->A new scheduling diagnostic for moderator movement dependency loops.
->Dependency graph construction across adjacent time blocks.
->Branch-aware chain traversal so cases where one moderator has multiple replacements are fully explored.
->Loop detection for chains that return to the originating moderator.
->Severity classification based on loop presence and dependency chain length.

Diagnostic output with:
->Current block
->Next block
->Moderator
->Dependency chain
->Dependency count
->Severity level
->Loop flag
->Branching Dependency Handling
This implementation includes branch-aware dependency tracing so that when a moderator has multiple possible replacements, all dependency branches are explored. This prevents missing loops that may exist on alternate replacement paths.

Tests
New automated tests were added for:

->Simple loop case: verifies direct moderator dependency loops are detected.
->Non-loop dependency case: ensures valid dependency chains are not incorrectly flagged.
->Branching dependency case: verifies multiple replacement branches are explored and loops are detected when present.
->Test imports were updated where necessary.

Related Issue
Closes #3307

Notes
This implementation follows the existing Scheduling Diagnostics workflow used in the project for identifying scheduling issues.

Severity is currently heuristic-based:

->High severity for detected loops.
->Lower severity levels based on dependency chain length.
->Type of Change
->New feature

Testing
->Existing tests pass: not verified locally.
->New tests added: yes.

AI Disclosure
AI assistance (Copilot) was used to help understand the issue requirements and review the existing scheduling-check workflow. The implementation, testing, and final changes were completed by me, and all code was reviewed before committing.

Checklist
->Self-reviewed the code.
->Updated documentation where applicable: not required for this change.
->No new editor diagnostics were introduced in changed files.

### DIFF
--- a/esp/esp/random/tests.py
+++ b/esp/esp/random/tests.py
@@ -1,0 +1,126 @@
+"""
+Tests for esp.random.views
+Source: esp/esp/random/views.py
+
+Tests good_random_class(), main view, and ajax view.
+"""
+import json
+from unittest.mock import patch, MagicMock
+
+from django.db.models import Q
+from django.test import RequestFactory
+
+from esp.tests.util import CacheFlushTestCase as TestCase
+from esp.random.views import good_random_class, main, ajax
+
+
+class GoodRandomClassTest(TestCase):
+    """Tests for good_random_class() filtering logic"""
+
+    @patch('esp.random.views.Tag')
+    @patch('esp.random.views.ClassSubject')
+    def test_no_constraints(self, mock_classsubject, mock_tag):
+
+        mock_tag.getTag.return_value = '{}'
+        mock_cls = MagicMock()
+        mock_classsubject.objects.random_class.return_value = mock_cls
+
+        result = good_random_class()
+
+        expected_q = Q()
+        mock_classsubject.objects.random_class.assert_called_once_with(expected_q)
+        self.assertEqual(result, mock_cls)
+
+    @patch('esp.random.views.Tag')
+    @patch('esp.random.views.ClassSubject')
+    def test_bad_program_name_excluded(self, mock_classsubject, mock_tag):
+
+        mock_tag.getTag.return_value = json.dumps({
+            'bad_program_names': ['TestProgram']
+        })
+        mock_cls = MagicMock()
+        mock_classsubject.objects.random_class.return_value = mock_cls
+
+        result = good_random_class()
+
+        expected_q = Q() & ~Q(parent_program__name__icontains='TestProgram')
+        mock_classsubject.objects.random_class.assert_called_once_with(expected_q)
+        self.assertEqual(result, mock_cls)
+
+    @patch('esp.random.views.Tag')
+    @patch('esp.random.views.ClassSubject')
+    def test_bad_title_excluded(self, mock_classsubject, mock_tag):
+        """Classes with bad titles should be excluded."""
+        mock_tag.getTag.return_value = json.dumps({
+            'bad_titles': ['Lunch Period']
+        })
+        mock_cls = MagicMock()
+        mock_classsubject.objects.random_class.return_value = mock_cls
+
+        result = good_random_class()
+
+        expected_q = Q() & ~Q(title__iexact='Lunch Period')
+        mock_classsubject.objects.random_class.assert_called_once_with(expected_q)
+        self.assertEqual(result, mock_cls)
+
+    @patch('esp.random.views.Tag')
+    @patch('esp.random.views.ClassSubject')
+    def test_multiple_constraints_combined(self, mock_classsubject, mock_tag):
+
+        mock_tag.getTag.return_value = json.dumps({
+            'bad_program_names': ['TestProgram'],
+            'bad_titles': ['Lunch Period']
+        })
+        mock_cls = MagicMock()
+        mock_classsubject.objects.random_class.return_value = mock_cls
+
+        result = good_random_class()
+
+        expected_q = Q() & ~Q(parent_program__name__icontains='TestProgram') & ~Q(title__iexact='Lunch Period')
+        mock_classsubject.objects.random_class.assert_called_once_with(expected_q)
+        self.assertEqual(result, mock_cls)
+
+
+class MainViewTest(TestCase):
+    """Tests for main() view."""
+
+    def setUp(self):
+        super().setUp()
+        self.factory = RequestFactory()
+
+    @patch('esp.random.views.good_random_class')
+    def test_main_returns_200(self, mock_good_random_class):
+
+        mock_cls = MagicMock()
+        mock_good_random_class.return_value = mock_cls
+
+        request = self.factory.get('/random/')
+        response = main(request)
+
+        self.assertEqual(response.status_code, 200)
+
+
+class AjaxViewTest(TestCase):
+    """Tests for ajax() view."""
+
+    def setUp(self):
+        super().setUp()
+        self.factory = RequestFactory()
+
+    @patch('esp.random.views.good_random_class')
+    def test_ajax_returns_json(self, mock_good_random_class):
+
+        mock_cls = MagicMock()
+        mock_cls.title = 'Test Class'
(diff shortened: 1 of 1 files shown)


esp/esp/program/modules/handlers/schedulingcheckmodule.py
```diff
@@ -705,3 +707,158 @@ def unavailable_moderators(self):
...
+
+    def _consecutive_timeslot_pairs(self):
+        pairs = []
+        timeslots = sorted(self.p.getTimeSlotList(), key=lambda ts: ts.start)
+        for i in range(len(timeslots) - 1):
+            current_slot = timeslots[i]
+            next_slot = timeslots[i + 1]
+            if current_slot.start.date() == next_slot.start.date():
+                pairs.append((current_slot, next_slot))
+        return pairs
```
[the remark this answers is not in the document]
hwatheod: @Boopana-M Please also make the other change that Copilot suggested, to check that contiguous timeslots are within a certain tolerance, to avoid large gaps.  The relevant tag is `timeblock_contiguous_tolerance`.  You might also want to use `Event.group_contiguous`. For an example, you can find these being used elsewhere in this file.

esp/esp/program/modules/handlers/schedulingcheckmodule.py
```diff
@@ -705,3 +707,158 @@ def unavailable_moderators(self):
...
+                elif candidate not in chain:
+                    stack.append((start_moderator, chain + [candidate]))
+                    extended = True
+                else:
+                    # Internal cycle not returning to the start moderator.
+                    chains.append((chain + [candidate], False))
+                    extended = True
+
+            if not extended:
+                chains.append((chain, False))
```
hwatheod: This code looks unnecessary. The only way `extended` could be `False` is if there were no `next_candidates` at all, but you already handled that case earlier.

esp/esp/program/modules/handlers/schedulingcheckmodule.py
```diff
@@ -705,3 +707,158 @@ def unavailable_moderators(self):
...
+            return 'High'
+        if chain_length >= 5:
+            return 'High'
+        if chain_length >= 3:
+            return 'Medium'
+        return 'Low'
+
+    def _trace_dependency_chains(self, start_moderator, first_dependency, dependency_graph):
+        """Trace all dependency branches from one starting dependency edge."""
+        stack = [(start_moderator, [start_moderator, first_dependency])]
```
hwatheod, line 760: `start_moderator` never changes throughout this entire function and it has the same value each time you push it onto the stack. So you don't need to track it in the stack.

esp/esp/program/modules/handlers/schedulingcheckmodule.py
Boopana-M: @hwatheod ,thank you  for pointing that out. You’re right, that block was unnecessary. I removed the redundant if not extended path and cleaned up the now-unused extended variable, while keeping behavior unchanged.

esp/esp/program/modules/handlers/schedulingcheckmodule.py
```diff
@@ -705,3 +707,158 @@ def unavailable_moderators(self):
...
+                            "Next Block": next_slot,
+                            self.p.getModeratorTitle().capitalize(): moderator,
+                            "Dependency Chain": chain_str,
+                            "Dependency Count": dependency_count,
+                            "Severity": self._dependency_severity(dependency_count, has_loop),
+                            "Loop": "Yes" if has_loop else "No",
+                        })
+
+                reported_starts.add(moderator)
+
```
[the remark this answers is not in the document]
hwatheod: @Boopana-M Copilot is right here. With the current logic, it seems like these lines are unnecessary.

esp/esp/program/modules/tests/schedulingcheckmodule.py
hwatheod, line 1: Can you add additional tests for some longer loops and chains?  Right now you only have tests for loops and chains of length 2.

Boopana-M: @hwatheod I’ll try adding tests covering longer chains and loops (length 3+), including both loop and non-loop paths now

esp/esp/program/modules/handlers/schedulingcheckmodule.py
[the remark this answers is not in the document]
Boopana-M: @hwatheod  I removed the unnecessary reported_starts tracking since each moderator is already iterated once from [dependency_graph.keys()], so the extra guard was redundant.
thank you

esp/esp/program/modules/handlers/schedulingcheckmodule.py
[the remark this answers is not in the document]
Boopana-M: @hwatheod I’ve implemented this by reading timeblock_contiguous_tolerance and using [Event.group_contiguous(...)] so dependency checks only run across contiguous time blocks and avoid linking across large gaps , thank you for the suggestion !

Boopana-M:
@hwatheod
I have implemented all the pending suggestions:

• Updated the consecutive-slot logic to use contiguous grouping with timeblock_contiguous_tolerance
• Optimized classroom checks to avoid the N+1 query by using ResourceType.get_or_create and res_type_id
• Removed redundant reported_starts and the extended / if-not-extended paths
• Added new tests for longer dependency chain and longer dependency loop scenarios (length ≥ 3)

Please let me know if any further changes are required. I’d be happy to update the PR accordingly.
thank you

hwatheod (approved):
Thank you for doing this!

Boopana-M:
You're welcome! Glad I could help.

PR #5084, opened 2026-03-14
Unit test admissions dashboard

### LINKED ISSUE #5067: [Bug]: Add unit tests for AdmissionsDashboard selective admissions logic (0% coverage)
### Description

`admissionsdashboard.py` — the teacher/admin Junction admissions interface —
has 0% test coverage. It handles complex role-based filtering, JSON mutation
of application statuses, and admin-only pathways.

### Key Risky Paths
- `apps()` JSON endpoint: uses `tl != 'manage'` to decide whether to filter
  by `admin_status=APPROVED`. No test verifies teacher vs. admin sees
  different data.
- `update_apps()`: silently skips `StudentClassApp` rows where the user
  lacks permission. No test confirms an unauthorized write attempt is
  correctly dropped rather than partially applied.
- `app()`: swallows `StudentClassApp.DoesNotExist` with a silent core
  redirect. No test checks the correct HTTP response code or redirect target.

### Impact
HIGH — controls selective admissions data for Junction-style programs.
A regression here could expose private applications or corrupt admit status.

utkarshgupta2804:
## Description

This PR adds unit tests for the AdmissionsDashboard module to improve coverage of critical admissions dashboard logic.
The tests focus on validating role-based filtering, safe handling of malformed input, and ensuring that unauthorized updates are ignored rather than applied.

Key scenarios covered:
1. Teachers only see applications with admin_status=APPROVED
2. Admin users bypass the approval filter and can view all applications
3. Missing StudentClassApp objects correctly redirect via goToCore
4. Invalid JSON in update_apps returns a structured error response
5. Unauthorized updates in update_apps are safely ignored

These tests help ensure that admissions data visibility and update behavior remain correct and secure.

## Related Issue
Closes #5067 

## Testing

- [x] Existing tests pass
- [ ] Manually verified

## AI Disclosure
Understood the code by reading and test is written by Claude.

## Checklist

- [ ] Self-reviewed the code


pushkarkumar06 (changes requested):
hii @utkarshgupta2804 ,  I went through the PR carefully and compared the new tests with the actual module logic. The direction of the PR is good and adding tests here definitely makes sense, but I noticed a few issues that should probably be addressed before merging.

**1. Test framework mismatch**

The new test files appear to be written in pytest style (function-based tests and fixtures such as `rf` and `mocker`). However, the repository currently runs tests using Django’s default unittest runner via `manage.py test`.

Because of this, these tests may not actually be discovered or executed in CI. If that is the case, the intended coverage increase would not really happen.

It would be safer to either:

* convert the tests to the existing `TestCase` style used elsewhere in the repository, or
* explicitly configure pytest in CI if the intention is to move to pytest.

**2. Parsing tests replicate implementation logic**

In `test_lotteryfrontendmodule.py`, the parsing test appears to reimplement parts of the parsing logic instead of calling the actual `lottery_execute` handler with a request object.

This means the test could still pass even if the production implementation changes or breaks.

It might be better to construct a request with `request.POST` values and call `lottery_execute` directly so the test verifies the real control flow.

**3. POST key parsing edge cases**

The new `key.startswith("lottery_")` with slicing (`key[len("lottery_"):]`) is definitely an improvement over the previous split-based logic. However, a few edge cases may still be worth considering:

* keys like `lottery_` (empty suffix)
* malformed keys such as `lottery__` or `lottery__x`
* POST keys containing `lottery_` but not actually intended as lottery options

It may be safer to ignore empty suffixes or validate against known option names before passing them into the controller.

**4. Numeric input validation**

The update to `is_float()` using `math.isfinite()` is a good improvement because it avoids accepting `NaN` and infinite values.

However, rejected numeric-like inputs may still remain as raw strings and get forwarded into the options dictionary. If those values are later used in numeric operations, that could still lead to runtime errors.

It might be worth either rejecting invalid numeric values earlier or validating option types before constructing the controller.

**5. Authorization test reliability**

Some of the admissions dashboard tests rely heavily on mocked permission checks. Because of this, the tests could potentially pass even if the authorization logic changes.

Using explicit role setups (teacher/admin users) instead of relying only on mocks would make the tests a bit more reliable.

---

Overall this PR is a good step toward improving coverage for the admissions dashboard and lottery handling logic. With the test framework alignment and a few of the edge cases addressed, the tests will provide much stronger regression protection.

For now I think it would be better to address the points above before merging.

utkarshgupta2804:
Okay, @pushkarkumar06 I will make these changes and raise a pr.

PR #5138, opened 2026-03-16
Fix: improve StudentRegistration admin search performance

### LINKED ISSUE #1541: Searching for StudentRegistrations in the admin panel is sloooooow
Reported by @mgersh; this was apparently an issue at Helpdesk at MIT Splash.  Unclear if we can do much to fix it, but modifying the search fields might help.

Sameer0297:
Closes #1541

**Summary**
Improved search performance in the Django admin panel for `StudentRegistration` and `StudentSubjectInterest` by replacing broad `CONTAINS` searches with indexed `STARTSWITH` (`^`) lookups and adding `list_select_related` to eliminate N+1 queries.

**Root Cause**
`default_user_search()` returns unqualified field names which Django translates to `ILIKE %query%` — a full table scan with no index usage. On large programs with thousands of `StudentRegistration` rows this causes the slow search reported at MIT Splash.

**Changes**

`program/admin.py`
- Replaced `default_user_search()` with explicit `^` prefixed fields — `STARTSWITH` uses the database index
- Removed deep relational fields like `section__parent_class__title` that caused expensive multi-table JOINs with `ILIKE`
- Added `list_select_related` to prefetch `user`, `section`, and `section__parent_class` — eliminates N+1 queries on the list page
- Same fix applied to `StudentSubjectInterestAdmin`

**Testing**
- Verified both `StudentRegistrationAdmin` and `StudentSubjectInterestAdmin` search fields updated correctly
- Existing filters, actions, and display fields preserved


esp/esp/program/admin.py
```diff
@@ -251,16 +251,38 @@ def renew_student_registrations(modeladmin, request, queryset):
...
+
+    def get_search_results(self, request, queryset, search_term):
+        if search_term and not search_term.isdigit():
+            original = self.search_fields
+            self.search_fields = tuple(f for f in original if not f.startswith('='))
+            try:
+                return super().get_search_results(request, queryset, search_term)
+            finally:
+                self.search_fields = original
+        return super().get_search_results(request, queryset, search_term)
```
willgearty, line 266: This method is identical in both admins here (and needed in `SectionAdmin` too). Suggest extracting a mixin to avoid three copies drifting apart.

esp/esp/program/admin.py
```diff
@@ -251,16 +251,38 @@ def renew_student_registrations(modeladmin, request, queryset):
 class StudentRegistrationAdmin(admin.ModelAdmin):
     list_display = ('id', 'section', 'user', 'relationship', 'start_date', 'end_date',)
     actions = [ expire_student_registrations, renew_student_registrations ]
-    search_fields = default_user_search() + ['id', 'section__id', 'section__parent_class__title', 'section__parent_class__id']
+    search_fields = ['=id', '=section__id', '^user__username', '^user__last_name', '^user__email', '=user__id']
```
willgearty, line 254: Why did you remove class title as a search field? Copilot flagged this and you just disregarded it.

esp/esp/program/admin.py
willgearty, line 297: Looks like this class also needs the `get_search_results` fix because of the `=` search modifiers.

willgearty (changes requested):
Thanks for this contribution, just a few comments that should be addressed.

willgearty (approved):

PR #5150, opened 2026-03-17
Fix: correct Content-Type header in ajax_autocomplete view

### LINKED ISSUE #5144: [Bug]: Invalid Content-Type header in ajax_autocomplete view
### Description

In esp/esp/ajax_select/views.py, the ajax_autocomplete view returns JSON data using json.dumps() but the Content-Type header is set to 'javascript/javascript' which is not a valid MIME type It should be 'application/json' Looks like an old typo that was never caught.

### Steps to Reproduce

1. Open esp/esp/ajax_select/views.py
2. Scroll to the bottom of the ajax_autocomplete function
3. Check the content_type value in the return statement

### Expected Behavior

Content-Type should be 'application/json' since the response 
body is JSON data built with json.dumps()

### Actual Behavior

Content-Type is set to 'javascript/javascript' which is not 
a valid MIME type

### Screenshots

_No response_

### Operating System

Windows

### Browser

Chrome

### Additional Context

This looks like an old typo from when the code was originally written. 
The fix is a one word change in the return statement.

An-Orchids-whisper:
## Description
The ajax_autocomplete view in esp/esp/ajax_select/views.py was returning 
JSON data but the Content-Type header was set to 'javascript/javascript' 
which is not a valid MIME type. Changed it to 'application/json'.

## Related Issue
Closes #5144

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing
- [x] Existing tests pass
- [x] Manually verified

## AI Disclosure
None

## Checklist
- [x] Self-reviewed the code


RITVIKKAMASETTY (commented):
Hi @An-Orchids-whishper ,
The `content_type` fix is correct — `javascript/javascript` is invalid, `application/json` is right.

Looking at the full file, a few more things worth raising:

**1. Use `JsonResponse` instead of `HttpResponse` + manual `json.dumps`**
```python
from django.http import JsonResponse
return JsonResponse({'result': output2})
```
`JsonResponse` sets `application/json` automatically and removes the need for `json.dumps` and the import. Idiomatic Django.

**2. `autocomplete_wrapper` uses `__func__.__code__.co_varnames` — fragile**
```python
if 'allow_non_staff' in function.__func__.__code__.co_varnames:
```
This inspects bytecode internals to check if a function accepts a parameter. It will silently fail if the function is a lambda, a built-in, or wrapped with a decorator. Use a sentinel attribute instead:
```python
if getattr(function.__func__, 'allow_non_staff', False):
```
And set `allow_non_staff = True` explicitly on functions that should allow it.

**3. `__import__` for dynamic model loading — unsafe**
```python
Model = getattr(__import__(model_module, ...), model_name)
```
`model_module` comes directly from `GET` params with no whitelist validation. An attacker can pass any importable module path. This is a pre-existing issue but worth flagging — at minimum, validate `model_module` against an allowlist before importing.

**4. 400 response uses `HttpResponse('Malformed Input')` with no `content_type`**
Inconsistent with the JSON response. Should be:
```python
return JsonResponse({'error': 'Malformed Input'}, status=400)
```

**5. No test added**
A simple test asserting `response['Content-Type'] == 'application/json'` would prevent re-regression.

 Fix is correct. Points 2 and 3 are pre-existing but worth a follow-up issue. Points 1 and 4 are easy cleanup while you're in the file.

An-Orchids-whisper:
Thanks for the review! Working on points 1 and 4 now - will update 
the PR with JsonResponse and fix the 400 response as well.

Points 2 and 3 are pre-existing issues, happy to open separate 
issues for those if that would help.

An-Orchids-whisper:
Updated the PR with points 1 and 4 from your review. Replaced 
HttpResponse + json.dumps with JsonResponse, and fixed the 400 
error response as well.

esp/esp/db/views.py
```diff
@@ -1,6 +1,6 @@
 from django.contrib.auth.decorators import login_required, user_passes_test
 from django.core import serializers
-from django.http import HttpResponse
+from django.http import HttpResponse, JsonResponse
```
RITVIKKAMASETTY: if HttpResponse not used
please remove it

An-Orchids-whisper: done

RITVIKKAMASETTY:
lgtm

willgearty (approved):
Looks good, thanks for cleaning this up!

PR #5162, opened 2026-03-17
Fix: replace fragile co_varnames check with getattr in autocomplete_wrapper

### LINKED ISSUE #5159: [Bug]: Fragile use of __func__.__code__.co_varnames in autocomplete_wrapper
### Description

In esp/esp/ajax_select/views.py, the autocomplete_wrapper function checks if a function accepts allow_non_staff by inspecting bytecode internals which can silently fail in certain cases.

### Steps to Reproduce

1. Open esp/esp/ajax_select/views.py
2. Look at the autocomplete_wrapper function
3. Check the line that inspects __func__.__code__.co_varnames

### Expected Behavior

Should use getattr(function.__func__, 'allow_non_staff', False) to safely check if the attribute exists

### Actual Behavior

Currently uses function.__func__.__code__.co_varnames which will silently fail if the function is a lambda, built-in, or wrapped with a decorator

### Screenshots

_No response_

### Operating System

Windows

### Browser

Chrome

### Additional Context

This was flagged during review of PR #5150. Pre-existing issue in the codebase.

An-Orchids-whisper:
## Description
`autocomplete_wrapper` decided whether a non-staff user may call a model's
`ajax_autocomplete` by looking for `allow_non_staff` in
`__func__.__code__.co_varnames`. That reads bytecode internals: `co_varnames`
lists local variables alongside parameters, and it describes the outermost
function, so a decorated autocomplete reports the decorator's
`(*args, **kwargs)` rather than the parameters it forwards.

The opt-in is now an explicit marker. `@allow_non_staff_autocomplete` sets an
attribute on the function and the wrapper reads it with `getattr`, so the grant
is greppable and independent of how the callable is spelled — it survives
`@classmethod` binding and `functools.wraps`, and works for callables with no
introspectable signature. `inspect.signature` is still used, but only to decide
whether the callee accepts `request`.

Declaring an `allow_non_staff` **parameter** no longer grants access. That
distinction matters: `ajax_func` is caller-supplied, and
`ESPUser.ajax_autocomplete_student` and its siblings take `**kwargs`, so any
inference from the signature risks exposing student and teacher records to any
logged-in user.

**Files changed**
- `esp/esp/db/autocomplete.py` (new) — the `@allow_non_staff_autocomplete` decorator
- `esp/esp/db/views.py` — gate reads the marker
- `esp/esp/users/models/__init__.py` — marks `K12School.ajax_autocomplete`, the one opted-in autocomplete
- `esp/esp/db/test_ajax_autocomplete.py` — tests

**Upgrade note:** a site-local model that opted in by declaring an
`allow_non_staff` parameter must add `@allow_non_staff_autocomplete`, or it will
return `[]` for non-staff users.

## Related Issue
Closes #5015

Note: the issue proposes `getattr(function.__func__, 'allow_non_staff', False)`.
Nothing in the tree sets that attribute, so reading it alone would have denied
every non-staff autocomplete, including `K12School`. This PR adds the decorator
that makes the attribute real, which is the contract the issue was reaching for.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing
- [x] Existing tests pass
- [x] Manually verified

Adds coverage for the gating decision: the marker through a plain function, a
bound classmethod, a `functools.wraps` wrapper and an uninspectable callable;
an `allow_non_staff` parameter without the marker; `**kwargs` without the
marker; and that `request` reaches only callees that declare it. At the view
level, a non-staff user gets nothing from the `**kwargs`-only `ESPUser`
autocompletes, paired with a staff control proving those records are otherwise
returnable, plus unpatched coverage that `K12School` stays reachable.

Verified that removing `@allow_non_staff_autocomplete` from `K12School` fails
the suite, and that restoring the parameter-based check fails it too.

## AI Disclosure
The current implementation and tests were revised by a maintainer using Claude Code.

## Checklist
- [x] Self-reviewed the code

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/esp/esp/db/autocomplete.py
+++ b/esp/esp/db/autocomplete.py
@@ -1,0 +1,15 @@
+"""Opt-in marker for autocompletes that non-staff users are allowed to call."""
+
+
+def allow_non_staff_autocomplete(func):
+    """Mark an ajax_autocomplete as safe to expose to non-staff users.
+
+    Apply it directly to the function, inside @classmethod:
+
+        @classmethod
+        @allow_non_staff_autocomplete
+        def ajax_autocomplete(cls, data, **kwargs):
+            ...
+    """
+    func.allow_non_staff = True
+    return func
--- a/esp/esp/db/test_ajax_autocomplete.py
+++ b/esp/esp/db/test_ajax_autocomplete.py
@@ -8,6 +8,7 @@ from django.contrib.auth.models import Group
 from django.test import SimpleTestCase, TestCase
 from django.test.client import Client
 
+from esp.db.autocomplete import allow_non_staff_autocomplete
 from esp.db.views import autocomplete_wrapper
 from esp.users.models import ESPUser, K12School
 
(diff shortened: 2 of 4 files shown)


willgearty (approved):
OK, that should do it. Thanks for your original contribution, and apologies for the delay in reviewing it!

PR #5212, opened 2026-03-18
fix(usersearch): prefer exact numeric ID matches

### LINKED ISSUE #5004: [Bug]: Infinite loop risk in onsiteattendance.times_attending_class() when end_time <= start_time

moksha-hub:
## Summary
This fixes the admin /manage/usersearch flow for numeric user-ID searches.

When the search term is numeric, the current implementation mixes exact ID matching with fuzzy name matching. In a large test database, that can turn a direct user-ID lookup into a multi-result render path instead of the expected redirect. If one of those rendered users has no last program, users/userview_search.html also tries to access dates on None.

## Changes
- prefer exact-match results for numeric user-ID searches before falling back to fuzzy name matching
- keep the existing username/name overlap behavior for non-numeric searches
- guard the multi-result search template when a user has no last program
- add a regression test covering numeric ID search with a fuzzy first-name collision

## Testing
- python -m py_compile esp/esp/program/views.py esp/esp/program/tests.py
Fixes #5004


moksha-hub:
@oval please have a review as it directs to another open pr

PR #5225, opened 2026-03-19
Fix race condition in ClassSection.preregister_student causing over-e…

### LINKED ISSUE #5224: [Bug]: Race condition in `ClassSection.preregister_student` lets classes overfill during concurrent registrations
### Description

There’s a race condition in `ClassSection.preregister_student()` where the capacity check (`isFull()`) happens before creating the `StudentRegistration`, but there’s no locking in between.

Because of this, when multiple users try to register at the same time, they can all pass the capacity check and get enrolled, even if the class is about to be full.

This happens due to PostgreSQL’s default `READ COMMITTED` isolation, each request sees the same “not full yet” state before any of them commit, so they all proceed.

### Steps to Reproduce

1.Create a class section with a fixed capacity (for example, 30 students)
2.Make sure it’s almost full (e.g., 29/30 students enrolled)
3.Trigger multiple registrations at the same time (2+ users clicking enroll together)
4.Check the final number of enrolled students

### Expected Behavior

Once the class reaches its capacity, any additional registration attempts should fail. The number of enrolled students should never go beyond the limit.

### Actual Behavior

Multiple requests pass the `isFull()` check and all get enrolled, so the class ends up over capacity (e.g., 31/30 students).

### Screenshots

_No response_

### Operating System

_No response_

### Browser

_No response_

### Additional Context

_No response_

Anas-Blue:
## Description

Fixes a race condition in `ClassSection.preregister_student` where multiple users could get enrolled beyond capacity under concurrent requests. The capacity check was happening before the insert without any locking, which allowed over-enrollment.

## Related Issue

Closes #5224 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

## AI Disclosure

Used AI to help reason about the race condition and validate the approach.

## Checklist

- [x] Self-reviewed the code


willgearty (approved):
Looks good, thanks!

PR #5253, opened 2026-03-20
Feature/teacher calendar 5249

### LINKED ISSUE #5249: [Feature]: Calendar view for teacher event signup
### Problem Statement

Right now, signing up for a teacher event (training, interview, etc) involves selecting a choice from a dropdown. But this is an unintuitive signup interface for events that have specific times/dates and lengths.

### Proposed Solution

I had started something a little more visually appealing in https://github.com/learning-unlimited/ESP-Website/pull/2709. But ultimately, I think we want to fully replace the existing interface (/teach/.../.../event_signup) with an elegant calendar-based signup.

Other things to note:

- It's always bothered me that past teacher events don't show up on this page. All of the events should show up, and if someone was signed up for one in the past, it should still show that.
- Teacher interviews should only permit a single teacher to sign up for each slot. Functionality like calendly or Google Calendar appointments would be good for this.

### Alternatives Considered

_No response_

### Additional Context

_No response_

Satya-0023:
### Summary
Teachers currently use a legacy dropdown system for event and training signups, which lacks spatial awareness and becomes clunky as event lists grow. 

This PR introduces a responsive, interactive **FullCalendar.js** interface for teacher event signups. It replaces the legacy dropdowns with a visual grid while maintaining 100% compatibility with the existing `UserAvailability` model and POST submission flow.

Closes #5249  
References #2709

### Screenshots
image
image

### Architecture & Files Changed

| File | Change |
| --- | --- |
| `esp/esp/program/modules/handlers/teachereventsmodule.py` | New `calendar_data` JSON API returning explicitly authorized (401/403) event feeds with dynamic `is_teacher_type=True` mapping. |
| `esp/templates/program/modules/teachereventsmodule/event_signup.html` | Replaced dropdown UI with FullCalendar grid + JS-driven hidden Django form submission. |
| `esp/esp/program/modules/tests/test_teacherevents.py` | Added test cases covering API auth, payload structure, and edge cases. |
| `esp/esp/program/modules/forms/teacherreg.py` | Replaced naive datetime usage with `timezone.now()`. |

### Design Decisions & Fixes

- **Form Mapping:** JavaScript maps `extendedProps.event_type_id` to Django form fields (`id_event_type_<id>`) ensuring correct POST submission.
- **Strict API Auth:** Implemented explicit JSON responses (401/403) instead of relying on `@needs_teacher`, preventing data leakage.
- **Robust URL Handling:** Used `{{ prog.getUrlBase }}` to avoid routing issues.
- **HTML Validation:** Wrapped hidden `{{ form }}` inside a `<table>` to maintain valid DOM structure.

### Testing

- Verified UI behavior across:
  - Available (Blue)
  - Mine (Green)
  - Full (Red)
  - Past (Grey)

- Added unit tests for:
  - Authentication handling
  - JSON response structure
  - Event state logic

### AI Disclosure
AI tools were used to assist with development, testing, and code review. All design decisions and final implementations were manually reviewed and validated.

### Checklist
- [x] Self-reviewed code
- [x] Tests added and passing
- [x] Review feedback addressed


RITVIKKAMASETTY (commented):
**1. Two `<form>` tags rendered simultaneously — one will submit, one is hidden dead code**
The old form (`{{ form }}` with Submit button) and the new hidden form (`id="signup-form"` with Confirm Selection) both exist in the template. The old one has no `{% csrf_token %}` and will fail on POST. Remove it entirely — the hidden form is the one wired to the JS.

**2. `returnlink.html` included twice**
```html
{% include "program/modules/teacherregcore/returnlink.html" %}
```
Appears once inside `#program_form` and once outside. Remove the duplicate.

**3. `datetime.now()` is naive — should be `datetime.now(tz=timezone.utc)` or `timezone.now()`**
```python
is_past = event.start < datetime.now()
```
If `event.start` is timezone-aware (which Django model DateTimeFields are by default when `USE_TZ=True`), comparing against a naive `datetime.now()` will raise a `TypeError` at runtime. Use:
```python
from django.utils import timezone
is_past = event.start < timezone.now()
```

**4. `calendar_data` endpoint has no auth decorator**
`@aux_call` alone does not enforce login or teacher status. The existing `event_signup` view uses `@needs_teacher` and `@meets_deadline('/Events')`. The new JSON endpoint leaks event data (times, descriptions, signup counts) to unauthenticated users. Add `@needs_teacher` at minimum.

**5. `events: 'calendar_data'` is a relative URL — fragile**
FullCalendar will resolve this relative to the current page URL, which may break depending on routing. Use an absolute path:
```javascript
events: '/teach/{{ one }}/{{ two }}/calendar_data'
```
Or pass it as a template variable from the view.

**6. Interview "full" logic is incomplete**
```python
is_full = category == 'interview' and not is_mine and entries.count() > 0
```
This marks a slot as full if anyone has signed up, which is correct for single-signup interviews. But training slots are multi-signup — they will never be marked full even if there's a capacity limit. If capacity limits exist in the model, this should respect them. If not, at least document the assumption.

**7. No tests added**
The PR description says "Verified UI behavior...in local environment" — that's not a substitute for automated tests. At minimum, a test for `calendar_data` asserting correct JSON structure, auth enforcement, and past/full/mine status logic should be added.

**8. FullCalendar loaded from CDN in `{% block content %}` not `{% block stylesheets %}`**
CSS and JS CDN links are injected inside `{% block content %}`. They should be in `{% block stylesheets %}` and `{% block scripts %}` respectively so the base template can control load order and avoid render-blocking.

the calendar UX is a clear improvement. The duplicate form, missing auth on the endpoint, naive datetime comparison, and missing tests need to be fixed before merge.

Satya-0023:
Hi @RITVIKKAMASETTY, thanks for the detailed and helpful review!

I’ve addressed all the points you raised and cleaned up the branch by rebasing it onto the latest main to remove unrelated commits.

Summary of updates:

1. Security:
   - Added @needs_teacher and @meets_deadline decorators to the JSON endpoint

2. Timezone Safety:
   - Replaced datetime.now() with timezone.now() to ensure correct timezone handling

3. Template Cleanup:
   - Moved CSS to {% block stylesheets %} and JS to {% block xtrajs %}
   - Removed duplicate form and redundant returnlink.html include
   - Updated calendar data URL handling for better reliability

4. Testing:
   - Added test coverage in esp/esp/program/modules/tests/test_teacherevents.py
   - Covers authentication, JSON structure, and event state logic (mine/full/past)

5. Clean History:
   - Rebased and simplified commits to keep the PR focused on this feature

Please let me know if any further changes are needed!

Satya-0023:
Hi @kkbrum and @RITVIKKAMASETTY,

Thank you for the review and the automated Copilot feedback! I have pushed a new commit fix(copilot): Address automated code review feedback that fully resolves all 10 issues raised by the Copilot AI bot.

Summary of fixes:

**JSON Endpoint Auth**: Removed @needs_teacher and implemented explicit custom auth returning 401 Unauthorized and 403 Forbidden JSON responses to maintain strict API contracts. (Updated the Edge-case Unit Tests to match).

**Dynamic Template Mapping:** The frontend now correctly locates the exact id_event_type_id hidden form DOM elements using extendedProps.event_type_id so accurate POST data is submitted.

**HTML Validation:** Wrapped the {{ form }} inside a hidden table  to prevent invalid standalone HTML rendering.

**Button State Resiliency: The** Confirm button now properly enables on load if {% if form.initial %} data is already present.

**URL Robustness:** Replaced fragile relative URLs with {{ prog.getUrlBase }}calendar_data.

**Dynamic Type Filtering:** Rewrote the backend API query to dynamically loop hisfsfEventType.objects.filter(is_teacher_type=True) instead of hardcoding string checking.

**QSD Hook Reinstatement:** Re-added the missing render_inline_program_qsd hook above the calendar.
The calendar_data API and frontend are now highly resilient. Looking forward to any further thoughts!

kkbrum:
@Satya-0023  can you fix the failing tests?

Satya-0023:
Hi @kkbrum, thanks for the feedback!

I’ve rebased onto the latest main and fixed the failing tests, lint issues, and migration checks. Also corrected the program reference and improved robustness around event handling.

CI should be passing now — please let me know if anything else needs improvement.

RITVIKKAMASETTY:
> Hi @kkbrum, thanks for the feedback!
> 
> I’ve rebased onto the latest main and fixed the failing tests, lint issues, and migration checks. Also corrected the program reference and improved robustness around event handling.
> 
> CI should be passing now — please let me know if anything else needs improvement.

Hi @Satya-0023 
It was instructed to not tag mentors after addressing a review 
you can instead request rereview by using the button on top right part of the screen

kkbrum:
@Satya-0023 can you fix the failing test?

kkbrum (commented):
Do you know why I'm getting this error?

...django.request:224] WARNING: Not Found: /teach/Splash/2026_Fallcalendar_data

Satya-0023:
Fixed! The issue was that prog.getUrlBase returns the base URL without a trailing slash (e.g., Splash/2026_Fall), so calendar_data was being concatenated directly against it. I've added the missing / separator. The URL now correctly resolves to /teach/Splash/2026_Fall/calendar_data.

kkbrum (commented):
Thanks for this! A couple notes:

1. There is a colon after "Training" or "Interview" on the calendar, with nothing after the colon. On your screenshot, it sometimes shows some text after the colon, but that's not showing for me, even when there is a non-blank description of the event I create. I would expect the description of the event to be displayed after the colon, and for the colon to be removed entirely if there is no description.
2. Perhaps days with no events should be greyed out entirely, and if a whole week doesn't have events, there should be an arrow saying "next event: XXX" where XXX is the date of the next event, and pressing the arrow goes to that day. And perhaps also there should be an arrow to the previous event as well. I wouldn't really expect you to code this from scratch, but perhaps there are other versions of the calendar you are using that allow this? It seems like a common feature, so I'd think it would be an option. If not, it's okay, but then I might suggest adding a list of events to the right or left of the calendar, where if you press an event name/date in the list, the calendar jumps to that date. (This is probably even more coding work...) The main thing is that if the events are far apart, scrolling through the calendar to find the events, which may not even exist, is a pain. So there should be some way of knowing what you're looking for, either in a list, or with arrows on the calendar.

kkbrum (changes requested):

Satya-0023:
Thanks for the feedback! I've addressed both points:

1. Title colon issue — Fixed in the backend. The title now uses event.description as the primary detail (falling back to event.name), and the colon is omitted entirely when neither is set. So you'll see Interview: Afternoon slot if a description exists, or just Interview if not.

2. Navigation — Added two custom toolbar buttons: ← Prev Event and Next Event → that jump directly to the week containing the previous/next event. The calendar also auto-jumps to the nearest upcoming event on initial load, so teachers no longer need to scroll through empty weeks. The standard < > week arrows remain for manual browsing.

kkbrum (changes requested):
Thanks for these updates! The new event titles look great.

For the next/prev event buttons, the buttons look great, but when I click on them, the calendar does not jump anywhere. My calendar also does not jump to the first upcoming event on initial load as you describe.

Also, can you make this page more mobile-friendly? Perhaps if the page is too small, revert back to the old event sign up page or something similar? Or switch to the "day" calendar mode automatically for narrow screens and align the buttons better?

Image

Satya-0023:
Fixed both issues!

1. Navigation buttons not jumping / no auto-jump on load The root cause was a race condition: eventsSet fires immediately with an empty array (before the async fetch from the server completes), which was locking the initialNavDone flag too early. When the real events arrived, the callback was already skipped. Fixed by adding a guard to skip empty-array calls (events.length === 0). Also added setTimeout(..., 0) around gotoDate so it runs after FullCalendar's current render cycle completes.

2. Mobile-friendly layout On screens narrower than 768px, the calendar now: 
    Automatically switches to day view instead of week view
    Toolbar reorganises into two clean rows (← Prev Event / Next Event → on row 1, then < > today on row 2)
    Button text and padding shrink via CSS media query
    Screenshots below (desktop week view and mobile day view):
    
screenshot_desktop_1778604749027
screenshot_mobile_final_1778605053035

kkbrum (changes requested):
Thanks! These fixes are great. I did find one width that I'm not happy with. It's at about 768-979 pixels: 

Image

All the other widths look clean!

Satya-0023:
Ah, I see what you mean! The toolbar items were squishing together because they didn't have enough room to sit side-by-side on that specific tablet width.

I've pushed a minimal CSS fix targeting exactly that width:

1. Added white-space: nowrap to the custom buttons so the "Prev Event" text no longer awkwardly breaks into multiple lines.
2. Added flex-wrap: wrap specifically to the toolbar for widths up to 979px, so the buttons and title will cleanly stack when horizontal space gets tight instead of overlapping.

Kept everything else exactly the same. Let me know if that looks better!
tablet_layout_fix_verification_1778910181057

kkbrum (approved):
Looks great, thanks!

PR #5255, opened 2026-03-20
Fix FixIEMiddleware Crash on Django 2.2

### LINKED ISSUE #4236: [Bug]: FixIEMiddleware crashes on Django 2.2 due to removed HttpResponse.mimetype attribute

RITVIKKAMASETTY:
## Description

FixIEMiddleware currently assumes that `HttpResponse` instances have a `mimetype` attribute. That attribute was removed in Django 2.2, so for IE-style user agents the middleware calls `response.mimetype.lower()` and raises an `AttributeError`, returning a 500 instead of the expected page.

This PR replaces the use of `response.mimetype` with logic based on the standard `Content-Type` header and adds a small set of regression tests to cover this behavior.

Fixes #4236

## Changes

**Update FixIEMiddleware to use Content-Type**
- Stopped calling `getattr(response, "content_type", "")` / `response.mimetype.lower()` in [esp/esp/middleware/fixiemiddleware.py](cci:7://file:///Volumes/Untitled/abab/lu/ESP-Website/esp/esp/middleware/fixiemiddleware.py:0:0-0:0).
- Now reads `response.get('Content-Type')`, strips any parameters (e.g. `; charset=utf-8`), and compares the normalized MIME type (lowercased, without parameters) against the existing `safe_mime_types` tuple.
- If `Content-Type` is missing for some reason, it bails out and returns the response unchanged instead of raising.

**Add regression tests for FixIEMiddleware**
- Added [esp/esp/middleware/tests.py](cci:7://file:///Volumes/Untitled/abab/lu/ESP-Website/esp/esp/middleware/tests.py:0:0-0:0) with [test_ie_user_agent_does_not_crash_for_default_html](cci:1://file:///Volumes/Untitled/abab/lu/ESP-Website/esp/esp/middleware/tests.py:15:4-24:53) which verifies that `FixIEMiddleware.process_response` returns a 200 response and does not alter the content.
- Added [test_ie_user_agent_strips_vary_for_unsafe_content_type](cci:1://file:///Volumes/Untitled/abab/lu/ESP-Website/esp/esp/middleware/tests.py:26:4-37:84) which uses an IE-style User-Agent with an `application/json` response and an existing `Vary: User-Agent` header, verifying that the middleware removes `Vary` and sets `Pragma` and `Cache-Control` headers.

## Type of Change

- [x] Bug fix

## Testing

- [x] New tests added
- [x] Manually reviewed code changes against Django 2.2 spec

## AI Disclosure

This PR was developed with AI assistance.

## Checklist

- [x] Self-reviewed the code


RITVIKKAMASETTY:
fixed code according to copilot suggestion

Oval17 (commented):
Do we still need these changes 
@willgearty

willgearty:
Given that the usage of IE is ~0.1% of all users worldwide AND the original cited bug report (https://support.microsoft.com/kb/824847/en-us?spid=8722&sid=global) doesn't exist anymore, I don't even think we need FixIEMiddleware anymore.

Oval17:
We can close this and issue.

PR #5296, opened 2026-03-21
fix: inconsistent class removal by removing cannotRemove() check (learning-unlimited#1881)

### LINKED ISSUE #1881: Onsite class changes grid does not check cannotRemove when removing classes
I think there aren't really any cases where cannotRemove is relevant anyway (should we just remove it? #822?), but since it is checked by the fillslot/clearslot interface, the grid should probably be consistent with it.

VipulMore11:
## Description

This PR removes the `cannotRemove()` check from the `fillslot` and `clearslot` logic to resolve inconsistent class removal behavior between interfaces.

Previously, the onsite class changes grid allowed removing classes without restriction, while the `fillslot` / `clearslot` logic enforced `cannotRemove()`, leading to inconsistent behavior.

Since `cannotRemove()` was intended for schedule constraints that are not currently implemented or used, this change makes class removal behavior consistent and permissive across all interfaces.

---

## Related Issue

Closes #1881

---

## Type of Change

- [x] Bug fix
- [x] Refactor / cleanup

---

## Testing

- [x] Existing tests pass  
- [x] New tests added  
- [ ] Manually verified  

---

## AI Disclosure

This PR was created with assistance from AI tools (ChatGPT) for guidance on code changes and documentation.

---

## Checklist

- [x] Self-reviewed the code  
- [ ] Updated documentation (not needed)  


AnushkaPandit-21:
I see the motivation for consistency with the onsite grid, but I think the old code had that hasattr(request.user, "onsite_local") check as a deliberate escape hatch specifically for onsite admins. The intent was that cannotRemove() only blocked student-initiated removals, not admin ones. Removing it entirely means students can now remove classes that schedule constraints were trying to protect. Even if cannotRemove() isn't actively used right now, it might be safer to just always bypass it (return early) rather than delete the check entirely, in case someone adds schedule constraints later. Worth flagging to the maintainers.

VipulMore11:
Thank you @AnushkaPandit-21 for the feedback and clarification. That makes sense regarding the intended use of the onsite_local check as an admin bypass rather than removing cannotRemove() entirely.

Based on your suggestion, I will update the implementation to bypass cannotRemove() for onsite/admin users instead of removing the check completely, so future schedule constraints remain supported. I will push the updated changes shortly.

VipulMore11:
Hi @willgearty ,

I’ve addressed the Copilot review comments also @AnushkaPandit-21's valuable insights and updated the implementation accordingly:

- Materialized cleared section IDs before unregistering students
- Improved the AJAX clearslot test coverage
- Optimized section filtering logic in tests
- Updated logic to bypass cannotRemove() for onsite/admin users instead of removing it entirely

Ready for review!

AnushkaPandit-21:
@VipulMore11 Glad that was useful! The onsite_local escape hatch is one of those things that's easy to miss if you haven't dug into the onsite flow before. Looks good now.

VipulMore11:
Hey @willgearty , I resolved the conflicts. Is there anything else I need to do?

willgearty:
This no longer addresses the original problem. Closing in favor of a new PR.

PR #5297, opened 2026-03-21
scheduler: show same-classroom sections in section info panel (fixes #1404)

### LINKED ISSUE #1404: Which classes can this swap with?
@emannes wrote:

> When I pick up a class that is currently scheduled, AJAX should show me which classes in the same classroom I can swap it with.

I would add that it would be useful to know _any_ classes with which it can be swapped, although for many use cases the same classroom is more useful.

Hari20032005:
## Description

When a scheduled section is selected in the AJAX scheduler, the section info panel now shows a list of **other sections currently assigned to the same classroom**, with clickable links to select each one. This makes it easy to spot swap candidates without manually scanning the grid.

- Adds `getSameRoomSections(section)` to `Sections.js` — returns all other sections whose `room_id` matches the selected section's room assignment
- Updates `SectionInfoPanel.js` `getContent()` to render the same-room list below the existing Ctrl+Click swap hint (only shown when the section is scheduled)
- Timeslot labels are resolved from the matrix where available, falling back to raw IDs
- Fixes pre-existing test environment issues (`moderators` missing from fixtures, misaligned constructor args, missing global stubs) that caused the entire `SectionsSpec` suite to fail

## Related Issue

Closes #1404

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

Unit tests: `getSameRoomSections` describe block in `SectionsSpec.js` — 4 specs, all passing.

Manual verification via DevTools console on `/javascript_tests`:
```javascript
var s = new Sections(section_fixture(), {}, {}, teacher_fixture(), {}, schedule_assignment_fixture(), new FakeApiClient());
s.getSameRoomSections(s.getById(3));        // [] — alone in room-2
s.scheduleAssignments[2] = { room_id: "room-2", id: 2, timeslots: [3] };
s.getSameRoomSections(s.getById(3));        // [{…}] — finds section 2
` ``

## AI Disclosure

This pull request was developed with assistance from Claude (Anthropic) via Claude Code.

## Checklist

- [x] Self-reviewed the code

### CODE REVIEW

File: esp/public/media/scripts/ajaxschedulingmodule/spec/FakeMatrix.js [hunk 1]
```diff
@@ -1,7 +1,15 @@
+// Global configuration variables normally injected by Django templates.
+// These stubs allow the test suite to run without the production page context.
+var contiguous_tolerance = "0";
+var has_moderator_module = "False";
```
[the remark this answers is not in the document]
Hari20032005: Fixed in the latest commit — added var has_autoscheduler_frontend = "False"; to FakeMatrix.js alongside the other global stubs.

esp/public/media/scripts/ajaxschedulingmodule/spec/SectionsSpec.js
```diff
@@ -2,8 +2,8 @@ describe("SectionsSpec", function() {
     var sections, matrix;
 
     beforeEach(function() {
-        sections = new Sections(section_fixture(), {}, teacher_fixture(),
-            schedule_assignment_fixture(), new FakeApiClient());
+        sections = new Sections(section_fixture(), {}, {}, teacher_fixture(),
+            {}, schedule_assignment_fixture(), new FakeApiClient());
```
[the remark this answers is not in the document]
Hari20032005: Fixed in the latest commit — updated all 4 assertions to use room_id instead of room_name. The production code (scheduleSectionLocal) has always written room_id; the test assertions were using the wrong key. Note: the mapping approach suggested would have masked the pre-existing bug rather than fixing it.

esp/public/media/scripts/ajaxschedulingmodule/ESP/SectionInfoPanel.js
```diff
@@ -171,6 +171,38 @@ function SectionInfoPanel(el, sections, togglePanel, sectionCommentDialog) {
         }
         contentDiv.append($j('<br><div><b>Click on another section while holding down "Ctrl"/"Cmd" to swap it with this section</b>'));
 
+        // Show sections in the same classroom that can be swapped
+        if (this.sections.isScheduled(section)) {
+            var sameRoomSections = this.sections.getSameRoomSections(section);
+            var swapDiv = $j('<div>');
+            swapDiv.append('<br><b>Other sections in the same classroom:</b>');
```
kkbrum: Does this also only show sections which 1) are scheduled during a timeslot the teacher of the current class is also available during and 2) whose teacher is available during the current timeslot. Is that correct? This is the logic that I presume is necessary for classes to be "swappable". If this is not already the case, could you make it so?

Given this, I think this text should say "Other sections in the same classroom that can be swapped with this one:"

esp/public/media/scripts/ajaxschedulingmodule/ESP/SectionInfoPanel.js
```diff
@@ -171,6 +171,38 @@ function SectionInfoPanel(el, sections, togglePanel, sectionCommentDialog) {
         }
         contentDiv.append($j('<br><div><b>Click on another section while holding down "Ctrl"/"Cmd" to swap it with this section</b>'));
```
kkbrum, line 172: For the sections listed in the swappable category below, I think this "Ctrl"/"Cmd" functionality should also do the swapping action. Right now, it doesn't work if you click on a section in the list below (only if you click on the section in the grid or among the unscheduled list).

kkbrum (changes requested):

esp/public/media/scripts/ajaxschedulingmodule/ESP/SectionInfoPanel.js
Hari20032005: Yep, that's exactly it. Both directions have to clear — the current section's teachers need to be free at the other's timeslots, and the other's teachers need to be free at the current timeslots. getAvailableTimeslots runs with both sections in the ignore list so their current assignments don't block each other.

Updated the label and added two specs via generateFakeMatrix for the include/exclude cases. Both pass.

esp/public/media/scripts/ajaxschedulingmodule/ESP/SectionInfoPanel.js
Hari20032005: Added evt.stopPropagation() to the swap list click handler. Ctrl/Cmd+click on an <a href="#"> would otherwise open a new tab before JS could handle it.

kkbrum:
@Hari20032005 can you address Copilot’s new feedback? Note that you don’t necessarily have to do what it says, but you should explain why it’s not a valid critique if so.

kkbrum:
Closing due to inactivity. Feel free to re-assign yourself to the issue and reopen if you'd like to finish working on it.

PR #5307, opened 2026-03-21
Fix: loginbox FOUC (#5304)

### LINKED ISSUE #5304: [Bug]: Login box visibility flicker (FOUC) on initial page load
### Description

The login box visibility on the base template is currently controlled via client-side JavaScript (`user_classes.js`). Because the HTML initially renders the container before the JS executes on `document.ready`, there is a visible flicker (Flash of Unstyled Content/UI) during the initial page load.

### Steps to Reproduce

1. Load any standard page on the site as an authenticated or unauthenticated user.
2. Watch the login box area during the first ~500ms of page load.
3. Observe the box flickering into its correct hidden/visible state after JS initialization.

### Expected Behavior

The initial visibility of the login box should be stable from the first render, without any JS-induced UI flicker or layout shift.

### Actual Behavior

The login box container flashes briefly on the screen before the client-side JavaScript executes and applies the `hidden` class, causing a jarring user experience.

### Screenshots

_No response_

### Operating System

All / OS Independent

### Browser

All / Browser Independent

### Additional Context

**Proposed Mitigation**: Use Django template logic (e.g., `{% if user.is_authenticated %}`) to control the initial CSS display state server-side. This prevents the flicker entirely and reduces reliance on client-side JS for basic layout rendering.

*(Note: Instructed by @willgearty to split this from the closed issue #5247. I have a PR ready for this).*

Satya-0023:
### **Summary**
This PR fixes the visible UI flicker (Flash of Unstyled Content) that occurs during initial page load because the login box visibility was relying entirely on client-side JavaScript.

### **Approach**
Replaced the static `hidden` CSS classes in `loginbox_content.html` with Django's `{% if user.is_authenticated %}` server-side template logic. This ensures the correct container is either shown or hidden from the very first paint, before any JS executes.

### **Impact**
- Eliminates the jarring ~500ms visual layout shift on page load.
- Reduces reliance on `user_classes.js` for fundamental layout state.
- Provides a much smoother, "snappier" perceive performance.

**Closes #5304**


Satya-0023:
Thank you for the review!

I have pushed a new commit fix(copilot): Safely centralize server-side rendering and avoid FOUC string flash that fully implements your suggestions.

The is_authenticated template condition has been cleanly centralized into a single if/else block, eliminating the duplicated/inverted CSS classes so the branches can never drift out of sync.
The authenticated user's name is now injected natively on the server-side ({{ user.get_short_name|default:user.first_name|default:user.username }}), fully preventing the Hello, ! string flash before client-side scripts execute.

willgearty:
Not working.

1. Doesn't cover all themes.
2. `user` is not a context variable, so the logic doesn't work.

willgearty (approved):
This needed a lot of work, but it should be good to merge now. Thanks for your contribution!

willgearty (approved):
OK, that should fix the flicker while also maintaining the cache.

PR #5371, opened 2026-03-23
Fix stale email cache in get_teachers (#3836)

### LINKED ISSUE #3836: Email doesn't update immediately
To replicate:

1. Update a class, receive class comments email
2. Change email address for teacher account
3. Update same class, receive class comments email at old email address
4. Flush cache
5. Update same class, receive class comments email at new email address

tanupriyasingh1:
Fixes #3836
Problem
When a teacher changes their email, class comment emails still go to the old address until the cache is manually flushed.
get_teachers() is cached via @cache_function and only invalidates when teachers are added or removed from a class:
pythonget_teachers.depend_on_m2m('program.ClassSubject', 'teachers', lambda subj, event: {'self': subj})
It doesn't know about profile updates, so the cached ESPUser objects keep the old email indefinitely.
Fix
Added a targeted depend_on_row invalidation scoped only to classes the updated user actually teaches:
pythonget_teachers.depend_on_row('users.ESPUser', lambda user: [{'self': cls} for cls in user.classsubject_set.all()])
This avoids broad cache invalidation and ensures only the relevant ClassSubject caches are busted when a user's row changes.
Testing
Added a regression test in ClassSubjectCacheTest that:

Calls get_teachers() to populate the cache
Updates the teacher's email and saves the user
Verifies the next get_teachers() call returns the updated email without requiring a manual cache flush

AI Usage Disclosure
Used Claude to help trace the root cause through the argcache dependency chain. Reviewed and verified all changes myself against existing codebase patterns.


tanupriyasingh1:
hey @willgearty, just a gentle reminder to review my PR when you get free time, no rush at all!

AnushkaPandit-21:
This fixes the bug, but lambda user: {} will invalidate the get_teachers cache for every ClassSubject whenever any user is saved. That's pretty aggressive. For a site with a lot of user activity, that could cause a bunch of unnecessary cache misses. Have you looked at whether you could narrow it down, maybe something like only invalidating for ClassSubjects that the modified user actually teaches? I know catalog_cached uses the broad pattern too, but that's a site-wide cache so it makes more sense there. The per-class get_teachers cache could probably be more targeted.

tanupriyasingh1:
> This fixes the bug, but lambda user: {} will invalidate the get_teachers cache for every ClassSubject whenever any user is saved. That's pretty aggressive. For a site with a lot of user activity, that could cause a bunch of unnecessary cache misses. Have you looked at whether you could narrow it down, maybe something like only invalidating for ClassSubjects that the modified user actually teaches? I know catalog_cached uses the broad pattern too, but that's a site-wide cache so it makes more sense there. The per-class get_teachers cache could probably be more targeted.

That's a fair point — I went with lambda user: {} because I wasn't sure if argcache supports returning a list of dicts from a depend_on_row lambda (something like lambda user: [{'self': cls} for cls in user.classsubject_set.all()]). I couldn't find the argcache source easily to verify. Every existing usage in the codebase returns a single dict.
If argcache does support returning an iterable, I'd be happy to update it. Could you point me to where in the argcache source I should look to confirm this, or let me know if there's a preferred pattern for this?

AnushkaPandit-21:
@tanupriyasingh1 Yeah, I think the targeted approach would be worth trying. Even just filtering to ClassSubjects where the user is a teacher would cut down the invalidation surface a lot. Let me know if you want to bounce ideas on the query structure.

tanupriyasingh1:
> @tanupriyasingh1 Yeah, I think the targeted approach would be worth trying. Even just filtering to ClassSubjects where the user is a teacher would cut down the invalidation surface a lot. Let me know if you want to bounce ideas on the query structure.

hey, I've updated the PR to use the targeted depend_on_row approach with user.classsubject_set.all(). Also added the regression test for it in tests.py. Does this query structure look better?

AnushkaPandit-21:
@tanupriyasingh1 Nice, using user.classsubject_set.all() is exactly what I had in mind. That scopes the invalidation to just the classes the user actually teaches, which is way better than blowing away the whole cache. Looks solid to me!

tanupriyasingh1:
Hey @willgearty, PR is updated and ready — let me know if anything needs changing whenever you get the time!

willgearty:
You've removed the actual fix in the latest commit. Please restore it.

tanupriyasingh1:
> You've removed the actual fix in the latest commit. Please restore it.

Hi @willgearty, thanks for catching that! Restored the `depend_on_row` line in the latest commit. Sorry for the accidental removal!

willgearty (approved):
Looks good, thanks! And thanks for adding a regression test!

PR #5420, opened 2026-03-26
Fix sorttable not re-sorting when data is updated after sorting

### LINKED ISSUE #3398: sorttable doesn't work properly if data in a sorted column is updated after sorting

Pjoshi0805:
Fixes #3398

When a column was sorted and data was subsequently updated, 
sorttable retained the sorttable_sorted class on the <th> tag. 
Clicking the column again would just reverse the rows instead 
of performing a fresh sort.

Fix: Remove the fast-path that called reverseSort() when the 
column was already sorted. Now a full sort is always performed 
on click, so updated data is correctly re-sorted.

Also removed unused assignment `info[col].known = table.rows.length`
which was never read anywhere in the codebase.

Tested locally by sorting a table, randomizing the data, and 
re-sorting — rows now sort correctly instead of just reversing.
Screenshot 2026-03-26 162701


hwatheod (approved):
Thank you for doing this!

PR #5451, opened 2026-03-27
Fix program creation integrity hardening

### LINKED ISSUE #2181: Programs sometimes created without accounting info
This can happen when someone creates a program from the admin panel instead of the form, but we've now seen it on two programs (Claremont's spring 2016 and Duke's fall 2016) where it didn't appear that they did that. It causes a bunch of stuff to error until we do manual cleanup.

dan-squared:
## Description

Fixes the program creation integrity issue in the `newprogram` workflow where partially created programs could be left behind in the database upon error. 
Previously, the global `ATOMIC_REQUESTS = True` setting was inert because Django only honors it inside the `DATABASES` configuration. This meant any `IntegrityError` triggered after `form.save()` would leave behind partially saved related records without a full transaction rollback.

This PR hardens the overall program creation integrity by:
1. Wrapping the entire database portion of program creation in an explicit `atomic` block.
2. Moving external side effects (like Mailman list creation) to `transaction.on_commit(...)` to ensure they don't execute unless the database commit truly succeeds.
3. Adding explicit `IntegrityError` handling: it redirects to an existing program if a concurrent duplicate is detected, or safely returns a non-field error indicating no changes were saved.
4. Changing the program account naming from relying on a fragile `slugify(program.name)` to a per-program unique identifier based on the program ID and URL.
5. Adding a client-side submit lock to the "Make Program Happen!" button to mitigate accidental double-submit races.
6. Removing the misleading, inert `ATOMIC_REQUESTS` global setting and adding an explanatory comment regarding our explicit-transaction approach instead.

## Related Issue

Closes #2181

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] New tests added (validates clean rollback on `IntegrityError` and verifies that same-named programs successfully get distinct accounts)
- [x] Manually verified

## AI Disclosure

This pull request was developed with the assistance of an AI coding agent. The code optimization and regression tests were subsequently reviewed to verify their logic, accuracy, and adherence to the project patterns.

## Checklist

- [x] Self-reviewed the code

### DIFF
--- a/esp/esp/accounting/controllers.py
+++ b/esp/esp/accounting/controllers.py
@@ -81,6 +81,10 @@ class ProgramAccountingController(BaseAccountingController):
         self.refund_items = ['Student refund']
         self.admission_items = ["Program admission", "Student payment"]
 
+    def program_account_name(self):
+        program = self.program
+        return f'program-{program.id}-{slugify(program.url)}'
+
     @transaction.atomic
     def clear_all_data(self):
         #   Clear all financial data for the program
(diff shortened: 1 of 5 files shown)


SuyashJain17 (commented):
@dan-squared please address copilot's requested changes

PR #5461, opened 2026-03-28
Fix loop scoping bug in sanitize_walkin to process all sections

### LINKED ISSUE #5460: [Bug]: Loop scope bug in sanitize_walkin skips registrations
### Description

I was reading through studentregsanity.py and noticed a variable scoping issue in the sanitize_walkin function. The loop that actually expires the registrations (for sr in srs:) is sitting outside the section loop. Because of this, it only ever processes the very last section.

### Steps to Reproduce

Since this is a backend logic bug, here is how to spot it in the code:

1. Open esp/esp/program/controllers/studentregsanity.py.

2. Scroll to the sanitize_walkin function.

3. Notice that the for sr in srs: block is outdented too far—it runs after the for sec in w.get_sections(): loop is completely finished.

4. If this runs on a program with multiple walk-in sections, earlier sections are completely skipped. Also, if there are 0 sections, the script will crash with an UnboundLocalError since srs is never assigned.

### Expected Behavior

The code should iterate through all sections and expire the walk-in registrations for every single section. (The sanitize_lunch function right below it handles this nesting correctly).

### Actual Behavior

It only cleans up the last section because the cleanup loop just uses the leftover srs variable from the finished section loop.

### Screenshots

_No response_

### Operating System

_No response_

### Browser

_No response_

### Additional Context

The fix is pretty straightforward—just indent the for sr in srs: block and the report.append line so they sit inside the for sec... loop. I would like to work on this issue

Vipeen-Kumar:
## Description

This PR fixes the scoping bug in `sanitize_walkin` where it was skipping earlier walk-in sections and only expiring registrations for the very last one. 

I fixed the indentation so the registration cleanup loop is properly nested inside the section loop. While I was in there, I also made a tiny optimization: I moved the `if not fake:` check outside the `for sr in srs:` loop so it only evaluates once per section instead of redundantly for every single student registration. 

## Related Issue

Closes #5460

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

## AI Disclosure

Used AI (Gemini) to help spot the exact lines for the scoping issue and format the bug report.

## Checklist

- [x] Self-reviewed the code

### DIFF
--- a/esp/esp/customforms/views.py
+++ b/esp/esp/customforms/views.py
@@ -2,10 +2,11 @@ from copy import deepcopy
 import json
 
 from django.db import transaction
-from django.shortcuts import redirect, HttpResponse
+from django.shortcuts import redirect
+from django.http import Http404, HttpResponse, HttpResponseRedirect, JsonResponse
 from django.urls import reverse
-from django.http import Http404, HttpResponseRedirect, JsonResponse
 from django.db import connection
+from django.core.exceptions import ObjectDoesNotExist
 from django.core.serializers.json import DjangoJSONEncoder
 
 from esp.customforms.models import *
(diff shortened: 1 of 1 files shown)


AnushkaPandit-21:
Good catch on the indentation bug. Classic Python gotcha where everything looks fine until you realize the loop body isn't actually inside the loop. Did you test this with multiple walk-in sections to confirm they all get processed now?

samalarussa (approved):

PR #5468, opened 2026-03-28
Fix inconsistent scheduling diagnostics by snapshotting section data

### LINKED ISSUE #511: Scheduling checks doesn't work if classes change during the run
Running the scheduling checks feature while other people are scheduling frequently causes a whoops error.

To reproduce:
*schedule a class
*start running the scheduling checks by going to /manage/Program/Year/scheduling_checks (on MIT Splash, these take about 3 minutes)
*unschedule the class
You should get a whoops error.

Probably reason:
The scheduling checks code creates a list of scheduled classes and then repeatedly iterates through them.  If one of those classes is unscheduled while the tests run, then any test which retrieves the rooms  or timeslots of a class will throw a whoops error because there are none.

Proposed solution:
Solution is probably to load all the data before running the tests, and that would also speed up the tests considerably by reducing the number of database queries.

aksht27:
## Description

Scheduling diagnostics could produce inconsistent results when section data (meeting times, resources, etc.) changed during execution.

This PR fixes the issue by introducing a snapshot-based approach:
- All relevant section data is fetched once during initialization
- Diagnostics operate on in-memory snapshots instead of live database queries
- Prevents inconsistencies caused by mid-run modifications

## Related Issue

Closes #511

## Type of Change

- [x] Bug fix
- [x] Refactor / cleanup

## Testing

- [x] Existing tests pass
- [ ] New tests added (not required for this fix)
- [x] Manually verified

## Reproduction & Verification
Reproduced the issue by modifying section meeting times during execution.

Before fix:
- Diagnostics output changed when data was modified mid-run

After fix:
- Output remains consistent due to snapshot-based execution

**Verification Output:**

AdobeExpressPhotos_9ecd346470a04018a44eaaa98b22a3d7_CopyEdited

Example:
Same output: True

## AI Disclosure

Used ChatGPT for debugging guidance and structuring the approach. Implementation and testing were done manually.

## Checklist

- [x] Self-reviewed the code
- [ ] Updated documentation (not required)


Oval17 (commented):
Sound idea.
Can anyone with scheduling-domain knowledge review this

PR #5480, opened 2026-03-28
Feat/1680 deactivate old class list

### LINKED ISSUE #1680: Deactivate old class lists
Sometimes very old class lists get spam, and they very rarely get used, so it would be useful to automatically deactivate them after a few years.  Ideally we might be able to manually reactivate them later, in case someone does want it, although it might be easier to just make such a list manually.

kartikJaiswal-hash:
(#1680 ) Refactored the annual class list cleanup task to prevent automated "cancellation" emails from being sent to students and teachers. Instead of calling cancel() on ClassSubject objects, the task now identifies the associated EmailList objects using the emailcode() pattern and sets admin_hold=True. This effectively "silences" old lists (causing them to bounce) without triggering the side effects of a full class cancellation.

The query has been optimized to use the direct meeting_times ManyToMany relationship on ClassSubject, and the logic is idempotent, ensuring that redundant runs of the cron job do not perform unnecessary database updates.

Related Issue
Closes #1680 

Type of Change
[x] Bug fix
[x] New feature
[x] Refactor / cleanup

Testing
[x] Existing tests pass
[x] Manually verified

Tested in the local Django shell by creating a mock ClassSubject with an Event dated 2000 days in the past. Verified that the task correctly identified the class, matched its EmailList, and toggled admin_hold. Confirmed that subsequent runs returned a count of 1 after admin_hold=True the count became 0 showing Idempotent behavior. 

AI Disclosure
No AI tools were used to resolve this issue.

Checklist
[x] Self-reviewed the code
[x] No new warnings or errors introduced


Dhirenderchoudhary:
hey @kartikJaiswal-hash I have pull and run tests locally on the changes
Here are few comments : missing [EmailList](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) import in tests
possible false-positive list matching in deactivate_old_class_lists(), and fragile Jan 1 2-3 AM cron window

kartikJaiswal-hash:
@Dhirenderchoudhary EmailList was already imported above so i didnt imported it again also you are right window at jan 1 time 2 to 3 AM will be fragile but i dont think many people will use the website during that time if you have a better idea for implementation i am ready to make changes and open to your suggestion

kartikJaiswal-hash:
should i make it that it only disable 1000 or 10,000 email (depending on how much server can handle) every 15 min. I wasn't sure how much load the server can take

kartikJaiswal-hash:
should i make it that it only disable 1000 to 10,000 email (depending on how much server can take). I wasn't sure how much server will take if a mentor can answer me. I am ready to make changes depending on how much limit i should set

Dhirenderchoudhary:
According to me : Yes, use batch processing every 15 minutes. Start with a conservative limit (around 1,000)
keep it idempotent, and monitor query/runtime impact before increasing. Also please avoid the Jan 1 2-3 AM window and use deterministic matching to prevent false positives.

kartikJaiswal-hash:
@Dhirenderchoudhary Thanks for your review it helped a lot. 
I should also start reviewing other people like you.

kartikJaiswal-hash:
when will this code going to get merge. If there is a error or blunder pls review the code

kartikJaiswal-hash (commented):
look ok

kkbrum:
A maintainer will need to look at this before it gets merged, and unfortunately there is a very long backlog. We will get to this when we can.

esp/dbmail_cron.py
```diff
@@ -5,6 +5,7 @@
 import fcntl
 import logging
 from io import open
+
```
Oval17, line 8: Hi, 
can you remove these stray lines present here and there

Oval17 (changes requested):
Hi, 
Can you fix imports and cleanup is required 
also is it possible to narrow the mailgate scope

PR #5529, opened 2026-03-30
Fix: catch IndexError instead of DoesNotExist in GroupTextModule

### LINKED ISSUE #5407: [Bug]: GroupTextModule catches wrong exception — IndexError goes uncaught
### Description

In `grouptextmodule.py` at line 154, `sendTextMessage()` fetches a user's contact info using `.filter()[0]`, but the except block catches `ContactInfo.DoesNotExist`. The problem is `.filter()[0]` never raises `DoesNotExist`  it raises `IndexError`. So when a user in the SMS list has no matching `ContactInfo`, the `IndexError` goes uncaught and crashes the loop.

https://github.com/learning-unlimited/ESP-Website/blob/main/esp/esp/program/modules/handlers/grouptextmodule.py#L152-L156

```python
contactInfo = None
try:
    contactInfo = ContactInfo.objects.filter(
        user=user, as_user__isnull=False).distinct('user')[0]
except ContactInfo.DoesNotExist:  # never fires for filter()[0]
    pass
```

### Steps to Reproduce

1. Open Docker Desktop → `esp-website-web-1` container → Exec tab
2. Run `cd /app/esp && python manage.py shell`
3. Run:
```python
from esp.users.models import ContactInfo
ContactInfo.objects.filter(user_id=999999, as_user__isnull=False).distinct('user')[0]
```
4. See `IndexError: list index out of range`

### Expected Behavior

Users without contact info should be skipped gracefully the `if not contactInfo: continue` block below already handles this nd remaining users should still receive their texts.

### Actual Behavior

`IndexError` escapes the try/except, causing a 500 error and aborting the entire send mid-loop.

### Screenshots

Image

### Operating System

macOS 16

### Browser

Chrome

### Additional Context

`.get()` raises `DoesNotExist`, but `.filter()[0]` raises `IndexError` different Django APIs. A fix would be replacing `filter(...)[0]` with `.first()`, which returns `None` on an empty queryset. The existing `if not contactInfo: continue` guard handles it without needing a try/except at all.

puri-adityakumar:
## Description

`filter()[0]` raises `IndexError` on empty querysets, not `DoesNotExist`. The except block in `sendMessages()` was catching the wrong exception, so when a user has no `ContactInfo`, the `IndexError` goes uncaught and crashes the entire SMS send loop.

Changed `except ContactInfo.DoesNotExist` to `except IndexError`.

## Related Issue

Closes #5407

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

Verified in Django shell:
- `except IndexError` catches the error and returns `None` — the existing `if not contactInfo: continue` guard handles it gracefully.
- `except ContactInfo.DoesNotExist` (old code) lets `IndexError` escape uncaught.

## Screenshots

**Before (IndexError crashes uncaught):**

before

**After (fix applied):**

image

## AI Disclosure

Bug was identified via static analysis and verified manually in Django shell inside Docker. Fix is a single-line exception type change. I understand the code path and can explain the fix independently

## Checklist

- [x] Self-reviewed the code


puri-adityakumar:
Hi, just checking in. Is there anything you would like me to change?

Copilot suggested switching to .first(), but I kept the fix minimal on purpose. The filter()[0] pattern is already used in the codebase here, and switching to .first() with the existing .distinct('user') could cause issues with PostgreSQL DISTINCT ON ordering.

Please, Let me know.
@willgearty

willgearty:

willgearty (approved):
Looks good, thanks!

PR #5543, opened 2026-03-30
Fix security issue in deadline_management

### LINKED ISSUE #5538: [Bug]: Security issue in deadline_management (GET used instead of POST)
### Description

There is a security issue in the deadline_management function.

Currently, request.GET is used for handling sensitive operations.

This can lead to security vulnerabilities.

It should use request.POST instead to ensure safer handling of data.

I have already fixed this issue and will submit a pull request.

### Steps to Reproduce

_No response_

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Screenshots

_No response_

### Operating System

_No response_

### Browser

_No response_

### Additional Context

_No response_

Tanush-Nk:
Closes #5538 

## Description
Fixed security issue in deadline_management by replacing request.GET with request.POST.

## Type of Change
- Security fix

### DIFF
--- a/esp/esp/program/modules/handlers/admincore.py
+++ b/esp/esp/program/modules/handlers/admincore.py
@@ -360,57 +360,57 @@ class AdminCore(ProgramModuleObj, CoreModule):
         message_bad = ''
 
         #   Handle 'open' / 'close' / 'delete' actions
-        if extra == 'open':
+        if request.method == "POST" and extra == 'open':
             #   If there are no permissions for this permission type, create one and open it now (open ended)
             #   If there are permission(s) for this type, take the most recent(?) and open it (open ended)
-            if 'group' in request.GET and 'perm' in request.GET:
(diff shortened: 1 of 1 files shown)


willgearty:
You've only changed the handler, you'll also need to change the template so that the data is submitted as POST.

HiddendarkViper07:
@willgearty ,Thanks for the review!

You're right — I’ve updated the handler to use POST, but the template still submits data using GET. I’ll update the template to use POST and ensure the form submission aligns with the backend changes.

I’ll push the updated changes shortly.

Dhirenderchoudhary:
According to me right now it breaks the page flow.

admincore now expects POST for open/close/delete, while [deadlines.html](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) still sends those actions via GET links, so those buttons won’t work.

willgearty:

PR #5576, opened 2026-03-31
Fix variable shadowing in ESPError: swap log and message assignment order

### LINKED ISSUE #5562: Bug: ESPError(True) silently downgrades to no-log due to variable shadowing in ESPError()
## Summary
In `esp/esp/middleware/esperrormiddleware.py`, the `ESPError()` factory function has a
variable shadowing bug in its legacy-argument-handling branch that causes the `log` flag
to always be set to `None` instead of the original bool argument.

## Affected File
`esp/esp/middleware/esperrormiddleware.py`, lines 79–83

## Code
```python
def ESPError(message=None, log=True):
    if isinstance(message, bool):
        message = None
        log = message   # BUG: message is already None here, not the original bool
```

## Root Cause
On line 82, message is reassigned to None. On line 83, log = message copies
this new None value, not the original bool argument that was passed in.

The intended logic was:

```
log = message    # capture the original bool
message = None   # then clear message
```
But the two lines are in the wrong order.

### Steps to Reproduce

```
from esp.middleware.esperrormiddleware import ESPError, ESPError_Log, ESPError_NoLog

e = ESPError(True)   # legacy usage: should return ESPError_Log
print(type(e))       # prints ESPError_NoLog wrong!
```

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Screenshots

_No response_

### Operating System

_No response_

### Browser

_No response_

### Additional Context

_No response_

rehantroy:
…rder

log = message was executed after message = None, causing log to always be None instead of the intended boolean. Swapped the two lines to fix the shadowing bug.

## Description

Fixes a variable shadowing bug where `message = None` was executed before `log = message`, causing `log` to always be `None` instead of the intended boolean value passed by the caller.

## Related Issue

Closes #5562

## Type of Change

✅ Bug fix

## Testing

✅ Manually verified

## AI Disclosure

AI tools were used for general research and understanding the codebase. The bug identification, fix decision, and code change were made and verified by me personally.

## Checklist

✅ Self-reviewed the code
✅ No new warnings or errors introduced

### DIFF
--- a/esp/esp/middleware/esperrormiddleware.py
+++ b/esp/esp/middleware/esperrormiddleware.py
@@ -79,8 +79,8 @@ def ESPError(message=None, log=True):
     if isinstance(message, bool):
         # trying to pass a bool argument: assume they meant log rather than message
         # this should become deprecated -lua 2013-02-15
-        message = None
         log = message
+        message = None
 
     if log:
         cls = ESPError_Log


deepaksoni47:
Tested locally ,looks good to go!

willgearty (approved):
Looks good, thanks!

PR #5584, opened 2026-03-31
fix: fixed GroupTextModule handling for users without contact info

### LINKED ISSUE #5566: [Bug]: GroupTextModule crashes when a matched user has no direct contact info
### Description
I noticed that the Group Text flow appears to break if one of the matched users does not have a direct `ContactInfo` record linked through
  `RegistrationProfile.contact_user`.

  In `esp/esp/program/modules/handlers/grouptextmodule.py`, the lookup currently uses queryset indexing (`[0]`) but the surrounding code seems
  to expect the missing-contact case to be handled by skipping that user and continuing.

### Steps to Reproduce
  1. Enable the Group Text module with valid Twilio settings.
  2. Select a user set where at least one matched user does not have direct contact info.
  3. Try to send a group text.
  4. Observe that the send flow errors instead of skipping that user.

### Expected Behavior

If a matched user has no direct contact info, the module should log that and continue sending to the rest of the matched users.

### Actual Behavior

 The flow appears to fail instead of treating the missing contact info as a non-fatal case.

### Screenshots

_No response_

### Operating System

Ubuntu 24.04

### Browser

Chrome

### Additional Context

Relevant code path:
 Relevant file:

  - `esp/esp/program/modules/handlers/grouptextmodule.py`

The current logic already logs `"Could not find contact info for ..."` and continues, so this seems like an implementation bug rather than intended behavior.

samar-703:
## Description

This fixes a bug in `GroupTextModule.sendMessages()` when a matched user does not have direct contact info.
The existing code already seems to intend that case to be non-fatal: it logs that contact info could not be found and moves on to the next user. But the lookup was using queryset indexing, which can fail before that fallback path runs.

This change switches that lookup to a non-raising pattern and adds regression tests for the surrounding behavior.

## Related Issue

Closes #5566 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [ ] Existing tests pass
- [ ] Manually verified

## Checklist

- [x] Self-reviewed the code


deepaksoni47:
Thanks for the fix and tests.

I found one blocker. The new query uses `.distinct('user').first()` which raises a Postgres `DISTINCT ON` / `ORDER BY` error, so `sendMessages()` can still crash.

I reproduced it by running the new `grouptextmodule` tests locally in Docker (they fail on this query). Please update the lookup (e.g., avoid `distinct('user')` here or add compatible ordering) and re-run tests.

samar-703:
> Thanks for the fix and tests.
> 
> I found one blocker. The new query uses `.distinct('user').first()` which raises a Postgres `DISTINCT ON` / `ORDER BY` error, so `sendMessages()` can still crash.
> 
> I reproduced it by running the new `grouptextmodule` tests locally in Docker (they fail on this query). Please update the lookup (e.g., avoid `distinct('user')` here or add compatible ordering) and re-run tests.

thanks for this, i got your point and i checked it, the error was coming, so I updated the lookup to use the user’s direct profile contact (`user.getLastProfile().contact_user`) instead of the `distinct('user').first()` query, which avoids the PostgreSQL `DISTINCT ON` / `ORDER BY` issue and better matches the intended semantics here.

deepaksoni47:
I pulled this locally and ran the focused GroupTextModule tests on my system. Looks good to me now.

willgearty:

PR #3460, opened 2021-12-04
Fix selenium tests

willgearty:
This fixes our selenium tests (front-end behavior tests) that haven't been running for ages (if not ever). We don't really have a lot of them (and some that were marked as such aren't actually selenium tests), but this opens the avenue to potentially create more in the future. Previously, the selenium tests had to be run separately using a separate management command, but now they are just included in our normal set of unit tests (so they should also be run with Github Actions every time a PR is created). We also have some JS tests for the ajax scheduler, but those are currently written in jasmine and are therefore outside of the scope of this PR.

PR #5586, opened 2026-03-31
Fix Selenium tests for MakeAClass and EditClass views (#3460)

Akritiiiiiiii:
This PR fixes the existing Selenium tests for MakeAClass and EditClass views as described in issue #3460.

- Updated element selectors to match current templates
- Added explicit waits for input fields to avoid timing issues
- Verified tests run successfully locally and on CI

Closes #3460


Akritiiiiiiii:
Fix Selenium tests for MakeAClass and EditClass views (#3460)

PR #5588, opened 2026-03-31
Refactor CacheClass to subclass PyLibMCCache instead of proxy

### LINKED ISSUE #1305: memcached_multikey.CacheClass should subclass memcached.PyLibMCCache
The current behavior is to save a `PyLibMCCache` instance to `self._wrapped_cache`, and to proxy all requests through there.

I'm not sure what the purpose of having a proxy is. It seems like the problem can be more easily solved with subclassing, and that this would be the preferred design pattern.

This class was originally introduced in commit 69602e4a. @aseering do you know if there is any benefit in this case to using a proxy over subclassing?

Relatedly, now that we manage our dependencies with pip requirements, we no longer need to conditionally choose which subclass / proxy class to use. We can always use `PyLibMCCache`.

This issue will no longer be applicable if #1306 is first addressed.

vaibhavpal7549:
## Description

This PR refactors `memcached_multikey.CacheClass` to subclass `memcached.PyLibMCCache` instead of using a proxy-based implementation.

Previously, `CacheClass` stored a `PyLibMCCache` instance internally and proxied method calls through `_wrapped_cache`. This added unnecessary indirection without providing clear benefits.

With the current dependency setup, we can directly subclass `PyLibMCCache`, resulting in a cleaner and more maintainable design that aligns with standard Django caching patterns.

---

## Related Issue

Closes #1305

---

## Type of Change

* [ ] Bug fix
* [ ] New feature
* [ ] Breaking change
* [ ] Documentation update
* [x] Refactor / cleanup
* [ ] CI/CD or build change

---

## Testing

* [x] Existing tests pass
* [x] Manually verified

---

## AI Disclosure

* Tool used: Codex
* Scope of assistance: Helped in understanding the refactoring approach and improving the structure of the solution and PR description
* Verification: All code changes were manually reviewed, tested, and validated before submission

---

## Checklist

* [x] Self-reviewed the code

### DIFF
--- a/esp/esp/program/views.py
+++ b/esp/esp/program/views.py
@@ -977,11 +977,8 @@ def flushcache(request):
     if request.POST:
         if "reason" in request.POST and len(request.POST['reason']) > 5:
             reason = request.POST['reason']
-            _cache = cache
-            while hasattr(_cache, "_wrapped_cache"):
-                _cache = _cache._wrapped_cache
-            if hasattr(_cache, "clear"):
-                _cache.clear()
+            if hasattr(cache, "clear"):
+                cache.clear()
                 mail_admins("Cache Flushed on server '%s'!" % request.META['SERVER_NAME'], "The cache was flushed by %s!  The following reason was given:\n\n%s" % (request.user.username, reason))
                 context['success'] = "Cache Cleared."
             else:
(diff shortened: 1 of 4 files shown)


vaibhavpal7549 (commented):
Done

deepaksoni47:
Reviewed it locally on my system: no blockers from my side.

The subclass refactor looks correct, the double-keying issue is fixed (keys are no longer pre-transformed before calling `super()`), and `cache.clear()` callsites are consistent with the new design. Targeted cache/dependency tests passed locally.

Issue #3854, opened 2025-06-27
New module management UI


willgearty:
I mocked up a new module management UI today:

![Image](https://github.com/user-attachments/assets/604eafb6-39fb-43ad-b157-ae7fb8278098)

Part of this is influenced by this issue where certain modules sometimes need to be enabled/disabled over time: https://github.com/learning-unlimited/ESP-Website/issues/2895.

But this also stems from comments that I've heard about how the existing module management page is confusing (which I would argue is much better than having to manage the modules in the admin pages).

I think deadlines would still need to exist outside of this (although perhaps the defaults would correspond to the open/close of the modules) since you might need complex situations where deadlines open and close multiple times? And there also isn't a 1:1 of deadlines to modules. And because deadlines might need to exist for other user groups.

I think we could also combine this with the module questions interface that @milescalabresi developed ages ago, such that when you check or uncheck a box, those modules would appear/disappear in this interface so admins would see what things were being added right there. I also imagine that whatever page hosts this would have two of these, one for student modules and one for teacher modules.

willgearty:
I suppose, since the phase zero module is set up to always be enabled to ensure rejected students can't get in, you would need it to be enabled the entire time. And you also would want the custom form before the 2-phase lottery module since the 2-phase module has a blocker so higher sequence modules wouldn't be seen. But you get the picture.

willgearty:
Some more ideas copied from Slack in response to comments from @milescalabresi:

> maybe a UI like this could completely subsume the deadlines[/permissions] system (from an admin perspective). I had originally only envisioned two versions of this, one for students and one for teachers. But I could also envision a page with a dropdown at the top where you could just choose a user group (or even individual user) and then it would show the "module schedule" for that user/user group using ajax calls, kind of like the onsite grid-based class changes for class schedules?

> And upon submitting, we could then have the necessary deadlines generated automatically server-side (perhaps via a set of program module object class methods that can be overridden that can be overridden if a module needs custom stuff e.g., 2-phase class lottery)

> So when an admin views the module schedule for an individual student, for example, we would determine the start and end times based on inherited deadlines. Then if they make changes, it would generate deadlines just for that student. But none of this would be visible to the admin, it would just show this visual representation.

NobleCoder69:
Hi! I’m a GSoC 2026 aspirant interested in contributing to ESP Website.
I’d like to work on this issue. Could you please confirm if it’s still available and whether the mockups represent the final expected UI?
[...]

Issue #2895, opened 2020-01-14
Module opening/closing times


willgearty:
In many cases for larger programs, admins need to turn on/off different modules as they transition between different parts of their registration. For example, Stanford's student registration consists of three different phases: a student lottery, a class lottery, and FCFS (in which there's also financial aid applications, online payment, etc). In each of these phases, different modules need to be enabled, while others need to be disabled to prevent conflicts. All of this needs to be done manually in the admin pages. The upcoming program settings page (https://github.com/learning-unlimited/ESP-Website/pull/2816) will move all of this out of the admin pages, but it will still require manual changing of the settings possibly multiple times over the course of student/teacher registration.

The fix for this that I'm envisioning would make `ProgramModuleObj` objects similar in function to `StudentRegistration` objects, in that they could have opening/closing (start/expire) times/dates. Then, like with `StudentRegistration` objects, we could have a `valid_objects()` method that would return only the module objects that are open and have not closed/expired yet. We would just need to use this new method instead of `objects` when we query for the modules for a program (e.g. `studentregcore`).

By default all module objects would have no times listed, so they would be perpetually open. In the first iteration of this, admins could set these times in the admin pages. However, eventually we'd probably want some new user-friendly page where admins can add/remove module objects and set their times (maybe it would even display them as a timeline with multiple rows).

Requested by Stanford and MIT at SplashCon 2020.

adityars07:
@willgearty Hi!
I’d like to work on this issue and implement the module opening/closing time feature.
I’ll review the current ProgramModuleObj logic and propose a clean approach for adding scheduled open/close DateTime fields along with necessary tests.
Please assign this issue to me. Thanks!

willgearty:
@adityars07 Thanks for your interest in contributing to our project! I've assigned you to this issue. Please make sure to check out our [instructions for getting started](https://github.com/learning-unlimited/ESP-Website/wiki#getting-started) and follow our [contribution rules](https://github.com/learning-unlimited/ESP-Website/wiki#rules). Let me know if you have any questions.

imran-techvoyager:
Hi @adityars07 and @willgearty  I've been looking into both #2895 and #3854 since they seem closely related architecturally, and I wanted to share a few observations in case they're helpful as this moves forward.

[...]

PR #5594, opened 2026-03-31
Add flags to admin dashboard (#1354)

### LINKED ISSUE #1354: Add flags to dashboard
Flags which are designated should show up in the dashboard, and all flags, with their detail, should show up in the dashboard status pop-up.

PranavGadge01:
## Summary

This PR adds support for displaying class flags directly in the admin dashboard to improve visibility of important class metadata.

Flags are now accessible in two places:
1. Class list view (inline badges)
2. Class status popup (detailed view)

This improves the ability for admins to quickly identify flagged classes and inspect associated metadata without navigating away.

---

### Backend
- Extended class data API to include:
  - `dashboard_flags` → lightweight flag data for list view
  - `flags` → detailed flag information for status popup
- Each flag includes:
  - name
  - color
  - comment
  - modified_by
  - modified_time

### Frontend
- Added inline flag badges next to class titles in admin dashboard
- Implemented flag rendering in status popup:
  - Displays badge + metadata
  - Includes comments and modification details
- Introduced helper function to generate consistent badge UI

---

### Class List
- Flags with `show_in_dashboard=True` appear as colored badges
- Positioned next to class title for quick scanning

### Status Popup
- Displays all flags associated with a class
- Includes:
  - flag name (badge)
  - comment (if present)
  - modifier and timestamp

---

## Scope

This PR is intentionally limited to **flags display functionality (#1354)**.

It does NOT include:
- module scheduling changes (#3854 / #2895)
- unrelated test additions

This keeps the PR atomic and easy to review.

---

## Testing

- Manually verified:
  - Flags appear correctly in class list
  - Status popup shows complete flag details
- Existing functionality remains unaffected

---

## Related Issue

Closes #1354

### DIFF
--- a/esp/esp/program/modules/handlers/jsondatamodule.py
+++ b/esp/esp/program/modules/handlers/jsondatamodule.py
@@ -567,6 +567,7 @@ class JSONDataModule(ProgramModuleObj, CoreModule):
     class_subjects.cached_function.depend_on_row(ClassSubject, lambda cls: {'prog': cls.parent_program})
     class_subjects.cached_function.depend_on_cache(ClassSubject.get_teachers, lambda cls=wildcard, **kwargs: {'prog': cls.parent_program})
     class_subjects.cached_function.depend_on_row(ClassFlag, lambda flag: {'prog': flag.subject.parent_program})
+    class_subjects.cached_function.depend_on_row(ClassFlagType, lambda ft: {})
 
     @aux_call
     @json_response({
--- a/esp/esp/program/modules/handlers/jsondatamodule.py
+++ b/esp/esp/program/modules/handlers/jsondatamodule.py
@@ -497,7 +497,8 @@ class JSONDataModule(ProgramModuleObj, CoreModule):
         moderators = []
         classes = []
         qs = prog.classes().prefetch_related(
-            'category', 'sections', 'teachers')
+            'category', 'sections', 'teachers',
+            'flags__flag_type')
 
         for c in qs:
             class_teachers = c.get_teachers()
@@ -519,6 +520,11 @@ class JSONDataModule(ProgramModuleObj, CoreModule):
                 cls['difficulty'] = c.hardness_rating
                 cls['prereqs'] = c.prereqs
             cls['emailcode'] = c.emailcode()
+            cls['dashboard_flags'] = [
+                {'name': f.flag_type.name, 'color': f.flag_type.getColor()}
+                for f in c.flags.all()
+                if f.flag_type.show_in_dashboard
+            ]
             if c.duration:
                 cls['length'] = float(c.duration)
             cls['sections'] = [s.id for s in c.sections.all()]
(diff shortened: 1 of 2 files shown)


deepaksoni47:
Nice cleanup and scope is good now. 
I found one issue: dashboard badge data is cached but not invalidated when ClassFlagType fields change (name/color/show_in_dashboard), so admins may see stale flags; please add a cache dependency for ClassFlagType updates.

PranavGadge01:
Good catch — added a depend_on_row for ClassFlagType so the cache busts when flag name, color, or show_in_dashboard changes. Thanks for the review!

esp/esp/program/modules/handlers/jsondatamodule.py
```diff
@@ -560,6 +566,8 @@ def class_subjects(extra, prog):
         return {'classes': classes, 'teachers': teachers, 'moderators': moderators}
     class_subjects.cached_function.depend_on_row(ClassSubject, lambda cls: {'prog': cls.parent_program})
     class_subjects.cached_function.depend_on_cache(ClassSubject.get_teachers, lambda cls=wildcard, **kwargs: {'prog': cls.parent_program})
+    class_subjects.cached_function.depend_on_row(ClassFlag, lambda flag: {'prog': flag.subject.parent_program})
+    class_subjects.cached_function.depend_on_row(ClassFlagType, lambda ft: {})
```
Oval17, line 570: `lambda ft: {}` invalidates the cache for ALL programs whenever any `ClassFlagType` changes. Since `ClassFlagType` is shared across programs via M2M, this is hard to scope narrowly 
umm what we can do is consider using `depend_on_model(ClassFlagType)` instead (same effect, clearer intent, matches the pattern used by `sections_admin` on line 444) if possible

esp/esp/program/modules/handlers/jsondatamodule.py
```diff
@@ -519,6 +520,11 @@ def class_subjects(extra, prog):
                 cls['difficulty'] = c.hardness_rating
                 cls['prereqs'] = c.prereqs
             cls['emailcode'] = c.emailcode()
+            cls['dashboard_flags'] = [
+                {'name': f.flag_type.name, 'color': f.flag_type.getColor()}
+                for f in c.flags.all()
+                if f.flag_type.show_in_dashboard
+            ]
```
Oval17, line 527: `c.flags.all()` uses the prefetch cache from `flags__flag_type` which is  correct. But this filters `show_in_dashboard` in Python after fetching all flags. For programs with many flags, filter at the DB level instead: use a dedicated prefetch with `Prefetch('flags', queryset=ClassFlag.objects.filter(flag_type__show_in_dashboard=True).select_related('flag_type'))`.
Like from db optimization pov this is ideal somewhat can you explore once if its possible?

Oval17 (commented):
Backend logic seems correct and prefetch is in place, cache dependencies cover both ClassFlag and ClassFlagType changes, and the null guard on `modified_by` is appropriately defensive. 
There are few optimisation suggested that can be explored if possible.

willgearty:

Issue #995, opened 2014-01-29
Junction admissions backend enhancements


lenaqr:
(this refers to #865 which hasn't been merged yet, but I collected this list of issues last year and want to put it somewhere public)

Bugs:
- The list view is supposed to auto-sort itself as you rank apps, but equivalent apps sometimes switch around randomly. Use app ID or something as a tiebreaker to prevent this.
- Small UI bug from teacher: if I make the comment for a student " ", then I can no longer click to edit that comment.
- Missing cache dependencies
- Javascript performance issues?

List view UI enhancements:
- Admin-side: Allow viewing all applications for the program at once, not just for one class at a time.
- Admin-side: Add a column showing timestamps of when each app was received.
- Teacher-side: Add a column showing whether the class is the student's 1st, 2nd, or only choice.
- Teacher-side: An option for teachers to redact names of applicants.
- Teacher-side: Drag and drop capability.
- More flexible sorting and filtering (basically, allow sorting by any column).
- More seach options for the admin panel (search by student name, etc) --> in admin_py_stuff
- Highlight the row in the list for the application that is currently being viewed.
- Alternate row colors for readability.
- If a comment is too long for the list view, use an ellipsis or fade-out to indicate that there's more.

New features:
- Allow admins to explicitly reject students from classes. (The current schema for "decision status" only has "accepted", "waitlisted", and "not accepted"; the last one should be split into "rejected" and "unreviewed".)
- Add a page for students to view their decisions?
- For admins, add a stats view that shows counts of the number of applicants, number of apps rated green/yellow/red/unrated, number of accepted applicants (with breakdown by class?)
- Spreadsheet export?
- Financial aid: track finaid requests, finaid received.
- Class assignment script.

Boopana-M:
Hi @willgearty  , I'd like to work on this issue .I can start by reviewing the listed bugs and UI enhancements and work on implementing the necessary backend and admin panel improvements
ThankYou !

willgearty:
@Boopana-M Thanks for your interest in contributing to our project! I've assigned you to this issue. Please make sure to check out our [instructions for getting started](https://github.com/learning-unlimited/ESP-Website/wiki#getting-started) and follow our [contribution rules](https://github.com/learning-unlimited/ESP-Website/wiki#rules). Let me know if you have any questions.

Boopana-M:
@willgearty Thank you for assigning this issue to me! I’m actively working on it and will post an update shortly.

[...]

PR #4517, opened 2026-02-28
Add cache invalidation for DBListCount when ESPUser changes

Boopana-M:
Description
This PR adds cache invalidation for DBListCount values used in the DBList.count() method.
Currently, the cached count of users can become stale because it is never invalidated when ESPUser objects change. This change ensures the cache stays consistent with the database.

Changes
->Added Django post_save and post_delete signals for the ESPUser model.
->When an ESPUser is created, updated, or deleted, cache keys prefixed with DBListCount: are cleared.
->Changes are limited to esp/esp/users/models/__init__.py.

Related Issue
Closes #5213

Type of Change
->Bug fix
->New feature
->Breaking change
->Documentation update
->Refactor / cleanup
->CI/CD or build change

Testing
->Existing tests pass
->New tests added (not required for this change)
->Manually verified

AI Disclosure
AI tools were used to help understand the existing caching implementation and identify where cache invalidation should be added. All code changes were reviewed and finalized by me.

Checklist
->Self-reviewed the code
->Updated documentation (not needed)
->No new warnings or errors introduced

PR #5602, opened 2026-03-31
Fix DBListCount cache staleness by adding dependency registration and invalidation (#5213)

### LINKED ISSUE #5213: [Bug]: Issue 995 sub-issue: missing cache dependencies for DBListCount
### Description

Parent: #995

Problem summary
DBListCount cache can become stale after user create/update/delete.

Proposed fix / scope
Register DBListCount cache keys when written
Invalidate those keys on user save/delete
Add regression test

Expected result
DBList count refreshes correctly after user writes.

### Steps to Reproduce

_No response_

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Screenshots

_No response_

### Operating System

_No response_

### Browser

_No response_

### Additional Context

_No response_

MRRanjan82:
## Description

This PR fixes stale cache issues in `DBListCount` that occur after user create, update, or delete operations.

Previously, cache entries generated by `DBListCount.count()` were not associated with any invalidation mechanism. As a result, changes to user data did not refresh cached counts, leading to inconsistent and outdated results.

This PR introduces cache dependency tracking by registering cache keys at write time and ensures proper invalidation through Django signals (`post_save` and `post_delete`) on the user model.

Additionally, regression tests have been added to verify that cached counts are correctly updated after user modifications.

## Related Issue

Closes #5213  
Parent issue: #995

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

## AI Disclosure

    No AI tools were used in the development of this pull request.

## Checklist

- [x] Self-reviewed the code

### DIFF
--- a/esp/esp/users/models/__init__.py
+++ b/esp/esp/users/models/__init__.py
@@ -1291,6 +1291,22 @@ class ESPUser(User, BaseESPUser):
 class AnonymousESPUser(BaseESPUser, AnonymousUser):
     pass
 
+def invalidate_DBListCount_caches():
+    keys = cache.get('DBListCount_keys', [])
+    for key in keys:
+        cache.delete(key)
+    cache.delete('DBListCount_keys')
+
+@dispatch.receiver(signals.post_save, sender=ESPUser,
+                     dispatch_uid='invalidate_DBListCount_caches_save')
+def invalidate_DBListCount_caches_on_save(sender, instance, **kwargs):
+    invalidate_DBListCount_caches()
+
+@dispatch.receiver(signals.post_delete, sender=ESPUser,
+                     dispatch_uid='invalidate_DBListCount_caches_delete')
+def invalidate_DBListCount_caches_on_delete(sender, instance, **kwargs):
+    invalidate_DBListCount_caches()
+
 @dispatch.receiver(signals.post_save, sender=User,
                      dispatch_uid='make_admin_save')
 def make_admin_save(sender, instance, **kwargs):
@@ -2286,15 +2302,25 @@ class DBList(object):
 
         retVal   = cache.get(cache_id) # get the cached result
         if self.QObject: # if there is a q object we can just
-            if not self.totalnum:
+            if not self.totalnum or cache.get(cache_id) is None:
                 if override:
                     self.totalnum = ESPUser.objects.filter(self.QObject).distinct().count()
                     cache.set(cache_id, self.totalnum, 60)
+                    # register cache key
+                    keys = cache.get('DBListCount_keys', [])
+                    if cache_id not in keys:
+                        keys.append(cache_id)
+                        cache.set('DBListCount_keys', keys, 60*60*24*7)
                 else:
                     cachedval = cache.get(cache_id)
                     if cachedval is None:
                         self.totalnum = ESPUser.objects.filter(self.QObject).distinct().count()
                         cache.set(cache_id, self.totalnum, 60)
+                        # register cache key
+                        keys = cache.get('DBListCount_keys', [])
+                        if cache_id not in keys:
+                            keys.append(cache_id)
+                            cache.set('DBListCount_keys', keys, 60*60*24*7)
                     else:
                         self.totalnum = cachedval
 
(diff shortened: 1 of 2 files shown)


deepaksoni47:
Great fix direction, but can we hook invalidation to User as well (or instead of only ESPUser) so we don’t miss real save/delete paths and end up with stale counts?

esp/esp/users/models/__init__.py
```diff
@@ -2271,15 +2302,25 @@ def count(self, override = False):
...
-            if not self.totalnum:
+            if not self.totalnum or cache.get(cache_id) is None:
                 if override:
                     self.totalnum = ESPUser.objects.filter(self.QObject).distinct().count()
                     cache.set(cache_id, self.totalnum, 60)
+                    # register cache key
+                    keys = cache.get('DBListCount_keys', [])
+                    if cache_id not in keys:
+                        keys.append(cache_id)
+                        cache.set('DBListCount_keys', keys, 60*60*24*7)
```
Oval17, line 2313: There is a possibility of race condition here: two concurrent requests both read `DBListCount_keys`, append different keys, one overwrites the other. 
Using an atomic cache operation will help here.

esp/esp/users/models/__init__.py
```diff
@@ -1291,11 +1286,27 @@ def makeAdmin(self):
...
+                     dispatch_uid='invalidate_DBListCount_caches_save')
+def invalidate_DBListCount_caches_on_save(sender, instance, **kwargs):
+    invalidate_DBListCount_caches()
+
+@dispatch.receiver(signals.post_delete, sender=ESPUser,
+                     dispatch_uid='invalidate_DBListCount_caches_delete')
+def invalidate_DBListCount_caches_on_delete(sender, instance, **kwargs):
+    invalidate_DBListCount_caches()
+
 @dispatch.receiver(signals.post_save, sender=User,
```
Oval17, line 1310: Agreed with @deepaksoni47 signal is only on ESPUser. A direct User.objects.filter(...).update(...) bypasses ESPUser.post_save entirely, leaving the cache stale. Add the same signal on sender=User (matching the existing make_admin_save pattern.

esp/esp/users/models/__init__.py
```diff
@@ -2271,15 +2302,25 @@ def count(self, override = False):
 
         retVal   = cache.get(cache_id) # get the cached result
         if self.QObject: # if there is a q object we can just
-            if not self.totalnum:
+            if not self.totalnum or cache.get(cache_id) is None:
```
Oval17, line 2305: There is possibility of a bug here when lets say `self.totalnum` is already set and cache was just invalidated, `override` is False so DB is not re-queried then `cache.set(cache_id, self.totalnum, 60)` as we can see below writes the stale value back into cache, undoing the invalidation. Must reset `self.totalnum` unconditionally when cache miss is detected, not just when `override=True`.

Oval17 (changes requested):
Tests only cover ESPUser save/delete. Add a test using `User.objects.filter(...).update(...)` to cover the signal gap.

esp/esp/users/tests.py
```diff
@@ -198,6 +199,53 @@ def test_run(self):
...
+        count_after_create = dblist.count()
+        self.assertEqual(count_after_create, count_before + 1,
+                        "Count should increase by 1 after user creation")
+        
+        # Delete the user
+        user.delete()
+        count_after_delete = dblist.count()
+        self.assertEqual(count_after_delete, count_before,
+                        "Count should return to baseline after user deletion")
+
```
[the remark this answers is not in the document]
MRRanjan82: Thank you for the review! I will fix this in the next commit.

hwatheod:
fixes same issue as #4517

willgearty:
Duplicate of https://github.com/learning-unlimited/ESP-Website/pull/4517

PR #5618, opened 2026-04-01
Fix ValueError by passing RecordType instance instead of string in createBit Function

### LINKED ISSUE #5617: [Bug]: ValueError in Student Registration (createBit method)
### Description

I was testing the student registration flow for Splash 2007 and ran into a ValueError when submitting the form. It looks like the code is trying to save a string ('attended') directly into a Foreign Key field (event), which Django doesn't allow—it expects an actual RecordType object instance.

### Steps to Reproduce

1. Navigate to the "New Student Registration" page (e.g., /create/splash/2007/onsite_create).
2. Fill out the student details (First Name, Last Name, Email, etc.).
3. Select a Grade and check the required boxes (Paid, Medical, Liability).
4. Click "Save Registration".
5. The application crashes with a ValueError.

### Expected Behavior

The student should be registered successfully, and a new record bit should be created in the database.

### Actual Behavior

The system returns a ValueError:

Cannot assign "'attended'": "Record.event" must be a "RecordType" instance.

### Screenshots

Image

Image

### Operating System

_No response_

### Browser

_No response_

### Additional Context

After digging into the stack trace, I found that the issue is in esp/esp/users/models/__init__.py inside the createBit method.

The code currently does this:
```
cls.objects.create(user=user, event=extension.lower(), program=program)
```

Since extension.lower() returns a string (like 'attended'), and the event field is a ForeignKey to RecordType, Django fails because it needs the actual object from the database, not just the name of the event.

The Solution should look like this:
```
# Fetch the object first
event_obj = RecordType.objects.get(name=extension.lower())
cls.objects.create(user=user, event=event_obj, program=program)
```

I initially thought this was a local setup issue, so I tried a fresh docker-compose up and ran migrations, but the error persists because it's a logic error in the model method.

@willgearty @kkbrum , could you please confirm if this fix aligns with how you want the RecordType handled? If so, could you please assign this issue to me?

kishorprajapati1212:
## Description

While testing the student registration flow for Splash 2007, I encountered a ValueError that prevented new registrations from being saved. After investigating the stack trace and the code in esp/esp/users/models/__init__.py, I found that the createBit method was attempting to pass a raw string (e.g., 'attended') into the event field.

Because the event field is defined as a ForeignKey to the RecordType model, Django requires an actual model instance rather than a string. I have updated the logic to fetch the correct RecordType object before calling create, which resolves the crash.

## Related Issue

Closes #5617 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [ ] Existing tests pass
- [x] Manually verified

## AI Disclosure

I used Gemini to help format this PR description and to double-check the technical reasoning behind the Django ForeignKey assignment error.

## Checklist

- [x] Self-reviewed the code


kishorprajapati1212:
Hi @willgearty , please take a look at this PR when you have a chance. Let me know if any changes are needed and I'll update it.

Oval17 (approved):
small fix, lgtm

PR #5626, opened 2026-04-02
Fix: Filter out cancelled sections

### LINKED ISSUE #845: Lottery deals incorrectly with rejected classes
If a class has some sections rejected, and students have marked a StudentSubjectInterest for the class, the lottery will still get their preferences for it, but won't know about the sections in question.  This will cause some array lookups to return -1, which the lottery will later pass as an array index, where it will be interpreted as the last section the lottery knows about.  (For Splash, this was a walk-in, which was why we noticed.)  Both the incorrect query causing this and the unhandled error should be fixed.  I will take a look at this over break, but I'm putting it here in case I don't.

Gsecurity1:
Fixes #845

## Problem
The lottery system includes cancelled sections when fetching sections for classes.
This can lead to invalid lookups and incorrect assignments.

## Solution
Filtered out cancelled sections using `isCancelled()` so only valid sections are processed.

## Result
Prevents invalid section selection and improves correctness of the lottery process.

## Note
Additional safeguards for invalid index handling (-1) can be added in the lottery assignment logic.


deepaksoni47:
I reviewed this PR locally, filtering cancelled sections at the model level keeps things clean, all tests pass, and it’s good to go from my side.

samalarussa (changes requested):
You can filter sections directly using the class's status, eliminating the need for a loop in the first place. This current implementation is fairly inefficient.

The issue #845 also mentions that the error handling for this case needs to be updated as well, so it would nice to see that in this PR too.

Gsecurity1:
I really appreciate the review. 
I’ll work more on it.
Thank you.

PR #5631, opened 2026-04-02
Adding unittests for onsiteclasslist.py

### LINKED ISSUE #5156: [Bug]: Missing Unit Tests for OnSiteClassList (onsiteclasslist.py)
## Summary
`esp/esp/program/modules/handlers/onsiteclasslist.py` contains large amounts of onsite registration and JSON API logic, but it does not appear to have dedicated unit tests.

## Why this matters
This module provides operational data during onsite registration, including:
- catalog status JSON
- enrollment status JSON
- student status data
- class/section serialization helpers
- filtering around open classes and section capacity state

Since onsite registration is time-sensitive, regressions in these responses can be especially disruptive.

## Suggested test coverage
Add tests for:
- `catalog_status()` returning valid JSON with classes, sections, timeslots, and categories
- `enrollment_status()` returning the expected enrolled user/section tuples
- `section_data()` returning the expected serialized fields
- behavior when classes or sections are filtered by status
- authorization behavior for onsite-only access

## Acceptance criteria
- A dedicated test module is added
- Tests assert JSON structure and key fields, not just HTTP 200
- Tests cover at least one permission-protected endpoint

wickedlydev:
## Description

Added tests for section_data in onsiteclasslist.py

## Related Issue

Closes #5156 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

## AI Disclosure

AI assisted with writing code and bridge gaps while writing tests

## Checklist

- [x] Self-reviewed the code


wickedlydev:
added tests for catalog_status

esp/esp/program/modules/handlers/tests/test_onsiteclasslist.py
```diff
@@ -69,3 +71,474 @@ def test_unknown_user_id_returns_400(self):
...
+    def test_known_student_appears(self):
+        """self.students[0].id must appear in the result."""
+        resp = self._call()
+        data = json.loads(resp.content)
+        ids = [e[0] for e in data]
+        self.assertIn(self.students[0].id, ids)
+
+    def test_profile_students_sorted_first(self):
+        """Entries with has_profile == True must all come before entries with has_profile == False."""
+        resp = self._call({'q': 'student'})
```
moksha-hub, line 526: This ordering check only becomes meaningful when the search response contains both `has_profile=True` and `has_profile=False` rows. In this fixture, `setUp()` creates just five students and immediately calls `add_user_profiles()`, so the `q='student'` search is made up entirely of profiled fixture students unless some unrelated test data happens to match the same token. In the all-True case the `if false_indices and true_indices:` guard skips the assertion entirely, so the test would still pass even if `data.sort(key=lambda x: not x[3])` were removed. Could we seed at least one matching student without a profile so this actually exercises the sort behavior?

esp/esp/program/modules/handlers/tests/test_onsiteclasslist.py
```diff
@@ -69,3 +71,474 @@ def test_unknown_user_id_returns_400(self):
...
+    def test_search_returns_matching_student(self):
+        """Searching by a student's first name prefix must return that student."""
+        student = self.students[0]
+        query = student.first_name[:4]
+        resp = self._call({'q': query})
+        data = json.loads(resp.content)
+        ids = [e[0] for e in data]
+        self.assertIn(student.id, ids)
+
+    def test_search_limits_to_20_results(self):
```
moksha-hub, line 547: This test never forces the truncation path to do any work. The fixture only creates five students, so the `q='student'` search cannot produce more than 20 matches; `len(data) <= 20` will stay true even if the `data = data[:20]` cap in `students_status()` is removed. To cover the behavior the docstring describes, this needs more than 20 matching students in the search result before asserting on the capped length.

esp/esp/program/modules/handlers/tests/test_onsiteclasslist.py
wickedlydev: yup, I understood what you told, and will add one student without profile.
i thought of 2 approaches, please correct me if I'm wrong in any, one is , add user profiles to only 4 of 5 created students self.add_user_profiles(to=4) OR I create another student separately without user-profiles.

wickedlydev: done that.

esp/esp/program/modules/handlers/tests/test_onsiteclasslist.py
wickedlydev: done

Oval17 (approved):
Tests are good, some tests are shape-only assertions but its fine overall.
LGTM

PR #5632, opened 2026-04-02
Fix #1843: Scale lottery utility by class duration and update regression tests

### LINKED ISSUE #1843: Lottery "utility" metric doesn't account for class duration
If a student ranks a lot of classes in each timeblock, and ends up getting a lot of 2- or 3-hour classes, the student can get a bad screwedness stat even if they have a nearly full schedule.

vipulpandey21:
## Description

This PR fixes the lottery utility calculation to scale with total class duration instead of counting the number of sections.

Previously, student utility weights were computed based on the number of selected sections, which caused longer classes to contribute the same utility as shorter ones. This resulted in unfair allocation.

This change computes `student_utility_weights` after `section_lengths` are available and uses duration-based weighting so that longer classes contribute proportionally more utility.

## Related Issue

Closes #1843

## Type of Change

* [x] Bug fix
* [ ] New feature
* [ ] Breaking change
* [ ] Documentation update
* [ ] Refactor / cleanup
* [ ] CI/CD or build change

## Testing

* [x] Existing tests pass
* [x] New tests added (if applicable)
* [x] Manually verified

## AI Disclosure

Some assistance from AI tools was used for guidance on implementation structure and debugging, but all code changes were written, tested, and validated manually.

## Checklist

* [x] Self-reviewed the code


vipulpandey21:
Hi Mr. Will @willgearty! The failing test (ViewUserInfoTest::testUserIDSearchOneResult) appears to be unrelated to this PR.
From inspecting the test, it includes a note indicating it may fail intermittently, suggesting it is somewhat flaky. The failure (KeyError: 'location') also points to a redirect issue in user search logic, which is unrelated to the lottery utility changes in this PR.
All changes here are confined to lottery utility computation and its associated tests, so this seems to be a pre-existing intermittent issue.
Please let me know if you'd like me to investigate further.

vipulpandey21:
Hi @anandmahadev, /assign works for issues and not for pull requests. This issue (#1843) is currently assigned to me.
You may want to look for another issue, but if you’re specifically interested in this one, we can also collaborate on it together. 🙂

vipulpandey21:
Hi Mr. Will @willgearty! All checks are passing now. Here's a quick summary of what this PR does:

The root issue was that student_utility_weights was counting number of registered sections per student, while student_utilities was already accumulating per timeslot (in hours). Because of this, the screwedness ratio (1+utility)/(1+weight) was not consistent — students who registered for many long classes had higher utility compared to their weight, so they looked less screwed than they actually are.
Fix: I moved the student_utility_weights calculation after section_lengths is populated, and changed it to use matrix multiply (interest @ section_lengths + priority @ section_lengths) so weight is also in total hours, same as utility.
Also updated testStats to compute student_weight using total meeting times instead of section count, so the regression test stays consistent with the new logic.
(Pardon my English, it’s not that perfect)

esp/esp/program/controllers/lottery.py
```diff
@@ -282,8 +282,7 @@ def initialize(self):
                 self.ranks[self.student_indices[student_id], self.section_indices[section_id]] = ESPUser.getRankInClass(student_id, self.parent_classes[self.section_indices[section_id]])
 
 
-        #   Set student utility weights. Counts number of classes that students selected. Used only for computing the overall_utility stat
-        self.student_utility_weights = numpy.sum(self.interest.astype(float), 1) + sum([numpy.sum(self.priority[i].astype(float), 1) for i in range(1, self.effective_priority_limit+1)])
+        #   student_utility_weights is set later (after section_lengths is populated)
```
willgearty: This comment is no longer needed.

willgearty (changes requested):
Thanks for making this change. Just one minor cleanup required, then we should be good to merge.

vipulpandey21:
Sorry for late Mr. Will @willgearty, Done! Removed the stale comment. Please take a look when you get a chance.

willgearty (approved):
Thanks for removing that! Good to merge 👍

PR #5639, opened 2026-04-02
Fix issues in long answer survey question handling.

### LINKED ISSUE #3932: Surveys long answer question has issues
Apparently you can set a long answer question to have negative rows! This should not be allowed. Also, the long answer response box is super narrow to start (you can drag it larger). It should start wider.

Hemanntnagar:
Closes #3932 

## 1. Backend: Question validation (esp/esp/survey/models.py)
- Long Answer questions use the first pipe-separated parameter value as the HTML rows attribute (see get_params() and long_answer.html). Validation is on the model so it runs whenever Question is validated (including QuestionForm and Django admin), because ModelForm._post_clean() calls instance.full_clean().  
- Empty/missing first value, non-integers, 0, and negative values all raise ValidationError on _param_values.Closes #

## 2. UI: full-width, responsive textarea
- New stylesheet: esp/public/media/styles/survey.css — textarea.survey-long-answer uses width / max-width: 100%, box-sizing: border-box, and a small min-height for readability.
- Template: esp/templates/survey/questions/long_answer.html — added class="survey-long-answer" on the <textarea>.
- Linked survey.css where those questions are shown or previewed:
        
            - esp/templates/survey/survey.html
            - esp/templates/program/modules/surveymanagement/manage.html
            - esp/templates/program/modules/surveymanagement/import.html
            - esp/templates/program/modules/studentonsite/survey.html
            - esp/templates/program/modules/teacheronsite/survey.html (via {% block head %} on the webapp templates)

## 3. Tests (esp/esp/survey/tests.py)

- Added LongAnswerQuestionValidationTest covering valid positive rows, negative, zero, empty params, and non-numeric values via full_clean()

Note: Existing rows in the database that are already invalid will start failing validation on the next edit/save until an admin sets Rows ≥ 1. I could not run Django tests here because the environment has no Django install; run them locally with:
        python manage.py test esp.survey.tests.LongAnswerQuestionValidationTest

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

## Checklist

- [x] Self-reviewed the code


esp/esp/survey/models.py
SuyashJain17, line 1: Looks good. Just a quick thought, validation in clean() won’t run with .save() or .create(), so maybe worth noting. Also comparing with "Long Answer" directly might be a bit fragile, and accessing it here could add an extra query.

esp/public/media/styles/survey.css
SuyashJain17, line 1: Nice improvement Adding resize: vertical could help avoid layout issues.

esp/esp/survey/tests.py
SuyashJain17, line 1: Tests look nice. Maybe add a couple of edge cases like whitespace or "3.5"

Hemanntnagar: That makes sense. I'll add edge case tests for whitespace-only input and float values like "3.5" to strengthen validation coverage.

esp/esp/survey/models.py
Hemanntnagar: I’ll update the model to call full_clean() in save() so validation is always enforced.
I’ll also replace the direct "Long Answer" comparison with a constant/choice value to make it less fragile, and review the access pattern to avoid any unnecessary queries.

esp/esp/survey/tests.py
Hemanntnagar: Test case added

Hemanntnagar:
@SuyashJain17
Made all the changes. Please review the pr and let me know if any other changes are needed.

Oval17 (approved):
LGTM.
Need to verify css changes once.

Hemanntnagar:
Hiii @willgearty ,
my pr got approved by @Oval17 so can you please merge it so I can get assign new issue.

willgearty:
I'll take a look when I get a chance. Thanks for your patience (and no need to poke me).

kkbrum (approved):

PR #5651, opened 2026-04-03
Fix: add weighted room scoring and match threshold slider to AJAX scheduler room filters

### LINKED ISSUE #4962: [Feature]: AJAX Scheduler room filter weights
### Problem Statement

Pulled out from #458 about filtering rooms in the AJAX Scheduler:

> ...rather than fully filtering out classes that don't match (we frequently can't find a classroom that matches perfectly) we should have weights for each filter parameter and sort the classrooms by closeness to the filter parameters (probably a simple weighted sum of normalized distances squared). We can highlight at the top any classrooms that fully match the filter query.

### Proposed Solution

Implement new options in the room filters in the AJAX scheduler to enable weightings.

### Alternatives Considered

_No response_

### Additional Context

_No response_

Sameer0297:
Closes #4962
Summary
Replaces strict room filtering in the AJAX scheduler with a weighted scoring system and adds a match threshold slider, so admins can see the closest available rooms even when no perfect match exists.
Changes
ajaxschedulingmodule/ESP/Matrix.js

Added roomMatchScore() — computes a match score (0–1) for each room using weighted normalized distances: capacity filters use squared distance, resource/name filters use binary mismatch
Updated filteredRooms() — rooms are now sorted by score descending (perfect matches first) and filtered by the threshold slider value
Updated updateRooms() — shows a color-coded match badge on each visible room (green=100%, amber=60–99%, red=<60%) when any filter is active
Added slider event binding — threshold slider updates label and triggers room-filters-changed in real time

ajaxschedulingmodule/ajax_scheduling.html

Added Match Threshold slider below existing room filters
Slider range 0–100%, defaults to 0% (show all rooms)
Label updates live as slider moves

How It Works

Score = 1 − (weighted penalty / total weight)
Capacity: penalty = normalized distance squared
Resource/Name: penalty = 1 if mismatch, 0 if match
Rooms with score = 1.0 appear at top with green badge
Moving the slider to e.g. 70% hides all rooms below 70% match

Testing

Verified scoring logic handles all active filter combinations
Verified slider updates room list in real time
Verified badge colors update correctly per score range


Sameer0297:
Hi @willgearty @kkbrum ,

Addressed all Copilot review comments division-by-zero guard on capacity filters, parseInt radix, O(1) room id lookup replacing includes(), DOM reordering so highest-scoring rooms appear first, amber badge contrast fix, and trailing whitespace. All 7 conversations resolved. Ready for review.

PR #5656, opened 2026-04-04
Feat/1682 print schedule button do not auto check student init

### LINKED ISSUE #1682: Onsite grid's "Print schedule" button doesn't automatically check students in
We could modify `OnSiteClassList.printschedule_status` to do it. If we do, we should probably change the text of the "Print schedule" button to say something like "Check in student and print schedule", just to be explicit about the side effect.

Today, UChicago is using the script below to check in everyone who appears on the printer queue.

(The solution may share some functionality with `OnSiteCheckinModule.rapidcheckin` and/or `StudentRegCore.students`, so it might be worth factoring out any common code into a controller.)

```py
import time
from datetime import datetime
from esp.utils.models import PrintRequest
from esp.program.models import Program, ESPUser, Record

TODAY=datetime(2015, 05, 02)
PROGRAM=Program.by_prog_inst('Splash', '2015_Spring2')

while True:
    time.sleep(1)
    for x in PrintRequest.objects.order_by('time_requested').filter(time_requested__gte=TODAY).values_list('user_id', flat=True):
        record, created = Record.objects.get_or_create(user_id=x, event='attended', program=PROGRAM)
        if created:
            print '%s checked in' % ESPUser.objects.get(id=x).username
```

kartikJaiswal-hash:
Description
This PR automates the student check-in process during onsite registration (#1682 ). Previously, printing a student's schedule and marking them as "attended" were two separate manual steps.

Changes:

Modified OnsitePrintSchedules.printschedules to automatically create an attendance Record when a PrintRequest is executed.

Implemented a RecordType lookup for the "attended" event to ensure database integrity.

Added a comprehensive test suite in esp/program/modules/handlers/tests/test_onsiteprintschedules.py using mocks to verify both record creation and print request status updates.

Related Issue
Closes #1682 

Type of Change
[x] New feature
[x] Refactor / cleanup
Testing
[x] Existing tests pass
[x] New tests added (if applicable)
[x] Manually verified
Test Execution Output:
Plaintext
Ran 2 tests in 0.069s
OK
AI Disclosure
I used Gemini to review my written code.

Checklist
[x] Self-reviewed the code
[x] Updated documentation (if needed)
[x] No new warnings or errors introduced


Dhirenderchoudhary:
This currently auto-checks students in when printing, which seems opposite to the issue intent.
Also, instructions.html now has duplicate/conflicting “ready” actions, and the new test setup looks invalid ([Program](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) created without required [url](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)).

willgearty:
> This currently auto-checks students in when printing, which seems opposite to the issue intent.

No, issue https://github.com/learning-unlimited/ESP-Website/issues/1682 is stating that printing a student's schedule should check them in but it doesn't currently.

Dhirenderchoudhary:
Sorry for the previous one i got confused.
You’re right on #1682  printing should check students in.

Two suggestions :

Instructions page has duplicate/conflicting “ready” links now
New test creates Program without required url, so setup looks invalid

moksha-hub:
I tested this path with a local mock harness around `printschedules()`. If `ProgramPrintables.get_student_schedules()` raises, this branch has already marked the `PrintRequest` executed **and** saved the new `attended` record first. In my reproduction, both `req.save()` and `Record(...).save()` were called before the render failure was raised.

That means a transient schedule-generation failure now falsely checks the student in and consumes the pending print request, even though no schedule was actually produced. The attendance side effect should happen only after schedule generation succeeds, or all of this should be wrapped so the state change rolls back on failure.

kartikJaiswal-hash:
@willgearty if you can spare little time, can you pls review this

willgearty:
We've been working our way through the backlog of PRs. Will get to this when we can.

esp/esp/program/modules/handlers/onsiteprintschedules.py
```diff
@@ -89,6 +91,36 @@ def printschedules(self, request, tl, one, two, module, extra, prog):
...
+                            'time_executed': str(req.time_executed),
+                        }
+                        resp = HttpResponse(content_type='application/json')
+                        json.dump(data, resp)
+                        return resp
+                    else:
+                        return response
+
+            except Exception as e:
+                return HttpResponse(f"Error generating schedule: {e}", status=500)
```
Oval17, line 122: {e} shouldnt be used its not good practice in general. please remove them at all the places.

Oval17 (changes requested):
Please drop the stray test edit and added one imp comment.

PR #5659, opened 2026-04-04
Bugfix 5431: Fix statistics query

### LINKED ISSUE #5431: [Bug]: "What other programs have the students attended?" throws an exception
### Description

On the statistics page `/manage/statistics`, the question "What other programs have the students attended?" throws an exception:

`django.core.exceptions.FieldError: Cannot resolve keyword 'program_type' into field.`

It's due to the change #4037 .  The `program_type` is a property, not a database field.  It used to be accessed via `program.program_type`, but #4037 changed it to accessing it via `program__program_type` which would only work if `program_type` were a database field.

### Steps to Reproduce

1. Login to an admin account.
2. Go to `/manage/statistics`
3. Select the question "What other programs have the students attended?"  
4. Click "Go!"
5. See the exception above

### Expected Behavior

The statistics query should run and display the results.

### Actual Behavior

Exception described above is thrown

### Screenshots

_No response_

### Operating System

_No response_

### Browser

_No response_

### Additional Context

_No response_

logicwithharshal:
Fixes #5431

- Corrected statistics query that incorrectly used program__program_type.
- program_type is a property, not a database field.
- Replaced with program__type to resolve FieldError.
- This ensures the "What other programs have the students attended?" statistics query runs without exceptions.


Dhirenderchoudhary:
The fix in statistics.py is still incorrect because program__type is not a valid ORM field path, so the query will fail at runtime. Please update the query to use real DB fields and keep this PR focused on #5431 by removing unrelated CI and module-test additions.

hwatheod:
@logicwithharshal You have unrelated changes in this PR.  Please use one branch per PR, and only include the changes related to this particular issue in the PR.

esp/esp/program/statistics.py
```diff
@@ -265,7 +265,7 @@ def repeats(form, programs, students, profiles, result_dict=None):
             user__in=students,
             event__name='reg_confirmed',
         )
-        .values_list('user_id', 'program__program_type')
+        .values_list('user_id', 'program__url')
```
hwatheod: This is not correct. `url` is too specific.  `url` corresponds to one specific instance of a program, such as Splash Fall 2025 or Splash Spring 2026.  In this example, "type" just means Splash.  There can be many different instances of one type of program over time.

logicwithharshal: Thanks for catching that. You’re right — including program__url was too specific and doesn’t align with the intent of this query. I’ll update the PR to use program__program_type instead, so the statistics reflect the broader program type (e.g., Splash) rather than a single instance.
Also, I’ll clean up the branch to remove unrelated changes so this PR only addresses the statistics query fix.

kkbrum:
@logicwithharshal can you fix the failing tests?

PR #5667, opened 2026-04-05
Fix mutable default argument in `zip_download`

### LINKED ISSUE #5666: [Bug]: Mutable default argument in `zip_download` utility
### Description

The zip_download function in esp/esp/utils/web.py uses a mutable list ([]) as a default argument.

```def zip_download(files = [], zipname = 'files'):```

In Python, default arguments are evaluated once at module import time. If the list is ever inadvertently modified in one call, that change will persist to all future calls that use the default value. While currently only being read, this is a dangerous pattern that can lead to data leakage or state bugs if the function is ever expanded.

**Impact:** Potential for state leakage and unpredictable behavior in file downloads.

**Suggested Fix:** Change the default value to None and initialize it inside the function:
```def zip_download(files=None, zipname='files'):
    if files is None:
        files = []
```

please assign this issue to me

sanjana2505006:
Description:
This Pull Request corrects the use of a mutable list `([])` as a default argument in the `zip_download` utility function.

As discussed in the linked issue, using mutable defaults in Python is a dangerous pattern because the default list is shared across all function calls. I've updated the signature to use `None` and safely initialize the list inside the function body.

Closes #5666 

**Changes:**
Modified `esp/esp/utils/web.py` to initialize the `files` parameter within the function instead of using a default list.


deepaksoni47:
I reviewed it locally, good to go from my side.

willgearty (approved):
Looks good, thanks!

PR #5669, opened 2026-04-06
feat: Add multiweek_schedule template tag for repeating-class prograss (#3826)

### LINKED ISSUE #3826: Template tags for multi-week programs
Yale requested that we make a reusable schedule template for Sprout, but this could apply to any program with repeating classes. Doing so requires some custom Django template tags. Worth some discussion about how general to make this capability and which program(s) it should apply to

Oval17:
## Description

Adds a new `classes_by_day` template tag that expands multi-week ClassSections into one schedule row per distinct calendar date, so programs like Yale Sprout no longer need a full studentschedule.tex override just to show repeating classes on each week they meet.

- New templatetag: esp/esp/program/templatetags/schedule.py
  - `classes_by_day` tag groups meeting_times by date; single-date classes produce one row (unchanged), multi-date classes produce one row per week with that week's specific time string
  - Compulsory Event objects pass through unmodified
  - Output sorted by event start time

- Updated studentschedule.tex to load the new tag and branch on the `multiweek_schedule` TagDict program tag:
  - Tag set → uses classes_by_day, time column shows per-week date
  - Tag not set → existing loop unchanged (no regression)

Enables the opt-in by setting Tag key=multiweek_schedule on the program in the Django admin.

### How to enable
- In the Django admin, add a Tag for the program:

```
Key: multiweek_schedule
Value: True
```
That's all whats needed to be done instead of overriding the template.

## Related Issue

Closes #3826 

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [ ] Existing tests pass
- [ ] Manually verified

## Checklist

- [ ] Self-reviewed the code


Oval17:
Ready for Review and inputs.
Thanks

esp/esp/program/templatetags/schedule.py
SuyashJain17, line 1: This looks solid overall, but a few things might be worth tightening up. Calling `meeting_times` inside the loop could turn into an N+1 query if it’s not prefetched. Also, sorting only by `event.start` might give slightly inconsistent ordering when multiple classes start at the same time. One more thing to double-check is the use of `datetime.min`, as it could cause issues if `event.start` is timezone-aware. Nothing major, just a few things to make it more robust.

esp/templates/program/modules/programprintables/studentschedule.tex
SuyashJain17, line 1: The changes here look good, but I think there could be issues with compulsory events. In the multiweek branch, fields like `initial_rooms`, `prettyrooms`, and `emailcode` are accessed directly, which might not exist on `Event` objects and could cause errors. Adding a small guard like in the original logic would make this safer. Also, both branches are quite similar, so keeping them in sync later might be a bit tricky.

esp/esp/program/templatetags/schedule.py
Oval17: umm called `.meeting_times.all()` inside the loop for that reason only??

esp/templates/program/modules/programprintables/studentschedule.tex
Oval17: Its a tradeoff between simplicity and DRY, will see this once more review is done if PR actually serves whats required.

esp/esp/program/templatetags/schedule.py
SuyashJain17: Ahh I think there’s a small mix-up here

Calling `.all()` inside the loop still runs a separate query for each class, so it can turn into an N+1 situation. What I meant was prefetching `meeting_times` before this loop so everything is fetched in one go.

The logic here is fine, it just depends on how the queryset is prepared

Oval17:
Thanks for Review @SuyashJain17 !!

esp/templates/program/modules/programprintables/studentschedule.tex
SuyashJain17: I was just thinking that for compulsory events, fields like `initial_rooms` and `emailcode` might not exist, so those rows could end up showing empty values. Maybe adding a small guard there would keep things consistent.

esp/esp/program/templatetags/schedule.py
Oval17: This is standard Django, the entire point of prefetch_related is to make `.all()` calls inside loops free.

esp/esp/program/templatetags/schedule.py
```diff
@@ -0,0 +1,87 @@
...
+        for event in meeting_times:
+            by_date[event.start.date()].append(event)
+
+        is_repeating = len(by_date) > 1
+
+        if is_repeating:
+            # Yield one row per week; collapse same-day events into one time string
+            for date in sorted(by_date.keys()):
+                day_events = sorted(by_date[date], key=lambda e: e.start)
+                collapsed = Event.collapse(day_events)
```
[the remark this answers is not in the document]
Oval17: Changed to tol=datetime.timedelta(minutes=15) to match the tolerance used in ClassSection.friendly_times.

esp/esp/program/templatetags/schedule.py
```diff
@@ -0,0 +1,87 @@
...
+        else:
+            first_event = min(meeting_times, key=lambda e: e.start)
+            rows.append({
+                'cls': cls,
+                'event': first_event,
+                'day_time': ', '.join(cls.friendly_times(include_date=True)),
+                'is_repeating': False,
+            })
+
+    rows.sort(key=lambda r: r['event'].start if r['event'] else datetime.datetime.min)
```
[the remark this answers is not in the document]
Oval17: Replaced datetime.datetime.min sentinel with a tuple-based sort key (0, event.start) if event else (1,). This avoids any naive/aware comparison entirely and naturally pushes no-event rows to the end.

esp/esp/program/templatetags/schedule.py
```diff
@@ -0,0 +1,87 @@
...
+
+    Returns a list of dicts sorted by meeting start time:
+      cls          -- the ClassSection (or Event for compulsory blocks)
+      event        -- the specific Event representing this day's occurrence
+      day_time     -- human-readable time string for this occurrence,
+                      always including the date (e.g. "Sat, Oct 05, 10am--12pm")
+      is_repeating -- True when the class spans multiple distinct calendar dates
+    """
+    rows = []
+
```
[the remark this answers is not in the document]
Oval17: Added esp/esp/program/templatetags/tests.py with a ClassesByDayTagTest class covering: single-date section, multi-date section (one row per date), compulsory Event passthrough, section with no meeting times, chronological sort order, and no-event rows sorting last

Oval17:
Have addressed Review comments here.
Thanks.

esp/esp/program/templatetags/schedule.py
```diff
@@ -0,0 +1,87 @@
...
+                'is_repeating': False,
+            })
+            continue
+
+        # Group events by calendar date to detect repeating classes
+        by_date = defaultdict(list)
+        for event in meeting_times:
+            by_date[event.start.date()].append(event)
+
+        is_repeating = len(by_date) > 1
```
kkbrum, line 61: I _think_ you are assuming here that classes that repeat are assigned to multiple time slots. While logical, that's not really the case. What Yale does is that it makes one timeslot that stands in for multiple times, and assigns core (repeating) classes to that one timeslot. Here's an example of what their timeslot looks like: 
Image

However, #4852 seems to add functionality to assign a class to multiple times in the scheduler (but does not address studentschedule.tex as your PR here does). Once that is merged, I believe this PR will be more relevant (but will perhaps need some adjustments). I will come back to review this PR once the merge conflicts in #4852 are resolved and it is reviewed and merged.

Oval17: yes after 4852 we can go ahead with this.

PR #5146, opened 2026-03-16
Fix incorrect tab color rendering on identical and partial word URL m…

divyansh0-ai:
## Description
This PR fixes a bug in the fruitsalad theme where the base tab_0 color inappropriately overrides the correct tab color, as reported in #1087.

The root cause was that count_matching_chars in esp/esp/web/templatetags/main.py used naive character matching. This caused two problems:
1. Exact URLs fell back to the base tab color (because header links were strictly given priority over exact links ties).
2. Partial words (like index.html and ideas.html both having i) falsely matched, breaking directory logic.

**Fix Implemented**
- Refactored tab-matching algorithm to respect exact URL matches and directory boundaries.
- Added automated test cases (TabMatchingTest in esp/esp/web/tests.py) to confirm correct tab highlighting guarantees.

## Related Issue
Closes #1087 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] New tests added (if applicable)
- [ ] Manually verified

## Checklist

- [x] Self-reviewed the code
- [ ] Updated documentation (if needed)
- [x] No new warnings or errors introduced

PR #5674, opened 2026-04-06
Add regression tests for fruitsalad tab color fix

### LINKED ISSUE #2165: fruitsalad: first tab color not used for rest of page

Oval17:
## Description

When a top tab header_link and its first sub-link point to the same URL, the old logic defaulted to tabcolor0 (logo color) because the sub-link match was never strictly greater. Now defaults to tabcolor1 when the header_link is fully consumed, so the first tab gets its section color.

Adds regression tests for extract_theme covering the fix, deeper sub-tab matching, unmatched URLs, and the empty header_link edge case.

## Related Issue

Closes #2165

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [ ] Existing tests pass
- [ ] Manually verified

## Checklist

- [ ] Self-reviewed the code


Dhirenderchoudhary:
Tests cover the main scenarios. Checks pass locally. Looks good to merge.

esp/esp/web/templatetags/main.py
```diff
@@ -78,7 +78,10 @@ def extract_theme(url):
         num_chars_matched = count_matching_chars(url, category['header_link'])
         if num_chars_matched > max_chars_matched:
             max_chars_matched = num_chars_matched
-            tab_index = 0
+            if category['header_link'] and num_chars_matched >= len(category['header_link']) and category['links']:
```
moksha-hub: This broadens the behavior beyond the duplicate-header-link case from #2165. As written, any URL that fully matches header_link now flips to 	abcolor1 whenever the category has at least one sub-link, even if none of those sub-links match more specifically. For example, with header_link='/learn/' and first sub-link /learn/classes/, a page like /learn/randompage now returns 	abcolor1 instead of the old 	abcolor0. If the intent is only to handle the 'header_link and first sub-link are the same URL' case, this condition needs to key off that exact overlap instead of every full header-link match.

Oval17: I'll narrow it to:
```
if category['links'] and category['links'][0]['link'] == category['header_link']:
    tab_index = 1
```
This makes it trigger only when header_link and the first sub-link point to the same URL, which is exactly the overlap case the fix was intended for. 
Thanks for pointing out.

Oval17:
Hi , 
The bug seems to be resolved in #5146 by tie breaking logic, 
and its already merged and working properly.
I have only kept test cases here.
If we want we can see the TCs only and close the issue by merging this PR
or if we dont want TCs right now then we can close this pr and close the issue as well
But in either case the bug is resolved and we should close the issue.
Thanks
cc : @willgearty

kkbrum (approved):
Thanks!

PR #5682, opened 2026-04-08
Fix admin UI overflow when Tag value is too long

### LINKED ISSUE #5681: [Bug]:  Admin UI breaks when Tag value is too long (overflow issue)
## Bug Description
When a Tag has a very long value, the Django admin interface breaks.

Specifically:
- The header stretches horizontally
- The layout overflows outside the viewport
- The change page becomes partially unusable

## Root Cause
The `__str__` method of the Tag model returns the full value and target, which can be extremely long. This causes layout overflow in the Django admin UI.

## Impact
- Poor user experience in admin panel
- Difficult to view/edit tags with long values

### Steps to Reproduce

1. Go to /admin/tagdict/tag/
2. Create a new tag
3. Enter a very long string (200+ characters) in the value field
4. Save the tag
5. Open the tag change page

Result: The UI layout stretches horizontally and breaks

### Expected Behavior

The admin interface should properly contain long text values without breaking layout.

Long tag values should wrap within the container or be truncated, ensuring the page remains fully usable and visually stable.

### Actual Behavior

When a tag has a very long value, the admin UI breaks.

- The header stretches horizontally
- The layout overflows beyond the screen
- Content becomes misaligned
- Horizontal scrolling appears

This makes the page difficult to use.

### Screenshots

Image

Image

### Operating System

Windows 11 Home

### Browser

Google Chrome (latest)

### Additional Context

This issue occurs specifically when very long strings are used as tag values.

The problem is related to UI rendering and lack of proper text wrapping in the Django admin interface.

A possible fix involves limiting displayed text length in admin list view and ensuring proper word wrapping in form fields.

suryaparua-official:
## Description
This PR fixes a UI overflow issue in the Django admin interface when Tag values are too long.

Previously, long tag values caused layout breakage:
- Admin header stretched horizontally
- Page overflowed beyond viewport
- Change page became difficult to use

## Related Issue
Closes #5681

## Type of Change
- [x] Bug fix

## Testing
- [x] Manually verified

Tested by:
- Creating tags with very long values (200+ characters)
- Verifying admin list view, change page, and form behavior

## AI Disclosure
I used AI assistance for understanding and refining the solution.

## Checklist
- [x] Self-reviewed the code


deepaksoni47:
I think the `Tag.__str__()` change should be avoided here. This PR is fixing an admin UI overflow issue, but modifying `__str__()` changes the global representation of `Tag` objects.

Truncating the value and removing the target could make logs and debugging less informative. It would be better to keep the UI-specific fix in `admin.py` and leave the model’s string representation unchanged.

suryaparua-official:
Thanks for the suggestion @deepaksoni47

You're right — modifying `__str__()` would affect the global representation of `Tag`, so I reverted that change.

Initially, I handled the issue by truncating the value in `__str__()`, which fixed the UI overflow but introduced side effects in logs and other places.

Now, I’ve updated the fix to be UI-specific:

* Added CSS overrides to ensure long text wraps properly in the admin header, breadcrumbs, and messages
* Kept the model unchanged to avoid affecting global behavior

This ensures the admin interface remains usable without altering core model logic.

@willgearty  — please let me know if this approach looks good or if any further adjustments are needed.

deepaksoni47:
Thanks for the update. I rechecked the latest changes locally, and there’s still one blocker: `Tag.__str__()` is modified in `esp/esp/tagdict/models.py`, which affects the global string representation of `Tag`, not just the admin UI. This fix should stay scoped to the admin/CSS layer, so `__str__()` should remain unchanged.

suryaparua-official:
Thanks for pointing that out @deepaksoni47 — this was my mistake.

While pushing the changes earlier, I accidentally missed reverting the update in models.py as I was a bit rushed at the time. I've now corrected it and restored the original **str**() implementation.

Really appreciate you catching this and taking the time to review!

moksha-hub:
[P1] This overrides deployment-local STATICFILES_DIRS instead of providing a default

I tested this branch by importing `esp.settings` locally, and `STATICFILES_DIRS` is always forced to `[os.path.join(PROJECT_ROOT, 'esp/static')]` on the PR branch. The reason is visible in `settings.py`: `from .local_settings import *` happens at line 52, but this new assignment is at lines 280-282, so it runs *after* local settings are loaded and overwrites any deployment-specific `STATICFILES_DIRS` customization.

That makes this more than a UI-only fix; it changes settings precedence for every environment. If the goal is to provide a default static directory, this should only set `STATICFILES_DIRS` when it is undefined (or live before the local-settings import), not unconditionally overwrite local config.

suryaparua-official:
Thanks for catching this @moksha-hub

I’ve removed the unintended settings change and applied the fix via the existing media path so it works without affecting any configuration.

Really appreciate your careful review — thanks a lot!

esp/esp/tagdict/admin.py
```diff
@@ -2,7 +2,15 @@
 from django.contrib import admin
 from esp.admin import admin_site
 
+
 class TagAdmin(admin.ModelAdmin):
     list_display = ('key', 'value', 'target', )
     list_filter = ('key', 'object_id', 'content_type__model', )
+
+    class Media:
```
moksha-hub, line 52: This hard-codes the stylesheet under `/media/...`, so it bypasses `STATIC_URL` entirely. On deployments that serve static files from a subpath or CDN, Django's admin `Media` helper will emit this literal URL and the fix will never load. Could we keep this as a static-relative path (for example `styles/tag_fix.css`) so the admin picks it up through the configured staticfiles pipeline instead of assuming `/media/`?

suryaparua-official: Thanks for the suggestion @moksha-hub

I’ve updated this based on your feedback — the stylesheet is now referenced using a static-relative path and moved into the app’s static directory so it works through Django’s staticfiles pipeline.

Feel free to share any suggestions if you have.

willgearty (approved):
Works great, thanks for fixing this up!

PR #5690, opened 2026-04-10
Fix user authentication when unsubscribing

samalarussa:
## Description

Fixes a user call that erroneously allowed any user to unsubscribe any other

## Related Issue
No issue. Hotfix for important authentication issue

Closes #4310 (bogus issue to get github to let this thru)

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [ ] Existing tests pass
- [ ] Manually verified

## Checklist

- [ ] Self-reviewed the code


samalarussa:
TODO: fix redirect loop when token is wrong

esp/esp/users/views/__init__.py
```diff
@@ -225,7 +225,7 @@ def unsubscribe(request, username, token, oneclick = False):
 
     # otherwise show them a confirmation button
     # if they are logged into the correct account or the token is valid
-    if ( (request.user.is_authenticated and request.user == user) or user.check_token(token)):
+    if ( (request.user.is_authenticated and request.user == user) or request.user.check_token(token)):
```
moksha-hub: I verified this with a local token harness using `ESPUser.make_token()` / `check_token()`: this now validates the token against `request.user` instead of the unsubscribe target `user`. With the old code, `/unsubscribe/target/<attacker-token>/` is rejected because `target.check_token(attacker_token)` is false. With this PR, the same request is accepted because `request.user.check_token(token)` is true for the attacker. That changes the security property of the view: the token must continue to prove ownership of the account being unsubscribed.

moksha-hub:
`esp/esp/users/views/__init__.py` around lines 220-223

I tested the control flow here and the one-click unsubscribe vulnerability is still present. `unsubscribe_oneclick()` still calls `unsubscribe(..., oneclick=True)`, and this branch deactivates the target account immediately before any token or account validation happens:

`user.is_active = False`
`user.save()`

That means the changed condition below is never reached for one-click requests, so the reported auth bug is still exploitable. The validation needs to happen before deactivation, not after it.

samalarussa:
Also TODO: assess whether this needs to be an "and" or an "or" for authentication (leaning towards "and" in order to ensure token is checked, but want to test this myself first).

Oval17 (changes requested):
partial fix; the one-click path must validate the unsubscribe token as well.
Please add tests as well.

PR #5691, opened 2026-04-10
Add ability to wrap existing query with AND/OR condition

### LINKED ISSUE #1621: Query builder add a boolean filter at the top
Right now if you put in some query, and then realize you want to AND/OR it with something else, you have to select the AND/OR, then rebuild it, since you always build from the top down.  It would be cool if you could add an AND/OR "above" an existing filter.

Prm01:
Closes #1621

Currently, the query builder only allows users to construct filters in a top-down manner. If a user wants to add a new condition at a higher level (for example, converting A AND B into (A AND B) OR C), they must rebuild the entire query from scratch.

This PR introduces a simple UI enhancement that allows users to wrap the existing query with an AND or OR condition at the top level.

A new dropdown has been added below the generated sentence. When selected, it:

Wraps the existing query in parentheses
Combines it with a new condition using AND or OR
Updates the displayed sentence dynamically

This improves usability and reduces the need to rebuild queries manually.


Prm01:
Hi! I’ve updated the PR description to link this to issue #1621.

This PR adds a UI-based solution to allow wrapping an existing query with AND/OR at the top level, so users don’t need to rebuild queries from scratch.

Looking forward to feedback!

userAdityaa:
Hey @Prm01, there seems to be many smaller commits doing the same thing, can you please squash those commits to make a cleaner history ? 

(and) also it would be good to make the description more clearer, currently it looks very hard to read.

Prm01:
Thanks for the feedback! @userAdityaa 

Sure, I’ll squash the commits to clean up the history and update the PR description to make it clearer and easier to read.

I’ll push the changes shortly.

deepaksoni47 (changes requested):
I agree with @userAdityaa here. 
The main blocker from my side is that this PR still includes unrelated Docker/build changes (`.dockerignore`, `Dockerfile`, `docker-compose.yml`) alongside the actual query-builder UI change in `create_list.html`. Could you please remove those unrelated files from this PR and keep it scoped just to the AND/OR wrapping feature?
I also see a functional regression in `create_list.html`: this change removes the existing “search within this list” / “subset of this list” submit actions and also drops the hidden `submit_user_list` / `keys` fields that the current flow still uses in `esp.users.views.usersearch`. Could you keep the new wrap-with-AND/OR UI without removing those existing form actions/fields?

Prm01:
Thanks for the detailed feedback!

You're right — I’ll remove the unrelated Docker/build changes to keep this PR focused only on the query builder UI improvement.

Also, I didn’t intend to remove the existing “search within this list” / “subset of this list” functionality. I’ll restore those submit actions and the hidden fields (submit_user_list and keys) while keeping the new AND/OR wrapping feature.

I’ll push the updated changes shortly.

esp/templates/users/create_list.html
```diff
@@ -75,6 +74,22 @@
    generateSentence();
 }
 
+/* NEW FEATURE: wrap existing query */
+function wrapSentence() {
```
moksha-hub, line 78: [P1] This wrap-with-AND/OR flow is only cosmetic right now

I traced the submit path after these latest commits. `wrapSentence()` rewrites only the human-readable `finalsent` string after prompting for an extra condition, but it never creates any matching `operator_*`, `not_*`, or list-key inputs for that new condition. The backend in `esp.users.views.usersearch.get_user_list()` does not parse `finalsent` to build the filter; it still constructs the query entirely from `base_list`, `keys`, `operator_*`, and `not_*`. So the UI shows the user a wrapped top-level AND/OR query, but on submit the extra condition is ignored and the saved filter still reflects only the old structured selections.

Oval17 (changes requested):
There is need to split the docker workout and then rebase.
Then we can review this.

Oval17:
If author is not available due to inactivity we can close the pr.

PR #5695, opened 2026-04-10
Fix #1213: Show specific error message for non-activated account login

### LINKED ISSUE #1213: Logging into a non-activated account gives the general incorrect login page
You get:
"""
You need to log in before you can access /myesp/register/information.

Both the username and password are case-sensitive: capitalization matters!
"""

This should instead be something specific to a non-activated accounts, so users know that they should check their email for the activation link.

Manishtechbee:
## Description
When `require_email_validation` is `True`, newly registered accounts have
`is_active=False` until the user clicks the activation link sent to their email.
Previously, attempting to log in before activating showed the generic
"The password you entered is not valid" error, which gave users no indication
of what was wrong. This fix detects unactivated accounts in `get_context_data()`
and shows a specific message directing users to check their email for the
activation link.

## Related Issue
Closes #1213

## Type of Change
- [x] Bug fix

## Testing
- [x] Existing tests pass
- [x] Manually verified

`TestInactiveAccountLogin` — 3 tests, all pass
`AccountCreationTest` — 2 tests, all pass (no regression)

## AI Disclosure
Used Claude to generate regression tests.

## Checklist
- [x] Self-reviewed the code


moksha-hub:
`esp/esp/users/views/__init__.py` around lines 102-106

I also tested the wrong-password case here. Because this branch only looks up the username and account state, it does not distinguish "inactive account with correct credentials" from "inactive account with a bad password".

On this PR branch, submitting the wrong password for an inactive account still shows the activation banner and suppresses the generic password error. That makes the new message misleading, since the user can activate the account and still fail to log in because the password was wrong all along.

moksha-hub:
`esp/esp/users/views/__init__.py` around lines 102-104

I tested this branch locally with the new `get_context_data()` logic. This now treats every inactive account as "awaiting activation" whenever `require_email_validation` is enabled, because the predicate is only `not user.is_active`.

In this codebase, `is_active = False` is also used for other states like unsubscribed/deactivated accounts, so those users will now be told to check for an activation email that may not exist. We need a more specific predicate than `not user.is_active` before showing the activation banner.

Manishtechbee:
Thanks for the detailed feedback! You're right on both counts.

I think we can have the predicate to use `re.search(r'_\d+$', user.password)` 
which is more specific than just checking `not user.is_active`.

The reasoning:
- `registration.py` appends `"_%d" % userkey` (underscore + random integer)
  to the password hash ONLY when `require_email_validation=True` at 
  registration time:
```python
  user.password += "_%d" % userkey
  user.is_active = False
```
- A normal Django password hash (e.g. `pbkdf2_sha256$...`) does contain 
  underscores, but never ends with `_<digits>` — so `re.search(r'_\d+$')` 
  specifically matches only this activation marker
- Deactivated/unsubscribed accounts have normal password hashes with no 
  trailing `_<digits>`, so they correctly fall through to the generic 
  wrong-password error and won't see the activation banner

So the updated predicate is:

```python
if not user.is_active and re.search(r'_\d+$', user.password):
    context['inactive_account'] = True
```

I also removed the `Tag.getBooleanTag('require_email_validation')` check 
since the password marker is self-contained proof of the account state. 
Keeping the tag check could cause edge cases — for example if the tag is 
toggled OFF after registration, a genuinely unactivated user would still 
see the wrong-password error instead of the activation message.

The password marker alone correctly identifies unactivated accounts 
regardless of the current tag value.


Shall I go ahead and commit these changes?

Oval17 (commented):
cant find the remote branch , once we resolve the conflicts 
It should be good to go.

Manishtechbee:
sure

PR #5696, opened 2026-04-10
Fix: Handle invalid userid gracefully in ProgramPrintables onsite paths

### LINKED ISSUE #5684: [Bug]: ProgramPrintables onsite schedule and financial export paths raise 500 on invalid userid
### Description

`esp/esp/program/modules/handlers/programprintables.py` contains onsite-only paths that directly call `ESPUser.objects.get(id=request.GET['userid'])` without handling missing or invalid user IDs.

## Why this matters
These printable/export endpoints are used in admin and onsite workflows. A stale, malformed, or tampered `userid` should not result in a raw 500, especially during live event operations.

The same pattern appears in more than one onsite path, including student schedule and financial export flows.

## Steps to Reproduce
1. Visit an onsite printable/export route that uses `userid` in `ProgramPrintables`.
2. Supply a non-existent or invalid `userid` in the query string.
3. Observe that the handler attempts:
   `ESPUser.objects.get(id=request.GET['userid'])`
   directly.
4. If the user does not exist, the lookup can raise `ESPUser.DoesNotExist`.

## Expected Behavior
If `userid` is invalid or stale, the handler should fail gracefully with a user-facing error message.

## Actual Behavior
The module directly performs the user lookup and may raise an unhandled exception, resulting in a 500 response.

## Evidence
Relevant code:
- `esp/esp/program/modules/handlers/programprintables.py`
- Onsite branches in methods such as `student_financial_spreadsheet()` and `studentschedules()` fetch `ESPUser` directly from `request.GET['userid']`

## Suggested Fix
Add validation around `userid` parsing and catch `ESPUser.DoesNotExist` explicitly. Return a clean `ESPError` or similar admin-friendly response.

## Additional Context
Because this pattern appears in multiple onsite printable/export routes, fixing it may be best done consistently across all affected `userid`-based onsite branches in `ProgramPrintables`.

### Steps to Reproduce
2. Search for `request.GET['userid']`.
3. Review the onsite-only branches that fetch `ESPUser` directly.
4. Note the lack of defensive handling for invalid IDs.

### Expected Behavior
Onsite printable/export routes should handle invalid `userid` values gracefully.

SurbhiAgarwal1:
## Description

Fixed invalid userid handling in `student_financial_spreadsheet` and `studentschedules` methods in `esp/esp/program/modules/handlers/programprintables.py`. Added validation and error handling to catch missing, invalid, or non-existent user IDs. Instead of raising a 500 server error, the code now returns clean user-facing error messages when invalid userid is provided in onsite printable/export routes.

## Related Issue

Closes #5684

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

## AI Disclosure

No AI tools were used for this pull request.

## Checklist

- [x] Self-reviewed the code
- [x] No new warnings or errors introduced``


deepaksoni47 (changes requested):
Nice fix. It would be great to add a small regression test for the onsite `userid` cases touched here, especially missing/invalid/non-existent `userid` in `student_financial_spreadsheet` and `studentschedules`, so this behavior stays protected going forward.

deepaksoni47 (changes requested):
I reviewed this PR again after the latest changes. The intent is solid and the added regression tests are a good step forward, but there are still a few blockers.

* The `if onsite:` branches in the handlers are effectively unreachable via current routing. The dispatcher doesn’t pass `onsite=True`, so this logic won’t run. Recommend deriving it inside the handler using `onsite = (tl == 'onsite')`.
* The new tests are still hitting `/manage/` paths, which trigger non-onsite logic. They should use `/onsite/` URLs and ensure proper onsite permissions in `setUp`.
* There’s inconsistency in validation coverage. `OnsitePrintSchedules` directly calls `get_student_schedules`, so similar validation should be applied there as well.

Also, tests are still failing, so this should remain in **draft** until CI is green.

Test status:

* `esp.program.modules.tests.programprintables.TestOnsiteUseridValidation`: failed (6 failures)

deepaksoni47 (commented):
Nice work on fixing the `onsite` reachability by deriving it from the `tl` parameter! This ensures the validation logic is now reachable via onsite routes.

However, we still have a blocker: the regression tests were removed in the latest update. It is important to keep these tests to protect this behavior. Please re-add the tests and update them to use `/onsite/` URLs instead of `/manage/` URLs. For example:

```python
response = self.client.get('/onsite/%s/studentschedules?userid=invalid' % self.program.getUrlBase())
self.assertIn(b'Invalid userid format', response.content)

**Comment by SurbhiAgarwal1** [comment]
Thanks for pointing that out, @deepaksoni47! The tests were accidentally removed during refactoring. I'll restore them and update the URLs to use `/onsite/` as suggested — will have it up in the next commit.

**Review by deepaksoni47 — COMMENTED** [review]
I re-reviewed after the latest update, but I don’t see any changes addressing the missing regression tests.

If this was pushed for review accidentally, no worries. Otherwise, this is still a blocker. Please re-add the tests using `/onsite/` paths to cover this behavior.

File: esp/esp/program/modules/tests/programprintables.py [hunk 1]
```diff
@@ -229,3 +229,82 @@ def test_class_rooms_format(self):
...
+
+    def test_student_financial_spreadsheet_onsite_success(self):
+        """Test student_financial_spreadsheet returns success with valid userid in onsite path"""
+        self._login_admin()
+        user = self.students[0]
+        response = self.client.get('/onsite/%s/student_financial_spreadsheet?userid=%s' % (self.program.getUrlBase(), user.id))
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response['Content-Type'], 'text/csv')
+
+    def test_onsiteprintschedules_userid_success(self):
```
moksha-hub, line 364: This only covers the happy path for the new `printschedules?userid=...` branch. The handler change here now routes through `ProgramPrintables.get_onsite_student()`, which has three new failure modes (`missing`, `invalid`, and `not found` userid), but none of those are exercised for `printschedules`. That means the exact error-handling behavior added in this PR can regress here without any test failing, even though the same negative cases are already covered for `studentschedules` and `student_financial_spreadsheet`. Could we add the three matching negative tests for this path as well?

SurbhiAgarwal1: I've added the matching negative tests for the printschedules path in ProgramPrintablesModuleTest. These now cover when a userid is missing, invalid, or simply not found in the DB. I also updated the handler to ensure these validation failure modes are properly exercised.

deepaksoni47 (commented):
Thanks for picking up on all the feedback and pushing the fixes through.

I went through the latest changes and checked the flows, including the edge cases. Everything looks stable now and behaves as expected.

All good from my side 👍

Oval17 (approved):
resolved conflicts , LGTM

PR #5702, opened 2026-04-10
Fix seed_dummy_data QuestionType names and add template validation

### LINKED ISSUE #5675: [Bug]: seed_dummy_data uses invalid survey `QuestionType`s
### Description

https://github.com/learning-unlimited/ESP-Website/pull/4448 added a management command to seed standardized dummy data into a development environment. 

As part of this, it adds survey question types to the database:
https://github.com/learning-unlimited/ESP-Website/blob/5a1ca3bb6224ee9800a1fcddc5ee08ab229cb012/esp/esp/program/management/commands/seed_dummy_data.py#L277-L278

However, these survey question types don't match the templates that we have in the codebase: https://github.com/learning-unlimited/ESP-Website/tree/main/esp/templates/survey/questions

`QuestionType`s must match 1-to-1 to templates:
https://github.com/learning-unlimited/ESP-Website/blob/5a1ca3bb6224ee9800a1fcddc5ee08ab229cb012/esp/esp/survey/models.py#L174-L176

So when you go to fill out or edit a survey from one of the seeded programs, you get a `TemplateDoesNotExist` error.

Pjoshi0805:
Fixes #5675 

Problem:
seed_dummy_data.py was creating QuestionType objects with names that
didn't match any existing templates in survey/questions/, causing a
TemplateDoesNotExist error when filling out or editing seeded surveys.

Changes:

1. Fix seed data (seed_dummy_data.py)
   Updated _seed_lookup_data and _create_survey to use correct names:
   - 'yes/no' -> 'yes-no response'
   - 'rating' -> 'numeric rating'
   - 'open response' -> 'long answer'
   - 'multiple choice' unchanged (already correct)

2. Add model-level validation (survey/models.py)
   - Added clean() to QuestionType that checks a matching template
     exists before saving, raising ValidationError if not found.
   - Added save() override to call full_clean() so validation runs
     on every save, not just form submissions.

Testing:
- Run seed_dummy_data and verify no errors
- Navigate to a seeded survey and verify it renders correctly
- Try creating a QuestionType with an invalid name and verify
  a ValidationError is raised

### DIFF
--- a/esp/esp/program/management/commands/seed_dummy_data.py
+++ b/esp/esp/program/management/commands/seed_dummy_data.py
@@ -274,7 +274,7 @@ class Command(BaseCommand):
             defaults=dict(admin_title='Teacher Profile Editor',
                           seq=0, required=True, choosable=1))
 
-        for name in ['yes/no', 'rating', 'open response', 'multiple choice']:
+        for name in ['yes-no response', 'numeric rating', 'long answer', 'multiple choice']:
             QuestionType.objects.get_or_create(name=name)
 
     # ── navbar ────────────────────────────────────────────────────────────────
@@ -619,8 +619,8 @@ class Command(BaseCommand):
             name=f'Dev Survey - {program.name}', program=program,
             defaults=dict(category='learn'))
 
-        rt = QuestionType.objects.get(name='rating')
-        ot = QuestionType.objects.get(name='open response')
+        rt = QuestionType.objects.get(name='numeric rating')
+        ot = QuestionType.objects.get(name='long answer')
 
         # Question has: survey, name, question_type (FK), seq, per_class
         # There is no 'question' text field — the name IS the question label.
--- a/esp/esp/program/management/commands/seed_dummy_data.py
+++ b/esp/esp/program/management/commands/seed_dummy_data.py
@@ -274,7 +274,7 @@ class Command(BaseCommand):
             defaults=dict(admin_title='Teacher Profile Editor',
                           seq=0, required=True, choosable=1))
 
-        for name in ['yes-no response', 'numeric rating', 'long answer', 'multiple choice']:
+        for name in ['Yes-No Response', 'Numeric Rating', 'Long Answer', 'Multiple Choice']:
             QuestionType.objects.get_or_create(name=name)
 
     # ── navbar ────────────────────────────────────────────────────────────────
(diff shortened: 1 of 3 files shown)


deepaksoni47 (changes requested):
I reviewed this PR locally. The fix in `seed_dummy_data.py` looks correct and maps `QuestionType` names to the right templates.

However, there’s a blocker in the tests. `QuestionTest` and `AnswerTest` are failing due to a `TypeError` caused by using `@patch` at the class level, which passes a mock into `setUp()` incorrectly.

Suggested fix: move the patch to individual test methods or handle it inside `setUp` using `self.patcher.start()` and stop it in `tearDown`.

Also, please keep the PR in draft while tests are failing and move it to ready only after CI passes.

Pjoshi0805:
@willgearty  The Migration Consistency Check failure is a CI infrastructure issue the deadsnakes PPA failed to install Python 3.7 before reaching our code. Our `models.py` changes only add methods and imports, no field changes, so no migration is needed.

deepaksoni47:
Verified locally. The seed data mapping and template validation are correct. The test fixes resolve the `setUp()` conflict, and all  tests pass. Looks good to me👍

esp/esp/program/management/commands/seed_dummy_data.py
```diff
@@ -274,7 +274,7 @@ def _seed_lookup_data(self):
             defaults=dict(admin_title='Teacher Profile Editor',
                           seq=0, required=True, choosable=1))
 
-        for name in ['yes/no', 'rating', 'open response', 'multiple choice']:
+        for name in ['yes-no response', 'numeric rating', 'long answer', 'multiple choice']:
```
moksha-hub: These new names no longer match the canonical `QuestionType.name` values used elsewhere in the survey stack. The existing templates / JS still branch on title-cased names like `Yes-No Response`, `Multiple Choice`, `Numeric Rating`, and `Labeled Numeric Rating` (for example `templates/survey/answers.html` and `public/media/scripts/program/modules/survey_manage.js`). Seeding lowercase variants here means the dummy-data surveys can take a different code path from real data, even though the template filenames still resolve. It would be safer to keep the seeded names aligned with the existing title-cased question types instead of introducing parallel lowercase names.

PR #5705, opened 2026-04-10
added unit tests for TeachrPreviewModule

### LINKED ISSUE #5168: [Bug]: Missing Unit Tests for TeacherPreviewModule (teacherpreviewmodule.py)
## Summary
`esp/esp/program/modules/handlers/teacherpreviewmodule.py` contains logic for teacher printables and schedule previews, but it does not have dedicated unit tests.

## Why this matters
This module does more than display a static page:
- resolves the printables module dynamically
- builds teacher schedule handout data
- supports admin impersonation via `?user=`
- separates teacher-only and teacher-or-moderator schedule flows
- raises an error when no printables module is available

That makes it a good target for focused tests.

## Suggested test coverage
Add tests for:
- teacher schedule generation when a printables module exists
- moderator-inclusive schedule generation
- admin `?user=` behavior
- error behavior when no printables module can be resolved
- filtering out unscheduled or inactive classes from schedule output

## Acceptance criteria
- A dedicated test file is added under `esp/esp/program/modules/tests/`
- Tests cover both success and failure paths
- Tests verify schedule contents, not just response status

rafiya618:
This pull request adds a new test file with 14 comprehensive test methods to cover the TeacherPreviewModule functionality, Closes #5168 

CHANGES MADE

New file: esp/esp/program/modules/tests/teacherpreviewmodule.py

This test file includes the following:

Test Class: TeacherPreviewModuleTest (extends ProgramFrameworkTest)

14 Test Methods covering all acceptance criteria:

test_teacher_schedule_generation_with_printables_module - Verifies teacher schedule is generated correctly when a printables module exists. Checks response status, scheditems count, required keys, and context flags.

test_moderator_inclusive_schedule_generation - Verifies moderator-inclusive schedule includes both taught and moderated classes with correct role assignments. Validates role field contains either Teacher or moderator title.

test_admin_impersonation_via_user_query_param - Verifies admin can impersonate another teacher using the ?user= query parameter. Confirms schedule contains impersonated teacher's data.

test_admin_cannot_impersonate_without_user_param - Verifies admin receives their own schedule when ?user= parameter is omitted.

test_non_admin_cannot_impersonate_other_teacher - Ensures non-admin teachers cannot bypass security and impersonate other users via query parameter.

test_error_when_no_printables_module_resolved - Confirms ESPError is raised with appropriate message when no printables module is found.

test_error_when_multiple_printables_modules_exist - Confirms ESPError is raised when multiple printables modules are found (ambiguous state).

test_filters_out_unscheduled_classes - Verifies classes without meeting_times are excluded from schedule output.

test_filters_out_classes_without_resource_assignments - Verifies classes without resource assignments are excluded from schedule output.

test_filters_out_inactive_classes - Verifies classes with status less than or equal to 0 are excluded from schedule output.

test_schedule_items_sorted_by_class - Verifies schedule items are returned in sorted order.

test_moderator_schedule_filters_correctly - Verifies moderator schedule applies the same filtering rules as teacher schedule.

test_moderator_only_schedule - Verifies moderator-only schedule displays correctly with only moderated classes and appropriate context flags.

test_schedule_context_has_required_fields - Verifies all required context fields are present and have correct types.

Helper Methods:

_assign_resources_to_classes - Assigns basic resources to classes to enable scheduling
_make_mock_request - Creates mock HTTP requests with user and query parameters

COVERAGE

All 14 tests verify actual schedule content and context data, not just HTTP response status codes. Tests include:

Response status validation
Context key presence verification
Schedule item count validation
Field value verification (teacher names, class IDs, roles)
Sorting order validation
Filtering correctness validation
Error message validation
Boolean flag validation
Type checking

REQUIREMENTS FULFILLED

Requirement 1: Dedicated test file added in correct location (esp/esp/program/modules/tests/)
Requirement 2: Tests cover both success and failure paths (9 success, 5 failure/edge case tests)
Requirement 3: Tests verify schedule contents, not just response status (all tests verify context data)

METHODS TESTED

teacherhandout - 11 dedicated tests covering all code paths
teachermoderatorhandout - 2 dedicated tests
moderatorhandout - 1 dedicated test

HOW TO RUN

python -m pytest esp/program/modules/tests/teacherpreviewmodule.py -v

Or using Django test runner:
python manage.py test esp.program.modules.tests.teacherpreviewmodule --verbosity=2


deepaksoni47 (changes requested):
I reviewed this PR locally. The test coverage is strong and well thought out, especially around edge cases like impersonation and filtering.

However, there are blocking issues:

* All tests are failing due to `AttributeError` from using `response.context` when calling handlers directly. Since this bypasses Django’s test client, context isn’t populated. Recommend switching to `self.client` and hitting actual URLs so both context and routing are properly tested.
* There are redundant imports (`ProgramModule`, `ProgramModuleObj`) inside `setUp` causing a lint (F401) failure.

Minor:

* Mocking `isAdmin` on `request.user` may be brittle. Prefer using a real admin user from the test setup.

Also, please keep the PR in **draft** while tests are failing and move it to ready only after CI passes.

deepaksoni47 (commented):
I reviewed the latest changes again. The test fixes look solid and all previous blockers are properly addressed.

However, I have a concern regarding the infrastructure changes introduced to fix the `psycopg2` build issue. Downgrading the CI environment and modifying install scripts feels like a heavy workaround for what is essentially a dependency problem.

A cleaner and more maintainable approach would be to avoid infra changes altogether and switch to the pre-compiled binary. This removes the need for system-level dependencies (`libpq-dev`) and should work seamlessly on the latest CI runners without modifying workflows or install scripts.


This keeps the PR focused, reduces maintenance overhead, and avoids coupling the project to a specific CI environment.

Let me know your thoughts on this approach.

rafiya618:
@deepaksoni47 
I agree the workflow/install changes are heavier than this PR needs. A cleaner approach is to move the PostgreSQL driver to psycopg2-binary in requirements and remove the CI-specific build workaround, if the project is okay with that dependency policy. If not, I’ll keep the current source-build fix

deepaksoni47:
That sounds good.

It would be best to confirm this with the maintainers on Slack before proceeding.

If they’re okay with `psycopg2-binary`, I’d prefer going with that and reverting the CI/install changes since it keeps the fix simpler and avoids infra coupling.

Otherwise, the current approach works as a fallback.

Oval17 (changes requested):
@rafiya618 
Can we have a look over CI here they are unrelated?? also there are conflicts in migration as well.

rafiya618:
@Oval17 Yeah my bad. I have resolved the CI and migration conflicts.

Oval17 (approved):
fixed file name and tests.
Once all the tests pass
we can merge this
LGTM

PR #5707, opened 2026-04-11
Fix lunch selection form not pre-filling correctly for all days

### LINKED ISSUE #1453: Lunch period selection form doesn't populate all previous choices
When I go to https://esp.mit.edu/learn/Splash/2014/select_lunch, select two lunch periods, click submit, and then return to the page by clicking on "Select Lunch Period," my lunch period choice for Saturday is not pre-filled in the form, although my choice for Sunday is.

khushpreet78141:
## Description
Assigned to this issue and now  submitting fix.
I fix an issue where the lunch selection form did not correctly pre-populate previously selected lunch periods for all days.

The root cause was comparing only the day component of datetime values (`start.day == self.day.day`), which ignores month and year and can lead to incorrect filtering.

This PR updates the logic to use full date comparison (`start.date() == self.day`) across all relevant methods to ensure accurate matching and correct pre-filling of selections.

---

## Related Issue

Closes #1453

---

## Type of Change

* [x] Bug fix
* [ ] New feature
* [ ] Breaking change
* [ ] Documentation update
* [ ] Refactor / cleanup
* [ ] CI/CD or build change

---

## Testing

* [ ] Existing tests pass
* [x] Manually verified (logic and code-level validation)

---

## AI Disclosure

Used AI assistance for guidance in debugging approach, identifying the root cause, and structuring the fix. I manually implemented and understand the changes.

---

## Checklist

* [x] Self-reviewed the code
* [ ] No new warnings or errors introduced


khushpreet78141:
Hi @willgearty,
I would like to work on this issue. I have already gone through the code and understand the root cause of the problem (incorrect filtering due to start.day == self.day.day).
I have also implemented a local fix using full date comparison (start.date() == self.day) and verified the logic flow across the related methods.
Could you please assign this issue to me so I can proceed with the PR following the proper workflow? I will be happy to make any changes based on review feedback.
Thank you

willgearty:
@khushpreet78141 you can post `/assign` on the issue to get assigned.

khushpreet78141:
Assigned to this issue and submitting fix.

khushpreet78141:
Hi  @willgearty, I’ve updated my branch and ensured everything is in sync with main. However, the required checks (Lint and Tests) are still in "Expected" state and not running. Could you please take a look or enable the workflows? Thanks!

moksha-hub:
I verified this branch with a focused local reproduction around `StudentLunchSelectionForm`. Using a target day of `2026-03-05`, I patched the lunch event queryset to include one real March 5 event and one unrelated April 5 event with the same day-of-month. On current `main`, both events are treated as matches because the code only compares `.day`; on this PR branch, only the March 5 event is offered in the form choices because the comparison uses the full date. I did not find a blocker here.

khushpreet78141:
Hi @moksha-hub, thanks for validating the fix!
Could you please approve the PR if everything looks good on your side?

khushpreet78141:
@willgearty , I’ve addressed the review comments:
Moved filtering into queryset using start__date
Replaced unsafe [0] access with ordered .first() checks
also Cleaned up unused variables
Please take another look when you have time .
Thanks.

esp/esp/program/modules/handlers/studentlunchselection.py
```diff
@@ -56,17 +56,32 @@ def __init__(self, program, user, day, *args, **kwargs):
 
         #   Set choices for timeslot field
         #   [(None, '')] +
-        events_all = Event.objects.filter(meeting_times__parent_class__parent_program=self.program, meeting_times__parent_class__category__category='Lunch').order_by('start').distinct()
-        events_filtered = [x for x in events_all if x.start.day == self.day.day]
+        
+        events_filtered = Event.objects.filter(
```
moksha-hub, line 60: The date-based comparison here looks like the right fix, but this PR still lands without a regression test for the exact bug it is addressing. The repo already has `StudentLunchSelectionTest` in `esp/esp/program/modules/tests/studentregmodules.py`, and right now it only checks `isCompleted()`. Without a fixture that creates two lunch slots on different calendar dates sharing the same day-of-month, we can still regress the prefill/save behavior later and never notice. Could we extend the existing lunch-selection tests with one cross-month case that proves the form picks the correct lunch period by full date?

khushpreet78141:
Hi @moksha-hub and @willgearty 
Added a regression test in StudentLunchSelectionTest to cover the cross-month date bug.
The test creates two lunch events on different calendar dates (March 5 and April 5) that share the same day-of-month, and verifies that only the correct date is included in the timeslot choices and initial selection.
This should prevent regressions where filtering only compares .day instead of the full date.

Oval17 (approved):
LGTM

Oval17:
There is a small lint error.

PR #5710, opened 2026-04-11
Replace memcached_multikey.CacheClass with Django's built-in cache ba…

### LINKED ISSUE #1306: memcached_multikey.CacheClass should be replaced with Django's
Since Django 1.3 there has been some form of support for cache key prefixes and cache key functions.

In Django 1.4, we can make use of these functionalities by adding `KEY_PREFIX` and `KEY_FUNCTION`, respectively, to the `CACHES` setting (see https://docs.djangoproject.com/en/1.4/topics/cache/#cache-arguments).

We can easily make use of this, and then completely remove our custom `CACHE_PREFIX` functionality. If we also completely remove `idebug` functionality (see discussion on pull request #1223), then we can completely remove `memcached_multikey`. In the latter case, #1305 would also be closed for no longer being applicable.

Gourigupta2004:
## Description
Replaces the custom `esp.utils.memcached_multikey.CacheClass` with Django's 
built-in `PyLibMCCache` backend. Since Django 1.3+, Django natively supports 
cache key prefixing via `KEY_PREFIX` in the `CACHES` setting, making the 
custom implementation unnecessary.

## Related Issue
Closes #1306

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactor / cleanup
- [ ] Documentation update
- [ ] CI/CD or build change

## Testing
- [x] Existing tests pass
- [x] Manually verified

## AI Disclosure
AI tools were used to assist in understanding the codebase, identifying 
affected files, and writing test cases for this change.

## Checklist
- [x] Self-reviewed the code


esp/esp/utils/tests.py
SuyashJain17, line 1: The new `test_cache_long_key` expects an error for long keys, while `MemcachedKeyLengthTestCase` seems to rely on long keys working.
Maybe worth aligning these two so the expected behaviour is consistent (either support long keys via hashing or update the older test accordingly).
Also noticed the date change in `test_datetime_input` (Nov 30 → Nov 1). This feels a bit unrelated to the cache refactor.
Might be cleaner to move this into a separate PR to keep things focused.

esp/esp/tests/util.py
SuyashJain17, line 1: Updating `cache.key_prefix` directly might not always affect the underlying backend depending on Django version.
Using `override_settings` or reinitializing the cache could make this more reliable.

esp/esp/local_settings.py.ci
SuyashJain17, line 1: The `PyLibMCCache` backend path might be worth double-checking against the Django version used in CI, just to avoid compatibility issues.
`TIMEOUT` isn’t set here, so it will fall back to Django’s default.
Maybe we can align it with the project’s expected timeout to keep behaviour consistent across environments.

SuyashJain17:
Also fix the failing Unit Test and Migration Consistency Check

Gourigupta2004:
Thanks @SuyashJain17 for your feedback. Let me fix everything and get back to you.

Gourigupta2004:
Hi @SuyashJain17 @willgearty, 

Please check and for any feedback do let me know!

Thanks

deploy/config_templates/local_settings.py
```diff
@@ -31,3 +30,11 @@
 SECRET_KEY = '%(secret_key)s'
```
moksha-hub, line 30: This template no longer overrides `CACHE_PREFIX`, but `esp/esp/settings.py` still uses `CACHE_PREFIX` to build the per-site temp directory name (`esptmp__<CACHE_PREFIX>_<uid>`). With this change, installs generated from this template fall back to the default `CACHE_PREFIX = "ESP"`, so multiple sites on the same machine can end up sharing the same tempdir again. The cache backend migration makes sense, but we still need a site-specific `CACHE_PREFIX` value here unless the tempdir logic is moved off that setting too.

Gourigupta2004:
Thanks @moksha-hub for the feedback. I'll handle a site-specific CACHE_PREFIX value for the tempdir logic.

PR #5711, opened 2026-04-11
REFACTOR(BaseESPUser): Replace the cmp comparison logic with python3 compatible logic.

### LINKED ISSUE #5701: [Feature]: Modernize BaseESPUser comparison methods to Python 3
### Problem Statement

The BaseESPUser class relies on Python 2's deprecated __cmp__ method and redundant manual comparison overrides to sort user objects. This creates unnecessary technical debt and uses sorting logic that is not natively supported or optimized in 
modern   Python 3

### Proposed Solution

-  Adding the @functools.total_ordering decorator to the class.
- Replacing __cmp__ with explicit __eq__ and __lt__ methods that cleanly handle comparisons (including AnonymousESPUser hierarchies) using native tuple comparison.
- Removing the deprecated __cmp__ method and the redundant manual boilerplate overrides for __gt__, __ge__, __le__, and __ne__.

### Alternatives Considered

_No response_

### Additional Context

_No response_

mn3m-24:
## Description

- Replace the usage of `__cmp__` magic method and `cmp` function for comparison in `BaseESPUser` class with modern comparison magic methods that are python3 compliant (`__lt__`, etc.).
- Fix the Legacy bug of having separate behavior in equality & hashing (names-based equality and PK-based hashing).
- Add regression tests covering all known edge cases and scenarios for the `_get_sort_key` function.

## Related Issue

Closes #5701

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

## Checklist

- [x] Self-reviewed the code

### DIFF
--- a/esp/esp/users/models/__init__.py
+++ b/esp/esp/users/models/__init__.py
@@ -335,33 +335,32 @@ class BaseESPUser(object):
         """
         return ESPUser.email_sendto_address(self.email, self.name())
 
-    def __cmp__(self, other):
-        if other is None:
-            return 1
-        # two anonymous users are equal
-        if isinstance(self, AnonymousESPUser) and isinstance(other, AnonymousESPUser):
-            return 0
-        # otherwise, rank the signed-in user higher
-        elif isinstance(other, AnonymousESPUser):
-            return 1
-        elif isinstance(self, AnonymousESPUser):
-            return -1
-        lastname = cmp(self.last_name.upper(), other.last_name.upper())
-        if lastname == 0:
-           return cmp(self.first_name.upper(), other.first_name.upper())
-        return lastname
+    def _get_sort_key(self):
+        not_anon = not isinstance(self, AnonymousESPUser)
+        # safe fallback to empty string if user is anonymous or normal user with last_name or first_name as None
+        last_name = getattr(self, 'last_name', '') or ''
+        first_name = getattr(self, 'first_name', '') or ''
+        return (not_anon, last_name.upper(), first_name.upper())
+
     def __lt__(self, other):
-        return self.__cmp__(other) < 0
-    def __gt__(self, other):
-        return self.__cmp__(other) > 0
-    def __eq__(self, other):
-        return self.__cmp__(other) == 0
+        if not isinstance(other, BaseESPUser):
+            return NotImplemented
+        return self._get_sort_key() < other._get_sort_key()
+
     def __le__(self, other):
-        return self.__cmp__(other) <= 0
+        if not isinstance(other, BaseESPUser):
+            return NotImplemented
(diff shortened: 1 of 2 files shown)


deepaksoni47 (commented):
I reviewed this PR locally. Moving to `total_ordering` and removing `__cmp__` is the right direction, and the implementation is clean overall.

However, there’s a major regression in sorting logic. The new `_get_sort_key` places `AnonymousESPUser` after regular users because `is_anon=True` evaluates higher than `False`. This reverses the previous behavior where anonymous users were ranked lower.

You’ll want to flip this ordering (e.g., use `not is_anon` or adjust the tuple) to match the original logic.

Also worth noting: existing tests don’t catch this mixed-type ordering case, so adding a regression test here would be important.

Test suite passes, but this specific behavior is currently incorrect.

SuyashJain17 (commented):
Looks like a clean and nice refactor towards Python 3. Just a couple of small suggestions — the type check in `__eq__`/`__lt__` could be simplified to `isinstance(other, BaseESPUser)`, and it might be worth quickly verifying that `AnonymousESPUser` is still hashable. Also, adding a few basic tests for sorting/comparisons would help make the behaviour clearer. Nothing blocking from my side.

esp/esp/users/models/__init__.py
```diff
@@ -1280,6 +1271,18 @@ class Meta:
...
+        return self._get_sort_key() == other._get_sort_key()
+
+    def __lt__(self, other):
+        if not isinstance(other, type(self)) and not isinstance(other, BaseESPUser):
+            return NotImplemented
+        return self._get_sort_key() < other._get_sort_key()
+
+    __hash__ = User.__hash__
+
     def makeAdmin(self):
```
moksha-hub, line 1356: This re-enables hashing for `ESPUser` while equality is still based only on the name sort key. Two different users with the same first/last name now compare equal via `_get_sort_key()`, but `User.__hash__` hashes by model identity/pk, so equal objects can end up with different hashes. That breaks basic set/dict invariants and is a behavioral change from the unhashable state we had before. If we want name-based ordering only, it would be safer to leave these objects unhashable than to restore a hash that doesn't match `__eq__`.

mn3m-24: This conflation of ordering identity with equality identity behavior existed before my PR, I didn't introduce it.
But yeah, I should fix the original bug and revert to PK-based equality to separate ordering from equality.

Oval17 (approved):
LGTM,, resolved the conflicts

willgearty (approved):
Thanks!

PR #5718, opened 2026-04-11
Fix duplicate error messages and improve required field indication for heard_about

### LINKED ISSUE #278: Account registration error message issues
When filling out the profile after creating an account, I get:
"""
heard_about:

```
If 'Other...', please provide details
```

**all**:

```
Please provide either a day phone or cell phone number in your
```

personal contact information.
"""
duplicated twice, once in <dl> or something, and once in red.
This seems ugly.

Also, apparently the website thinks "How did you hear of this program?"
is required, but it isn't displayed as such.

Irfan762:
Removed form.errors to prevent duplicate error messages. 
Added a red asterisk to the "heard_about" field to indicate it may be required. 
Improves UI clarity and avoids repeated validation messages.

Closes #278

### DIFF
--- a/esp/esp/users/forms/user_profile.py
+++ b/esp/esp/users/forms/user_profile.py
@@ -171,7 +171,7 @@ class StudentInfoForm(FormUnrestrictedOtherUser):
     dob = forms.DateField(widget=SplitDateWidget(min_year=datetime.now().year-20))
     studentrep = forms.BooleanField(required=False)
     studentrep_expl = forms.CharField(required=False)
-    heard_about = DropdownOtherField(required=False, widget=DropdownOtherWidget(choices=list(zip(HEARD_ABOUT_ESP_CHOICES, HEARD_ABOUT_ESP_CHOICES))))#forms.CharField(required=False)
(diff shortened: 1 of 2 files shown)


deepaksoni47 (commented):
Thanks for fixing the duplicate errors and label focus issue.

One blocker: the required field indicator is inconsistent. The template shows a red asterisk for “How did you hear of this program?”, but the field is still `required=False`.

This is misleading and breaks consistency with the existing form logic.

Suggested fix:

* If it should be required → set `required=True` and remove the manual asterisk
* If optional → remove the asterisk from the template

Irfan762:
Thanks for pointing this out.

You're right ,having the red asterisk while the field is `required=False` is misleading. I’ve removed the asterisk from the template to keep it consistent with the current form logic.

Let me know if you’d prefer making the field required instead happy to update accordingly.

deepaksoni47 (commented):
Looks good now, thanks for addressing this.

Removing the asterisk keeps it consistent with the current form behavior. Good to go from my side.

Irfan762:
Thanks for the review and confirmation! Appreciate your help

Venkata-Sai-Vishwanath-robo:
LGTM

willgearty (approved):
Looks good, thanks!

PR #5722, opened 2026-04-13
Fix CSRF failure view path parsing and pass program context

### LINKED ISSUE #5692: [Bug]: CSRF failure view does not correctly resolve or pass program context
### Description

While looking into the `csrf_failure` view, I noticed that it tries to extract a `Program` object from the request path, but the logic doesn’t seem to behave as expected.

The regex used for parsing doesn’t account for the leading `/` in `request.path`, so it may fail to match in normal cases. On top of that, even if a program is successfully resolved, it isn’t added to the template context, so it doesn’t actually affect the rendered response.

I also noticed that `prog` isn’t currently used anywhere in `403_csrf_failure.html`, which makes this logic feel a bit incomplete or possibly leftover from an earlier implementation. Would it make sense to pass `prog` into the template and use it there, or should this lookup be removed entirely if it’s no longer needed?

One thought I had was to simplify the parsing by using `request.path.strip('/').split('/')`, which is easier to read and maintain. That said, I realize the current regex does provide some basic validation (restricting characters before hitting the database), so it might be preferable to keep the regex and just adjust it to handle the leading `/` properly.

Overall, just wanted to flag this in case it’s unintended behavior or something that could be cleaned up 👍

### Steps to Reproduce

1. Trigger a CSRF failure on a URL that includes a program path (e.g. /something/program1/session1/...)
2. Observe the behavior inside csrf_failure view
3. Check whether:
         The regex successfully matches the path
         The resolved prog is passed to the template

### Expected Behavior

The program should be correctly extracted from the request path
The resolved Program object should be passed into the template context
The CSRF failure page should have access to program-specific data (if needed)

### Actual Behavior

The regex may fail due to the leading / in request.path
Even when a program is found, it is not added to the context
The computed prog variable is effectively unused

### Screenshots

_No response_

### Operating System

Windows 11

### Browser

Chrome

### Additional Context

A simple improvement could be:

Replacing regex parsing with request.path.strip('/').split('/')
Adding prog to the context (c['prog'] = prog)

Happy to work on a fix if this approach looks good.

Vipeen-Kumar:
## Description

This PR fixes an issue in csrf.py where the view wasn't picking up the program from the URL during a CSRF failure.

I just tweaked the prog_re regex to handle the leading / in request.path, and made sure the prog object actually gets added to the context dict c.

Just a heads-up based on the issue discussion: prog isn't actually being used in 403_csrf_failure.html right now. I left the template alone, but the backend is wired up correctly now if someone wants to use it later!

## Related Issue

Closes #5692

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

## AI Disclosure

Used AI assistance to help review the code changes and validate the approach.

## Checklist

- [x] Self-reviewed the code

### DIFF
--- a/esp/esp/web/views/csrf.py
+++ b/esp/esp/web/views/csrf.py
@@ -23,16 +23,17 @@ def csrf_failure(request, reason=""):
         from esp.utils.web import render_to_response
         from esp.program.models import Program
 
-        prog_re = r'^[-A-Za-z0-9_ ]+/([-A-Za-z0-9_ ]+)/([-A-Za-z0-9_ ]+)'
+        prog_re = r'^/?[-A-Za-z0-9_ ]+/([-A-Za-z0-9_ ]+)/([-A-Za-z0-9_ ]+)'
         match = re.match(prog_re, request.path)
+        prog = None
         if match:
             one, two = match.groups()
             try:
                 prog = Program.by_prog_inst(one, two)
             except Program.DoesNotExist:
                 prog = None
-        else:
-            prog = None
+
+        c['prog'] = prog
 
         response = render_to_response('403_csrf_failure.html', request, c)
         response = HttpResponseForbidden(str(response.content, encoding='UTF-8'),


willgearty (approved):
Looks good, thanks!

PR #5724, opened 2026-04-13
[Feature] Improve test coverage for AJAX changelog unschedule entries

### LINKED ISSUE #5717: [Feature]: Improve test coverage for AJAX changelog unschedule entries in testChangeLogUnscheduledClasses
### Problem Statement

**File:** `esp/esp/program/modules/tests/ajaxschedulingmodule.py` — Line 192

`testChangeLogUnscheduledClasses` currently only asserts that the changelog contains exactly one entry after unscheduling a class, but does not verify the *content* of that entry — specifically that it correctly records an unschedule action (i.e., `room_name == ""` and `timeslots == []`).

The `AJAXChangeLogEntry.toDict()` method in `module_ext.py` (lines 331–347) already exposes these fields in the JSON response. A future bug in `ajax_schedule_deletereg` — for example, accidentally logging a schedule action instead of an unschedule action — would not be caught by the current test.

---

### Proposed Solution

Extend the test with 4 additional assertions to validate the changelog entry's content, not just its presence.

**Before:**
```python
changelog_response = self.client.get(self.changelog_url, {'last_fetched_index': 1})
changelog = json.loads(changelog_response.content)["changelog"]
self.assertTrue(len(changelog) == 1, "Change log did not contain the unscheduled class: " + str(changelog))
#TODO: more detailed testing here
```

**After:**
```python
changelog_response = self.client.get(self.changelog_url, {'last_fetched_index': 1})
changelog = json.loads(changelog_response.content)["changelog"]
self.assertTrue(len(changelog) == 1, "Change log did not contain the unscheduled class: " + str(changelog))

entry = changelog[0]
self.assertEqual(entry['id'], section.id, "Changelog entry has wrong class section ID")
self.assertTrue(entry['is_scheduling'], "Changelog entry should be a scheduling entry")
self.assertEqual(entry['timeslots'], [], "Timeslots should be empty for an unscheduled class")
self.assertEqual(entry['room_name'], "", "Room name should be empty for an unscheduled class")
```

These assertions directly address the `#TODO` comment and ensure that the changelog entry is semantically correct — not just present.

---

### Alternatives Considered

None.

v0idheaven:
## Summary
Extends `testChangeLogUnscheduledClasses` to validate the *content* of the changelog entry after unscheduling a class, not just its presence.

## Problem
The test previously only asserted that the changelog contained exactly one entry after unscheduling a class. It did not verify that the entry was semantically correct, i.e., that it properly recorded an unschedule action with `room_name == ""` and `timeslots == []`.

A bug in `ajax_schedule_deletereg` such as accidentally logging a schedule action instead of an unschedule action, or leaving stale room/timeslot data, would go completely undetected.

## Changes
- Removed `#TODO: more detailed testing here`
- Added 4 assertions to validate the changelog entry's content:
  - `entry['id'] == section.id` - confirms the entry refers to the correct section
  - `entry['is_scheduling'] == True` - confirms the entry is tagged as a scheduling-type action
  - `entry['timeslots'] == []` - confirms no timeslots remain after unscheduling
  - `entry['room_name'] == ""` - confirms the room was cleared after unscheduling

## Testing
No new logic introduced pure test improvement. The `AJAXChangeLogEntry.toDict()` method (lines 331–347 of `module_ext.py`) already exposes all fields being asserted.

Closes #5717


Oval17 (approved):
LGTM

PR #5725, opened 2026-04-13
feat: Add Class Description Images with per-program toggles and modern catalog support

### LINKED ISSUE #546: Images in class descriptions
How cool would it be if teachers could include a little picture in their class description? It would be cool if it was as sexy as JD's Workshop Weekend https://workshopweekend.net/catalog/ww-nov-2012 , but we could also just have something very simple.

It would need some kind of program toggle so directors could not use it if they didn't want too (maybe having a catalog full of pictures for MIT splash would be too much), but I think it would be a nice touch for smaller programs.

Irfan762:
This PR implements a highly requested feature that allows teachers to include images in their class descriptions. It provides a complete, professional solution that works across all catalog views while giving directors full control.

Key Features:
Image Uploads: Adds a picture field to the ClassSubject model, allowing teachers and directors to upload images (up to 500x500) directly through the registration and edit forms.
Program-level Toggle: Directors can enable or disable this feature for specific programs using the enable_class_description_images boolean tag.
Comprehensive UI Support: Ensures images are rendered beautifully across all versions of the catalog:
Classic/Standard Catalog: Updated Django templates for the main catalog.
Modern Catalog: Updated the JSON API (JSONDataModule) and Knockout.js model (catalog-new.js) to support images in the modern interactive registration interface.
Mobile Webapp: Updated the webapp/onsite catalog view for a great mobile experience.
Polished Styling: Added CSS styling to ensure images have professional borders, subtle shadows, and proper alignment (floating right) for a premium look.
Windows Support: Includes a minor fix to django_settings.py (socket.gethostname) to allow the project to run in Windows development environments.

Files Changed:
models/class_.py: Database fields for images.
tagdict/__init__.py: Added the feature toggle.
forms/teacherreg.py: Added the image field to teacher registration.
jsondatamodule.py: API support for images.
catalog-new.js: JavaScript support for the modern catalog.
catalog.css: Visual styling for images.
Various HTML templates for rendering.
Verified: The code follows existing patterns and handles both template-based and JSON-based React/Knockout views correctly.

Closes #546


esp/templates/inclusion/program/class_catalog_core.html
```diff
@@ -29,6 +29,9 @@
     <div id="class_{{ class.id }}_content">
     <br />
     <div class="class_content">
+        {% if enable_images and class.picture %}
```
moksha-hub, line 26: [P1] The classic catalog path never receives `enable_images`

This template now checks `{% if enable_images and class.picture %}`, but the classic catalog renders it through `{% render_class_core class %}` in `inclusion/program/class_catalog.html`. `render_class_core()` still returns only the existing catalog context and never includes `enable_images`, so in this path the new image block is always falsy and will never render even when the tag is enabled and the class has an uploaded picture. The image toggle was added to `_render_class_helper()`, but this template path does not use that helper.

Irfan762: Good catch, thanks
I've updated render_class_core() to pass enable_images so the classic catalog now renders images correctly.

esp/esp/program/templatetags/class_render.py
```diff
@@ -28,7 +28,8 @@ def render_class_core(cls):
             'colorstring': colorstring,
             'show_enrollment': scrmi.visible_enrollments,
             'show_emailcodes': scrmi.show_emailcodes,
-            'show_meeting_times': scrmi.visible_meeting_times}
+            'show_meeting_times': scrmi.visible_meeting_times,
+            'enable_images': Tag.getBooleanTag('enable_class_description_images', prog)}
```
moksha-hub, line 41: `enable_images` is now sourced from a program tag inside this cached helper, but the cache dependencies below still only cover class/section/module rows. If `enable_class_description_images` gets toggled after this fragment is cached, the classic catalog can keep rendering the old state until some unrelated class dependency invalidates it. Could we add a `depend_on_model(Tag)` (or another tag-specific dependency) here, the same way the JSON endpoints in this PR do?

Oval17 (commented):
General : 
Pr description is very cluttered took a lot of time to decipher.

PR issues and Conflicts that I resolved here : 
1 - Migration 0038 will collide with other in-flight 0038_* migrations 
2 -Multipart enctype on the teacher-reg <form> not shown verify uploads actually post files.
3 -Added tests for the new field/toggle.
4- There is unrelated socket.gethostname() change bundled in (minor).

PR #5732, opened 2026-04-15
Fix paid_list crash on non-integer filter values #5731

### LINKED ISSUE #5731: [Bug]: ProgramPrintables.paid_list() crashes when filter contains a non-integer value
### Description

`esp/esp/program/modules/handlers/programprintables.py` can crash in `paid_list()` when the `filter` query parameter contains a non-integer value.

## Why this matters
The paid-list printable is an admin workflow used to review program charges and payments. A malformed or tampered `filter` value should fall back safely or raise a clean user-facing error, not crash the page.

Right now, the handler tries to parse `request.GET.getlist('filter')` as integers. If any value fails integer conversion, it sets `ids = None`, but then stores the fallback queryset in `transfers` instead of `lineitems`. A few lines later, the code always iterates over `lineitems`, which is undefined in that branch.

This means a bad `filter` value can trigger an exception instead of rendering the printable.

## Evidence
Relevant code:
`esp/esp/program/modules/handlers/programprintables.py`

Current control flow in `paid_list()`:

- parse `request.GET.getlist('filter')`
- on `ValueError`, set:
  - `ids = None`
  - `single_select = False`
- then in the `ids is None` branch assign:
  - `transfers = ...`
- but later always do:
  - `for lineitem in lineitems:`

Because `lineitems` is not assigned in that fallback branch, the function can raise an exception before rendering.

## Steps to Reproduce
2. Locate `paid_list()`.
3. Follow the branch where:
   - `'filter' in request.GET`
   - integer conversion raises `ValueError`
4. Observe that the fallback queryset is assigned to `transfers`, not `lineitems`.
5. Observe that the function later iterates over `lineitems` unconditionally.

A request like this should hit the bad path:
`...?filter=abc`

## Expected Behavior
If `filter` contains an invalid value, the handler should either:
- fall back to the unfiltered paid list, or
- show a clean user-facing error

It should not crash.

## Actual Behavior
An invalid `filter` value can leave `lineitems` undefined and cause the handler to fail before rendering the page.

## Impact
This is a real crash path in an admin printable workflow. It can be triggered by malformed query parameters, stale links, or manual URL edits.

## Suggested Fix
In the invalid-filter fallback branch, assign the queryset to `lineitems` instead of `transfers`, or otherwise normalize the branches so the later code always works with the same variable.

## Additional Context
Verified against current `main` by inspecting `paid_list()` in:
`esp/esp/program/modules/handlers/programprintables.py`

workonlly:
## Description

This PR fixes a crash in the Program Printables paid list view caused by a malformed filter query parameter (for example, `?filter=abc`). 

When filter parsing raised a `ValueError`, the fallback branch mistakenly assigned the result to `transfers` instead of `lineitems`. Later code expected to iterate over `lineitems`, causing an exception when it was left undefined. This updates `programprintables.py:139` so the invalid-filter fallback branch correctly assigns the queryset to `lineitems`.

## Related Issue

Closes #5731

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass *(Note: A broader `ProgramPrintables` test class fails in my environment due to missing `pdflatex`, but this is unrelated to the fix.)*
- [x] New tests added (if applicable) *(Added focused regression test `test_programprintables_paid_list.py`)*
- [x] Manually verified *(Ran `docker compose exec web python manage.py test esp.program.modules.handlers.tests.test_programprintables_paid_list` with successful result)*

## AI Disclosure

Drafted with the assistance of an AI (Gemini).

## Checklist

- [x] Self-reviewed the code
image

### DIFF
--- a/esp/esp/program/modules/handlers/programprintables.py
+++ b/esp/esp/program/modules/handlers/programprintables.py
@@ -128,17 +128,19 @@ class ProgramPrintables(ProgramModuleObj):
         exclude_line_items = ["Sibling discount", "Program admission", "Financial aid grant", "Student payment"]
         pac = ProgramAccountingController(prog)
         if 'filter' in request.GET:
-            try:
-                ids = [ int(x) for x in request.GET.getlist('filter') ]
-                single_select = ( len(ids) == 1 )
-            except ValueError:
-                ids = None
-                single_select = False
+            ids = []
+            for value in request.GET.getlist('filter'):
+                try:
+                    ids.append(int(value))
+                except (TypeError, ValueError):
+                    continue
 
-            if ids is None:
-                lineitems = pac.all_transfers().exclude(line_item__text__in=exclude_line_items).order_by('line_item', 'user').select_related()
-            else:
+            single_select = ( len(ids) == 1 )
+
+            if ids:
                 lineitems = pac.all_transfers().filter(line_item__id__in=ids).order_by('line_item', 'user').select_related()
+            else:
+                lineitems = pac.all_transfers().exclude(line_item__text__in=exclude_line_items).order_by('line_item', 'user').select_related()
         else:
             single_select = False
             lineitems = pac.all_transfers().exclude(line_item__text__in=exclude_line_items).order_by('line_item', 'user').select_related()
--- a/esp/esp/program/modules/handlers/programprintables.py
+++ b/esp/esp/program/modules/handlers/programprintables.py
@@ -136,7 +136,7 @@ class ProgramPrintables(ProgramModuleObj):
                 single_select = False
 
             if ids is None:
-                transfers = pac.all_transfers().exclude(line_item__text__in=exclude_line_items).order_by('line_item', 'user').select_related()
+                lineitems = pac.all_transfers().exclude(line_item__text__in=exclude_line_items).order_by('line_item', 'user').select_related()
             else:
                 lineitems = pac.all_transfers().filter(line_item__id__in=ids).order_by('line_item', 'user').select_related()
         else:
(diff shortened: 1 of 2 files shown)


Sameer-00001:
Hey @workonlly, fantastic work on this

I just reviewed the diffs. Changing the variable assignment from transfers to lineitems in the fallback block is a perfectly clean and precise fix for the unbound variable crash.

Plus for including the RequestFactory regression test to specifically cover the non-integer ?filter=abc case. The mocking is clean, and this will definitely prevent regressions in the future.

The code looks incredibly solid. Approved!

esp/esp/program/modules/handlers/programprintables.py
```diff
@@ -136,7 +136,7 @@ def paid_list(self, request, tl, one, two, module, extra, prog):
                 single_select = False
 
             if ids is None:
-                transfers = pac.all_transfers().exclude(line_item__text__in=exclude_line_items).order_by('line_item', 'user').select_related()
+                lineitems = pac.all_transfers().exclude(line_item__text__in=exclude_line_items).order_by('line_item', 'user').select_related()
```
moksha-hub: This avoids the crash, but it also changes the malformed-input behavior in a pretty broad way: if the request contains one valid line-item id and one bad value (for example `?filter=12&filter=abc`), the `ValueError` now drops into this unfiltered fallback and shows *all* transfers instead of the selected subset. That feels risky for an admin report, because a partially bad filter expands the result set rather than rejecting the request or ignoring only the malformed entry. It would be safer to validate each value individually and keep the valid ids, or return a clean error for mixed invalid input.

workonlly: Thanks, good catch. I agree the current behavior is risky for admin reporting. I will change this so any malformed filter value causes a clean validation error rather than broadening the query, and add tests for mixed valid/invalid filter input.

workonlly: done some changes please go through it and if you found some new error please share it with me 
thanks @moksha-hub

PR #5745, opened 2026-04-21
fixed the open class list

### LINKED ISSUE #2080: The open class list should not use the class cap multiplier

rafiya618:
## Description

**What I changed:**

- Added a dedicated section helper that checks fullness while ignoring cap multiplier/offset changes _(class_.py)_
- Updated the onsite scrolling list template to use that helper instead of the default fullness check _(classlist.html)_
- Updated the public open-classes fragment template the same way _(allclass_fragment.html)_

**Why this fixes the issue**

- The old template condition used class.isFull, which ultimately uses adjusted capacity (multiplier/offset) by default.
- The new condition uses class.isFullIgnoreChanges, which calls isFull(ignore_changes=True), so it compares enrollment against unadjusted capacity.
- **Result:** open class list visibility now reflects actual/base capacity, not temporary registration-throttle settings.

## Related Issue

Closes #2080 

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [ ] Existing tests pass
- [X] Manually verified

## Checklist

- [X] Self-reviewed the code


deepaksoni47 (commented):
The change looks correct and well-scoped.

A regression test for this behavior would be a good follow-up.

Good to go from my side.

Oval17 (approved):
LGTM, added tests

PR #5766, opened 2026-04-25
Fix stale filterid value in deactivationmodule.py

### LINKED ISSUE #5683: [Bug]: DeactivationModule crashes on stale or invalid filterid in deactivatefinal()
### Description

`esp/esp/program/modules/handlers/deactivationmodule.py` assumes that `request.GET['filterid']` always references a valid `PersistentQueryFilter`. If the filter is stale, deleted, or tampered with, `deactivatefinal()` can crash during a mass deactivation action.

## Why this matters
`DeactivationModule` performs bulk account deactivation. Because this is a high-impact admin action, the handler should fail safely when the saved filter is no longer valid.

Currently, the module checks that `filterid` exists in the request, but it does not guard the actual database lookup.

## Steps to Reproduce
1. Start the mass deactivation flow and reach the final confirmation step with a valid `filterid`.
2. Delete the related `PersistentQueryFilter`, or replace the `filterid` in the URL with a non-existent value.
3. Submit the confirmation form.
4. Observe that the handler calls:
   `PersistentQueryFilter.objects.get(id=request.GET['filterid'])`
   without catching missing-filter errors.

## Expected Behavior
If the selected filter no longer exists, the module should show a clear `ESPError` and avoid deactivating any users.

## Actual Behavior
A stale or invalid `filterid` can raise `PersistentQueryFilter.DoesNotExist` and result in a 500 error.

## Evidence
Relevant code:
- `esp/esp/program/modules/handlers/deactivationmodule.py`
- `deactivatefinal()` performs a direct filter lookup from `request.GET['filterid']`

## Suggested Fix
Add defensive handling around the filter lookup:
- validate that the filter exists
- catch `PersistentQueryFilter.DoesNotExist`
- return a user-friendly error instead of a raw exception

## Additional Context
This appears to be another instance of the same stale-filter issue class present in a few admin bulk-action handlers.

### Steps to Reproduce
1. Open `esp/esp/program/modules/handlers/deactivationmodule.py`.
2. Locate `deactivatefinal()`.
3. Observe the direct `PersistentQueryFilter.objects.get(...)` call.
4. Note the lack of graceful missing-filter handling.

### Expected Behavior
Mass deactivation should stop cleanly with an understandable admin-facing error if the filter is invalid.

mn3m-24:
## Description

Implements defensive handling for `PersistentQueryFilter` lookups within the `DeactivationModule`. 

Previously, `deactivatefinal()` performed a direct `objects.get()` lookup using the `filterid` from the request. If the filter was deleted, expired, or tampered with between the search and confirmation steps, the application would raise an uncaught `PersistentQueryFilter.DoesNotExist` exception, resulting in a 500 Internal Server Error. 

This change wraps the lookup in a `try...except` block, catching the specific `DoesNotExist` exception and raising a user-friendly `ESPError`. This ensures the administrative action fails gracefully with a clear message rather than a raw server crash.

## Related Issue

Closes #5683

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [ ] New tests added
- [x] Manually verified

## Checklist

- [x] Self-reviewed the code
- [ ] Updated documentation (N/A)

### DIFF
--- a/esp/esp/program/modules/handlers/deactivationmodule.py
+++ b/esp/esp/program/modules/handlers/deactivationmodule.py
@@ -64,7 +64,11 @@ class DeactivationModule(ProgramModuleObj):
             raise ESPError()('You must confirm that you want to deactivate these users.')
 
         # get the filter to use and text message to send from the request; this is set in grouptextpanel form
-        filterObj = PersistentQueryFilter.objects.get(id=request.GET['filterid'])
+        try:
+            filterObj = PersistentQueryFilter.objects.get(id=request.GET['filterid'])
+        except PersistentQueryFilter.DoesNotExist:
+            raise ESPError()('The selected filter no longer exists. Please go back and reselect a valid user group.')
+
         users = filterObj.getList(ESPUser)
 
         if not users:


esp/esp/program/modules/handlers/deactivationmodule.py
```diff
@@ -64,7 +64,11 @@ def deactivatefinal(self, request, tl, one, two, module, extra, prog):
             raise ESPError()('You must confirm that you want to deactivate these users.')
 
         # get the filter to use and text message to send from the request; this is set in grouptextpanel form
-        filterObj = PersistentQueryFilter.objects.get(id=request.GET['filterid'])
+        try:
+            filterObj = PersistentQueryFilter.objects.get(id=request.GET['filterid'])
```
moksha-hub, line 68: This still misses malformed ids. If `filterid` is something like `abc`, `get(id=...)` raises before `DoesNotExist`, so the page still 500s. Please catch invalid-value errors here too so bad query strings fail the same safe way.

PR #5767, opened 2026-04-25
Fix scheduling checks for consecutive walk-in classes

### LINKED ISSUE #1597: Checks don't handle walkins timing correctly
If someone is teaching two consecutive walk-ins, consistency checks (what you'll see if you go to one of the walk-ins's manage pages) will think they conflict, as will the teachers-teaching-two-classes-at-once scheduling check, which was recently changed to include walk-ins. (Pretty sure it's still correct to include walk ins, though, because we want to catch walk-ins which conflict with normal classes and that should work correctly right now.) We could probably do something like allowing walk-ins to overlap on one timeblock without the system complaining.

pushkarkumar06:
## Description

This PR fixes scheduling conflict handling for walk-in teacher assignments by using strict interval overlap checks.

What is fixed:
- Walk-in sections remain included in teacher conflict diagnostics.
- Consecutive classes (where one ends exactly when the next starts) are no longer treated as conflicts.
- Real overlaps are still correctly flagged.
- Added regression coverage for:
  - consecutive intervals (allowed),
  - overlapping distinct intervals (blocked),
  - open class time block handling in scheduling checks.

## Related Issue

Closes #1597

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

Details:
- Ran focused module tests successfully:
  - django test esp.program.modules.tests.ajaxschedulingmodule --settings=esp.settings --noinput

## AI Disclosure

We took a little help from AI only to refine parts of the code.  
All changes were manually reviewed and checked multiple times, and we ensured behavior matches expected outcomes.

## Checklist

- [x] Self-reviewed the code


esp/esp/program/modules/handlers/schedulingcheckmodule.py
```diff
@@ -301,28 +301,68 @@ def unapproved_scheduled_classes(self):
...
+        Two sections are consecutive (and therefore NOT a conflict) if one ends
+        exactly when the other begins (end == start).  We therefore use strict
+        less-than comparisons so that consecutive walk-in classes taught by the
+        same teacher are not incorrectly flagged.
+        """
+        s1_times = list(s1.meeting_times.all())
+        s2_times = list(s2.meeting_times.all())
+        if not s1_times or not s2_times:
+            return False
+        s1_start = min(t.start for t in s1_times)
```
moksha-hub: This helper turns every multi-block section into one big min→max interval. If a section meets 9-10 and 11-12, another class at 10-11 will be treated as overlapping even though the actual events never intersect. This needs pairwise event overlap checks instead of the outer envelope.

Sameer-00001:
Hey @pushkarkumar06, great work tackling these scheduling checks!

To build on @moksha-hub's excellent point about the 'outer envelope' issue, you'll need to drop the min() and max() aggregations and instead compare every individual block against each other.

You can replace that block of code with a standard pairwise overlap check. It would look something like this:
Screenshot 2026-04-27 013122

for t1 in s1_times:
    for t2 in s2_times:
        # Two intervals overlap if the max of their starts is strictly less than the min of their ends
        if max(t1.start, t2.start) < min(t1.end, t2.end):
            return True # Conflict found!
return False

This way, a teacher with a split schedule (like 9-10 and 11-12) will safely pass the check if a new class is scheduled in their 10-11 gap. Let us know if you need any help implementing this!

pushkarkumar06:
Hii @willgearty  , Now this PR is ready to have your review .. , after the some suggested fix by the other contributors ..
Thanks to give you time.

willgearty:

PR #5772, opened 2026-04-27
Fix crashing of BatchClassRegModule on stale or invalid filterid in batchclassregfinal()

### LINKED ISSUE #5733: [Bug]: BatchClassRegModule crashes on stale or invalid filterid in batchclassregfinal()
### Description

`esp/esp/program/modules/handlers/batchclassregmodule.py` assumes that `request.GET['filterid']` always points to a valid `PersistentQueryFilter`. If that filter is stale, deleted, or tampered with, `batchclassregfinal()` can crash instead of failing gracefully.

## Why this matters
`BatchClassRegModule` performs bulk student registration into a selected section. This is an admin-facing mass action, so a stale saved filter should produce a recoverable error message rather than a 500.

Right now, the handler validates the selected section carefully, but does not apply the same defensive handling to the filter lookup.

## Evidence
Relevant code in `esp/esp/program/modules/handlers/batchclassregmodule.py`:

- `batchclassregfinal()` requires `filterid` in `request.GET`
- it then does:
  `PersistentQueryFilter.objects.get(id=request.GET['filterid'])`
- there is no `DoesNotExist` handling around that lookup

## Steps to Reproduce
1. Start the batch class registration flow and advance to the final step with a valid `filterid`.
2. Delete the underlying `PersistentQueryFilter`, or manually replace the `filterid` in the URL with a non-existent value.
3. Submit the final form.
4. Observe that the handler attempts the direct filter lookup without graceful error handling.

## Expected Behavior
If the filter is stale or invalid, the module should show a user-facing error explaining that the selected filter is no longer available.

## Actual Behavior
A stale or invalid `filterid` can raise `PersistentQueryFilter.DoesNotExist` and result in a 500 error.

## Suggested Fix
Wrap the filter lookup in validation and exception handling:
- validate that `filterid` is present
- catch `PersistentQueryFilter.DoesNotExist`
- return a clean `ESPError` instead of letting the exception propagate

## Additional Context
I checked the current issue tracker with searches around `batchclassregmodule`, `batchclassregfinal`, and `PersistentQueryFilter` and did not find an existing bug issue for this crash path.

TheGreatMostafa:
## Description
Fixed a server error crash in the batch class registration process `batchclassregfinal`. If the passed filter id in the URL query was either stale, deleted, or altered, the direct database lookup using `PersistentQueryFilter.objects.get()` method raised an unhandled `DoesNotExist` exception.
This patch is done by a try and except clause for catching any errors caused by non-existing PersistentQueryFilter objects or incorrect ID types like the non-numeric ones or etc. In such cases, it now returns a user-friendly ESPError object with the following message: *"Selected Filter is either invalid or available no more"*.
A new automatic test case `test_batchclassregfinal_invalid_filterid` has also been implemented to detect all invalid scenarios.

## Related Issue
Closes #5733 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [ ] Manually verified

## AI Disclosure

Gemini 3.1 Pro for better Understanding of Bug

## Checklist

- [x] Self-reviewed the code

### DIFF
--- a/esp/esp/program/modules/handlers/batchclassregmodule.py
+++ b/esp/esp/program/modules/handlers/batchclassregmodule.py
@@ -99,9 +99,12 @@ class BatchClassRegModule(ProgramModuleObj):
         except (ClassSection.DoesNotExist, ValueError):
             raise ESPError()('Invalid class section selected')
 
-        filterObj = PersistentQueryFilter.objects.get(id=request.GET['filterid'])
+        try:
+            filterObj = PersistentQueryFilter.objects.get(id=request.GET['filterid'])
+        except (PersistentQueryFilter.DoesNotExist, ValueError):
+            raise ESPError()('Selected Filter is either invalid or available no more')
+        
         override_full = 'override_full' in request.POST
-
         result = self.batch_register(filterObj, section, override_full)
 
         context = {
--- a/esp/esp/program/modules/handlers/batchclassregmodule.py
+++ b/esp/esp/program/modules/handlers/batchclassregmodule.py
@@ -102,7 +102,7 @@ class BatchClassRegModule(ProgramModuleObj):
         try:
             filterObj = PersistentQueryFilter.objects.get(id=request.GET['filterid'])
         except (PersistentQueryFilter.DoesNotExist, ValueError):
-            raise ESPError()('Selected Filter is either not valid or available no more')
+            raise ESPError()('Selected filter is invalid or no longer available.')
         
         override_full = 'override_full' in request.POST
         result = self.batch_register(filterObj, section, override_full)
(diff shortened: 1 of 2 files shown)


deepaksoni47 (commented):
Nice fix on the handler, the try/except around the filter lookup looks correct.

There are a few issues in the test:

* **Error message mismatch**:
  Handler raises `"Selected filter is invalid or no longer available."` but the test expects `"Selected Filter is either invalid or available no more"`, so assertions will fail.

* **Incorrect `assertRaisesMessage` usage**:
  `ESPError()` returns a class (e.g., `ESPError_Log`), not an instance. Passing `ESPError()` here is incorrect; the test should use the actual exception class.

* **Test scope mismatch**:
  The docstring mentions “invalid or missing filterid”, but the missing `filterid` case does not exercise the new try/except logic added in this PR.

Once these are aligned, this should be good to go.

TheGreatMostafa:
Thanks for the review! i added the changes to fix the issues you wrote

deepaksoni47 (commented):
Thanks for addressing the feedback. I rechecked the latest changes and everything looks good now.

Good to go from my side.

Issue #3944, opened 2025-11-04
Automatic account deactivation


milescalabresi:
Suggestion (up for discussion): deactivate accounts that haven't been logged into for x many years. Helps the issue in #3935 and avoid mass emails to participants who haven't logged in for years.

willgearty:
Also related: #3850 (deactivate accounts when emails bounce)

yadavsidd:
Hi, I would like to work on this issue. Could you please confirm if this is open for contribution and provide guidance on where this logic should be implemented?

willgearty:
@yadavsidd Thanks for your interest in contributing to this project! I think this issue still needs a little discussion, but we hope you'll find another issue to work on!

SuyashJain17:
Thanks for the clarification!

To help move the discussion forward, would it make sense to implement this as a management command (e.g., deactivate accounts inactive for X years) rather than automatic logic on login? That way it could be run manually or via cron and remain configurable.

We could also consider adding safeguards like:

- Only targeting accounts with no recent activity (login + program participation),

- Sending a warning email before deactivation,

- Logging deactivated accounts for easy reactivation.

Happy to help think through the design if that would be useful.

willgearty:
I think a cron job probably is the best way to address this.

Supriyo760:
Hi @willgearty
I agree that a management command run via a cron job is the cleanest and safest way to handle this. I'd love to take this issue on!

Here's my proposed implementation plan:

**1. Configuration Tags:** I'll add new configuration tags to `esp/esp/tagdict/__init__.py`:
   - `inactive_account_years_threshold`: Years before an account is considered inactive
   - `inactive_account_warning_days`: Days to wait between sending the warning email and actually deactivating the account

**2. Tracking Table:** I'll create a lightweight model (e.g., `AccountDeactivationWarning`) in `esp/esp/users/models.py` to track who has been sent a warning email and when, so we don't accidentally spam them or deactivate them immediately.

**3. Management Command (`deactivate_stale_accounts.py`):** I'll create a Django management command in `esp/esp/users/management/commands/` that performs a two-pass check:
   - **Pass 1 (Warning):** Find `ESPUser`s where `is_active=True` and `last_login` exceeds the inactivity threshold, but who haven't received a warning yet. Send them an email asking them to log in to keep their account active, and log the warning timestamp in `AccountDeactivationWarning`.
[...]

PR #5814, opened 2026-05-29
Feature/deactivate bouncing emails 3850

### LINKED ISSUE #3850: Deactivate accounts when emails bounce?
I _think_ we could use the SendGrid API to identify when emails bounce and then to automatically deactivate the account associated with the email message (or _any_ accounts associated with the bounced email address). The downside for this could be that sometimes bounces occur for reasons other than an account not existing anymore (e.g., spam filters, full inbox, network issues, etc). However, based on the weekly emails we get, it looks like we could possibly only do this for certain bounce reasons (e.g., "550-5.2.1 The email account that you tried to reach is inactive").

willgearty:
This is a new version of #4464 (which was reverted in #5813)

## Description

This PR implements the automatic account deactivation pipeline for emails that hard bounce, as discussed in Issue #3850. Because bounced emails can negatively impact our sending reputation, this routinely polls the SendGrid API and turns off accounts associated with invalid or unrouteable inboxes.

### Changes Included:
* **Management Command:** Added [esp/esp/dbmail/management/commands/deactivate_bouncing_emails.py](cci:7://file:///c:/Users/Dange/Downloads/ESP-Website-main/ESP-Website-fresh/esp/esp/dbmail/management/commands/deactivate_bouncing_emails.py:0:0-0:0). 
    * This uses the existing `django-sendgrid-v5` environment settings (`SENDGRID_API_KEY`) to initialize a `SendGridAPIClient`. 
    * It calls the `suppression.bounces.get()` API to fetch the global bounce list.
* **Hard-Bounce Filtering:** To avoid accidentally deactivating accounts due to temporary network issues or full inboxes (soft bounces), the script parses the `reason` string returned by SendGrid. It strictly filters for exact hard-bounce indicators (e.g., `5.1.1`, `5.2.1`, `user unknown`, `inactive`, `does not exist`, `invalid`).
* **Cron Injection:** Modified [esp/dbmail_cron.py](cci:7://file:///c:/Users/Dange/Downloads/ESP-Website-main/ESP-Website-fresh/esp/dbmail_cron.py:0:0-0:0) to invoke this management command at the very start of the email processing cycle. I isolated it inside a `try/except` guard so that if the SendGrid API times out or fails, it will gracefully log the error and permit the standard email delivery thread to proceed unimpeded.

## Related Issue

Closes #3850

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Tested locally (confirmed syntax and Django management environment hooks are valid)
- [x] Ensured graceful exiting when `SENDGRID_API_KEY` is not configured, protecting developer environments without API access.

## AI Disclosure

**Tools used:** AI coding assistant

**Scope of assistance:** Used exclusively for minor syntax suggestions while adding the SendGrid API loop logic and formatting the Django management command template.

**Human verification:** I independently scoped the requirements, reviewed the [dbmail_cron.py](cci:7://file:///c:/Users/Dange/Downloads/ESP-Website-main/ESP-Website-fresh/esp/dbmail_cron.py:0:0-0:0) entrypoint, designed the safety try/catch mechanism around the SendGrid polling, manually authored the hard-bounce string isolation logic, and reviewed the final implementation.

## Checklist

- [x] Self-reviewed the code


willgearty:
Posting my message from #5813 just so it's here as well:

> The maintainer team has realized we need to think about user deactivation a bit more before merging this (for example, see #3944 and #4604), so I'm reverting this for now
> 
> I've also realized there's some undesirable behavior given that this process is occurring separately for each production site. Here's a quick example:
> 
> 1. A user has an account on the Yale site.
> 2. An email is sent to the user through the Yale site. It bounces. SendGrid records this as a hard bounce.
> 3. The Berkeley site is the next site to run dbmail_cron. It gets the list of bounces of Sendgrid.
> 4. While processing the list, the Berkeley dbmail_cron finds no accounts with the bounced email address. It then delete the bounce record from SendGrid.
> 5. Yale send another email to the user the next day. It bounces again. SendGrid records this as a hard bounce.
> 6. etc etc

PR #5818, opened 2026-05-29
Fixes #5815: Align boundary logic in ExpirableModel.is_valid() with valid_objects()

### LINKED ISSUE #5815: [Bug]: Inconsistent boundary check between ExpirableModel.is_valid() and valid_objects()
### Description

The boundary logic for validity checks in ExpirableModel is inconsistent:
- is_valid() uses strict start/end comparisons for start_date and end_date
- valid_objects() includes rows exactly at the boundary

This inconsistency means that for a given timestamp (when), valid_objects(when) can return a module whose start_date or end_date matches the timestamp, but obj.is_valid(when) returns false for the same object.

This can cause subtle bugs for any feature that relies on both APIs being interchangeable or assumes their results are aligned.

See discussion: https://github.com/learning-unlimited/ESP-Website/pull/5810#discussion_r3311444132

Note: Fixing boundary logic in ExpirableModel is high impact because the class is widely inherited (including StudentRegistration, Permission, etc.), so this likely requires careful codebase-wide review.

### Steps to Reproduce

1. Create an ExpirableModel object with a specific start_date or end_date.
2. Call valid_objects(when) using a 'when' datetime equal to the start_date or end_date.
3. Call obj.is_valid(when) for the same object and 'when' datetime.
4. Observe that the object may be included in valid_objects(when) but is_valid(when) returns false.

### Expected Behavior

The two APIs should align on boundary logic. Either both should include or exclude objects whose validity timestamps match the query boundary.

### Actual Behavior

valid_objects() includes objects with start_date or end_date equal to 'when', while is_valid() excludes them.

### Screenshots

_No response_

### Operating System

_No response_

### Browser

_No response_

### Additional Context

_No response_

Supriyo760:
### Description
Addresses #5815 by resolving a boundary logic inconsistency in `ExpirableModel` that was identified during the review of #5810.

### Changes

* **Method alignment:** Changed strict inequalities (`<`, `>`) to inclusive inequalities (`<=`, `>=`) in `ExpirableModel.is_valid()`. This safely aligns the Python-level object checks with the database-level `valid_objects()` method (which already used inclusive boundaries via `__lte` and `__gte`), ensuring they behave consistently across `Permission`, `StudentRegistration`, and other child models.

### Related Issue
Fixes #5815

### Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

### Testing

- [x] Existing tests pass
- [ ] Manually verified

### AI Disclosure
AI was used for suggestions only.

### Checklist

- [x] Self-reviewed the code

### DIFF
--- a/esp/esp/utils/expirable_model.py
+++ b/esp/esp/utils/expirable_model.py
@@ -62,8 +62,8 @@ class ExpirableModel(models.Model):
     def is_valid(self, when=None):
         if when is None:
             when = datetime.now()
-        return (self.start_date is None or self.start_date < when) and \
-               (self.end_date is None or self.end_date > when)
+        return (self.start_date is None or self.start_date <= when) and \
+               (self.end_date is None or self.end_date >= when)
 
     @classmethod
     def valid_objects(cls, when=None):
--- a/esp/esp/utils/tests.py
+++ b/esp/esp/utils/tests.py
@@ -831,39 +831,67 @@ class ChooseProgramTestCase(unittest.TestCase):
 
 
 class ExpirableModelBoundaryTest(DjangoTestCase):
-    """Tests to ensure ExpirableModel.is_valid() and valid_objects() have consistent boundary logic (#5815)."""
-    
+    """Tests if ExpirableModel bounds have consistent logic (#5815)."""
+
     def test_boundary_consistency(self):
         from esp.users.models import Permission
-        
+
         # 1. Setup specific boundaries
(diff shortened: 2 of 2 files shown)


willgearty:
Thanks for putting together this quick fix @Supriyo760. Could you please add a simple test to make sure the two strategies return the same result?

willgearty (approved):
Looks good, thank you!

PR #5823, opened 2026-05-30
Fix duplicate test collection in program/modules/tests

### LINKED ISSUE #5820: `program/modules` tests are collected 2–3×
Every test class under `esp/esp/program/modules/tests/` is collected at least twice, and `test_*.py`-resident classes three times. Confirmed via `pytest --collect-only`.

Three overlapping collection routes:

- `esp/esp/program/modules/tests/conftest.py` has a `pytest_collect_file` hook that collects **every** `.py` file in the directory (added so pytest finds the module-named files like `existence.py` that don't match `python_files`).
- `esp/esp/program/modules/tests_modules.py` is `from esp.program.modules.tests import *`; it matches `python_files` (`tests_*.py`), so pytest collects it as a module and finds every class the package `__init__.py` re-exports.
- pytest's built-in `python_files` collection independently picks up the `test_*.py` files.

pytest does not dedup by path across `pytest_collect_file` implementations, so module-named files = hook + aggregator = 2×, and `test_*.py` files = built-in + hook + aggregator = 3×.

**Suggested fix:** keep the conftest hook (it's explicit, scoped, and handles the `ajaxstudentreg.py`/selenium skip) and remove the redundancy:

1. Delete `tests_modules.py` (grep confirms nothing else imports it or `from esp.program.modules.tests import`).
2. Make the conftest hook skip files matching `python_files` so it only collects the module-named files, eliminating the `test_*.py` double-collection.

vipulpandey21:
## Description

Every test under `esp/program/modules/tests/` was being collected 2–3× by pytest because three routes all pointed to the same tests:

- The `conftest.py` hook collected **all** `.py` files, including `test_*.py` ones
- `tests_modules.py` (`from esp.program.modules.tests import *`) matched pytest's `tests_*.py` pattern, re-exporting every test class a second time
- pytest's built-in collection **also** picked up `test_*.py` files independently

**Two changes fix this:**

1. **Deleted `tests_modules.py`** — a single-line wildcard re-export with no other importers. Removing it eliminates one collection route entirely.

2. **Updated `conftest.py` hook** — added `and not file_path.name.startswith("test_")` so the hook only collects the non-standard named files like `programprintables.py` and `studentreg.py`. The `test_*.py` files are left to pytest's built-in collection, which already handles them.

After this fix, every test file is collected exactly once.

## Related Issue

Closes #5820

## Type of Change

- [x] Bug fix

## Testing

- [x] Existing tests pass
- [x] Manually verified

## AI Disclosure

Used Kiro (AI coding assistant) to help verify this implementation.

## Checklist

- [x] Self-reviewed the code

### DIFF
--- a/esp/esp/program/modules/tests/conftest.py
+++ b/esp/esp/program/modules/tests/conftest.py
@@ -7,8 +7,11 @@ standard test_*.py convention. Django's test runner discovers TestCase
 subclasses regardless of filename, but pytest only collects files that
 match the python_files patterns in pytest.ini.
 
-The pytest_collect_file hook below explicitly collects all .py files in
-this directory so pytest finds the same tests as manage.py test.
+The pytest_collect_file hook below explicitly collects the non-standard
+named .py files in this directory so pytest finds the same tests as
+manage.py test. Files matching pytest's built-in python_files pattern
+(test_*.py) are intentionally excluded here since pytest collects those
+natively, avoiding double-collection.
 """
 import pytest
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


willgearty (approved):
Looks good, thanks! Just ran it locally and it no longer runs any of the tests multiple times. Also looks like it shaves ~5 minutes off of our test suite runtime on GitHub 🙌

PR #5829, opened 2026-06-01
Revert form rendering to Django 4 style

willgearty:
This reverts form rendering to the Django 4 default style of using tables (the default in Django 5 uses `div`s). This fixes, for example, the rendering of the forms on the resources page. Any uses of `{{ form }}` should now behave like they did in Django 4 and earlier, and any uses of `{{ form.as_table }}` or other specific renderers should still work.


hwatheod (approved):
Looks good now.

PR #5831, opened 2026-06-01
Fix the program modules form field

willgearty:
This fixes the program modules field in the program settings form. It broke because the structure of forms changed in Django 5. Now we use jQuery instead of brittle Javascript pathing to get the correct elements.

### DIFF
--- a/esp/public/media/scripts/program/modules/modulequestions.js
+++ b/esp/public/media/scripts/program/modules/modulequestions.js
@@ -50,13 +50,13 @@ function modulesToQuestions() {
     // check if ALL the modules associated with a given question are selected
     // if so, check that box
     for (var i=0; i < questions.length; i++) {
-        var q = questions[i];
-        var qvalslist = q.firstChild.firstChild.value.split(",");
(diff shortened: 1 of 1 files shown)


hwatheod (approved):
Is there more similar broken javascript like this?

willgearty:
I'm not sure, I could try having Claude or Copilot do a sweep, I guess. Ultimately, I think we'll just need to do thorough manual testing before the next stable release.

willgearty:
Copilot suggests that some of these similar scripts might have broken: https://github.com/search?q=repo%3Alearning-unlimited%2FESP-Website+%28firstChild+OR+parent%28%29.parent%28%29+OR+textContent+OR+childNodes%29&type=code

I can take a look at these and see if any of them are indeed broken.

willgearty:
OK, I've checked every item in that list and the custom things are still working because they act on custom UIs, not Django forms (e.g., class search, custom forms, teacher checkin, extracosts, comm panel) and everything else is vendored.

PR #5834, opened 2026-06-02
Phase 2(Week 2): Add time-filtering to getModules() for non-admins (Issue #4935)

### LINKED ISSUE #4935: [Feature]: getModules_cached() does not invalidate on time-based transitions, modules may serve stale results after start_date/end_date passes
### Problem Statement

Program.getModules_cached() uses depend_on_row hooks from the argcache library to invalidate on database writes. This works correctly when a ProgramModuleObj record is explicitly saved.

However, if a module has a time-bounded validity window (e.g., start_date / end_date via ExpirableModel), the cache will not invalidate when that window opens or closes, because no DB write occurs at that moment. The cached result will continue returning the same set of modules until the next explicit save, which may be hours or days later.

This means getModules() could return modules that have already expired, or not yet return modules that have just become active, purely due to stale cache state.

### Proposed Solution

Apply time-based filtering in getModules() (the post-processing layer that wraps the cached call) rather than relying on cache invalidation to handle it. Since getModules_cached() is intentionally unbounded, the validity check should be done after the cached lookup:

# In getModules(), after calling getModules_cached():
modules = [m for m in modules if m.is_valid()]
This keeps the cache hit rate high while ensuring time-based transitions are always respected at call time.

### Alternatives Considered

- Short TTL on the cache: Would reduce the stale window but add DB load and doesn't eliminate the problem.
- Invalidating cache on a scheduled task: Adds infrastructure complexity; the post-processing filter is simpler and correct.

### Additional Context

ExpirableModel (esp/utils/expirable_model.py) already provides is_valid() and valid_objects() for exactly this purpose. ProgramModuleObj currently extends it (via ModuleBase). The fix is a one-liner in getModules() that uses the existing pattern.

Supriyo760:
### Description
Addresses #4935 by implementing Phase 2 of the Module Management Timeline UI. This update enforces the `start_date` and `end_date` boundary logic on the backend so that expired or upcoming modules are hidden from students and teachers in the dashboard, but remain fully visible to administrators.

### Changes
* **Time-Aware Filtering:** Updated the `getModules()` function in `esp/esp/program/models/__init__.py` to filter modules using `m.is_valid()`. 
* **Role-Based Bypass:** Included an explicit `user.isAdmin(self)` check so that administrators completely bypass the time constraints and can always see all modules to manage them.
* **Cache Safety:** Cast the `django-argcache` result to a new `list()` before filtering to ensure the shared global cache array isn't accidentally mutated.
* **Unit Testing:** Added a comprehensive test suite (`test_module_time_filtering.py`) that tests boundary scenarios across student, admin, and system roles for past, future, and active program modules.

### Related Issue
Fixes #4935

### Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

### Testing

- [x] Existing tests pass
- [x] Manually verified

### AI Disclosure
AI was used for suggestions only.

### Checklist

- [x] Self-reviewed the code

### DIFF
--- a/esp/esp/program/models/__init__.py
+++ b/esp/esp/program/models/__init__.py
@@ -1266,7 +1266,11 @@ class Program(models.Model, CustomFormsLinkModel):
 
     def getModules(self, user = None, tl = None, old_prog = None):
         """ Gets modules for this program, optionally attaching a user. """
-        modules = self.getModules_cached(tl, old_prog)
+        modules = list(self.getModules_cached(tl, old_prog))
+
+        if user and not user.isAdmin(self):
+            modules = [m for m in modules if m.is_valid()]
+
         if user:
             for module in modules:
                 module.user = user
--- a/esp/esp/program/models/__init__.py
+++ b/esp/esp/program/models/__init__.py
@@ -1265,7 +1265,7 @@ class Program(models.Model, CustomFormsLinkModel):
     getModules_cached.depend_on_row('modules.StudentClassRegModuleInfo', lambda modinfo: {'self': modinfo.program})
 
     def getModules(self, user = None, tl = None, old_prog = None):
-        """ Gets modules for this program, optionally attaching a user. """
+        """ Gets modules for this program, optionally attaching a user. Only open modules are included for non-admins. """
         modules = list(self.getModules_cached(tl, old_prog))
 
         if user and not user.isAdmin(self):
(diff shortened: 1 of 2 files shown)


esp/esp/program/models/__init__.py
```diff
@@ -1266,7 +1266,11 @@ def getModules_cached(self, tl = None, old_prog = None):
 
     def getModules(self, user = None, tl = None, old_prog = None):
         """ Gets modules for this program, optionally attaching a user. """
```
willgearty: Could you update this docstring to mention that only open modules are included for non-admins?

willgearty (approved):
Looks great, thanks! Just one nitpick before merging...

PR #5805, opened 2026-05-25
Week 1: Add shared test infrastructure: factories and ModuleHandlerTestMixin

vipulpandey21:
## Description

Adds two files of shared test infrastructure that all subsequent test PRs in this GSoC project will build on.

**File 1 — `esp/esp/tests/factories.py`** (new file)
Three factory functions for creating test objects in one call:
- `make_user(role)` — creates an ESPUser + Group membership
- `make_program(**kwargs)` — creates a full Program with timeslots, rooms, and permissions
- `make_class(program, teacher)` — creates a ClassSubject + ClassSection + assigns teacher

These replace the copy-paste setup pattern scattered across 8+ test files.

**File 2 — `esp/esp/program/modules/tests/support/mixins.py`** (new file)
`ModuleHandlerTestMixin` with 7 methods:
- `login_as(role)` — log in as admin/teacher/student in one call
- `get_module_url(tl, view)` — build handler URLs without manual string construction
- `assert_view_ok(url)` — assert GET returns 200
- `assert_view_forbidden(url)` — assert GET returns 302 or 403
- `post_to_module(tl, view, data)` — POST to a handler view and return response
- `get_module_obj(handler_name)` — retrieve ProgramModuleObj for a handler
- `logout()` — log out the current session

**File 3 — `esp/esp/program/modules/tests/support/__init__.py`** (modified, 1 line added)
Exports `ModuleHandlerTestMixin` from the support package.

Writing a test for any module handler currently requires manually building URLs,
copy-pasting login boilerplate, and figuring out the request pattern — 30-60 minutes
of setup before writing a single assertion. That friction is the primary reason
high-consequence handlers like `finaidapprovemodule.py` and `adminreviewapps.py`
have 0% coverage today. This infrastructure reduces that to 2-3 lines permanently,
for every future contributor beyond this GSoC project.

This is the foundation for:
- PR2: ClassCreationController direct unit tests (including `is not` identity bug regression)
- PR3: LotteryAssignmentController tests (first ever for lottery.py — 1000+ lines, 0 tests)
- PR4: finaidapprovemodule, adminreviewapps, lotterystudentregmodule behavioral tests

## Related Issue

Part of #3780

## Type of Change

- [x] New feature

## Testing

- [x] Existing tests pass
- [ ] New tests added (if applicable)
- [x] Manually verified

## AI Disclosure

Used Kiro (AI coding assistant) to help verify this after my implementation.

## Checklist

- [x] Self-reviewed the code
[...]

Issue #3780, opened 2024-10-18
Create more unit tests


milescalabresi:
Go through one of our test/deployment documents and create a ticket for each item that isn't currently tested in our automatic pipeline

Diya910:
Sure, test cases are the subtle but very core component in preventing code clashes. I would love to deep dive in the code and write some necessary test cases. I will proceed with the ones that are being tagged first. Can I proceed?

willgearty:
Hi @Diya910, yes please good ahead and make PRs for small subsets of tests. The ones that are already tagged would be a great place to start.

Diya910:
Hey @willgearty , I've been looking into #3977 and wanted to share my understanding of the test cases needed before I start writing them — please correct me if I'm off anywhere!

Here's what I'm planning to test:

**Recipient Resolution**
- learningu.org alias resolves to the correct recipients
- PlainRedirect lookup returns the correct destination
- ESPUser lookup finds the right user by email
- Empty recipient list after filtering is handled gracefully
- Null redirect destinations are handled (no crash)

**Sender Validation**
- Email from a valid site user gets forwarded successfully
- Email from an unknown sender (no account on site) gets rejected and a bounce is sent
- Malformed or unparseable From address is handled gracefully
- Sender with an email not associated with any learningu.org account gets rejected

**Multi-Account Sender Logic**
- When a sender has multiple accounts, the correct one is picked

**Attachment Handling**
- Email with attachments gets correctly extracted and forwarded
- An attachment that fails extraction gets skipped with logging and doesn't crash

**Edge Cases**
- Missing From header is handled gracefully
- Recipients list is empty after all filtering steps

A few questions before I proceed:
1. Should I mock SendGrid calls throughout, or is there a test helper already set up for that?
2. Where should the test file live — alongside existing mailgate files or somewhere else in the test suite?

Let me know if I'm missing anything or misunderstood any part of the logic.

Oval17:
Hey @Diya910 
Just to add my two cents to the discussion, you can view the PR that was raised by me..
Also as mostly those test cases would be written by AI, kindly make sure that when we do prompting we should mention that it should cover the codes(important functions) and should not write always truthy test cases as AI hallucinates a lot in these cases.
Please make sure that coverage is overall high.
Thanks

willgearty:
[...]

PR #5838, opened 2026-06-04
Week-1(Part-B) Refactor: make ProgramFrameworkTest delegate to factories (#5805 follow-up)

### LINKED ISSUE #5822: Make test factories the single source of truth for ProgramFrameworkTest and refactor duplication
### Problem Statement

There is significant code duplication between ProgramFrameworkTest and the new shared test factories (make_user, make_program, make_class) in the ESP test suite. While the current behavior is consistent, there is a risk that these two sources diverge over time, especially as test infrastructure evolves and new contributors make changes in only one place. Additionally, some implementation details (such as how the cache is handled, timeslot queries, and class 'accept' logic) are currently repeated in both locations or differ subtly.

### Proposed Solution

Have ProgramFrameworkTest call the shared factories for creation logic, making the new factories the single source of truth for test data setup. This should eliminate repeated implementation in two places and reduce the risk of behavior divergence. Ideally, after refactoring, all changes to how test resources are made for unit or integration tests should only need to be made in the factories. Also, review the following to be incorporated into the refactor:
- Ensure make_program resets ResourceType._get_or_create_cache before creating resources to handle persistent caches across test runs.
- Cache timeslot queries inside make_program to avoid multiple queries when num_rooms is large.
- Consider whether make_class should accept the class by default (optionally behind a parameter) or update its docs to clarify accepted/unaccepted state.
- Eliminate unnecessary boilerplate and migrate all test code to use the new factories where possible.

### Alternatives Considered

Continue maintaining both ProgramFrameworkTest setup logic and the new factories independently, with the risk that test setup behaviors diverge. Alternatively, only update documentation to warn future contributors not to diverge, but this does not guarantee consistent logic.

### Additional Context

- This is based on reviewer feedback from PR #5805.

vipulpandey21:
Follow-up to #5805 per Mr. Will's review comment: "Could we have this new set of factories be the new truth and then have ProgramFrameworkTest call the factories so that we don't have the same code in two places?"

`ProgramFrameworkTest.setUp()` was duplicating the same user/program/class creation logic as the new factories, creating a divergence risk. This PR makes the factories the single source of truth.

**`esp/tests/factories.py`**
- `make_program()`: add `admins` (list), `categories` (list), `sibling_discount`, and `start_time` params so `ProgramFrameworkTest` can pass its full settings through without inline duplication
- `make_class()`: add `class_info` and `sections` params to match `sections_per_class` and `class_info` behaviour in `ProgramFrameworkTest`

**`esp/program/tests.py`**
- `ProgramFrameworkTest.setUp()`: replace the inline user-creation loops, program-creation block, timeslot/room loops, and class-creation loop with calls to `make_user()`, `make_program()`, and `make_class()`
- `ResourceType` cache reset and `user_role_setup()` are now handled inside the factory functions; removed from `setUp` directly
- All public attributes (`self.students`, `self.teachers`, `self.admins`, `self.program`, `self.timeslots`, `self.rooms`, `self.event_type`) preserved with identical values; all helper methods unchanged

No production code modified. All existing tests continue to pass.

Part of #3780

Closes #5822

---

- [x] Refactor / cleanup
- [x] Existing tests pass
- [x] Self-reviewed the code


vipulpandey21:
Hi Mr. Ted @hwatheod , Mr. Will @willgearty 
The 3 failing tests (`ModuleManagementLinkTitleTest`) do not appear to be caused by this PR.
This refactor only modifies `factories.py` and `tests.py`; neither `admincore.py` nor `studentcertmodule.py` is touched. I verified that the same failures occur on upstream `main` before applying these changes.

The failures stem from `get_current_request()` returning `None` in `studentcertmodule.py` (around line 88) during `inModulesList()`, which seems unrelated to the factory-driven refactor. All other tests pass (1270 total), including those that exercise `ProgramFrameworkTest` directly.
let me know if you'd like me to investigate the pre-existing failure separately, but I believe it is outside the scope of this PR.

willgearty:
All tests are currently running successfully on main: https://github.com/learning-unlimited/ESP-Website/actions/runs/26884248121/job/79292016212. As far as I can tell, these errors have been introduced by this PR and should therefore be resolved in this PR.

vipulpandey21:
Mr. Will @willgearty, you were right, sorry about that. The root cause was that our refactor dropped the modules= kwarg when delegating to make_program(). Subclasses like ModuleManagementLinkTitleTest pass a restricted module list via super().setUp(modules=[...]) so only specific modules get attached to the program. With that kwarg missing, make_program() was attaching all ProgramModule rows including StudentCertModule, whose isStep() calls get_current_request() outside a request context and crashes. Fixed by adding a modules param to make_program() (defaults to ProgramModule.objects.all() for backward compatibility) and threading settings['modules'] through from ProgramFrameworkTest.setUp().

esp/esp/tests/factories.py
```diff
@@ -306,15 +331,15 @@ def make_class(program, teacher, title=None, category=None,
...
     )
 
     cls.makeTeacher(teacher)
 
-    # Add a section if none exists yet
-    if cls.get_sections().count() == 0:
-        cls.add_section(duration=duration)
+    for j in range(sections):
+        if cls.get_sections().count() <= j:
+            cls.add_section(duration=duration)
```
hwatheod: This code looks a little bit strange.  Why do you need the `if` statement here?

vipulpandey21:
Hi Mr. Will and Mr. Ted, the "Run Tests" step is passing (all tests green), but "Upload Coverage to Codecov" is failing with a GPG signature error — gpg: Can't check signature: No public key. how should I proceed?

hwatheod (approved):

PR #5842, opened 2026-06-11
Phase 4(Week 4): Implement special module lock handling and conflict detection (Issue #3854)

Supriyo760:
**Description**
Implements backend protection logic for Phase 4 of the New Module Management UI. This PR enforces strict scheduling rules for special internal modules — preventing arbitrary reordering and blocking concurrent scheduling with incompatible modules.

**Changes**
- **Module Lock Flags** — Added `always_enabled` and `seq_locked` class-level attributes to `ProgramModuleObj` in `esp/esp/program/modules/base.py` and included them in the timeline JSON API response.
- **Update Protection** — Hardened `module_schedule_update_api` to reject drag-and-drop reorder requests for locked modules. The API hydrates the subclass via `getFromProgModule()` and returns a 403 if a locked module is targeted.
- **Conflict API Route** — Registered a new URL endpoint in `esp/esp/program/urls.py` for `module_schedule_conflicts`.
- **Conflict Detection** — Implemented `module_schedule_conflicts_api` in `esp/esp/program/views.py`. Detects bi-directional conflicts declared via `conflicts_with` and calculates overlapping `start_date`/`end_date` bounds between active modules.

- **Tests** — Added `test_module_schedule_conflicts_api.py` with mocked locked and conflicting handlers, covering both 403 enforcement on `seq` modifications and overlap intersection logic.

**Related Issue**
Partially addresses #3854

**Type of Change**
- [x] New feature

**Testing**
- [x] Existing tests pass
- [x] New tests added
- [ ] Manually verified

**AI Disclosure**
AI used for suggestions only.

**Checklist**
- [x] Self-reviewed
- [ ] Documentation updated


esp/esp/program/modules/tests/test_module_schedule_conflicts_api.py
```diff
@@ -0,0 +1,129 @@
...
+        self.future = self.now + timedelta(days=2)
+        self.base_url = f"/manage/{self.program.program_type}/{self.program.program_instance}/module_schedule"
+
+    def test_seq_locked_blocks_update(self):
+        """Updating a seq-locked module should return 403."""
+        # Mock the handler class of pmo1 to be locked
+        original_class = self.pmo1.__class__
+
+        # We also need to patch getFromProgModule temporarily so the view sees it as locked
+        # But wait, we can just modify the class dynamically for the test!
```
willgearty: This is a weird comment. Guessing it was written by AI? Remember you are responsible for every line of your PRs.

esp/esp/program/views.py
```diff
@@ -1747,10 +1753,53 @@ def module_schedule_preview_api(request, program_type, program_term):
...
+                "seq_locked": mod.seq_locked,
                 "start_date": mod.start_date.isoformat() if mod.start_date else None,
                 "end_date": mod.end_date.isoformat() if mod.end_date else None,
                 "required": mod.required,
                 "required_label": mod.required_label,
             })
 
     return JsonResponse(data)
+
+@require_GET
```
willgearty, line 1766: I think this (and the other API endpoints) should all have @needs_admin, right?

esp/esp/program/modules/tests/test_module_schedule_conflicts_api.py
```diff
@@ -0,0 +1,129 @@
...
+        self.pmo2 = ProgramModuleObj.getFromProgModule(self.program, self.mod2)
+
+        self.now = datetime.now()
+        self.past = self.now - timedelta(days=2)
+        self.future = self.now + timedelta(days=2)
+        self.base_url = f"/manage/{self.program.program_type}/{self.program.program_instance}/module_schedule"
+
+    def test_seq_locked_blocks_update(self):
+        """Updating a seq-locked module should return 403."""
+        # Mock the handler class of pmo1 to be locked
```
willgearty, line 38: Shouldn't you use a formal mock, then? Something like `mock.patch.object(HandlerClass, 'seq_locked', True)`? Then the values should revert after the test is done.

esp/esp/program/modules/tests/test_module_schedule_conflicts_api.py
```diff
@@ -0,0 +1,129 @@
...
+            data = json.loads(response.content)
+            self.assertEqual(len(data["conflicts"]), 0)
+        finally:
+            original_class.conflicts_with = original_conflicts
+
+    def test_conflicts_api_with_overlap(self):
+        """Conflicting handlers with overlapping times return conflicts."""
+        self.client.force_login(self.admin)
+
+        original_class = self.pmo1.__class__
```
willgearty, line 84: Should formally mock here as well

esp/esp/program/modules/tests/test_module_schedule_conflicts_api.py
```diff
@@ -0,0 +1,129 @@
...
+        self.program.program_modules.clear()
+
+        # Grab or create two real modules linked to the program
+        mod_qs = ProgramModule.objects.order_by('id')
+        if mod_qs.count() >= 2:
+            self.mod1 = mod_qs[0]
+            self.mod2 = mod_qs[1]
+        else:
+            self.mod1 = ProgramModule.objects.create(admin_title="Test Mod 1", module_type="learn", handler="StudentRegTwoPhase", seq=1)
+            self.mod2 = ProgramModule.objects.create(admin_title="Test Mod 2", module_type="teach", handler="TeacherClassRegModule", seq=2)
```
willgearty: Why not choose two modules that we know should exist?

esp/esp/program/modules/tests/test_module_schedule_conflicts_api.py
```diff
@@ -0,0 +1,129 @@
...
+            self.assertIn("locked", data["error"])
+        finally:
+            original_class.seq_locked = original_locked
+
+    def test_conflicts_api_no_overlap(self):
+        """Conflicting handlers with no time overlap return no conflicts."""
+        self.client.force_login(self.admin)
+
+        original_class = self.pmo1.__class__
+        original_conflicts = original_class.conflicts_with
```
willgearty: Should formally mock here as well

willgearty (changes requested):
Looking pretty good, thanks. Just a handful of comments.

esp/esp/program/modules/tests/test_module_schedule_conflicts_api.py
Supriyo760: Switched to mock.patch.object(original_class, 'seq_locked', True)

esp/esp/program/modules/tests/test_module_schedule_conflicts_api.py
Supriyo760: Replaced the manual save/restore pattern with mock.patch.object

esp/esp/program/modules/tests/test_module_schedule_conflicts_api.py
Supriyo760: Replaced the arbitrary queryset lookup with explicit ProgramModule.objects.create()

esp/esp/program/modules/tests/test_module_schedule_conflicts_api.py
Supriyo760: Done same as above one.

esp/esp/program/views.py
Supriyo760: We can't use @needs_admin here because it only works on class methods, not on regular API views like this one.
However, these API views are still protected. They all call get_program_or_404() right at the beginning, which does the exact same admin check behind the scenes and blocks unauthorized users. 
Let me know if you'd still prefer a different way of doing this.

esp/esp/program/modules/tests/test_module_schedule_conflicts_api.py
Supriyo760: My bad, removed in the latest commit.

esp/esp/program/views.py
willgearty: Ah, that works then, thanks!

willgearty (approved):
Looks good, thanks!

PR #5844, opened 2026-06-15
Fix #1655: Comm panel combo queries generate bad filter

### LINKED ISSUE #1655: Comm panel generates a bad filter for some queries
While trying to find potential workarounds for #1654, I tried to pass the query "All teachers in the DB OR all onsite volunteers for [program]".  This generated 0 users, so I investigated further.  It looks like the generated Q object saved to the PQF is incorrect; it generated the Q object
```
(AND: (NOT (AND: )), (OR: (AND: ), ('volunteeroffer__request__program', <Program: Spark 2014>)), ('is_active', True), ('groups__name', u'Teacher'), ('is_active', True))
```
which evaluates to "teachers in the db AND onsite volunteers for [program]":
```
SELECT [fields]
FROM [join of various tables]
WHERE (("program_volunteerrequest"."program_id" = 105)
       AND "auth_user"."is_active" = true
       AND "auth_group"."name" = 'Teacher'
       AND "auth_user"."is_active" = true) LIMIT 21
```
(Oddly, I would actually expect the `NOT (AND: )` to evaluate to False no matter what, but it doesn't.)

No clue where the bug is, but if anyone wants to look at the PQF on esp.mit.edu, it's `PersistentQueryFilter.objects.get(id=11388)`.

Supriyo760:
### Description
Addresses a longstanding issue in the Communications Panel where combining an `all_X` base list (e.g., "All Teachers") with an `OR` condition using another list failed to produce a correct union, unexpectedly restricting the results to only the overlap. Resolves #1655.

### Changes
- **Fix Combo Queries for `all_X` Base Lists**: In `UserSearchController.query_from_postdata`, the logic for `combo_base_list` was setting `q_program` to an empty `Q()` when parsing an `all_X` list. This empty `Q()` was absorbed during `OR` combinations, collapsing `Q() | Q(list_b)` into just `Q(list_b)` and causing the query to intersect instead of union. Updated the assignment to explicitly use `ESPUser.getAllOfType(recipient_type, True)` so that the base set properly evaluates during boolean operations.
- **Added Regression Test**: Created `test_all_base_list_or_overlap` in `test_usersearch.py`. The test simulates submitting a request for "All Teachers" combined with an `OR` filter for `class_approved`. It confirms the query now properly returns a union of both teachers with and without an approved class.

### Related Issue
Fixes #1655

### Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [x] Test addition

### Testing
- [x] Existing tests pass
- [x] New regression test `test_all_base_list_or_overlap` added and passing

### AI Disclosure
AI was used for suggestions and test generation.

### Checklist
- [x] Self-reviewed the code

### DIFF
--- a/esp/esp/users/controllers/tests/test_usersearch.py
+++ b/esp/esp/users/controllers/tests/test_usersearch.py
@@ -93,3 +93,61 @@ class TestUserSearchController(ProgramFrameworkTest):
         # The other_user should NOT be found because they only have ONE of the records (AND logic)
         self.assertNotIn(other_user, results)
         self.assertEqual(results.count(), 1)
+
+    def test_all_base_list_or_overlap(self):
+        """
+        Verify that using 'all_X' as the combo_base_list correctly seeds q_program
+        with ESPUser.getAllOfType() instead of an empty Q(), so that OR conditions
+        correctly produce a union rather than an intersection.
+
+        Before the fix: Q() | Q(class_approved) collapsed to just Q(class_approved),
+        making "All teachers OR class_approved teachers" return only teachers with
+        an approved class — incorrectly excluding teachers without one.
+
+        After the fix: ESPUser.getAllOfType('Teacher') | Q(class_approved) correctly
+        returns ALL teachers in the DB regardless of whether they have an approved class.
+
+        Issue #1655
+        """
+        # teacher_only: in the Teacher group, but has NO approved class
+        teacher_only = ESPUser.objects.create_user(
+            username='teacher_only_1655', email='tonly@example.com', password='password')
+        teacher_only.makeRole('Teacher')
+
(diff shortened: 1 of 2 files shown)


willgearty (changes requested):
This doesn't work when the OR user group is different from the base user group. For example, "Teachers who have submitted a class OR students who are enrolled in a class" only returns users who are teachers who have submitted a class OR have enrolled in a class. It seems that the base user group is still causing some sort of override. I think this is due to this line:
https://github.com/learning-unlimited/ESP-Website/blob/0447697c2034f44022ee78921d69d7902ba51b63/esp/esp/users/controllers/usersearch.py#L338

willgearty (approved):
Thanks for figuring that out. Works great now!

PR #5853, opened 2026-06-22
Week 5: Behavioral tests for finaidapprovemodule, adminreviewapps, and lotterystudentregmodule

vipulpandey21:
## Description

Three module handlers — `finaidapprovemodule.py`, `adminreviewapps.py`, and `lotterystudentregmodule.py` — had 0% behavioral coverage despite being used on every program day. This PR adds direct HTTP-layer tests for each handler using `ModuleHandlerTestMixin` from PR1 (#5805), which reduces test setup from ~10 lines to 2–3.

**`test_finaidapprove.py`** — tests for `FinAidApproveModule.finaidapprove()`:
- Admin GET returns 200 with `requests` in context
- Non-admin GET is forbidden (302/403)
- POST with valid user ID and amount creates a `FinancialAidGrant`
- POST without `approve_blanks` skips requests with blank `household_income`
- POST with empty user checklist does not approve any request

**`test_adminreviewapps.py`** — tests for `AdminReviewApps.accept_student()` and `reject_student()`:
- `accept_student()` creates a `StudentRegistration` with `relationship='Accepted'`
- `reject_student()` expires the `Accepted` registration so `valid_objects()` returns 0
- Non-admin cannot call either endpoint

**`test_lotterystudentregmodule.py`** — tests for `timeslots_json()` and `viewlotteryprefs()`:
- `timeslots_json()` returns 200 with correct JSON content-type
- Response is a list of `[id, short_description]` pairs ordered by start time
- Returned IDs match the program's timeslots
- Unauthenticated `viewlotteryprefs` request redirects
- Student with `Priority/1` registration sees it on `viewlotteryprefs`

Note: `LotteryStudentRegTest` (already in `studentregmodules.py`) covers `students()` and `isCompleted()` — this PR only adds the HTTP-layer view tests that were missing.

## Related Issue

Refs #3780

## Type of Change
- [x] New feature

## Testing
- [x] Existing tests pass
- [x] Manually verified

## AI Disclosure

Used Kiro (AI coding assistant) to verify this implementation.

## Checklist
- [x] Self-reviewed the code


esp/esp/program/modules/tests/test_adminreviewapps.py
```diff
@@ -0,0 +1,79 @@
+"""
+Behavioral tests for AdminReviewApps (esp/program/modules/handlers/adminreviewapps.py).
+
+accept_student() and reject_student() create and expire StudentRegistration
+objects with relationship='Accepted'. Coverage was 0% before this PR.
```
hwatheod: You shouldn't write "Coverage was 0% before this PR" in the code itself - that won't really make sense to future readers of the code. Please remove from all the test files you added.  It is fine to point this out in the PR description.

esp/esp/program/modules/tests/test_adminreviewapps.py
```diff
@@ -0,0 +1,79 @@
...
+        self.student = self.students[0]
+        self.cls = ClassSubject.objects.filter(parent_program=self.program).first()
+        self.section = self.cls.get_sections()[0]
+        self.accepted_rt, _ = RegistrationType.objects.get_or_create(
+            name='Accepted', defaults={'category': 'student'}
+        )
+
+    def _accept_url(self):
+        return self.get_module_url('manage', 'accept_student') + \
+               f'?cls={self.cls.id}&student={self.student.id}'
```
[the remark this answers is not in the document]
hwatheod: Copilot's analysis seems right to me from my manual checking of the code. Yet it seems like tests have passed. Can you please look into this?

esp/esp/program/modules/tests/test_finaidapprove.py
```diff
@@ -0,0 +1,96 @@
...
+        blank_request = FinancialAidRequest.objects.create(
+            program=self.program,
+            user=blank_student,
+            household_income='',
+            extra_explaination='',
+        )
+        self.login_as('admin')
+        self.client.post(self._url(), {
+            'user': [str(blank_student.id)],
+            # approve_blanks NOT sent
```
hwatheod: should be `NOT set`

vipulpandey21:
Hi Mr. Ted,

Thanks for the detailed review! I've gone through all the feedback from both you and Copilot and pushed fixes for everything.

**What I changed:**

1. **Removed the "Coverage was 0%" lines** from all three docstrings — totally agree that detail belongs in the PR description, not in the code that future contributors will read.

2. **Fixed the URL structure bug** Copilot caught. You asked why the tests were passing despite the issue — turns out both `accept_student()` and `reject_student()` expect `extra` to be the class ID in the URL path (like `/manage/accept_student/123/`), not as a query parameter. I was building URLs like `?cls=123&student=456` when it should've been using the `extra=` parameter in `get_module_url()`.
   The tests passed before because ESP's error handling is pretty forgiving, but the actual URL routing was broken. I've updated both helper methods to construct the URLs correctly now.
   
3. **Other fixes:**
   - Changed "NOT sent" → "NOT set" in the finaidapprove comment
   - Made assertions more strict (only 302 for redirects, only 200 for authenticated views)
   - Simplified the timeslot sorting logic to be more readable
   
**Quick note on timing:** Sorry for the slight delay in getting these fixes pushed — I was tracing through the `review_students()` code path to understand why the tests were passing with incorrect URLs, and wanted to make sure the fix was right before pushing. The `extra` parameter gets passed all the way through the handler chain, so it was a bit buried.

Everything's pushed now and CI should be green. Let me know if anything needs adjustment!

hwatheod (approved):

PR #5856, opened 2026-06-29
Phase 5(Week 5&6) : Integrate Module Timeline UI into Django (Issue #3854)

Supriyo760:
**Description:**
Implements frontend UI replacement for Phase 5 of the New Module Management UI. This PR replaces the legacy drag-and-drop module column list with the interactive timeline grid, fully integrating it with the Django backend.

### Changes

- **Timeline UI Replacement** — Replaced the old sortable column HTML in `esp/templates/program/modules/admincore/modules.html` with a dynamic timeline container.
- **Dynamic JavaScript Rendering** — Created `public/media/scripts/program/modules/module_timeline.js` to fetch live data from the `module_schedule` API and render interactive timeline blocks with split student/teacher views.
- **Custom Styling** — Ported the mockup designs into `public/media/styles/module_timeline.css` using custom properties for role-based theming and responsive layouts, removing all Tailwind CDN dependencies.
- **Constraint Injection** — Updated `admincore.py` to inject `module_constraints` and `program.url` base as `json_script` metadata directly into the template so JavaScript can instantly read locked states without additional API calls.
- **API Extension** — Extended `module_schedule_update_api` in `views.py` to support `required`, `required_label`, and `link_title` updates, allowing the Slide-In Edit Panel to fully update modules via AJAX.
- **Add Module Drawer** — Built out the bottom drawer UI to dynamically categorize `Student Registration`, `Teacher Registration`, and `General` modules from the backend and save new additions automatically.
- **Test Fixes** — Updated legacy CSRF checks in `tests/admincore.py` and implemented test caching fixes across the system to ensure stable test runs.

### Related Issue
Partially addresses #3854

### Type of Change

- [x] New feature

### Testing

- [x] Existing tests pass
- [x] New tests added
- [x] Manually verified

### AI Disclosure
AI used for suggestions only.

### Checklist

- [x] Self-reviewed
- [x] Documentation updated


esp/templates/program/modules/admincore/modules.html
```diff
@@ -5,13 +5,13 @@
...
     <link rel="stylesheet" href="/media/styles/forms.css" type="text/css" />
-    <link rel="stylesheet" href="/media/styles/module_management.css" type="text/css" />
+    <link rel="stylesheet" href="/media/styles/module_timeline.css?v=1.0.6" type="text/css" />
     <link rel="stylesheet" href="/media/styles/expand_display.css" type="text/css" />
 {% endblock %}
 
 {% block xtrajs %}
     {{ block.super }}
-    <script type="text/javascript" src="/media/scripts/program/modules/module_management.js"></script>
+    <script type="text/javascript" src="/media/scripts/program/modules/module_timeline.js?v=1.0.6"></script>
```
willgearty: Why not just modify the existing files? Once we've migrated to this new UI, there won't be any need for the old files and they will just be dead code.

esp/public/media/scripts/program/modules/module_management.js
```diff
@@ -1,236 +1,567 @@
+/*
+ * module_timeline.js
+ * Interactive script for Phase 6 Module Management page.
```
willgearty: Phase 6 is meaningless to other contributors.
```suggestion
 * Script for interactive module management page.
```

esp/esp/program/views.py
willgearty: This description no longer matches the behavior of the function

willgearty: Is there any point for this endpoint anymore?

esp/public/media/styles/module_management.css
```diff
@@ -1,96 +1,879 @@
...
-
-.connectedSortable li {
-    padding: 5px;
-    margin-top: 5px;
-    margin-bottom: 5px;
-    font-size: 1.2em;
-    width: 100%;
+/* 
+ * module_timeline.css 
+ * Styles for the interactive module scheduler (Phase 6).
```
willgearty:
```suggestion
 * Styles for the interactive module management page.
```

willgearty (changes requested):
This is looking really great! I tested it a bunch and it works quite well, but I also noticed some issues:

- If you set a module to open far in the future, the start date is the same as the red line. I think there should always be at least a week before the red line (like there is always a week before the oldest module with a start date).
- The REQ pill doesn't show on "Always On" modules.
- I'm not a fan of using the infinity symbol (on the modules and in the edit panel)
- The "Always On" text within the modules often gets cut off, especially when showing all modules:

Image

- The Closes widget in the edit panel should be moved to a new line, otherwise it often gets cut off, forcing the user to scroll horizontally to use the widget.
- When you add/remove a module through the questions, it doesn't immediately add/remove it from the modules list (but shows the questions list as updated). I'm guessing this is a caching issue.
- I really like the splitting of the questions into student and teacher modules, but there appears to be one fallout from this. There are two questions that were designed to go one after another in a list: "Do students have to apply to individual classes?" and "If yes, can teachers admit them (as opposed to just admins)?". Because the questions are now split apart, the second question doesn't make any sense. Either the first question needs to be moved above the second question or the second question needs to be modified to make sense on its own.

Image

willgearty (approved):
Changes look good, and everything works great! Thanks for addressing my comments.

Issue #116, opened 2011-11-23
Onsite scrolling class list has stale data


aseering:
The scrolling list of available classes did not update itself according to changing enrollments.

This is possibly because the scrolling list of available classes doesn't update itself in mid-scroll; it scrolls for 5 minutes (roughly as long as it takes to scroll through the list of classes at MIT Splash) and then downloads the latest catalog and starts over.  So the list is guaranteed to be somewhat stale for the later time blocks, as long as reg is actively underway.

However, there may be additional reasons for this as well.

PR #5878, opened 2026-07-22
Reload onsite class list at end of scroll cycle instead of on a blind…

### LINKED ISSUE #318: Open class list: reload after end is reached
From MIT: The scrolling class list should be refreshed more frequently, and maybe a good time to refresh is when it reaches the end of one scrolling cycle.  Related to #116.

Yashika28284:
## Description

Fixes the onsite scrolling class list showing stale data during active registration.

Previously, the page used `<meta http-equiv="refresh" content="{{ refresh }}">` to reload on a fixed timer, regardless of where the scroll currently was — this could interrupt the list mid-scroll and, per the original report, still left data stale for later time blocks since it wasn't tied to actual scroll completion.

This PR removes the blind meta refresh and instead reloads the page only when:
1. The scroll reaches the end of a full cycle, AND
2. At least `refresh` seconds (converted to ms) have elapsed since the page loaded

This matches the suggestion on #318 — refreshing at a natural scroll boundary instead of an arbitrary fixed timer.

**Testing:** Verified via the automated test suite (`ClassListReloadTests`), which checks the rendered template output directly. I wasn't able to manually verify against a live onsite event display, since that requires a running program with the physical on-site screen setup — but all the relevant JS logic and conditions are covered by the new tests. Ran the full `test_onsiteclasslist` suite (58 tests) to confirm nothing else broke.

## Related Issue

Closes #318

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup


Yashika28284:
Fixed in the latest commit — thanks!

Yashika28284:
@willgearty sir if u could just review it once again ! Will be very helpful

willgearty:
Thanks for your submission. We've got quite a backlog of PRs to review, which we probably won't get to for a bit. Sorry for the delay, but we're a small volunteer team!

willgearty (approved):
Looks good, thanks! This definitely addresses #318 (refreshing at the end of the cycle), but it seems like the data can still get stale (#116), especially if there is a very long list of classes to cycle through. I'm not sure what the right solution for that would be (ajax calls to update the data mid-cycle?), but you're welcome to submit a separate PR to address it.

PR #5881, opened 2026-07-22
Phase 7 (Week 8, 9, & 10): Add module drawer animation, safety locks, and full accessibility pass (Issue #1827)

### LINKED ISSUE #1827: Improve required module workflow
Related to and possibly supersedes #335, #1496.

An MIT ESP brought up some concerns about the required module redirection workflow possibly being disorienting to students (and misleading about parents, since the medical form is a required step).

One suggestion, similar to #335:

> Add the checkboxes thingy that appears on student reg appearing on all the modules so that we can get rid of "registration steps" completely. 

Another, possibly orthogonal, idea that I've been thinking about is to have some way for each module to have prerequisites, rather than just each module being required or not; then if any module's prerequisites are not met, it tells you what those prerequisites are and redirects you to them.

A third idea: the whole idea of the registration mainpage in the first place was that you go there first and it gives you a checklist of steps. Maybe running with force_show_required_modules disabled should be a more supported workflow; we could do this by making the mainpage content more dynamic based on which modules the user has yet to complete.

Supriyo760:
**Description:**
Implements frontend UI polish, full accessibility support, and backend locking mechanisms for Weeks 8, 9, and 10 of the New Module Management UI. This PR enables a smoother admin experience by adding dynamic timeline updates without page reloads, bidirectional synchronization, protections for active student data, and robust keyboard navigation.

**Changes**
* **Dynamic Canvas Rendering** — Rewrote the success callback in `saveModules()` (`module_management.js`) to bypass `window.location.reload()`. It now triggers `loadModules()` asynchronously, allowing newly checked blocks to instantly and smoothly animate onto the timeline.
* **Active Student Safety Prompt** — Added a browser `confirm()` check in `module_management.js` that intercepts unchecking operations. If an administrator attempts to remove a module, they are warned that active student data tied to that module might be hidden or affected, requiring explicit confirmation before the AJAX POST proceeds.
* **Bidirectional DOM Synchronization** — Engineered a state-sync solution during the dynamic reload where `this.defaultChecked` is actively updated to match the successfully saved `this.checked` state. This ensures subsequent operations in the drawer calculate the correct diffs without requiring a hard refresh.
* **Drawer Lock Architecture** — Modified the modules view context in `admincore.py` to hydrate a new `always_enabled` boolean flag down into `module_questions_categorized`. Updated `modules.html` to accept this flag, applying `disabled`, `opacity-50`, and `cursor-not-allowed` attributes to structurally protect core modules from being unchecked by administrators.
* **Accessibility & Focus Management (Week 9)** — Implemented full keyboard navigation support, including correct `focus-visible` outlines for high-contrast visibility, explicit `aria-labels`, and screen-reader compliant `dialog` roles. Focus is now intelligently trapped and restored to the correct timeline block when modals close, and global Escape key handlers were added to easily dismiss modals.
* **Drawer UI Consistency** — Converted the "Add Module" configuration drawer from a bottom-sheet layout to a right-side panel layout, providing a unified and consistent user experience between the "Edit Module" and "Add Module" flows.
* **Scroll-Gap Bug Fix** — Engineered a custom scroll-lock event listener to fix a native browser rendering bug where keyboard focus transitions inside `overflow: hidden` wrappers would invisibly shift the container, leaving a white gap on the timeline grid.

**Related Issue**
Resolves #1827

**Type of Change**
- [x] New feature

**Testing**
- [x] Existing tests pass
- [ ] New tests added
- [x] Manually verified

**AI Disclosure**
AI used for suggestions only.

**Checklist**
- [x] Self-reviewed
- [x] Documentation updated

### DIFF
--- a/esp/public/media/scripts/program/modules/module_management.js
+++ b/esp/public/media/scripts/program/modules/module_management.js
@@ -32,13 +32,23 @@ $j(document).ready(function() {
     var $addOverlay  = $j('#addOverlay');
     var $addDrawer   = $j('#addDrawer');
 
+    // Prevent browser focus-scroll bugs from shifting the layout when modals open/close
+    $j('.tl-wrapper').on('scroll', function() {
+        if (this.scrollLeft !== 0) {
+            this.scrollLeft = 0;
+        }
+    });
+
     // ──────────────────────────────────────────────────────────────
     // Toast Notifications (replaces alert())
     // ──────────────────────────────────────────────────────────────
     function showToast(msg, type) {
         // type = 'success' | 'error'
-        var $toast = $j('<div>').addClass('tl-toast tl-toast-' + (type || 'success')).text(msg);
-        $j('body').append($toast);
+        var $toast = $j('<div>')
+            .addClass('tl-toast tl-toast-' + (type || 'success'))
+            .attr('role', 'status')
+            .attr('aria-live', 'polite')
+            .text(msg);        $j('body').append($toast);
         setTimeout(function() { $toast.addClass('tl-toast-visible'); }, 10);
         setTimeout(function() {
             $toast.removeClass('tl-toast-visible');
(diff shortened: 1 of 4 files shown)


esp/public/media/scripts/program/modules/module_management.js
```diff
@@ -32,13 +32,23 @@ $j(document).ready(function() {
...
     // ──────────────────────────────────────────────────────────────
     function showToast(msg, type) {
         // type = 'success' | 'error'
-        var $toast = $j('<div>').addClass('tl-toast tl-toast-' + (type || 'success')).text(msg);
-        $j('body').append($toast);
+        var $toast = $j('<div>')
+            .addClass('tl-toast tl-toast-' + (type || 'success'))
+            .attr('role', 'status')
+            .attr('aria-live', 'polite')
+            .text(msg);        $j('body').append($toast);
```
willgearty: The append here should be on its own line.

willgearty (changes requested):
Thanks for these additions, most of it works flawlessly!

I did notice one bug: the focus remains on a module when you "esc" (or click "cancel") out of the drawer, but when you click "save changes" the focus goes back up to the top of the module list, presumably due to rerendering of the module list. 

https://github.com/user-attachments/assets/aae8c23f-d8b3-4501-833d-5cd6d1bebb91

Please also address the Copilot comments above.

willgearty (changes requested):
The focus now works great, thanks!

I've noticed a regression with the side drawer. It now pushes everything to the left, leaving a gap on the right:

Image

willgearty (approved):
Works great, thanks for fixing that up!

PR #5894, opened 2026-07-26
Week 7: Enforce Codecov coverage targets and document how to write tests

vipulpandey21:
## Summary

Turns the Codecov project status into an actual gate, and adds
`docs/dev/testing.rst` describing how to write tests against ESP's existing test
infrastructure.

**Refs:** #3780

## 1. Codecov enforcement

The project status had no `target`, so it defaulted to `auto`: each pull request
was only compared against its own base commit, with a `0.1%` tolerance. Nothing
held the project to an absolute floor.

```yaml
coverage:
  status:
    project:
      default:
        target: 75%
        threshold: 0.1%
    patch: no
```

- `target: 75%` is the floor for overall coverage. `main` is at 78.32% today, so
  this does not fail anything currently on the branch.
- `threshold: 0.1%` stays as tolerance for run-to-run noise, which is what
  `threshold` actually controls — it is the permitted drop below target, not the
  target itself.
- `patch: no` keeps the patch status off, per review. Note that this key has to
  be set explicitly: if `patch` is absent, Codecov enables the patch status with
  a target of `auto`, so deleting the block would add a gate rather than remove
  one.

The project status shows up as the `codecov/project` check on the pull request.

## 2. Testing guide

`docs/dev/contributing.rst` explains how to *run* tests. It does not explain how
to *write* them, so each new contributor re-derives the fixtures and helpers by
reading existing test files. `docs/dev/testing.rst` covers:

- Narrowing a pytest run to one file, class, or method while iterating
- `ProgramFrameworkTest` fixtures (`self.students`, `self.teachers`,
  `self.admins`, `self.program`, `self.categories`) and its helpers
  (`add_user_profiles()`, `schedule_randomly()`)
- `ModuleHandlerTestMixin` and each helper it provides
- `esp/tests/factories.py` for tests that do not need a whole program
- Writing controller tests and behavioral invariant tests
- Where test files live for each kind of code, and the naming conventions

It also documents three things that are easy to get wrong and are not obvious
from reading the source:

- `@needs_admin` renders `errors/program/notanadmin.html` with HTTP 200 rather
  than returning 403 or redirecting, so access-control tests must assert on the
  template, not the status code.
- `cannotAdd()` needs `autocorrect_constraints=False` in tests, or it may
  resolve the constraint itself and report no error.
- `update_class_sections()` reads `cls.duration`, which the fixtures leave unset
  on `ClassSubject`.

Every example is taken from or checked against code currently on `main`.

## Files changed

```
codecov.yml          |   +5
docs/dev/testing.rst | +369
```

No production code is modified.

## Verification

- [x] `codecov.yml` parses as valid YAML
- [x] Both `.rst` files parse cleanly through docutils
- [x] Every class, method, fixture, and file path referenced in the guide was
      checked against `main`
- [x] Test-selection commands use classes that exist and that pytest collects
      under the patterns in `esp/pytest.ini`


vipulpandey21:
Hi Ted sir, @hwatheod ready for review whenever you have a moment.

One deliberate deviation from the proposal worth flagging up front: the proposal
describes this as "threshold 0.1% → 70%". In Codecov's schema `threshold` is the
permitted drop *below* target, not the target itself — `threshold: 70%` would have
allowed a 70-point regression and effectively disabled the check. So this sets
`target: 70%` and keeps `threshold: 0.1%` as tolerance for run-to-run noise, which
is what actually implements the enforcement the proposal describes. The
`codecov/project` check confirms it reads the new config: `77.10% (target 70.00%)`.

Patch coverage is at 60% as planned, and no production code is touched.

The earlier `Test (Python 3.12)` failure was `TeacherReviewAppsTest.test_prev_redirect`,
which selects its fixture data via `random.choice()` with no seed — unrelated to this
PR, and green on re-run. #5887 addressed part of that flakiness but not all of it.
Happy to open a separate issue if that's worth tracking.

willgearty:
@vipulpandey21 Good catch, please open an issue and then a PR for the flaky test, thanks!

docs/dev/testing.rst
```diff
@@ -0,0 +1,356 @@
...
+      response = self.post_to_module('manage', 'finaidapprove', {
+          'user': [str(self.student.id)],
+          'approve_blanks': 'on',
+          'amount_max_dec': '25.00',
+          'percent': '100',
+      })
+      self.assertIn(response.status_code, [200, 302])
+      self.assertTrue(
+          FinancialAidGrant.objects.filter(request=self.request).exists()
+      )
```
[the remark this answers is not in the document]
vipulpandey21: Good catch — fixed in 3291522.

The example now includes the `setUp()` that assigns `self.student` and creates the `FinancialAidRequest`, matching the real `test_finaidapprove.py`, so the snippet stands on its own.

docs/dev/testing.rst
```diff
@@ -0,0 +1,356 @@
...
+  esp/program/modules/handlers/finaidapprovemodule.py
+  esp/program/modules/tests/test_finaidapprove.py
+
+Naming
+~~~~~~
+
+- Test files: ``test_<module_name>.py``. Some older files in
+  ``esp/program/modules/tests/`` are named ``<module_name>.py`` instead and are
+  picked up through the ``tests_modules.py`` aggregator; new files should use
+  the ``test_`` prefix so pytest collects them directly.
```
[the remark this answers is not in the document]
vipulpandey21: Correct, and thanks for catching it — there is no `tests_modules.py` in the repo. I had taken that detail from a comment in `esp/pytest.ini`, which is itself inaccurate on this point.

Fixed in 3291522. Rather than substitute another mechanism I could not verify, the guide now states only what is checkable: those legacy filenames match none of the collection patterns in `esp/pytest.ini`, which is the actual reason to use the `test_` prefix on new files.

vipulpandey21:
okay Mr. Will @willgearty I will open the issue with its pr shortly.

codecov.yml
```diff
@@ -2,7 +2,13 @@ coverage:
...
       default:
+        # Overall coverage must stay at or above this level.
+        target: 70%
+        # Tolerance for run-to-run noise; a larger drop is a real regression.
         threshold: 0.1%
-    patch: no
+    patch:
+      default:
+        # New and changed lines in a pull request must be at least this covered.
+        target: 60%
```
hwatheod: Had a discussion with @willgearty about this.  We don't feel like we are ready for a patch target.  Let's remove it for now.  We would also like the overall target to be 75%.

vipulpandey21: Ok sir, Done in 0ff0a11  project target is now 75%, and the patch status is gone.

One implementation detail worth noting: I disabled the patch status with `patch: no` rather than deleting the key. If the `patch` key is absent entirely, Codecov enables the patch status by default with a target of `auto`, so removing the block would have left a patch gate in place instead of removing one. `patch: no` restores the behaviour the file had before this PR.

`main` is at 78.32% today, so the 75% floor has some headroom.

vipulpandey21: Small addendum, since I would rather not have you take the `patch: no` detail on trust, this repo's own history demonstrates it:

- `f12e5d22c` (2021-09-09) added `codecov.yml` with no `patch` key at all.
- `f7dd6aa19` (2021-10-27), titled *"Don't run patch coverage checks"*, is what introduced `patch: no`.

So patch checks were running while the key was absent, and turning them off required setting it explicitly. Consistent with that, PRs against `main` today report only `codecov/project` and no `codecov/patch`.

vipulpandey21:
@willgearty Turns out this got fixed while the PR was in review, so I haven't opened an issue for it.

ce2dbc98f (part of #5865, Jul 27) sorts the roster in the test by `added_class`, the same key the view sorts by. The test had been picking its prev/next pair out of `students_dict()`, which has no `ORDER BY` — when those two weren't adjacent in the view's order, it rendered the roster and returned 200 instead of 302.

Checked before writing it off: current `main` passed 12/12, and with just those sort lines removed it failed 3/12 with the same `Expected 302 redirect, got 200`.

Happy to still file the issue if you want it on record.

willgearty:
Thanks for looking into. No need to make an issue if it's already fixed.

hwatheod (approved):

PR #5895, opened 2026-07-26
Phase 8: Automatic permission synchronization and timeline grid clamping (Issue #1827)

### LINKED ISSUE #1827: Improve required module workflow

Supriyo760:
Description:
Implements automatic permission synchronization between program module schedules and backend permission records, along with frontend grid stability improvements for Weeks 11 and 12 of the New Module Management UI. This PR ensures that whenever an administrator schedules or updates a module's start and end dates on the timeline, its associated access permissions are automatically enabled and locked to that exact same date range, preventing unauthorized access to closed modules.

Changes

- **Automatic Permission Synchronization Engine** — Engineered `sync_permissions()` on `ProgramModuleObj` (`base.py`) which automatically finds or creates backend `Permission` records for Student, Teacher, and Volunteer roles, synchronizing their `start_date` and `end_date` to match the module's exact schedule.
- **Schedule Save Pipeline Integration** — Wired up `mod_hydrated.sync_permissions()` directly inside `module_schedule_update_api` (`views.py`) immediately following `mod.save()`. This ensures that whenever an administrator updates a module's schedule in the timeline UI, its associated permission records are updated simultaneously in the database without requiring manual intervention.
- **Comprehensive Module Handler Mapping (100% Coverage)** — Added default `permission_types = ()` and `get_permission_types()` on the base class, and applied explicit permission type mappings across all 14 core and optional registration module handlers (including `StudentClassRegModule`, `TeacherClassRegModule`, `StudentRegPhaseZero`, `RegProfileModule`, `VolunteerSignup`, `StudentSurveyModule`, `TeacherSurveyModule`, and more).
- **Timeline Scroll-Explosion Bug Fix** — Identified and resolved a UI rendering bug in `module_management.js` where out-of-range module start dates (e.g., an accidental year 2029 entry) would cause horizontal position calculations (`leftPct`) to exceed `100%`, stretching the browser's scrollable container width to 17 times the viewport and creating endless scrolling white space. Implemented robust mathematical clamping to keep block positioning safely between `0%` and `99.5%`.
- **Automated Test Coverage** — Added a dedicated unit test case `test_module_schedule_update_syncs_permissions` within `test_module_schedule_api.py` that verifies the creation and updating of permission records upon timeline edits, ensuring regression-free reliability.

Related Issue
Resolves #1827

Type of Change

- [x] New feature

Testing

- [x] Existing tests pass
- [x] New tests added
- [x] Manually verified

AI Disclosure
AI used for suggestions only.

Checklist

- [x] Self-reviewed
- [x] Documentation updated


willgearty (commented):
Thanks for implementing this, it works really seamlessly!

One broader thought:

- I noticed that the deadlines are only set when you edit a module. This makes sense; however, it means that most deadlines are set when a new program is created. Should we use the registration start and end date fields of the new program form (/manage/newprogram) to initialize the start/end dates for the modules and the associated permissions? (you can do this in a separate PR if you think it's a good idea but out of scope)

Image

And two things that definitely need fixing:

- The rows in the left panel seem to be slightly shorter than the rows in the right panel:

Image

- The question about the student acknowledgment module is now disabled?

Image

willgearty:
Ah, the row size mismatch is due to padding-bottom: 15px. Not a regression, we just missed it before. It should still be fixed in this PR.

willgearty (approved):
Looks like you accidentally reverted the padding fix in one of your recent commits. I've updated it again, and the two sides match up now. Thanks!

PR #5905, opened 2026-07-28
Week 8: Tests for dbmail sendto functions, and a fix for the CSRF failure page

vipulpandey21:
## Description

Week 8 of the regression-test expansion: coverage for the two remaining gaps named in the proposal, `esp/dbmail/sendto_fns.py` and `esp/web` views.

Along the way this PR picks up two small production fixes, both in code the new tests cover. Details in sections 2 and 3.

### 1. `esp/dbmail/test_dbmail_sendto_fns.py` — 36 tests, new file

`sendto_fns.py` decides which addresses actually receive a `MessageRequest`: the user's own, a guardian's, an emergency contact's, or a combination. It had no tests. Every function there returns `[]` on failure rather than raising, so a regression silently drops recipients instead of surfacing an error — which is why these tests assert on the address lists themselves rather than on "no exception was raised".

- **`send_to_self()`** — the returned pair, and the empty-list error path.
- **`send_to_guardian()` / `send_to_emergency()`** — the contact each one reads, a profile with no contact, a contact with a blank email, and a user with no profile at all. Also pins the `getattr()` fallback inside `_send_to_contact()`: an unrecognised contact kind silently mails the guardian, which is worth having written down.
- **`_send_to_combination()`** — concatenation order, duplicate addresses collapsed to the first pair, a missing component, and a raising component not discarding the addresses already collected from the others.
- **`MessageRequest` dispatch** — every entry in `SENDTO_FN_CHOICES` resolves to a callable, `''` maps to `send_to_self`, and both `ImproperlyConfigured` branches are reached by extending the choice list rather than by mocking the code under test.

### 2. `esp/dbmail/models.py` — the ESPError message was being dropped

`assert_is_valid_sendto_fn_or_ESPError()` called `ESPError(True, '<explanation>')`. `ESPError()` takes `(message, log)` and treats a bool first argument as the `log` flag while discarding the rest, so it returned the bare `ESPError_Log` class and the explanation was never attached. The middleware then logged an empty message, rendered an error page with no detail, and separately warned that raising the exception class is deprecated.

Passing the message first, as the rest of that module already does, gives `ESPError_Log(message)`. `log` already defaults to `True`, so logging behaviour is unchanged. `test_esperror_carries_the_explanation` asserts the rejected name, the "Invalid sendto function" text, the underlying `ImproperlyConfigured` reason and the support address all survive onto the raised exception.

### 3. `esp/web/views/csrf.py` — the ESP CSRF page was never being shown

`csrf.py` is the site's `CSRF_FAILURE_VIEW`. It renders `403_csrf_failure.html` and copies the result into an `HttpResponseForbidden`, reading `response.content` to do so.

Since 4dc8f53, `render_to_response()` returns an unrendered `TemplateResponse` instead of the rendered `HttpResponse` it returned before. Reading `.content` on an unrendered `TemplateResponse` raises `ContentNotRenderedError` — and this view wraps its whole body in `except Exception` so it can always show *something*, so the error was caught and converted into Django's default CSRF page.

Two consequences on `main` today:

1. The ESP-themed 403 page is never shown. Every CSRF failure falls back to Django's default page.
2. Django's default page prints the failure reason whenever `DEBUG` is on, so the reason reaches ordinary visitors even though this view deliberately gates it on `isAdministrator()`.

Verified end-to-end rather than by reading — a real rejection through the middleware:

```python
Client(enforce_csrf_checks=True).post('/accounts/login/')
```

returns Django's default 403 on `main` (no ESP page markers in the body), and the ESP-themed page after the fix.

The fix calls `render()` before reading the content. Django documents `render()` as a no-op once the content is baked, so no `is_rendered` check is needed. No other caller in the codebase reads `.content` straight off a `render_to_response()` result, so the blast radius of the underlying change appears limited to this one view — I have not touched `esp/utils/web.py`.

### 4. `esp/web/tests_csrf_view.py` — 17 tests, new file

The 403 status and rendered page, the no-referer explanation, the reason shown to administrators and withheld from anonymous and non-admin users under `DEBUG`, the program looked up from the request path, and the fallback to Django's view when rendering genuinely fails.

Three are regression guards for section 3. `403_csrf_failure.html` was adapted from Django's built-in CSRF page, so most of the wording is common to both and cannot tell them apart — assertions on that shared text pass whether our page renders or the view falls back. `test_serves_the_esp_page_and_not_djangos_default` asserts on text belonging to exactly one of the two pages, `test_middleware_rejection_serves_the_esp_page` goes through `CsrfViewMiddleware` with `enforce_csrf_checks`, and `test_renders_any_render_capable_response` covers a response defining `render()` but not `is_rendered`.

## Related Issue

Refs #3780

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

Locally, against this branch:

- `pytest esp/dbmail esp/web esp/program/modules/tests/commpanel.py -n 4` — **398 passed**, so the new tests are xdist-safe.
- Full suite, `pytest -n 4` — **1844 passed, 15 failed**. All 15 failures reproduce identically on pristine `main` in the same environment and are environmental (LaTeX/npm assets absent; `DependenciesTestCase` sees `psycopg2-binary` substituted for `psycopg2` locally). None are in `dbmail` or `web`.
- Line coverage on the two modules this PR targets: `esp/dbmail/sendto_fns.py` 100% (47 stmts), `esp/web/views/csrf.py` 100% (28 stmts).
- `flake8` clean on all changed files.

I also mutation-tested the new tests rather than trusting a green run: breaking the duplicate-address check and the contact lookup in `sendto_fns.py` killed 6 tests; breaking the dispatch in `models.py` killed 4; widening the admin gate in `csrf.py` to plain `settings.DEBUG` killed 2; reverting the render fix killed 5; restoring the old `ESPError` argument order killed 1. They fail when the code they cover is wrong.

## AI Disclosure

Used Claude Code as an assistant while writing and verifying this work, including the mutation testing described above and the end-to-end reproduction of the CSRF regression. I reviewed and validated everything before submitting.

## Checklist

- [x] Self-reviewed the code


esp/esp/web/views/csrf.py
```diff
@@ -36,6 +36,12 @@ def csrf_failure(request, reason=""):
         c['prog'] = prog
 
         response = render_to_response('403_csrf_failure.html', request, c)
+        # render_to_response() hands back an unrendered TemplateResponse, whose
+        # content is not available until it has been rendered. Reading .content
+        # first raises ContentNotRenderedError, which the except clause below
+        # would quietly turn into Django's default error page.
+        if hasattr(response, 'render') and not response.is_rendered:
```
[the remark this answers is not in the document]
vipulpandey21: Good catch on the underlying problem — fixed in 9ab1963, though not with the suggested default.

`getattr(response, 'is_rendered', True)` defaults to "assume rendered, skip `render()`". For the case you describe — a render-capable response that does not define `is_rendered` — that skips the render and then reads `.content` off an unrendered response, which is the exact failure this change exists to prevent. The safe default would be `False`.

But the check turns out to be unnecessary either way. Django documents `render()` as a no-op once the content is baked:

```python
def render(self):
    """Render (thereby finalizing) the content of the response.

    If the content has already been rendered, this is a no-op.
    """
    retval = self
    if not self._is_rendered:
        ...
```

(`django/template/response.py`)

So the guard is now just:

```python
if hasattr(response, 'render'):
    response.render()
```

which is correct whether or not the response has already been rendered, and makes no assumptions beyond the `render()` method itself.

Added `test_renders_any_render_capable_response`, which patches in a response defining `render()` but not `is_rendered`. It passes with the version above, and fails with `getattr(..., True)` — verified both ways locally, along with the full removal of the fix, which now fails 5 tests in this file.

esp/esp/dbmail/test_dbmail_sendto_fns.py
```diff
@@ -0,0 +1,394 @@
...
+                MessageRequest.SEND_TO_SELF),
+            sendto_fns.send_to_self,
+        )
+
+    def test_assert_is_valid_raises_esperror_for_unknown_name(self):
+        """Invalid names surface as ESPError, not ImproperlyConfigured.
+
+        Note: the raise site calls ESPError(True, '<explanation>'), and
+        ESPError() treats a bool first argument as the `log` flag and drops the
+        rest, so the explanation never reaches the user - the exception is
```
hwatheod: Please fix this at the raise site.

vipulpandey21: Fixed at the raise site in c32aff4. The message now goes first, so `ESPError()` returns `ESPError_Log(message)` instead of the bare class, the reason a name was rejected now reaches the log and the error page, instead of being blank.

Replaced the docstring with `test_esperror_carries_the_explanation`, which asserts the message content so the argument order can't silently regress. Checked the rest of the tree — no other call sites with this mistake.

hwatheod (approved):

PR #5948, opened 2026-08-17
Fix case-insensitive resend activation lookup

### LINKED ISSUE #4798: [Bug]: resend_activation_view  crashes with DoesNotExist when username case differs from registration

Oval17:
## Description

`AwaitingActivationEmailForm` accepts usernames case-insensitively, but `resend_activation_view` then did a case-sensitive `ESPUser.objects.get(username=...)`. Submitting `JOHNDOE` for stored `johndoe` raised `DoesNotExist` and 500'd.

Lookup now uses `username__iexact`.

## Related Issue

Closes #4798

## Type of Change

- [x] Bug fix
- [ ] New feature

## Testing

- [x] New tests added

## AI Disclosure

AI-assisted.

## Checklist

- [x] Self-reviewed the code


Oval17:
Resolved merge conflicts with `main` (HMAC-token activation flow) and addressed Copilot review:

- Replaced bare `ESPUser.objects.get(username__iexact=...)` with filtered queryset matching `AwaitingActivationEmailForm` (`exclude(password='emailuser')` + `awaiting_activation_Q()`) + `.first()`, raising `ESPError` instead of 500 on 0/multiple matches or race.
- Updated `ResendActivationCaseTest` to new `PendingActivation` flow (was legacy `password+_actkey`) and added rejection test for missing pending row. Kept upstream `DisableAccountPostOnlyTest`.

kkbrum (approved):

PR #5951, opened 2026-08-17
Fix mutable default arguments in Formstack API methods

### LINKED ISSUE #4907: [Bug]: Mutable default arguments used in Formstack API methods

Oval17:
## Description

Formstack API helpers used `args={}` as a default. That shared dict leaked `id` / `api_key` across calls.

Defaults are now `None` and a new dict is created per call.

## Related Issue

Closes #4907

## Type of Change

- [x] Bug fix
- [ ] New feature

## Testing

- [x] New tests added

## AI Disclosure

AI-assisted.

## Checklist

- [x] Self-reviewed the code


Oval17 (commented):
This should be good to merge.

willgearty (approved):
Looks good, thanks!

PR #5954, opened 2026-08-17
Validate sections query param in update_schedule_json

### LINKED ISSUE #4268: [Bug]: Update_schedule_json: validate "sections" query param (must be JSON list of section IDs)
### Problem 

update_schedule_json does json.loads(request.GET['sections']) and assumes the result is a list of section IDs.
But json.loads can return dict/int/string/etc., and later set-math assumes a list of ints.

### Examples
- sections="{}" parses to dict => treated like empty set => can remove all sections (dangerous)
- sections="5" parses to int => set(5) TypeError => 500 crash
- sections='["abc", 12, 12]' includes junk + duplicates

The issue is currently in  esp/esp/program/modules/handlers/onsiteclasslist.py -> update_schedule_json

### Acceptance
- sections="{}" -> HTTP 400 + JSON error message
- sections="5"  -> HTTP 400 + JSON error message
- valid list works normally end-to-end

### Steps to Reproduce

1. We have a Program with onsite module enabled.
2. We have a valid student user ID (replace <USER_ID>).
3. The student is enrolled in at least one section (for the “schedule wipe” repro).
4. Replace <HOST> and <PROGRAM_URL> with your environment.

### Expected Behavior

**Expected**
Return HTTP 400 with JSON error if sections is not a JSON list.
Sanitise list: cast to int, drop invalid, dedupe, and enforce max length.

### Actual Behavior

Right now the code does:

1. desired_sections = json.loads(request.GET['sections'])
2. Later, it does set(desired_sections) and assumes it’s a list of ints
3. But json.loads() can return a dict, int, string, etc. — and that can:
4. wipe schedules (e.g. {} becomes empty set)
5. 500 crash (e.g. "5" becomes int → set(5) crashes)

This is visible in the current code inside update_schedule_json

### Screenshots

_No response_

### Operating System

MAC OS

### Browser

Chrome latest

### Additional Context

_No response_

Oval17:
## Description

`update_schedule_json` accepted any JSON value for `sections`. `{}` became an empty set and could wipe enrollments; `5` caused `set(5)` and a 500.

Non-list values and non-integer IDs now return HTTP 400 with a JSON error.

## Related Issue

Closes #4268

## Type of Change

- [x] Bug fix
- [ ] New feature

## Testing

- [x] New tests added

## AI Disclosure

AI-assisted.

## Checklist

- [x] Self-reviewed the code


Oval17:
Addressed Copilot strictness + side-effect comments:

- `sanitize_section_ids` now strict: rejects `bool` (subclass of int) and floats incl. `1.0`; accepts only true `int` or decimal integer strings; dedupes preserving order.
- Malformed `sections` JSON now returns 400 immediately without `check_in` side effects; missing param still means no schedule change (check_in-only).
- Added regression tests: malformed JSON 400, floats/bools rejected, mixed `["abc", id]` sanitized, all-floats 400 without wiping.

PR #5957, opened 2026-08-17
Handle stale comm panel filter IDs in mailer warnings

### LINKED ISSUE #5116: [Bug]: Communications Portal can fail on stale/invalid filter id during warning generation
## Description
When using the Communications Portal, the mailer warning logic can fail hard if the stored filter ID is stale/invalid. Instead of a user-facing error, this can raise an assertion/exception path from `PersistentQueryFilter.getFilterFromID`, which breaks the flow while composing a comm panel message.

This seems especially likely if an admin has an old tab open, resumes a partially completed flow, or submits with outdated hidden form values.

## Steps to Reproduce
1. Open a program's comm panel and proceed to Step 2 (message composition) so a filter ID is in play.
2. Use a stale/invalid `filterid` value in the request path (for example by resubmitting an old tab/session, or modifying the hidden `filterid` before submit).
3. Trigger the path that computes mailer warnings.
4. Observe that the page can error instead of handling the invalid filter gracefully.

## Expected Behavior
The comm panel should handle invalid/stale filter IDs gracefully and show a user-facing error message (or redirect to restart list selection), rather than raising a server-side assertion/exception.

## Actual Behavior
The code path calls `PersistentQueryFilter.getFilterFromID` directly during warning generation. Invalid input can hit the assertion/exception path and break the request.

## Screenshots
N/A

## Operating System
macOS

## Browser
Chrome

## Additional Context
Related but not duplicate:
- #1655 is about incorrect query construction in generated filters.
- #3893 is about missing default "from" address in an older checklist path.

This issue is specifically about stale/invalid filter ID handling in comm panel warning flow.

Oval17:
## Description

`get_mailer_warnings` called `PersistentQueryFilter.getFilterFromID` unsafely. A stale or non-integer `filterid` raised `AssertionError` / `DoesNotExist` and broke the comm panel.

Invalid filters now add a user-facing warning and skip the grade-range check.

## Related Issue

Closes #5116

## Type of Change

- [x] Bug fix
- [ ] New feature

## Testing

- [x] New tests added

## AI Disclosure

AI-assisted.

## Checklist

- [x] Self-reviewed the code


Oval17:
Resolved conflict with `main` (`_q_has_grade_filter` robust grade check) and addressed both Copilot threads:

- Combined safe stale-filter lookup (try/except -> warning) with upstream Q-object grade detection: `if filter_obj is not None and not _q_has_grade_filter(filter_obj.get_Q())`.
- Kept `get_valid_filter_or_error` for `commprev`/`commfinal` so stale IDs raise helpful `ESPError` instead of 500 (Copilot moderate).
- Tests already in DB-backed `ProgramFrameworkTest` (second commit fixed SimpleTestCase issue).

PR #5963, opened 2026-08-21
feat: implement module schedule export and import functionality

### LINKED ISSUE #5942: Module schedule templates
It would be nice to be able to save a schedule for a set of modules and then import it for a future program. When importing, you would set a starting date/time, and then the modules would match the same relative open/close dates from the saved schedule. This should be integrated into the new program modules interface.

rafiya618:
### Module Schedule Export & Import Implementation

1. **Backend Endpoints (`esp/esp/program/views.py` & `urls.py`)**:

   * Added `module_schedule_export_api` (`GET /manage/<type>/<term>/module_schedule/export`): Serializes module settings (`seq`, `required`, `link_title`, etc.) and calculates relative `start_offset_seconds` and `end_offset_seconds` against an anchor timestamp.
   * Added `module_schedule_import_api` (`POST /manage/<type>/<term>/module_schedule/import`): Takes a JSON template and target start datetime, calculates new absolute dates for all modules, updates database records inside an atomic transaction, and invokes `sync_permissions()` to keep permissions synchronized.
   * Added missing imports (`timezone`, `parse_datetime`, `require_GET`).

2. **Frontend UI & CSS (`modules.html`, `module_management.css`, `module_management.js`)**:

   * Added **Save Schedule** and **Import Schedule** buttons in the module management toolbar.
   * Added `exportModal` and `importModal` popups in `modules.html`.
   * Updated CSS in `module_management.css`:

     * Centered form fields inside the modal body.
     * Added `box-sizing: border-box`, `white-space: pre-wrap`, `word-break: break-word`, and `overflow-x: hidden` to eliminate horizontal scrollbars completely.
   * Handled JSON clipboard copy, file download, file upload parsing (`FileReader`), and AJAX request dispatching in JavaScript.

3. **Automated Testing & Documentation**:

   * Added `test_export_schedule_api` and `test_import_schedule_api` in `test_module_schedule_api.py`.
   * Verified tests pass (`12 tests OK`).
   * Updated `docs/admin/module_schedule.rst` with endpoint specifications and testing notes.

## Related Issue

Closes #5942 

## Type of Change

* [ ] Bug fix
* [x] New feature
* [ ] Breaking change
* [x] Documentation update
* [ ] Refactor / cleanup
* [ ] CI/CD or build change

## Testing

* [x] Existing tests pass
* [x] New tests added (if applicable)
* [ ] Manually verified

All unit tests in `esp.program.modules.tests.test_module_schedule_api` passed successfully in Docker.

## Checklist

* [x] Self-reviewed the code
* [x] Updated documentation (if needed)

## Some Screenshots

image
image
image
image
image

### DIFF
--- a/esp/esp/program/views.py
+++ b/esp/esp/program/views.py
@@ -62,10 +62,13 @@ from django.core.cache import cache
 from django.urls import reverse
 from django.forms.models import model_to_dict
 from django.views.decorators.csrf import csrf_exempt
-from django.views.decorators.http import require_POST
+from django.views.decorators.http import require_POST, require_GET
+from django.utils import timezone
+from django.utils.dateparse import parse_datetime
 from django.template.loader import render_to_string
 from django.http import HttpResponse
 from django import forms
+
 
 from esp.program.modules.module_ext import ClassRegModuleInfo, StudentClassRegModuleInfo
 from esp.program.models import Program, TeacherBio, RegistrationType, ClassSection, StudentRegistration, VolunteerOffer, RegistrationProfile, ClassCategories, ClassFlagType, StudentSubjectInterest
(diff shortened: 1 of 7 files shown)


Supriyo760 (commented):
Hi @rafiya618, great work on this feature.
I tested your branch locally in my browser and had a few notes from testing and reading the code
You can check for these after addressing the copilot reviews:

1. Timeline view after import:
When testing the import modal, the modal closes and shows "Imported schedule successfully", but the timeline does not change at all and dates in the database remain unchanged.

2. Anchor date clarity:
The export API calculates offsets starting from the earliest module date (`min(valid_start_dates)`). In many programs, this is an early teacher setup module that opens weeks before student registration. 
When an admin sees the input field labeled "New Program Starting Date", they will likely think of the main program or student registration start date. This can cause all dates to shift further into the future than they intended. 
It would help to show a small note in the modal explaining what date the schedule was originally anchored to, or clarifying what the "New Program Starting Date" field represents.

3. File upload error handling:
In `handleImportFile`, adding a `reader.onerror` handler would be good so that if a user accidentally picks a corrupted or unreadable file, a clear error toast is displayed.

4. Unit tests:
The current tests cover the happy path nicely. It would be great to add two more test cases:
- A test where one module has an invalid date (`start_date >= end_date`), verifying that the entire import rolls back and no earlier modules are partially saved.
- A test checking that non-admin users get a 403 Forbidden when calling the export and import endpoints.

5. Module list and timeline row alignment: 
The module items on the left sidebar and the corresponding timeline rows on the right should stay parallel and vertically aligned with each other so the labels always match up cleanly with their timeline bars.

Supriyo760 (commented):
Wasn't expecting it to look like this
plz test locally before pushing any changes

Image

Image

esp/esp/program/views.py
```diff
@@ -1872,3 +1876,194 @@ def module_schedule_reorder_api(request, program_type, program_term):
...
+        })
+
+    return JsonResponse({
+        "success": True,
+        "program_id": prog.id,
+        "program_name": prog.name,
+        "anchor_date": anchor_dt.isoformat(),
+        "schedule": {
+            "version": "1.0",
+            "modules": module_templates
```
Supriyo760, line 1934: move anchor_date inside the schedule

Supriyo760 (commented):
In module_schedule_import_api, strings from the JSON are saved directly without checking their length. The database field max length is 64 characters. 
Suppose if someone imports a JSON with a longer string, Django will throw a DataError which will cause a 500 Internal Server Error instead of a clean 400 Bad Request with a helpful message.
The module_schedule_update_api already does this check correctly, so the import API should do the same.

esp/esp/program/views.py
```diff
@@ -2053,12 +2053,15 @@ def module_schedule_import_api(request, program_type, program_term):
...
+                raise ValueError("No valid module configurations were found or updated from the provided schedule payload.")
+
         return JsonResponse({
             "success": True,
             "updated_count": updated_count,
             "target_start_date": target_anchor_dt.isoformat()
         })
-    except ValueError as e:
+    except (ValueError, TypeError) as e:
         return JsonResponse({"success": False, "error": str(e)}, status=400)
```
Supriyo760, line 2077: Split these 2 in seperate handlers cause TypeError messages can contain internal Python details that shouldn't be shown to end users.
Try this way
except ValueError as e:
    return JsonResponse({"success": False, "error": str(e)}, status=400)
except TypeError:
    return JsonResponse({"success": False, "error": "Invalid data format in schedule payload"}, status=400)

rafiya618:
@Supriyo760 Just to clarify, do you want the vertical scroll bar removed and want the previous layout for timeline and modules? in the previous implementation, when more modules were added, the timeline bars shrunk a bit to fit them properly, but the recent update doesn't do that. 
Also, The button interface looks completely fine on my screen, but I will add browser compatibility styles to the CSS. You might be seeing a different layout due to browser rendering differences.

Previously: 
Screenshot From 2026-09-02 13-18-30

Recent Change (I did):
image

Supriyo760:
@rafiya618 
No need to remove the vertical scroll bar.
But the size of the bars of both side should the same as it was earlier.
Cause at present the size of the bars are too big and one has to scroll again and again for switching between them and they can't have a broader view of the timeline...
Plus the left side bars should be in parallel with the postions of their right side bars respectively.

rafiya618:
Updated the timeline rendering logic to dynamically adjust row heights based on the number of modules: rows remain comfortably sized for smaller module counts, progressively shrink as modules increase so up to 12 modules can fit within the available viewport, and then stop shrinking at a compact minimum height. Once the module count exceeds 12, vertical scrolling is introduced while maintaining the compact row size, with the horizontal scrollbar remaining fixed in position.

@Supriyo760, does this approach work for you, or would you prefer that we switch back completely to the previous version with no vertical scrollbar at all?

Supriyo760:
@rafiya618 u said 12 modules can fit and after that if more modules are added then scroll bar will appear but i can only 8 complete bars
image

rafiya618:
@Supriyo760 Thats working exactly as I said on my side. Plus I am also seeing the UI that is being shown to you is a bit different from my side. Any ideas where the issue is coming from?

Supriyo760:
@rafiya618 can u share pictures of what it looks (at present) like in your side.

rafiya618:
@Supriyo760 

Less Than 12 Modules:
image

More than 12:
image

Supriyo760:
Screenshot (1226)
Screenshot (1225)

@rafiya So the problem is a hardcoded assumption that disables the vertical scrollbar whenever a program has $\le 12$ modules, assuming 12 rows will always fit on any screen (which isn't always true). As seen in testing, at 75% scaling all 12 rows are visible, but at 100% scaling only 8 rows fit within the 72vh container, leaving any remaining modules clipped.

To fix this, instead of relying on a hardcoded count threshold, You can dynamically toggle the scrollbar by checking if the total rendered row height exceeds the container's available pixel height.

Supriyo760 (commented):
Hi @rafiya,
Thanks for the update.
The dynamic scrollbar logic works well to prevent clipping, but I wanted to clarify the design goal:
The intention is for all 12 rows to fit on the screen at a glance without needing vertical scrolling, at both 75% and 100% zoom.

Currently, at 100% zoom, only about 8.5 rows fit in the viewport because the container height is restricted while the rows have a minimum height limit. This forces the remaining rows into a vertical scrollbar.

Could you allow the timeline container to expand taller so that all 12 modules fit cleanly on screen at 100% zoom?

Current Status:
at 75%
Image
at 100%
Image

Supriyo760:
Hi @rafiya,
Thanks for the update.
The dynamic scrollbar logic works well to prevent clipping, but I wanted to clarify the design goal:
The intention is for all 12 rows to fit on the screen at a glance without needing vertical scrolling, at both 75% and 100% zoom.

Currently, at 100% zoom, only about 8.5 rows fit in the viewport because the container height is restricted while the rows have a minimum height limit. This forces the remaining rows into a vertical scrollbar.

Could you allow the timeline container to expand taller so that all 12 modules fit cleanly on screen at 100% zoom?

Current Status:
at 75%
Image
at 100%
Image

rafiya618:
@Supriyo760 sure, It would be better to let the timeline container expand so all 12 rows fit at 100% zoom.

One question though: the buttons I am seeing are not fitting well within the container. Do you think I should add a dropdown option to save/import the schedule to make the interface more cleaner? It may look something like this (thats just a prototype for now):
a26df232-eac8-4d85-aced-66174a53c706

Supriyo760:
@rafiya618 hmm...dropdown option looks better.

PR #5840, opened 2026-06-08
Week 2: Direct unit tests for ClassCreationController

vipulpandey21:
## Description

Adds direct unit test coverage for `ClassCreationController` in `classreg.py`.

Existing coverage in `test_class_creation.py` validates the HTTP layer (`makeaclass` / `editclass` views, form validation, co-teacher management), but `ClassCreationController` itself contains 16+ methods with no direct unit tests. This leaves business logic regressions undetected when view-level tests still return HTTP 200.

This PR introduces direct controller-level tests to validate core behavior independently of the HTTP layer.

#### `GetCustomFieldsTest`

Tests `get_custom_fields()`:

* Returns an empty dict when no tag is configured
* Raises `AttributeError` when the configured form class does not exist
* Returns an empty dict for an empty JSON field list

#### `SetClassDataTest`

Tests `set_class_data()`:

* Standard fields (`class_info`, `grade_min`, etc.) are written to `cls.dict`
* `title` is assigned through `cls.title`
* `category` is retrieved by ID and assigned correctly
* Fields prefixed with `section_` are excluded from `cls.dict`

  * Regression guard against the previous identity-comparison bug
* Custom fields are routed to `cls.custom_form_data`
* `custom_form_data` remains empty when no custom fields exist
* `cls.duration` is converted to `Decimal`
* Explicitly excluded fields (`resources`, `viable_times`, etc.) are not stored in `cls.dict`

#### `TeacherTimeTest`

Tests `teacher_has_time()` and `require_teacher_has_time()`:

* Returns `True` when the teacher has sufficient capacity
* Returns `False` when requested hours exceed program limits
* Does not raise when capacity constraints are satisfied
* Raises `ESPError` when limits are exceeded
* Uses "you" in error messages for the current user
* Includes the teacher's name when validating another user

#### `UpdateClassSectionsTest`

Tests `update_class_sections()`:

* Creates sections up to the requested count
* Removes excess sections when count is reduced
* Updates all section durations to match `cls.duration`
* Performs no changes when requested and existing counts match
* Correctly handles reduction to zero sections

### Test Summary

* 4 test classes
* 22 controller-level unit tests
* Existing test suite passes
* New tests verified locally

### Why This Matters

These tests protect critical business logic that HTTP-layer tests cannot reliably validate, ensuring regressions in field routing, capacity checks, custom field handling, and section management are caught immediately.

### Related Issues

Refs #3780
Refs #4480

### AI Disclosure

[...]

PR #5965, opened 2026-08-22
Fix test docstring referencing removed code

vipulpandey21:
## Description

One of the test docstrings in `test_classreg_controller.py` (added in #5840) talked about a bug in code that isn't there anymore, a string comparison using `is not` instead of `!=`. That code was already fixed before this test file existed, so the docstring was just describing something nobody reading it could actually go check.

Rewrote it to say what the test actually checks now: fields starting with `section_` don't get written into `cls.__dict__`. Fixed the same kind of line in the file's top docstring too.

No test logic changed, just the comments.

## Related Issue

Refs #3780

## Type of Change

- [x] Documentation update

## Testing

- [x] Existing tests pass (`pytest esp/program/controllers/test_classreg_controller.py`, 22 passed)
- [x] flake8 clean


willgearty (approved):
Thanks!

PR #5968, opened 2026-08-25
fix(middleware): guard against missing request.session in process_response

### LINKED ISSUE #5967: 'WSGIRequest' object has no attribute 'session'
We are getting  `'WSGIRequest' object has no attribute 'session'` at [this line](https://github.com/learning-unlimited/ESP-Website/blob/aac171586147ab76324a3794dfea426b5834eee8/esp/esp/middleware/espauthmiddleware.py#L150).  (This permalink is pointing to stable-release-16 since that's what current production servers are running, where we see the error. The code still exists in `main` although it was moved to a [different place](https://github.com/learning-unlimited/ESP-Website/blob/main/esp/esp/middleware/espauthmiddleware.py#L82) in the file by [this PR](https://github.com/learning-unlimited/ESP-Website/pull/5884).)

The missing session attribute is probably related to the [accepted answer here](https://stackoverflow.com/questions/11783404/wsgirequest-object-has-no-attribute-session).  Although we aren't using the `www` or `/` flags mentioned in the answer, it is the case that our process_request has a guard for session not existing but our process_response doesn't.

I'm not really sure what why the `session` is missing. We could add a guard in `process_response` but I'm not sure what we do in the case where it is missing.  In the latest code in `main`, the comment was changed to say that it has to do with `Vary: Cookie`. I don't know the details of how it all flows through the middleware but it is important that the cache doesn't serve cached pages from one user to a different user (or to unauthenticated users), which could happen if we don't set `Vary: Cookie` properly.

Janvi-kapoor:
### Description
This pull request addresses #5967, resolving an `AttributeError: 'WSGIRequest' object has no attribute 'session'` that occurs when requests lack a `session` attribute during the `process_response` lifecycle (e.g., requests handled before `SessionMiddleware` executes or custom error responses).

### Key Changes
* Added defensive `hasattr(request, 'session')` guard before setting `request.session.accessed = request.session.modified`.
* Safely handled `request.session` retrieval when extracting `user_morph` data to prevent `AttributeError`.
* Added comprehensive unit tests in `esp/esp/tests/test_middleware.py` covering both anonymous and authenticated requests that lack a `session` attribute.

### Related Issue
Closes #5967

### Checklist
- [x] Code strictly contains functional changes without unrelated edits.
- [x] Unit tests added and verified.
- [x] All CI checks passing.

### DIFF
--- a/esp/esp/middleware/espauthmiddleware.py
+++ b/esp/esp/middleware/espauthmiddleware.py
@@ -79,7 +79,8 @@ class ESPAuthMiddleware(AuthenticationMiddleware):
         ## the no_set_cookies early-return so that cacheable views (decorated with
         ## @disable_csrf_cookie_update) also benefit from the workaround.
         ## -- aseering 11/1/2010, moved before no_set_cookies check
-        request.session.accessed = request.session.modified
+        if hasattr(request, 'session'):
+            request.session.accessed = request.session.modified
 
         ## This gets set if we're not supposed to modify the cookie
         if getattr(response, 'no_set_cookies', False):
@@ -105,10 +106,9 @@ class ESPAuthMiddleware(AuthenticationMiddleware):
                     "%a, %d-%b-%Y %H:%M:%S GMT"
                 )
             ret_title = ''
-            try:
-                ret_title = request.session['user_morph']['retTitle']
-            except KeyError:
-                pass
+            ret_title = ''
+            if hasattr(request, 'session'):
+                ret_title = request.session.get('user_morph', {}).get('retTitle', '')
 
             # URL-encode some data since cookies don't like funny characters. They
             # make the chocolate chips nervous.
--- a/esp/esp/tests/test_middleware.py
+++ b/esp/esp/tests/test_middleware.py
@@ -94,3 +94,30 @@ class ESPAuthMiddlewareProcessResponseTest(TestCase):
         self.assertIn('cur_username', cookie_names)
         self.assertIn('cur_userid', cookie_names)
         self.assertIn('cur_email', cookie_names)
+
+    def test_process_response_without_session_attribute_anonymous(self):
+        """Should handle requests without session attribute for anonymous users without error."""
+        request = self.factory.get('/')
+        request._cached_user = AnonymousESPUser()
+        request.COOKIES = {}
+        response = HttpResponse()
+        result = self.middleware.process_response(request, response)
+        self.assertIs(result, response)
+
+    def test_process_response_without_session_attribute_authenticated(self):
+        """Should handle requests without session attribute for authenticated users without error."""
+        user = ESPUser.objects.create_user(
+            username='nosessionuser',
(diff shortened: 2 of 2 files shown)


Janvi-kapoor:
Hi @hwatheod , the defensive guards for `request.session` and unit tests have been implemented, and all 10 CI checks are passing green. This is ready for review whenever you have a moment!

Janvi-kapoor:
Hi @hwatheod @willgearty, all review feedback has been addressed (UTF-8 BOM removed and session retrieval simplified), and all CI checks/unit tests are passing green. This PR is ready for your review and approval!!

hwatheod:
I'm not sure this is right. The comment says that the purpose of that `session` line is to prevent SessionMiddleware from adding "Vary: Cookie" just because the session was read. Now we are just skipping that whenever `session` is not present.  So it sounds like `Vary: Cookie` will get added whenever the `session` field is not present. Is that the right thing to do? Before we can answer this question I think we need to understand why the session field is not present.

Janvi-kapoor:
> I'm not sure this is right. The comment says that the purpose of that `session` line is to prevent SessionMiddleware from adding "Vary: Cookie" just because the session was read. Now we are just skipping that whenever `session` is not present. So it sounds like `Vary: Cookie` will get added whenever the `session` field is not present. Is that the right thing to do? Before we can answer this question I think we need to understand why the session field is not present.

Thanks for raising this @hwatheod!

To clarify why `request.session` is absent and how `Vary: Cookie` behaves:

1. **Why `request.session` is missing:**
   In Django's classic middleware chain (`MIDDLEWARE_CLASSES` / `MIDDLEWARE`), if an earlier middleware (such as `SecurityMiddleware` for SSL redirect or `CommonMiddleware` for `APPEND_SLASH`) returns an `HttpResponseRedirect` directly during `process_request`, subsequent request middlewares—including `SessionMiddleware`—are skipped. However, Django still unwinds the response chain in reverse order, calling `process_response` on outer middlewares (like `ESPAuthMiddleware`) where `request.session` was never attached. This matches the behavior described in `process_request` which already had a guard for missing session.

2. **Why `Vary: Cookie` is not incorrectly added:**
   `SessionMiddleware.process_response` only adds `Vary: Cookie` if `request.session.accessed` or `request.session.modified` is `True`. If `request.session` does not exist on the request, `SessionMiddleware.process_response` also does nothing and does not add `Vary: Cookie`. Therefore, skipping the `accessed = False` reset when there is no session object is safe and will not cause unwanted caching or unwanted `Vary: Cookie` headers on redirected/session-less responses.

Let me know if this clarifies the flow or if you'd like me to add an explicit integration test covering the early redirect scenario! can you tell me what changes i have to do ???
hey @kkbrum can you review this pr.

Janvi-kapoor:
Hey,  @hwatheod can you give any feedback on this PR? If it is not useful, I will close it because I have already received too many PR tags

PR #5975, opened 2026-08-30
Fix TypeError on /archives/teachers/ and /archives/programs/

### LINKED ISSUE #5974: [Bug]:/archives/teachers/ and /archives/programs/ return a 500 on live sites
### Description

Hi Mentors @willgearty ,

Visiting /archives/teachers/ or /archives/programs/ on a live ESP site does not load an archives page. The request hits archives() in esp/web/views/main.py, which always calls the selected handler with four arguments: the request, category, options, and sort parameters. The class handler accepts that extra argument, so /archives/classes/ works. The teacher and program handlers only accept three arguments, so the view raises TypeError: takes 4 positional arguments but 5 were given and Django returns a 500.

This is not only a local code issue: https://esp.mit.edu/archives/teachers/ currently shows the public “Whoops! It appears the server just made an error” page. Existing unit tests miss it because they call archive_teachers and archive_programs directly with three arguments and never go through the URL dispatcher. The smallest fix is to add an optional sortorder=None argument to those two handlers so they match archive_classes, even if they still ignore sorting for now.

### Steps to Reproduce

Run the site locally.
Open /archives/teachers/ or /archives/programs/.
Observe a 500 / TypeError instead of the archives template.
Suggested fix

### Expected Behavior

running site flawless backend integrated frontend

### Actual Behavior

errors 

### Screenshots

Web:
Image

Local:
/archives
Image
/programs
Image

### Operating System

Ub

### Browser

Chrome

### Additional Context

Fix: Give archive_teachers and archive_programs the same optional sortorder=None parameter as archive_classes, even if they ignore it for now. Alternatively, only pass sortparams when the selected handler needs it.

sanketkokate712:
## Summary
- `/archives/teachers/` and `/archives/programs/` 500 on live sites (for example https://esp.mit.edu/archives/teachers/) because `archives()` always passes `sortparams` and those handlers did not accept it.
- Give `archive_teachers` and `archive_programs` the same optional `sortorder=None` argument as `archive_classes`.
- Add dispatcher tests so a future signature mismatch fails in CI instead of only on the URL.

## Test plan
- [ ] Hit `/archives/teachers/` and `/archives/programs/` locally; both should return 200 (stub pages), not a 500.
- [ ] Confirm `/archives/classes/` still works.
- [ ] Run `pytest esp/web/tests_archives.py` (or the project equivalent in Docker).

Closes #5974.

### DIFF
--- a/docs/dev/walkthroughs/esp/esp/web-README.MD
+++ b/docs/dev/walkthroughs/esp/esp/web-README.MD
@@ -97,8 +97,8 @@
 | `filter_archive` | Function | `views/archives.py` | Applies a list of `ArchiveFilter` objects as Django ORM filters to an `ArchiveClass` queryset. |
 | `title_heading` | Function | `views/archives.py` | Returns the first character of a title string for use as a jump-to heading, or `'[No Title]'` if empty. |
 | `archive_classes` | Function | `views/archives.py` | Renders a filterable, sortable, paginated list of historical `ArchiveClass` records. |
-| `archive_teachers` | Function | `views/archives.py` | Stub: sets `selection`, `category`, `options` in context and renders `program/archives.html` — **crashes with `TypeError`** when called via `archives()` because the dispatcher passes more positional arguments than this function accepts. |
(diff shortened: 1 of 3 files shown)


willgearty (approved):
Looks good, thanks for your contribution!

PR #5984, opened 2026-09-04
Add search filter to admin class list page

### LINKED ISSUE #5469: [Feature]: Add search filter to class list page
### Problem Statement

When managing programs with many classes, administrators need to quickly locate specific classes in the class list view. 
**Currently the interface lists all classes without a built-in search mechanism, which makes navigation difficult when the number of classes is large.** This becomes especially problematic during onsite program management when administrators need to quickly access specific classes.

### Proposed Solution

**Add a search input field above the class list table that allows administrators to filter classes in real time.**

The search filter will:
- Filter classes as the user types
- Perform case-insensitive matching
- Match text within the table row (class title, teacher name, etc.)

**The feature can be implemented entirely on the frontend using JavaScript without requiring backend changes.**

### Alternatives Considered

An alternative approach would be implementing server-side search through Django views.
However, for this use case a client-side filter is simpler and provides instant feedback without additional server requests.

### Additional Context

This feature would improve usability when administrators manage large numbers of classes and could also support future work on onsite administration tools.

Example implementation approach:
- Add search input above class table
- Filter rows using JavaScript

laypatel13:
## Description

Adds a live search filter above the admin "Manage Classes" table so administrators can quickly narrow down classes by title, teacher, or code as they type, without needing to scroll through the full list.

## Related Issue

Closes #5469

## Type of Change

- [x] New feature

## Testing

- [x] Manually verified

Screenshot 2026-09-04 at 7 41 30 AM

Manually verified in a local Docker developer environment with a test program containing 8 sample classes:

- Filtering by title, class code, and teacher name too
Screenshot 2026-09-04 at 8 58 32 AM

Screenshot 2026-09-04 at 8 59 43 AM

- Case-insensitive matching (e.g. "PyTHon" matches "Intro to Python Programming")
Screenshot 2026-09-04 at 9 01 12 AM

- A query with no matches shows the "no results" message instead of an empty table
Screenshot 2026-09-04 at 9 02 20 AM

- Clearing the search box restores all rows, Verified no JS console errors while typing.

> No automated JS test harness exists for this template in the current codebase, so no new tests were added; happy to add coverage if there's a preferred pattern to follow.

## AI Disclosure

AI assisted

## Checklist

- [x] Self-reviewed the code

### DIFF
--- a/esp/templates/program/modules/adminclass/listclasses.html
+++ b/esp/templates/program/modules/adminclass/listclasses.html
@@ -29,6 +29,54 @@ json_data = {};
   // Start the AJAX request to get the class data and display it
   json_fetch(["class_subjects"], fillClasses, json_data, function() { alert("Error loading class data"); });
 
+  // Class list search filter
+  function filterClassRows() {
+    var $tbody = $j('#classes_anchor');
+
+    // fillClasses() rebuilds the tbody's contents on load, which can wipe
+    // out our "no results" row, so re-create it here if it's missing.
+    var $noResultsRow = $j('#no_results_row');
+    if ($noResultsRow.length === 0) {
+      $noResultsRow = $j(
+        '<tr id="no_results_row" style="display: none;">' +
+        '<td colspan="7"><em>No classes match your search.</em></td>' +
+        '</tr>'
+      );
+      $tbody.append($noResultsRow);
+    }
+
+    var filter = $j('#classSearch').val().trim().toLowerCase();
+    var visibleCount = 0;
+
+    $tbody.children('tr').not('#no_results_row').each(function() {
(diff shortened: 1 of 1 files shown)


laypatel13:
I have pushed the changes Copilot flagged.!

> Colspan is calculated based on whether the moderator module is on, instead of being hard-coded.

> And for the search filter, I left out the Management Options column so button text like `"Edit"/"Manage"` doesn't mess up the search results.

laypatel13:
Bumping this one - addressed the Copilot feedback a while back (dynamic colspan and excluding the Management Options column from search matching), just needs a look whenever someone has a minute 😁

PR #5997, opened 2026-09-06
Don't assume `classreg` exists in `students()`

willgearty:
CI tests in #5988 revealed that depending on the state of the database, this dictionary pull could error. This adds a small check to prevent it (and some associated regression tests).

Done by Claude Code.

### DIFF
--- a/esp/esp/program/models/__init__.py
+++ b/esp/esp/program/models/__init__.py
@@ -686,10 +686,8 @@ class Program(models.Model, CustomFormsLinkModel):
     def _classreg_students(self):
         """Students registered for at least one class.
 
-        The 'classreg' key is contributed by StudentClassRegModule, so it is
-        absent for a program that does not use that module. Such a program has
-        no class registrations to count, so treat it as none rather than
-        raising KeyError out of the program cap checks below.
+        StudentClassRegModule contributes the 'classreg' key, so a program
+        without that module has none to count.
         """
         return self.students().get('classreg', ESPUser.objects.none())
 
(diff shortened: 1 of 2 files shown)


willgearty:
Comment cleanup done in f0381d4 — trimmed the `_classreg_students` docstring and the test comment, and dropped a redundant inline one. No behaviour change; flake8 clean, `ProgramCapTest` 5 passed, and `ModuleExistenceTest.test_studentreg` 12/12.

One thing worth recording, since I nearly pushed a wrong change on top of it. Your f46c018 dropped the `self.program.program_size_max = 3` line along with the `objects.get` fix, and I started to restore it, reasoning that `user_can_join()` needs a cap set or it returns early via `else: return True` and never reaches the count — which would have made the final loop vacuous.

That reasoning was wrong, and I checked rather than shipping it. `ProgramFrameworkTest` already sets `program_size_max = 3000`; I had inferred `None` from the model field's `null=True` instead of looking at the fixture. Confirmed by instrumenting the test:

```
PROBE program_size_max = 3000
PROBE size_by_grade tag = None
E   KeyError: 'classreg'   at models/__init__.py:707
```

So your version already exercises the full path — `user_can_join` → `_user_can_join_at_all` → `_student_is_in_program` — and I reverted my restoration. Nothing needed here; noting it only because it means the loop is doing real work, not just asserting `True`.

Also confirming Copilot's `objects.get` finding was correct and your fix is the right shape: `handler` isn't unique, so a global `get()` could raise `MultipleObjectsReturned`. Filtering the program's own modules and asserting non-empty is strictly better than what I wrote.

---
_Generated by [Claude Code](https://claude.ai/code)_

kkbrum (approved):

PR #6024, opened 2026-09-10
Fix: Use passwd_recover for admin password reset (#5998)

### LINKED ISSUE #5998: [Bug]: Admin login page never shows the "Forgotten your password?" link
### Description

`admin/login.html` reverses a URL name that does not exist in this project, so the
password recovery link on the Django admin login page is never rendered. The site
has a working password recovery page and the user-facing login links to it, but
the admin login page — where a locked-out admin actually lands — shows nothing.

[esp/templates/admin/login.html#L52-L57](https://github.com/learning-unlimited/ESP-Website/blob/844310f1b6adc979f19665d86c33e2c3a1e213d5/esp/templates/admin/login.html#L52-L57):

```django
{% url 'admin_password_reset' as password_reset_url %}
{% if password_reset_url %}
<div class="password-reset-link">
  <a href="{{ password_reset_url }}">{% translate 'Forgotten your password or username?' %}</a>
</div>
{% endif %}
```

`admin_password_reset` comes from `django.contrib.admin`'s own URLs, which this
project never includes — there is no such name anywhere in the URLconf. And the
`as var` form of `{% url %}` is [documented](https://docs.djangoproject.com/en/5.2/ref/templates/builtins/#url)
to swallow `NoReverseMatch` and set the variable to an empty string rather than
raising. So the `{% if %}` is permanently false, the block is dead, and there is
no error, no 500, and no log line to notice.

Meanwhile the recovery view exists and works, as
[`passwd_recover`](https://github.com/learning-unlimited/ESP-Website/blob/844310f1b6adc979f19665d86c33e2c3a1e213d5/esp/esp/users/urls.py#L23) -> `/myesp/passwdrecover`. It is
already linked from `contact.html`.

I found this by extracting every `{% url '...' %}` name used in the templates (47
of them) and reversing each against the real URLconf. `admin_password_reset` was
the only one that did not resolve.

### Steps to Reproduce

1. Open `/admin/login/` on any deployment.
2. Look for the "Forgotten your password or username?" link.

### Expected Behavior

The link renders and points at the site's password recovery page, the same one
the user-facing login at `/myesp/login/` links to.

### Actual Behavior

The link is absent. The rendered page contains no `password-reset-link` element
at all.

Confirmed on production at `https://esp.mit.edu/` (public pages, no login):

| URL | Result |
| --- | --- |
| `/myesp/login/` | 200 — has "Forgot your password?" -> `/myesp/passwdrecover/` |
| `/admin/login/` | 200 — no `password-reset-link`, no "Forgotten your password" |
| `/myesp/passwdrecover/` | 200 — recovery flow is live and working |

That deployment is serving this template rather than an override: the page carries
this project's `href="/manage/programs"` notice, which stock Django's admin login
does not have. The complete link list on the production admin login page is
`/manage/programs` plus four stylesheets.

### Screenshots

Image
Image

### Operating System

Win 11

### Browser

Chrome

### Additional Context

Alternative fix, if the intent was to use Django's own admin password reset flow
rather than this project's: include `django.contrib.auth`'s password reset URLs so
that `admin_password_reset` resolves. That is a larger change and duplicates a
recovery flow the site already has, so pointing at `passwd_recover` seems like the
right call — happy to go the other way if maintainers prefer.

sanketkokate712:
This change ensures that the "Forgotten your password or username?" link is rendered properly and routes locked-out admins to the existing password recovery page (/myesp/passwdrecover), consistent with the user-facing login form.

02-after-link-renders
Fixes #5998

### DIFF
--- a/esp/templates/admin/login.html
+++ b/esp/templates/admin/login.html
@@ -49,7 +49,7 @@
     {{ form.password.label_tag }} {{ form.password }}
     <input type="hidden" name="next" value="{{ next }}">
   </div>
-  {% url 'admin_password_reset' as password_reset_url %}
+  {% url 'passwd_recover' as password_reset_url %}
   {% if password_reset_url %}
   <div class="password-reset-link">
     <a href="{{ password_reset_url }}">{% translate 'Forgotten your password or username?' %}</a>


Oval17 (commented):
check once copilot review 
rest code looks good to me

sanketkokate712 (commented):
checked.

Oval17 (approved):

willgearty:
Please revert the unrelated test changes. The flaky test was fixed in https://github.com/learning-unlimited/ESP-Website/pull/5997.

sanketkokate712:
I have reverted the unrelated test changes, as the issue regarding the flaky test was addressed in #5997. 
Please let me know if any further fixes are required.

willgearty (approved):
Thanks!

PR #6025, opened 2026-09-10
Fix: Handle use_student_apps option in lottery assignment (#5996)

### LINKED ISSUE #5996: [Bug]: Lottery returns HTTP 500 when the `use_student_apps` option is enabled
### Description

Turning on the "use student application ranks" option on the Lottery Frontend page
makes the lottery run fail with a `TypeError` before a single assignment is
computed. The option is a plain text box on the admin page, so any admin can set
it, and `lottery_execute` only catches `LotteryException`
([lotteryfrontendmodule.py#L66-L71](https://github.com/learning-unlimited/ESP-Website/blob/844310f1b6adc979f19665d86c33e2c3a1e213d5/esp/esp/program/modules/handlers/lotteryfrontendmodule.py#L66-L71)),
so the `TypeError` escapes as a 500 instead of the JSON error the page knows how
to display.

The option as rendered on the admin page
([lotteryfrontendmodule.py#L32](https://github.com/learning-unlimited/ESP-Website/blob/844310f1b6adc979f19665d86c33e2c3a1e213d5/esp/esp/program/modules/handlers/lotteryfrontendmodule.py#L32),
[lottery.html#L26](https://github.com/learning-unlimited/ESP-Website/blob/844310f1b6adc979f19665d86c33e2c3a1e213d5/esp/templates/program/modules/lotteryfrontendmodule/lottery.html#L26)):

![Lottery admin page showing the use_student_apps option]

Setting it to `True` and clicking "Generate Student Schedules":

![TypeError 500 page]

There are three separate defects stacked on this one path, so the feature cannot
have worked for a long time.

**1. `QuerySet + QuerySet` — [lottery.py#L305](https://github.com/learning-unlimited/ESP-Website/blob/844310f1b6adc979f19665d86c33e2c3a1e213d5/esp/esp/program/controllers/lottery.py#L305)**

```python
for (student_id, section_id) in interest_regs_sr + interest_regs_ssi:
```

Both operands are querysets from `.values_list(...).distinct()`. Django's
`QuerySet` has no `__add__`, so this raises before iterating anything:

```
TypeError: unsupported operand type(s) for +: 'QuerySet' and 'QuerySet'
```

**2. numpy integer defeats the type guard in `getRankInClass`**

```python
ESPUser.getRankInClass(student_id, self.parent_classes[self.section_indices[section_id]])
```

`self.parent_classes` is a numpy array, so the second argument is a
`numpy.int64`, not an `int`. [getRankInClass](https://github.com/learning-unlimited/ESP-Website/blob/844310f1b6adc979f19665d86c33e2c3a1e213d5/esp/esp/users/models/__init__.py#L1335-L1339)
begins with `if isinstance(subject, int): subject = ClassSubject.objects.get(id=subject)`,
and `isinstance(numpy.int64(1), int)` is `False`, so the id is never resolved to
a `ClassSubject` and the raw numpy scalar is used as a model object further down.

**3. `subject.teachers()` is not callable — [users/models/__init__.py#L1347](https://github.com/learning-unlimited/ESP-Website/blob/844310f1b6adc979f19665d86c33e2c3a1e213d5/esp/esp/users/models/__init__.py#L1347)**

`teachers` is a `ManyToManyField` on `ClassSubject`, so `subject.teachers` is a
manager and calling it raises:

```
TypeError: ManyRelatedManager.__call__() missing 1 required keyword-only argument: 'manager'
```

This one is not lottery-specific — `getRankInClass` is also called while rendering
the class change request page
([classchangerequestmodule.py#L93](https://github.com/learning-unlimited/ESP-Website/blob/844310f1b6adc979f19665d86c33e2c3a1e213d5/esp/esp/program/modules/handlers/classchangerequestmodule.py#L93)),
so the same crash is reachable there. It only fires once a subject actually has
`StudentAppQuestion`s and the student has an application with a non-empty
response; before that the function returns early at the `10` / `1` short circuits,
which is why programs that don't use student applications never see it.

## Proposed Fix

```diff
+import itertools
...
-            for (student_id, section_id) in interest_regs_sr + interest_regs_ssi:
+            for (student_id, section_id) in itertools.chain(interest_regs_sr, interest_regs_ssi):
```

plus `int(...)` around `self.parent_classes[...]` at both call sites, and:

```diff
-        rank = max(list(StudentAppReview.objects.filter(..., reviewer__in=subject.teachers()) ...
+        rank = max(list(StudentAppReview.objects.filter(..., reviewer__in=subject.get_teachers()) ...
```

[get_teachers()](https://github.com/learning-unlimited/ESP-Website/blob/844310f1b6adc979f19665d86c33e2c3a1e213d5/esp/esp/program/models/class_.py#L1734-L1741) is the existing
documented accessor on `ClassSubject` and is already `@cache_function`-backed.

With the fix, the identical request returns HTTP 200:

![After the fix: HTTP 200 with lottery stats]

### Steps to Reproduce

1. Open "Run the Lottery Assignment Thing" for a program with the lottery module.
2. Set `use_student_apps` to `True`.
3. Click "Generate Student Schedules".

Same request with the option left at `False` returns HTTP 200, so the option is
the only variable.

### Expected Behavior

The lottery runs, `self.ranks` is populated from application review scores, and
`compute_assignments()` makes its two passes over ranks `(10, 5)` as the rest of
the option's code intends.

### Actual Behavior

HTTP 500 from `LotteryAssignmentController.__init__` -> `initialize()`, before any
assignment happens. Fixing that one exposes the next, and so on.

### Screenshots

Image
Image

### Operating System

win 11

### Browser

Chrome

### Additional Context

There is currently no test coverage for `use_student_apps` anywhere in the suite,
which is how three stacked crashes survived. Happy to send a PR with the fix plus
three regression tests in `test_lottery_controller.py`:

- controller initializes with the option enabled;
- a reviewed application produces the expected entry in `ranks`;
- `compute_assignments()` completes with the option enabled.

All three fail on `main` with exactly the three errors above and pass with the
fix. The existing 18 lottery controller tests still pass, as do all 302 tests in
`esp/users` and `esp/program/controllers`.

sanketkokate712:
The Problem: 

When administrators enabled the "use student application ranks" option for the class lottery, the lottery would immediately crash and return a 500 Server Error instead of generating student schedules. This was caused by three separate bugs in the code:

It tried to add two database querysets together using a + sign (which isn't allowed in Django).
It passed a numpy integer instead of a standard Python integer when looking up a class, causing the lookup to fail.
It incorrectly tried to call .teachers() as a function when checking for application reviewers.

The Fix:

Replaced the + sign with itertools.chain() to safely combine the database queries.
Wrapped the numpy value in int() so the class lookup succeeds.
Changed .teachers() to the correct .get_teachers() method.
Added new automated tests to ensure this specific lottery feature doesn't break again in the future.

Fixes #5996

### DIFF
--- a/esp/esp/program/controllers/lottery.py
+++ b/esp/esp/program/controllers/lottery.py
@@ -55,6 +55,7 @@ from django.db import transaction
 from django.db.models import Min
 import os
 import operator
+import itertools
 import zlib
 import base64
 from io import BytesIO
@@ -301,9 +302,9 @@ class LotteryAssignmentController(object):
         if self.options['use_student_apps']:
             for i in range(1, self.effective_priority_limit+1):
                 for (student_id, section_id) in priority_regs[i]:
-                    self.ranks[self.student_indices[student_id], self.section_indices[section_id]] = ESPUser.getRankInClass(student_id, self.parent_classes[self.section_indices[section_id]])
-            for (student_id, section_id) in interest_regs_sr + interest_regs_ssi:
-                self.ranks[self.student_indices[student_id], self.section_indices[section_id]] = ESPUser.getRankInClass(student_id, self.parent_classes[self.section_indices[section_id]])
+                    self.ranks[self.student_indices[student_id], self.section_indices[section_id]] = ESPUser.getRankInClass(student_id, int(self.parent_classes[self.section_indices[section_id]]))
+            for (student_id, section_id) in itertools.chain(interest_regs_sr, interest_regs_ssi):
+                self.ranks[self.student_indices[student_id], self.section_indices[section_id]] = ESPUser.getRankInClass(student_id, int(self.parent_classes[self.section_indices[section_id]]))
 
 
         #   Populate section schedule
--- a/esp/esp/program/controllers/test_lottery_controller.py
+++ b/esp/esp/program/controllers/test_lottery_controller.py
@@ -20,6 +20,12 @@ from esp.program.models import (
     StudentRegistration,
     StudentSubjectInterest,
 )
+from esp.program.models.app_ import (
+    StudentAppQuestion,
+    StudentAppResponse,
+    StudentAppReview,
+    StudentApplication,
+)
 from esp.program.tests import ProgramFrameworkTest
 from esp.users.models import ESPUser, StudentInfo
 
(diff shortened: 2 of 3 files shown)


Oval17 (approved):
lgtm

PR #6042, opened 2026-09-15
Fix 5412 duplicate timeslot

### LINKED ISSUE #5412: [Bug]: Duplicate timeslot entries allowed on manual save (no validation in TimeslotForm)
### Description

There was no validation for duplicate timeslots on manual add/edit. As a result, identical rows were inserted into the Event table, creating redundant timeslots. Import has a duplicate check, but manual entry didn’t, so the UI allowed the same data to be saved repeatedly.

```
def add_or_edit_timeslot(self, form):
        """ form is a TimeslotForm object   """
        if form.cleaned_data['id'] is not None:
            new_timeslot = Event.objects.get(id=form.cleaned_data['id'])
        else:
            new_timeslot = Event()
        form.save_timeslot(self.program, new_timeslot)
        return new_timeslot
```

### Steps to Reproduce

1. Go to Manage → Resources → Timeslots.
2. Add a timeslot with a given name, description, start time, duration, type, and group.
3. Submit the same exact values again via the manual form.

### Expected Behavior

The form should reject an identical timeslot and show a validation error (non‑field error) without inserting a duplicate.

### Actual Behavior

A duplicate is inserted into Event, resulting in multiple identical timeslots.

### Screenshots

_No response_

### Operating System

_No response_

### Browser

_No response_

### Additional Context

Import path already checks duplicates via Event.objects.filter(program, start, end).exists  in resources_timeslot_import. Manual add/edit uses TimeslotForm.save_timeslot() without duplicate validation.

DarshanS-Dev:
## Description

Fixes duplicate `Event` timeslot rows being silently created when adding/editing timeslots manually via Manage → Resources → Timeslots. `TimeslotForm` had no duplicate check at all, so submitting the same start/end time twice created two separate `Event` rows for the same slot.

`TimeslotForm.__init__` now stores `self.program` (previously a local variable only used for `help_text`), making it available in `clean()`. `clean()` adds a duplicate check alongside the existing openclass/compulsory check: it queries `Event.objects.filter(program=self.program, start=start, end=end)`, excluding the current row's pk when editing, and raises a `forms.ValidationError` if a match is found. The duplicate criterion matches the project's existing precedent in `resources_timeslot_import`.

The handler's `addedit` branch now imports and catches `django.db.IntegrityError` around `controller.add_or_edit_timeslot(form)`, converting the rare race-condition case into a normal form error rather than a raw 500.

**Database & Migrations:**
- `Event` (in `esp/esp/cal/models.py`) gets a `UniqueConstraint` on `['program', 'start', 'end', 'event_type']`, named `unique_event_program_start_end`. Including `event_type` scopes the invariant safely so non-timeslot events (like Volunteers) can continue to coexist. This serves as a DB-level backstop for the check-then-save race condition.
- **Migration 0006 (`0006_dedupe_events_before_constraint.py`)**: A data migration that finds existing duplicate overlapping events and merges their related foreign keys (ClassSection, ClassSubject, VolunteerRequest, UserAvailability, etc.) into a single keeper record before deleting the duplicates. This ensures the constraint can be cleanly applied to production databases.
- **Migration 0007 (`0007_event_unique_constraint.py`)**: Adds the constraint.
- **Migration Patch (`program/migrations/0027_auto_20240220_2124.py`)**: Added a missing `dbmail` dependency to this 2-year-old migration. The addition of our new `cal` migrations altered Django's topological sort graph, pulling `program/0027` earlier in the execution plan during `--create-db` and unmasking a latent crash where it attempted to query `dbmail.PlainRedirect` before the app was loaded. 

Note: the import path (`resources_timeslot_import`) handles duplicates by silently skipping them rather than raising. This form raises an explicit validation error instead, since silently discarding a single interactive submission would leave the user with no explanation why nothing happened — worse UX than the bulk-preview context where skip-and-show-in-preview makes sense.

## Related Issue

Closes #5412

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] CI/CD or build change

## Testing

- [x] Existing tests pass
- [x] Manually verified

Added three tests to the existing `ResourceModuleTest` class in `esp/esp/program/modules/tests/resourcemodule.py`, following that file's convention of POSTing through the real handler URL rather than calling the controller/ORM directly:

- `testTimeslotDuplicateRejected` — submitting an identical timeslot twice is rejected with a form validation error.
- `testTimeslotEditWithoutChangeDoesNotSelfFlag` — editing an existing timeslot without changing start/end does not falsely flag it as a duplicate of itself.
- `testTimeslotUniqueConstraintHoldsAtDbLevel` — bypassing the form and calling `Event.objects.create()` twice with identical `program`/`start`/`end` raises `IntegrityError` on the second call.

Ran the full suite locally via Docker:

```
docker compose exec -w /app/esp web pytest
```

2412 passed, 3 skipped (pre-existing/environment: SQLite `FOR UPDATE` limitation, no Firefox, opt-in selenium), 0 failures.

## AI Disclosure

I used AI for code generation assistance. All design decisions, testing, and verification were done by me.

## Checklist

- [x] Self-reviewed the code


DarshanS-Dev:
Hey @willgearty , just wanted to flag that I've pushed an updated 
commit addressing the Copilot review feedback:

- Scoped the `UniqueConstraint` to include `event_type` so non-class 
  events (e.g. Volunteers) can coexist safely
- Added a data migration (`0006`) to deduplicate any existing rows 
  before the constraint is applied
- Fixed the missing `IntegrityError` import in the handler
- Also patched a latent missing dependency in `program/0027` that was 
  unmasked by our new migration graph

PR description has been updated to reflect all of this. Happy to make 
any further changes if needed!

